PR #41, opened 2019-03-27
JSON api description: Support for fixed-size arrays and tuples

### LINKED ISSUE #40: JSON api description: Support for fixed-size arrays and tuples
Our current JSON api description implementation is kept very minimal and can only support the Rust primitives bool, u{8,16,32,64,128} and i{8,16,32,64,128} as well as the SRML primitives Address and Balance.

To make the api more versatile we should also add tuples and arrays of fixed sizes.

## JSON

This section covers the JSON representation of the api description.

### Tuples

If we have an API like the following:
```rust
pub(external) fn get_pair() -> (bool, i32) { ... }
```
We want to generate JSON for the return type of this api description in the following way:
```json
ret_ty: [ "bool", "i32" ]
```

API description data structure extension with the following variant:

```rust
enum TypeDescription {
    ...
    Tuple {
         elems: Vec<TypeDescription>,
    }
    ...
}
```

### Arrays

If we have an API like the following:
```rust
pub(external) fn receive_array(vec3: [i32; 3]) { ... }
```
We want to generate JSON for the parameter of this api description in the following way:
```json
params: [
    {
        "name": "vec3",
        "ty": {
            "[T;n]": {
                "T": "i32",
                "n": 3
            }
        }
    }
]
```

API description data structure extension with the following variant:

```rust
enum TypeDescription {
    ...
    Array {
        inner: Box<TypeDescription>,
        arity: u32
    }
    ...
}
```

ascjones:
Fixes #40.

- Splits the top level `TypeDescription` enum into `Primitive`, `Array`, and `Tuple` variants.
- Adds some tests


lang/src/api.rs
```diff
@@ -69,34 +69,72 @@ pub enum TypeDescription {
     Address,
     /// The SRML balance type.
     Balance,
+    /// The tuple type
+    Tuple {
```
Robbepop: I just got the idea that we can even improve this further by doing something similar to how `syn` handles all its complex `enum` types: https://docs.rs/syn/0.15.29/syn/enum.Type.html

So we could define a `TupleTypeDescription` and `ArrayTypeDescription` where all the tuple- and array type description specific implementations are going. They would both implement their own `TryFrom` for `syn::TypeTuple` and `syn::TypeArray` respectively and we could have a much simpler `TryFrom` impl for `TypeDescription`.

lang/src/api.rs
```diff
@@ -69,34 +69,72 @@ pub enum TypeDescription {
...
+                    )
+                }
+            }
+        };
+        match ty {
+            syn::Type::Tuple(tuple) => {
+                let elems = tuple
+                    .elems
+                    .iter()
+                    .map(primitive)
```
Robbepop: Interesting approach with the primitives map. However, as far as I can see this restricts us from having recursive types like `(u32, (bool, u32))` or `[[u32; 5]; 5]`. I think this is a simple fix by again moving the primitives into the root match and together with my suggestion above just simply call `TypeDescription::try_from` recursively in the `TupleTypeDescription` and `ArrayTypeDescription` respectively.

ascjones:
Implemented your suggestions and added some tests. The json ones are failing, I will need to customise the serde serialization for those variants to match the format specified in the issue.

ascjones:
To achieve the specified json representation I have:
- Extracted primitive types into their own enum (so they remain tagged)
- Annotated the top level `TypeDescription` enum so that it is `untagged`
- Made `ArrayDescription` a single case enum to allow renames

Note that in the last case the resulting json will be a json object e.g.
```
{
    "[T;n]": {
        "T": "u32",
        "n": "5"
    }
}
```
Assume this is correct because it will be the value of a type in the higher order structure.

Robbepop:
lgtm!

Now only the tuple-array mixed tests are missing for a merge. :)

PR #43, opened 2019-04-01
Remove +nightly qualifier because of rust-toolchain

pepyakin:
This repository already has rust-toolchain file, but at the present in all build.sh there are inconsisency:

the first cargo invocation is without a qualifier
then the second cargo invocation is with +nightly qualifier

this means that build.sh will typically download the required toolchain, but then will use `nightly`. This might be a problem if `nightly` version is outdated.

### DIFF
--- a/cli/src/cmd/new.rs
+++ b/cli/src/cmd/new.rs
@@ -149,7 +149,7 @@ fn build_sh_contents(name: &str) -> String {
 PROJNAME={}
 
 CARGO_INCREMENTAL=0 &&
-cargo +nightly build --release --target=wasm32-unknown-unknown --verbose &&
+cargo build --release --target=wasm32-unknown-unknown --verbose &&
 wasm2wat -o target/$PROJNAME.wat target/wasm32-unknown-unknown/release/$PROJNAME.wasm &&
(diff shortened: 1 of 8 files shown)


Robbepop:
We should not forget to also change this in the rudimentary CLI tool.
Could you also pls rename the commit as noted in the [contribution guidelines](https://github.com/pepyakin/pdsl/blob/master/CONTRIBUTING.md)? :)

pepyakin:
> We should not forget to also change this in the rudimentary CLI tool.

What do you mean? If `cli/src/cmd/new.rs` then it is already in : )

> Could you also pls rename the commit as noted in the contribution guidelines? :)

Sure, `[chores]`?

Robbepop:
> 
> 
> > We should not forget to also change this in the rudimentary CLI tool.
> 
> What do you mean? If `cli/src/cmd/new.rs` then it is already in : )
> 
> > Could you also pls rename the commit as noted in the contribution guidelines? :)
> 
> Sure, `[chores]`?

Ah you are right I didn't see it!
`[chores]` sounds good for that but I have the feeling that in reality we are missing some further categories that we might want to add. Thanks!

PR #52, opened 2019-04-15
[core,model] Add `ext_deposit_event` runtime call bindings

shawntabrizi:
A litle bit of guessing how to integrate this into PDSL using the [commit for adding println](https://github.com/Robbepop/pdsl/commit/8517e0b9a97132e714cb9d894bab35daa9385e33) and [commit for ext_deposit_event](https://github.com/paritytech/substrate/commit/e748c33c8ef40bd51dc122b7f11fcf26b8354957)

Right now, I have it so test_env prints the event data when it is called. Is this right?

Is anything here right?

Fixes: https://github.com/Robbepop/pdsl/issues/51


shawntabrizi:
@thiolliere you may want to check it out too

Robbepop:
We also might want to have @pepyakin take a view on it for the Substrate side.

Also I am not sure about this but could be this advantageous also for calling events in Substrate?
https://github.com/Robbepop/pdsl/issues/42#issuecomment-477219821
Right now we use an interpreted buffer (the `Vec<u8>`) but maybe we can do better?

core/src/env/api.rs
```diff
@@ -65,11 +65,16 @@ pub fn println(content: &str) {
     ContractEnv::println(content)
 }
 
+/// Deposits an event
+pub fn deposit_event(data: Vec<u8>) {
```
Robbepop: For byte buffers at this stage you want to use `&[u8]` instead.
Also the calls in `api.rs` are meant to be a bit more high level if possible.
So I would prefer a solution that is more abstract over events and help users better to get the encoding right (best if no encoding would be needed ofc.)
Might this be improved by: https://github.com/Robbepop/pdsl/issues/42#issuecomment-477219821 ?

core/src/env/srml/srml_only/impls.rs
```diff
@@ -125,4 +125,10 @@ where
             sys::ext_println(content.as_ptr() as u32, content.len() as u32)
         }
     }
+
+    fn deposit_event(data: Vec<u8>) {
```
Robbepop: Same here you want to use a `&[u8]`

core/src/env/srml/srml_only/sys.rs
```diff
@@ -54,6 +54,9 @@ extern "C" {
     /// This must only be used in `--dev` chain environments!
     pub fn ext_println(str_ptr: u32, str_len: u32);
 
+    /// Deposits an event
+    pub fn ext_deposit_event(data_ptr: u32, data_len: u32);
```
Robbepop: Please improve the description and for comments please respect our (to be honest) very pedantic contribution guidelines: https://github.com/Robbepop/pdsl/blob/master/CONTRIBUTING.md

core/src/env/test_env.rs
```diff
@@ -263,6 +263,11 @@ impl TestEnvData {
     pub fn println(&self, content: &str) {
         println!("{}", content)
     }
+
+    /// Deposits an event
+    pub fn deposit_event(&self, data: Vec<u8>) {
+        println!("{:?}", data)
```
Robbepop: For testing purposes we definitely are not interested in printing out received events but more into storing them in a vector so that after an execution we can check if we received the correct events. For this you want to add a new field into the `TestEnv` data struct and provide it with getters to the vec (that returns a slice).

core/src/env/traits.rs
```diff
@@ -78,4 +78,7 @@ pub trait Env: EnvTypes + EnvStorage {
     ///
     /// Usable only in development (`--dev`) chains.
     fn println(content: &str);
+
+    /// Deposits a contract event through Contracts module
+    fn deposit_event(data: Vec<u8>);
```
Robbepop: Also `&[u8]` instead of `Vec<u8>` here

core/src/env/api.rs
shawntabrizi: Would want to talk to you about the second half of your comment in person :)

core/src/env/test_env.rs
shawntabrizi: I have done this I think, but did not introduce a getter. Where would that live?

core/src/env/srml/srml_only/impls.rs
```diff
@@ -121,8 +121,10 @@ where
     }
 
     fn println(content: &str) {
-        unsafe {
-            sys::ext_println(content.as_ptr() as u32, content.len() as u32)
-        }
+        unsafe { sys::ext_println(content.as_ptr() as u32, content.len() as u32) }
```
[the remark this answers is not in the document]
shawntabrizi: Yes it was: https://github.com/Robbepop/pdsl/pull/52/commits/21e48712123f890972c6e5cc6794734968aa19bf

core/src/env/test_env.rs
```diff
@@ -120,6 +120,8 @@ pub struct TestEnvData {
     total_reads: Cell<u64>,
     /// The total number of writes to the storage.
     total_writes: u64,
+    /// Deposited events of the contract invocation.
+    events: Vec<u8>,
```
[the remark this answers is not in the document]
shawntabrizi: Not quite sure this makes sense.

Right now, the only data we can send to `ext_deposit_event` is some `Vec<u8>`

This is because on the contracts module side we have defined an Event called `Contract(Vec<u8>)` which then gets placed inside of the `Events` enum with all the other possible events.

I could make a `ContractEvent` type of `Event::Contract(Vec<u8>)` or something, but I am not sure if this is right...

core/src/env/test_env.rs
```diff
@@ -28,6 +28,9 @@ use core::cell::{
     RefCell,
 };
 
+/// A wrapper for the generic bytearray used for data in contract events.
+struct EventData(Vec<u8>);
```
Robbepop: You might want to add `#[derive(Debug, PartialEq, Eq, ...)]` here since the purpose of the test-environment is to introspect it.

Robbepop:
I see no issues with this PR in regards to our CI so I will simply merge it now.

Robbepop:
Is this already based on master?

PR #54, opened 2019-04-15
[core,model] Add `ext_random_seed` runtime call bindings

shawntabrizi:
Adding the ability for contracts to get the random seed of the block.

Returning the value as a Vec<u8>, although within substrate it is treated as a `Hash`... should/how do I use the `Hash` type?

Fixes https://github.com/Robbepop/pdsl/issues/53


Robbepop:
I see that `fn random_seed` in Substrate is returning something like `<Block as BlockT>::Hash`.
So this is pretty similar in respect to how we handle `Address` and `Balance` currently.
The best for pDSL users is to not have returned a `Vec<u8>` but instead getting the same type (`<Block as BlockT>::Hash`) returned. We can achieve this by doing something like this: https://github.com/Robbepop/pdsl/issues/42#issuecomment-477219821

So I think we should really first base pDSL upon `node-template` (happy about your ideas around that) and then we can continue integrating those calls like `random_seed` with proper return types.

core/src/env/srml/types.rs
```diff
@@ -44,3 +45,17 @@ impl<'a> From<&'a [u8]> for Address {
 
 /// The default SRML balance type.
 pub type Balance = u64;
+
+/// The default SRML hash type.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Encode, Decode)]
+pub struct Hash([u8; 32]);
```
Robbepop: Are `node-runtime` hash types internally `[u8; 32]`?

core/src/env/srml/types.rs
```diff
@@ -44,3 +45,17 @@ impl<'a> From<&'a [u8]> for Address {
...
+
+impl<'a> From<&'a [u8]> for Hash {
```
Robbepop: With the approach of `TryFrom` we can do slightly better here by implementing a `impl<'a> From<&'a [u8; 32]> for Hash` instead. With `TryFrom` we can very simply convert from `&[u8]` into `&[u8;32]`.

core/src/env/test_env.rs
```diff
@@ -134,7 +134,7 @@ impl Default for TestEnvData {
             storage: HashMap::new(),
             caller: vec![0x0; 32],
             input: Vec::new(),
-            random_seed: Vec::new(),
+            random_seed: vec![0x0, 32],
```
Robbepop: Not sure if a random seed requires a fixed length vector.
I always think of a random seed (to initialize random data) as something like a very generic byte sequence.

But in this case we want to use `Hash` here.

core/src/env/test_env.rs
```diff
@@ -190,6 +198,11 @@ impl TestEnvData {
     pub fn set_input(&mut self, input_bytes: &[u8]) {
         self.input = input_bytes.to_vec();
     }
+
+    /// Sets the random seed for the next contract invocation.
+    pub fn set_random_seed(&mut self, random_seed_bytes: &[u8]) {
```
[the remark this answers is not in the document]
shawntabrizi: I only added it cause I have no idea how the random seed would be generated otherwise...

core/src/env/srml/types.rs
shawntabrizi: They are H256 and handled this way throughout.

core/src/env/srml/types.rs
shawntabrizi: Can we do this in another PR for both `Hash` and `Address`?

Robbepop: sure

core/src/env/srml/types.rs
Robbepop: then it is perfect

core/src/env/test_env.rs
Robbepop: `random_seed: Hash,`

shawntabrizi: Because of how similar this is to the changes in `Address`, I think we should also do this in the other PR. That way, if any issues arise as a cause of this change, we can cherry pick it out.

The current pattern matches what was there before.

core/src/env/test_env.rs
```diff
@@ -146,9 +146,9 @@ impl TestEnvData {
     /// Resets `self` as if no contract execution happened so far.
     pub fn reset(&mut self) {
         self.storage.clear();
-        self.caller.clear();
+        self.caller = srml::Address([0; 32]);
```
Robbepop: Can you please use `0x0` here instead of `0`?
It is just to signal that this is a byte blob.

core/src/env/srml/types.rs
```diff
@@ -31,7 +31,7 @@ impl EnvTypes for DefaultSrmlTypes {
 
 /// The default SRML address type.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Encode, Decode)]
-pub struct Address([u8; 32]);
+pub struct Address(pub [u8; 32]);
```
Robbepop: I am not sure about `pub` here (same in `Hash`). Generally for the sake of encapsulation and type safety I try to avoid using `pub` fields, especially in tuple structs. The reason for this is: If the inner field is private we cannot for example convert a `Hash` easily into an `Address` instance - thus avoiding the strict type system. If a users receives an `Address` he should only be able to use it as such. On the other hand this might be a bit restrictive in the general sense so as a backdoor for users that know what they are doing we would implement some methods to receive their internals.

shawntabrizi: I needed to do this in order to construct the `Address` and `Hash` type in the `test_env`.

Like so:
```
self.caller = srml::Address([0; 32]);
```

Otherwise, I guess I could implement a public constructor?

Robbepop: > Otherwise, I guess I could implement a public constructor?

shawntabrizi: Thanks for taking the time to work with me here :)

Latest commit adds public constructors `new()` and uses them to initialize these types from `[u8; 32]`

core/src/env/srml/types.rs
```diff
@@ -34,7 +34,13 @@ impl EnvTypes for DefaultSrmlTypes {
 
 /// The default SRML address type.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Encode, Decode)]
-pub struct Address(pub [u8; 32]);
+pub struct Address([u8; 32]);
+
+impl Address {
+    pub fn new(address: [u8; 32]) -> Address {
```
Robbepop: For implementations like these it is recommended to just implement `From<[u8;32]>`.

Robbepop: Same for `Hash`

Robbepop:
> LGTM

Robbepop:
I only see formatting issues in the CI not related to this PR so I will merge this now.

PR #60, opened 2019-04-17
[core] Introduce `ext_dispatch_call` and use node-runtime types

ascjones:
Still outstanding: 

- [ ] Implement `ext_dispatch_call` for `TestEnv`
- [ ] Test contract with dispatch call 
- [ ] Convert other types (`Balance`, `Hash`) to use `node-runtime` types?


Robbepop:
We probably also want a code review here from @pepyakin .

core/Cargo.toml
```diff
@@ -19,11 +19,15 @@ include = ["/Cargo.toml", "src/**/*.rs", "/README.md", "/LICENSE"]
 
 [dependencies]
 pdsl_utils = { path = "../utils/" }
-parity-codec = { version = "3.2", default-features = false, features = ["derive", "full"] }
+parity-codec = { version = "3.3", default-features = false, features = ["derive", "full"] }
 log = "0.4"
 wee_alloc = { version = "0.4", default-features = false }
+node-runtime = { git = "https://github.com/paritytech/substrate/", branch = "v1.0", package = "node-runtime", default-features = false }
```
Robbepop: So having `node-runtime` in our dependency tree means that we now have a shared dependency base with it which is huge I assume. Since `cargo` cannot properly unshare crate features of dependencies we have to be really careful about overlapping ones.

core/src/env/api.rs
```diff
@@ -39,6 +40,10 @@ pub fn caller() -> Address {
     ContractEnv::caller()
 }
 
+pub fn dispatch_call(value: node_runtime::Call) {
```
Robbepop: Missing documentation

core/src/env/srml/srml_only/impls.rs
```diff
@@ -94,7 +95,7 @@ where
 impl<T> Env for SrmlEnv<T>
 where
     T: EnvTypes,
-    <T as EnvTypes>::Address: for<'a> TryFrom<&'a [u8]>,
+    <T as EnvTypes>::Address: Decode,
     <T as EnvTypes>::Hash: for<'a> TryFrom<&'a [u8]>,
```
Robbepop: Those two lines look asymmetric. Would have suspected that `Hash` requires the same interface now. Also I am not sure if we are still in need of `EnvTypes` due to `node-runtime`.

core/src/env/srml/types.rs
```diff
@@ -34,24 +35,8 @@ impl EnvTypes for DefaultSrmlTypes {
...
-    fn try_from(bytes: &'a [u8]) -> Result<Address, TryFromSliceError> {
-        let address = <[u8; 32]>::try_from(bytes)?;
-        Ok(Address(address))
-    }
-}
+/// The default SRML address type
+pub type Address = node_runtime::Address;
 
 /// The default SRML balance type.
 pub type Balance = u64;
```
Robbepop: Can you convert all pDSL types to `node-runtime` types? So also `Hash` and `Balance` etc.?

core/src/env/traits.rs
```diff
@@ -62,6 +62,8 @@ pub trait Env: EnvTypes + EnvStorage {
     /// Returns the chain address of the caller.
     fn caller() -> <Self as EnvTypes>::Address;
 
+    fn dispatch_call(data: &[u8]);
```
Robbepop: Missing documentation

core/src/lib.rs
```diff
@@ -24,18 +24,15 @@
...
-#[global_allocator]
-static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
-
 #[cfg(all(test, feature = "std"))]
 #[macro_use]
 mod test_utils;
 
-#[cfg(not(feature = "std"))]
-mod panic_handler;
-
```
Robbepop: So I guess this functionality is now handled by `node-runtime`. Is there a way to control it from pDSL?

core/src/lib.rs
```diff
@@ -24,18 +24,15 @@
 // #[macro_use]
 extern crate alloc;
 
-// Use `wee_alloc` as the global allocator.
-#[global_allocator]
-static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
-
```
Robbepop: I assume `node-runtime` is using the same allocator?

examples/lang/runtime/.cargo/config
```diff
@@ -0,0 +1,5 @@
+[target.wasm32-unknown-unknown]
+rustflags = [
+	"-C", "overflow-checks=on",
+	"-C", "link-args=-z stack-size=65536 --import-memory"
+]
```
Robbepop: why this change?

examples/lang/runtime/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "runtime"
```
Robbepop: I would maybe call this example `calling_the_runtime` or something like that or `calling_back`

examples/lang/runtime/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+version = "0.1.0"
+authors = ["admin@parity.io"]
+edition = "2018"
+
+[dependencies]
+pdsl_core = { path = "../../../core" }
+pdsl_model = { path = "../../../model" }
+pdsl_lang = { path = "../../../lang" }
+parity-codec = { version = "3.2", default-features = false, features = ["derive"] }
```
Robbepop: I think above you used `parity-codec 3.3`

model/src/exec_env.rs
```diff
@@ -125,6 +126,10 @@ impl EnvHandler {
         env::caller()
     }
 
+    pub fn dispatch_call(&self, call: Call) {
```
Robbepop: Missing documentation

core/Cargo.toml
Robbepop: We can remove `wee_alloc` dependency here.
Also `pdsl_core` should not depend on `syn`.

ascjones:
Closing this and splitting into 2 PRs. First of which is #72

PR #63, opened 2019-04-21
[example] Trying to improve ERC20 example

shawntabrizi:
The functions here _should_ work... but don't.


Robbepop:
Closed because we already modernized and fixed the same example contract.

Issue #66, opened 2019-04-22
ink! mistreats Address as being the same as AccountId


shawntabrizi:
Right now, ink! uses the term `Address` synonymously with `AccountId`. However, this is not the case.

`Address` is a wrapper for `AccountId` and `AccountIndex`, with a 1 byte prefix to determine which one it represents.

This causes decoding issues within ink! because it assumes that `Address` is `AccountId`, ultimately misrepresenting the input address from contract calls, and giving bad results.

We should not add support for `Address`, because that would involve the Contracts layer having to query the Runtime to support index lookup. Instead, we should assume that we only support `AccountId` directly.

The fix here should be to replace all instances of the term `Address` from ink!, and replace it with `AccountId` to prevent any further confusion or bugs.

Robbepop:
Probably fixed by https://github.com/paritytech/ink/pull/67. Closed.

PR #67, opened 2019-04-22
[*] Rename Address to AccountId globally

Robbepop:
This should resolve confusions about Address and AccountId encoding/decoding.


Robbepop:
Fixes https://github.com/paritytech/ink/issues/66

PR #70, opened 2019-04-22
[cli] Update `new` cli for latest flipper

shawntabrizi:
- Fixes the test in the example contract
- Changes some hard coded instances of "Flipper" to use the desired variable name
- Calls println when the getter is called

Tested on MacOS

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cargo-contract"
 version = "0.1.0"
-authors = ["Robin Freyler <robin@parity.io>", "Parity Technologies <admin@parity.io>"]
+authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
 license = "GPL-3.0"
--- a/cli/README.md
+++ b/cli/README.md
@@ -1,0 +1,3 @@
+# Cargo plugin for Ink contracts
+
+A small CLI tool for helping setting up and managing WebAssembly smart contracts written with ink!.
--- a/cli/src/cmd/new.rs
+++ b/cli/src/cmd/new.rs
@@ -101,7 +101,11 @@ fn lib_rs_contents(name: &str) -> String {
     format!(
         r##"#![no_std]
 
-use ink_core::storage;
+use ink_core::{{
+    env::println,
+    memory::format,
+    storage,
+}};
 use ink_lang::contract;
 
 contract! {{
@@ -128,6 +132,7 @@ contract! {{
 
         /// Returns the current state.
         pub(external) fn get(&self) -> bool {{
+            println(&format!("Storage Value: {{:?}}", *self.value));
             *self.value
         }}
     }}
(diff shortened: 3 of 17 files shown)


cli/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "cargo-contract"
 version = "0.1.0"
-authors = ["Robin Freyler <robin@parity.io>", "Parity Technologies <admin@parity.io>"]
+authors = ["Parity Technologies <admin@parity.io>"]
```
Robbepop: I don't understand this change since I thought that this is already in master, isn't it?

cli/README.md
```diff
@@ -0,0 +1,3 @@
+# Cargo plugin for Ink contracts
+
+A small CLI tool for helping setting up and managing WebAssembly smart contracts written with ink!.
```
Robbepop: Same here ...

cli/src/cmd/new.rs
```diff
@@ -128,25 +132,26 @@ contract! {{
...
+            println(&format!("Storage Value: {{:?}}", *self.value));
             *self.value
         }}
     }}
 }}
 
 #[cfg(test)]
 mod tests {{
-    use super::Flipper;
+    use super::{};
```
Robbepop: maybe you mean to be using `use super::*;` here?

cli/src/cmd/new.rs
```diff
@@ -128,25 +132,26 @@ contract! {{
...
-        assert_eq!(flipper.get(), false);
+        let mut contract = {}::deploy_mock();
+        assert_eq!(contract.get(), false);
+        contract.flip();
+        assert_eq!(contract.get(), true);
     }}
 }}
 "##,
-        camel_name, camel_name, camel_name,
+        camel_name, camel_name, camel_name, camel_name, camel_name,
```
Robbepop: lol

cli/src/cmd/new.rs
shawntabrizi: That is probably more generic right? Can change.

cli/Cargo.toml
shawntabrizi: ¯\_(ツ)_/¯

cli/src/cmd/new.rs
shawntabrizi: Resolved

cli/Cargo.toml
Robbepop: Yeah so at the "Converation" tab it says that this branch is out of date and you should update it. Please do so. :S

cli/src/cmd/new.rs
```diff
@@ -161,12 +166,15 @@ fn build_sh_contents(name: &str) -> String {
...
-CARGO_INCREMENTAL=0 &&
-cargo build --release --features generate-api-description --target=wasm32-unknown-unknown --verbose &&
-wasm2wat -o target/$PROJNAME.wat target/wasm32-unknown-unknown/release/$PROJNAME.wasm &&
-cat target/$PROJNAME.wat | sed "s/(import \"env\" \"memory\" (memory (;0;) 2))/(import \"env\" \"memory\" (memory (;0;) 2 16))/" > target/$PROJNAME-fixed.wat &&
-wat2wasm -o target/$PROJNAME.wasm target/$PROJNAME-fixed.wat &&
-wasm-opt -Oz target/$PROJNAME.wasm -o target/$PROJNAME-opt.wasm"##,
+# cargo clean
+# rm Cargo.lock
+
+CARGO_INCREMENTAL=0 cargo build --release --features generate-api-description --target=wasm32-unknown-unknown --verbose
```
Robbepop: Why have you removed `&&` here?
I think our approach was to add `set -e`, wasn't it?

shawntabrizi: I copied what was in the build.sh script from the examples. Can you `suggest` an improvement to this?

shawntabrizi:
@ascjones Will try to help get this fixed...

cli/src/cmd/new.rs
Robbepop: Just add `set -e` on top of the script for now

Issue #42, opened 2019-03-27
Implement infrastructure for smart contract to call back into runtime


Robbepop:
Currently smart contracts are pretty isolated.
Wearing thick gloves one can call other remote smart contracts and even call back into the runtime, however, there is currently no support from the pDSL for doing that.

What we would like is a clean approach in calling into the runtime from the smart contract side.
This can be done by using [`ext_dispatch_call`](https://github.com/paritytech/substrate/blob/master/srml/contract/src/wasm/runtime.rs#L503) from the SRML contracts runtime module. It's documentation is currently unclear (at least to me) - further investigation needed.

Also we maybe need some infrastructure around it.

- [x] Find out how `ext_dispatch_call` is supposed to be called from a smart contract.
- [ ] Implement `ext_dispatch_call` rudimentary in `pdsl_core`
- [ ] Implement `ext_dispatch_call` in `pdsl_model`
- [ ] Optionally provide additional helper framework in order to improve the user experience while using `ext_dispatch_call` as good as possible for the current state of pDSL.

Robbepop:
For an MVP we want `pdsl_core` to depend on Substrate's `node-runtime` crate to have a common set of types to operate on. This also includes types such as `Call` which we can then use to make simple dispatches in pDSL to its underlying runtime.

PR #72, opened 2019-04-23
Use SRML types from node-runtime

ascjones:
Rel: https://github.com/paritytech/ink/issues/42. Split out from https://github.com/paritytech/ink/pull/60.

Still TODO:

- [ ] Fix `test-env` compilation errors
- [x] Remove temporary `syn` direct dependency
- [x] Remove direct dependencies on `srml-system` and `srml-balances` modules
- [ ] Test deploying and calling example contracts


core/src/env/api.rs
```diff
@@ -19,20 +19,23 @@ use crate::{
...
+    Runtime,
+};
 
 /// The environmental address type.
-pub type AccountId = <ContractEnv as EnvTypes>::AccountId;
+pub type AccountId = <Runtime as srml_system::Trait>::AccountId;
 
 /// The environmental balance type.
-pub type Balance = <ContractEnv as EnvTypes>::Balance;
+pub type Balance = <Runtime as srml_balances::Trait>::Balance;
```
xlc: could this be `srml_contract::BalanceOf<Runtime>` so it is more compatible with our runtime that doesn't use srml_balances module.

shawntabrizi: I agree here. The only module ink should depend on is the `contract` module. Any other types or dependencies need to come through the `contract` module to ensure that people can really customize their substrate chain in a modular way.

ascjones: Types in the `contract` module require an implementation of the `contract::Trait` which is why we import `node-runtime`.

I agree that this is suboptimal since chains will not necessarily have the same runtime module construction.

So we need to figure out the best way to provide a default `Runtime` but also allow a user supplied custom `Runtime`

Robbepop:
What is missing from this for the merge?

ascjones:
This needs rethinking properly. 

The essential problem with this current approach is that the types imported from `node_runtime` are only *guaranteed* to work on a chain which is running with the identical contract `Trait` [configuration](https://github.com/paritytech/substrate/blob/5091c4155ca502a4afb44d62f82bbc04adf02c33/node/runtime/src/lib.rs#L184)

Ideally we would allow the contract author to provide their own contract `Trait` impl. I ran into some difficulties here implementing this, which I will expand upon. 

I will open an issue about this where we can discuss the best approach, and come up with a fresh PR.

Issue #73, opened 2019-04-23
Issues with ERC20 Example


shawntabrizi:
* Transfer event is not in `transfer_impl`
* Tests should call the external functions, not internal ones

PR #75, opened 2019-04-23
[examples] Fixes to ERC20 contract

shawntabrizi:
Fixes https://github.com/paritytech/ink/issues/73

### DIFF
--- a/examples/lang/erc20/src/lib.rs
+++ b/examples/lang/erc20/src/lib.rs
@@ -184,6 +184,7 @@ contract! {
             }
             self.balances.insert(from, balance_from - value);
             self.balances.insert(to, balance_to + value);
+            Self::emit_transfer(from, to, value);
             true
         }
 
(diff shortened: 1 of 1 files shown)


examples/lang/erc20/src/lib.rs
```diff
@@ -257,7 +258,7 @@ mod tests {
         let bob = AccountId::try_from([0x1; 32]).unwrap();
         assert_eq!(erc20.total_supply(), 1234);
         assert_eq!(erc20.balance_of(alice), 1234);
-        assert_eq!(erc20.transfer_from(alice, bob, 234), true);
```
Robbepop: What is bad about `transfer_from` here? I think it is more explicit.

shawntabrizi: Cause then we never test the `transfer` function.

`transfer_from` is tested when you do the allowance stuff

shawntabrizi: I want to actually also add a failed test

Robbepop: Sounds fair!

Robbepop: With failure test you mean a test that `panics` or a test that returns `false` upon `transfer` ?

shawntabrizi: Just added. A test that tries to transfer more than Alice has, and returns false.

PR #76, opened 2019-04-23
[examples] Update ERC20 contract to use code from workshop

shawntabrizi:
This should fix the known bug, but may have some others in there...

Would rather have a more direct implementation: either simplified or directly copied from a known working sample.

### DIFF
--- a/examples/lang/erc20/src/lib.rs
+++ b/examples/lang/erc20/src/lib.rs
@@ -193,6 +193,7 @@ mod tests {
         assert_eq!(erc20.balance_of(bob), 234);
     }
 
+    #[test]
     fn allowance_works() {
         let mut erc20 = Erc20::deploy_mock(1234);
         let alice = AccountId::try_from([0x0; 32]).unwrap();
--- a/examples/lang/erc20/src/lib.rs
+++ b/examples/lang/erc20/src/lib.rs
@@ -110,16 +110,16 @@ contract! {
 
     impl Erc20 {
         /// Decreases the allowance and returns if it was successful.
-        fn try_decrease_allowance(&mut self, from: &AccountId, to: &AccountId, by: Balance) -> bool {
+        fn try_decrease_allowance(&mut self, from: &AccountId, by: Balance) -> bool {
             // The owner of the coins doesn't need an allowance.
             if &env::caller() == from {
                 return true
             }
-            let allowance = self.allowance_or_zero(from, to);
+            let allowance = self.allowance_or_zero(from, &env::caller());
             if allowance < by {
                 return false
             }
-            self.allowances.insert((*from, *to), allowance - by);
+            self.allowances.insert((*from, env::caller()), allowance - by);
             true
         }
 
@@ -160,7 +160,7 @@ contract! {
             if balance_from < value {
                 return false
             }
-            if !self.try_decrease_allowance(&from, &to, value) {
+            if !self.try_decrease_allowance(&from, value) {
                 return false
             }
             self.balances.insert(from, balance_from - value);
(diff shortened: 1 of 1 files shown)


examples/lang/erc20/src/lib.rs
```diff
@@ -135,16 +135,16 @@ contract! {
...
             if &env::caller() == from {
                 return true
             }
-            let allowance = self.allowance_or_zero(from, to);
+            let allowance = self.allowance_or_zero(from, &env::caller());
             if allowance < by {
                 return false
             }
-            self.allowances.insert((*from, *to), allowance - by);
+            self.allowances.insert((*from, env::caller()), allowance - by);
```
seunlanlege: unless this is already handled by the `Sub` implementation for balance, shouldn't this be a checked sub?

shawntabrizi: This configuration line:
https://github.com/paritytech/ink/blob/master/examples/lang/erc20/.cargo/config#L3

ensures that all math in the contract is "safe math" i.e. any overflow/underflow will revert the transaction.

Something we will need to teach to upcoming ink! developers.

Please look for other issues though!

seunlanlege: ah, nice one. :open_hands:

shawntabrizi:
There are more broken-ish things from ERC20, don't merge yet.

examples/lang/erc20/src/lib.rs
```diff
@@ -34,18 +34,16 @@ use ink_lang::contract;
 /// Events deposited by the ERC20 token contract.
 #[derive(Encode, Decode)]
 enum Event {
```
Robbepop: You can use the new event syntax. The remaining issue with it has been fixed.

examples/lang/erc20/src/lib.rs
```diff
@@ -86,162 +82,81 @@ contract! {
             total_supply
         }
 
-        /// Returns the balance of the given address.
+        /// Returns the balance of the given AccountId.
         pub(external) fn balance_of(&self, owner: AccountId) -> Balance {
-            let balance = *self.balances.get(&owner).unwrap_or(&0);
+            let balance = self.balance_of_or_zero(&owner);
             env.println(&format!("Erc20::balance_of(owner = {:?}) = {:?}", owner, balance));
```
Robbepop: Deploy lost its `println` so either we keep all or we remove all.
What suits you better.

examples/lang/erc20/src/lib.rs
```diff
@@ -86,162 +82,81 @@ contract! {
...
-            }
-            let allowance = self.allowance_or_zero(from, to);
-            if allowance < by {
-                return false
-            }
-            self.allowances.insert((*from, *to), allowance - by);
-            true
+        /// Returns the balance of the AccountId or 0 if there is no balance.
+        fn balance_of_or_zero(&self, of: &AccountId) -> Balance {
+            let balance = self.balances.get(of).unwrap_or(&0);
```
Robbepop: This can be reduce to a one-liner.

examples/lang/erc20/src/lib.rs
```diff
@@ -86,162 +82,81 @@ contract! {
...
 
         /// Returns the allowance or 0 of there is no allowance.
-        fn allowance_or_zero(&self, from: &AccountId, to: &AccountId) -> Balance {
-            let allowance = self.allowances.get(&(*from, *to)).unwrap_or(&0);
-            env::println(&format!(
-                "Erc20::allowance_or_zero(from = {:?}, to = {:?}) = {:?}",
-                from, to, allowance
-            ));
+        fn allowance_or_zero(&self, owner: &AccountId, spender: &AccountId) -> Balance {
+            let allowance = self.allowances.get(&(*owner, *spender)).unwrap_or(&0);
```
Robbepop: Can be reduced to a one-liner.

examples/lang/erc20/src/lib.rs
shawntabrizi: What does this mean? Sorry, I do not understand.

Robbepop: You removed the `env.println` call in `deploy` but not anywhere else. So what should we do here? Leave them? Remove all? I wouldn't keep a partial subset.

shawntabrizi: I made it so that println is only used when calling getter functions. In this situation it is no so much a "debugging tool" as it is the only way to really interact with your contract meaningfully.

For all the other functions, I believe an event is emitted when that stuff happens, so it does not need a println

examples/lang/erc20/src/lib.rs
shawntabrizi: Done :) (but not so pretty)

Issue #62, opened 2019-04-18
Design & implement user friendly events


Robbepop:
Currently we only support depositing events through `ext_deposit_event` that takes a generic byte blob. We want to improve this situation in the regard that a user is able to specify events in their contract and deposit those in a user friendly (smart contract developer friendly) fashion.

# Road Map

- [x] Define AST for `event Foo(bar: Bar);` syntax
- [x] Implement parser for `event Foo(bar: Bar);` syntax
- [x] Implement some semantic checks for event definitions
- [x] Implement code gen for event definitions and the infrastructure around them
- [x] Implement tests to verify that things roughly work

# Example Design

```rust
contract! {
    /// Documentation for balanced changed event.
    event BalanceChanged { from: Address, to: Address, amount: Balance }
    ...
    impl MyContract {
        fn transfer_balance(to: Address, amount: Balance) {
            let from = env.caller();
            ... // perform some checks etc.
            env.emit(BalanceChanged(from, to, amount));
        }
    }
}
```

# Codegen

The code generation routine would simply collect all event definitions and put them in an `Event` enum.
Upon depositing an event the code generation would simply construct the associated `Event` variant and encode it with the given values into `ext_deposit_event`.

For the above example the generated code could look like the following:

```rust
mod events {
    use super::*;

    #[derive(parity_codec::Encode, parity_codec::Decode)]
    enum Event {
        BalanceChanged(BalanceChanged),
    }

    impl From<BalanceChanged> for Event {
        fn from(event: BalanceChanged) -> Self {
            Event::BalanceChanged(event)
        }
    }

    /// Documentation for balanced changed event.
    #[derive(parity_codec::Encode, parity_codec::Decode)]
    pub struct BalanceChanged {
        pub from: Address,
        pub to: Address,
        pub amount: Balance,
    };

    /// Guard to protect the emit event trait to be implemented externally.
    trait Sealed;

    pub trait EmitEventExt: Sealed {
        fn emit<E>(&self, event: E)
        where
            E: Into<Event>,
        {
            ext::deposit_raw_event(
                event.into().encode().as_slice()
            )
        }

        /// Do not try to implement this trait externally!
        fn sealed(sealed: Sealed);
    }

    impl EmitEventExt for pdsl_model::EnvHandler {
        fn sealed(sealed: Sealed) {}
    }
}

use events::{BalanceChanged, EmitEventExt as _};

contract! {
    ...
    impl MyContract {
[...]

PR #78, opened 2019-04-23
[lang] Implement event syntax

Robbepop:
Implements https://github.com/paritytech/ink/issues/62/.

### DIFF
--- a/lang/src/ast.rs
+++ b/lang/src/ast.rs
@@ -84,7 +84,7 @@ pub enum Item {
 ///
 /// # Example
 ///
-/// This mirrors the syntax for: `event Foo(bar: Bar);`
+/// This mirrors the syntax for: `event Foo { bar: Bar };`
 #[derive(Debug, PartialEq, Eq, Hash)]
 pub struct ItemEvent {
     pub attrs: Vec<syn::Attribute>,
--- a/lang/src/ast.rs
+++ b/lang/src/ast.rs
@@ -80,7 +80,7 @@ pub enum Item {
     Event(ItemEvent),
 }
 
-/// An event declartion.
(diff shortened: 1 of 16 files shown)


lang/src/ast.rs
```diff
@@ -61,13 +61,44 @@ impl Contract {
...
 
 #[derive(Debug)]
 pub enum Item {
     State(ItemState),
     DeployImpl(ItemDeployImpl),
     Impl(ItemImpl),
+    Event(ItemEvent),
+}
+
+/// An event declartion.
```
pepyakin:
```suggestion
/// An event declaration.
```

lang/src/ast.rs
```diff
@@ -61,13 +61,44 @@ impl Contract {
...
     DeployImpl(ItemDeployImpl),
     Impl(ItemImpl),
+    Event(ItemEvent),
+}
+
+/// An event declaration.
+///
+/// # Example
+///
+/// This mirrors the syntax for: `event Foo(bar: Bar);`
```
ascjones:
```suggestion
/// This mirrors the syntax for: `event Foo { bar: Bar };`
```

lang/src/hir.rs
```diff
@@ -39,9 +43,63 @@ pub struct Contract {
...
 
 impl Contract {
+    /// Extracts all events from the contract.
+    ///
+    /// Performs some semantic checks on them as a whole.
+    ///
+    /// # Errors
+    ///
+    /// - If there are multiple events with the same names.
+    /// - If an event has
```
ascjones:
```suggestion
    /// - If an event has the same name as the contract
```

PR #86, opened 2019-05-06
[examples] Update to show `env.emit` inside non-message function

shawntabrizi:
Modified the example ever so slightly so that we can show how to do the `env` magic inside non-message functions.

### DIFF
--- a/examples/lang/events/src/lib.rs
+++ b/examples/lang/events/src/lib.rs
@@ -52,6 +52,12 @@ contract! {
         ///
         /// Also emits an event.
         pub(external) fn dec(&mut self) {
+            self.dec_internal(env);
+        }
+    }
+
+    impl CallCounter {
+        fn dec_internal(&mut self, env: &mut ink_model::EnvHandler) {
             self.count -= 1;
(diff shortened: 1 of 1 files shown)


Robbepop:
I think we do not need to put this extra in the contract.
I want this example to be minimalistic in showing event syntax.
We already have Erc20 that makes use of the work around.
We might want another example that shows explicit work around usage.
My optimum is having many but small examples where each show only a few new things.

shawntabrizi:
> See comment pls.

I don't disagree with your base goal, but I do think that this is within scope of teaching one new thing.

In this example we teach out to:
- emit an event within a message function
- emit an event within a non-message function

both common scenarios, but about emitting events.

That being said, certainly up to you :)

Robbepop:
> > See comment pls.
> 
> I don't disagree with your base goal, but I do think that this is within scope of teaching one new thing.
> 
> In this example we teach out to:
> 
>     * emit an event within a message function
> 
>     * emit an event within a non-message function
> 
> 
> both common scenarios, but about emitting events.
> 
> That being said, certainly up to you :)

Damn good point!

PR #87, opened 2019-05-06
Add ext_address and ext_balance calls

holygits:
Closes #58 and #59

PR #89, opened 2019-05-07
[core] [model] Impl ext_now

### LINKED ISSUE #84: Implement `ext_now`
Add the ability to get the current timestamp from `lang`.

Not sure if there's any current blockers to this as I'm still understanding the code base.

lsaether:
fixes #84 

Should be updated before merge but after merging #87 as these PRs touch the same files.

Tester contract will compile but it won't deploy to a Substrate dev chain at 1.0.0rc-1 throwing me the error here https://github.com/paritytech/substrate/blob/2899be9eef8821eb1b601812c038c69d7a177f93/srml/contract/src/wasm/prepare.rs#L266


core/src/env/srml/srml_only/impls.rs
```diff
@@ -145,6 +147,21 @@ where
             .expect("we can expect to receive a correctly sized buffer here")
     }
 
+    fn now() -> <Self as EnvTypes>::Moment {
```
Robbepop: Please change this to also use the new https://github.com/paritytech/ink/blob/master/core/src/env/srml/srml_only/impls.rs#L97 functionality.

Robbepop:
> fixes #84
> 
> Should be updated before merge but after merging #87 as these PRs touch the same files.
> 
> Tester contract will compile but it won't deploy to a Substrate dev chain at 1.0.0rc-1 throwing me the error here https://github.com/paritytech/substrate/blob/2899be9eef8821eb1b601812c038c69d7a177f93/srml/contract/src/wasm/prepare.rs#L266

I can see nothing in this PR that would cause contracts to be no longer deployable to Substrate. Have to investigate. Does it only happen with this PR? Have you used the accompanied `build.sh` script to build the contract with all the other tools?

Issue #79, opened 2019-04-24
Permission denied for ./build.sh for newly created contract


pepyakin:
After I created a new smart-contract with `cargo contract new` I tried to run `./build.sh` but it said `Permission denied`.

Maybe it is a good idea to `chmod +x` it as a part of the setup (or get rid of it completely).

Robbepop:
I couldn't found decent cross-platform std support for `chmod +x`. Is there some?

xlc:
Instead of create the file from rust code, how about clone an existing repo / unzip an archive and modify the output. Then git / zip should handle the permission flag.

shawntabrizi:
> Instead of create the file from rust code, how about clone an existing repo / unzip an archive and modify the output. Then git / zip should handle the permission flag.

Totally agree here. That was the general idea behind https://github.com/paritytech/ink/issues/69

Robbepop:
The big downside to this (imo) is that you then need a stable internet connection just to create a new project. To me this is a no-go honestly. But maybe that's just because I live in a developing country when it comes to internet access (germany).

xlc:
you can bundle a zip file in the rust executable which solves internet connection issue.
this also make possible to use 3rd parity provided templates.

Robbepop:
Yeah we could do that. But even though zip is ubiquitous we should really try to use as few dependencies for simple setup (and other) purposes.

xlc:
I am sure there are rust zip library available that can be static linked to avoid require user to install zip as dependency.

Robbepop:
I would totally accept such a solution.

ascjones:
I did a quick PoC in https://github.com/paritytech/ink/pull/90. It uses [zip-rs](https://github.com/mvdnes/zip-rs). However it will only set file permissions for `#[cfg(unix)]`, which I suppose is okay in this specific case of the `build.sh` file (i.e. that won't be executed on Windows).

ascjones:
Should be solved by #90

PR #90, opened 2019-05-07
[cli] contract new template

ascjones:
Replaces inline code templates with a actual files, which are embedded in the binary as a zip file.

**Overview**
- Template files exist in `./template` directory
- Build script zips them up using [zip-rs](https://github.com/mvdnes/zip-rs)
- cli unzips embedded template and sets permissions as per metadata (rel #79)
  - Sets default unix permissions of all files to `755` = `rwx`

**Advantages over current approach:**
- Handles script file permissions (only works for `#[cfg(unix)]` currently)
- Template files can be viewed and edited in place (with syntax highlighting etc.) rather than editing inline strings.
- CI could perform checks on template code, would prevent it getting out of date.


ascjones:
Requires testing on mac. @shawntabrizi?

shawntabrizi:
Can confirm working on mac. Sorry for delays

PR #94, opened 2019-05-10
Add shared_vec example

Robbepop:
It demonstrates usage of events, Option and Result in message interfaces.

### DIFF
--- a/examples/lang/shared_vec/src/lib.rs
+++ b/examples/lang/shared_vec/src/lib.rs
@@ -85,11 +85,11 @@ contract! {
     /// Fires whenever a new mutator is registered
     /// or when a mutators access rights are changed.
     event Register {
-        // / The mutator.
+        /// The mutator.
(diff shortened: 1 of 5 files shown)


examples/lang/shared_vec/src/lib.rs
```diff
@@ -0,0 +1,252 @@
...
+        mutator: AccountId,
+        // / The begin access index.
+        begin: Option<u32>,
+        // / The end access index.
+        end: Option<u32>,
+    }
+
+    /// Fires whenever a mutator pushes the vector successfully.
+    event Push {
+        // / The pushed value.
```
shawntabrizi: It makes sense to me to include the `mutator: AccountId` with these events... not for a practical reason, but more like an "imaginary UI/UX" reason

Robbepop: good idea, will do that!

PR #107, opened 2019-05-30
Contract Runtime Types

### LINKED ISSUE #98: Contract Runtime Types
This issue is to discuss and track how to approach replacing local `ink` implementations of srml runtime types (`AccountId`, `Balance`, `Hash`) with the actual types from the chain runtime.

### Background

The [mirroring of these types](https://github.com/paritytech/ink/blob/110d83a37416eaf86c1b66e35b580f6df04de601/core/src/env/srml/types.rs#L58) works for when the target chain's runtime types are compatible, which cannot be guaranteed.

Further, in order to implement https://github.com/paritytech/ink/issues/42 we need to introduce the `Call` type which would be more difficult to mirror, and would again require the runtime to be compatible.

Originally, the idea was to replace the mirrored types by importing `node_runtime` from substrate (see https://github.com/paritytech/ink/pull/72). However this still runs into the problem of being tightly coupled to the types there.

### Goal

Contract authors should be able to use the types defined in their own custom runtime.

Therefore we must not have a direct dependency on a specific type implementation in `ink_core`.

### Ideas

Sergei suggested the idea of creating some prepackaged libs:

> you need to remember that srml-contract is not the final product, it is just a module which will be used to assemble a final product
but there is another an important idea
ink! is also like that
so ink is not the final library that users will use
I'd expect that there will be libraries like:

>node_runtime_ink
xlc_runtime_ink
edgeware_ink
etc
and each of these libraries would depend on two libs:

>ink
the corresponding runtime
13:10
it would combine these two to instantiate ink library with concrete types

In order to achieve this we need to make those types generic in the ink `core` and `model` layers, and having the `lang` layer provide the type aliases for the concrete definitions. Here is a prototype of how that might look:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c8d2b9dd5ea8bbd0f3f317d65e60e580

I'm currently working on a PR that will attempt to incorporate that or similar into `ink`

Please contribute ideas and feedback here until I get the new PR up.

ascjones:
Resolves #98.

This PR enables contract authors to provide their own set of Env types compatible with their target chain contract runtime. e.g.
```
use ink_types_node_runtime::NodeRuntimeTypes;

contract! {
    type EnvTypes = NodeRuntimeTypes;
```
I've added a library `ink_types_node_runtime` which pulls in the types from substrate `node-runtime`. This library could potentially be moved out of this repo, removing the dependency on substrate from the build here.

**TODO**
- [ ] Fix lang tests
- [ ] Test examples still work on chain
- [ ] Move/rename `ink_types_node_runtime`?
- [ ] PR on substrate for `node_runtime` types from https://github.com/paritytech/substrate/compare/aj-ink-types
- [ ] Docs


ascjones:
Replaced by Draft PR #108

PR #108, opened 2019-05-30
Contract Runtime Types

### LINKED ISSUE #98: Contract Runtime Types

ascjones:
Resolves #98.

This PR enables contract authors to provide their own set of Env types compatible with their target chain contract runtime. 

Users can use set of predefined types e.g. https://github.com/paritytech/ink-types-node-runtime. 

```
use ink_types_node_runtime::NodeRuntimeTypes;

contract! {
    #![env = NodeRuntimeTypes]
```

Or define their own which are compatible with their runtime types, implementing the `EnvTypes` trait.

```
pub enum CustomRuntimeTypes {}

impl ink_core::env::EnvTypes for CustomRuntimeTypes {
    type AccountId = [u8; 32];
    type Balance = u128;
    type Hash = [u8; 32];
    type Moment = u64;
}

contract! {
    #![env = CustomRuntimeTypes]
```

~The precise syntax is still undecided, awaiting feedback from others, see https://github.com/paritytech/ink/pull/108#discussion_r290277529.~

Changed to `#![env = ink_core::env::DefaultSrmlTypes]`


examples/lang/erc20/src/lib.rs
```diff
@@ -75,16 +73,16 @@ contract! {
         /// Returns the balance of the given AccountId.
         pub(external) fn balance_of(&self, owner: AccountId) -> Balance {
             let balance = self.balance_of_or_zero(&owner);
-            env.println(&format!("Erc20::balance_of(owner = {:?}) = {:?}", owner, balance));
+            env.println(&format!("Erc20::balance_of(owner = {:?}) = {:?}", owner.0, balance));
```
Robbepop: I would rather introduce some getters for the underlying data (e.g. `as_bytes(&self) -> [&u8]` or something like that) instead of using `.0` syntax anywhere in a non-generic library environment. But that's a subjective meaning. Feel free to discuss.

examples/lang/erc20/src/lib.rs
```diff
@@ -130,7 +128,7 @@ contract! {
         }
 
         /// Transfers token from a specified AccountId to another AccountId.
-        fn transfer_impl(&mut self, env: &mut ink_model::EnvHandler, from: AccountId, to: AccountId, value: Balance) -> bool {
```
Robbepop: Why remove this path qualifier? I currently fail to see a proper import of `ink_model` to justify it.

lang/src/gen/build.rs
```diff
@@ -48,6 +49,21 @@ pub fn generate_code(tokens: &mut TokenStream2, contract: &hir::Contract) {
     codegen_for_event_mod(tokens, contract);
 }
 
+fn codegen_for_env_types(tokens: &mut TokenStream2, contract: &hir::Contract) {
+    let env_types = &contract.env_types_type;
+    tokens.extend(quote! {
+        use ink_core::env::{self, EnvTypes, SrmlEnv};
+
+        type Env = SrmlEnv<#env_types>;
+        type EnvHandler = ink_model::EnvHandler<Env>;
```
Robbepop: Why are you explicitely aliasing for these types?
Is it for the user or the later code gen?

model/src/contract.rs
```diff
@@ -29,6 +29,7 @@ use crate::{
 };
 use core::marker::PhantomData;
 use ink_core::memory::vec::Vec;
+use ink_core::env;
 // Copyright {d+}-{d+} Parity Technologies (UK) Ltd.
```
Robbepop: We should move this line back to the license header.

model/src/exec_env.rs
```diff
@@ -92,59 +93,61 @@ impl<State> ExecutionEnv<State> {
...
 
     /// Returns from the current smart contract execution with the given value.
-    pub unsafe fn r#return<T>(&self, val: T) -> !
+    pub unsafe fn r#return<V>(&self, val: V) -> !
     where
-        T: parity_codec::Encode,
+        V: parity_codec::Encode,
     {
-        env::r#return(val)
+        T::r#return(&val.encode()[..])
```
Robbepop: Can't we go for `T::r#return(val)` here, too?

types/node_runtime/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+categories = ["no-std", "embedded"]
+
+include = ["/Cargo.toml", "src/**/*.rs", "/README.md", "/LICENSE"]
+
+[dependencies]
+ink_core = { path = "../../core" }
+node-runtime = { git = "https://github.com/paritytech/substrate/", branch = "aj-ink-types", package = "node-runtime", default-features = false }
+srml-contract = { git = "https://github.com/paritytech/substrate/", branch = "aj-ink-types", package = "srml-contract", default-features = false }
+# ignore duplicate wasm lang items
+sr-std = { git = "https://github.com/paritytech/substrate/", branch = "aj-ink-types", package = "sr-std", default-features = false, features = ["no_global_allocator"] }
```
Robbepop: How about using `sr-std`'s `#[global_allocator]` and `sr-io`'s lang items?

examples/lang/erc20/src/lib.rs
Robbepop: Okay I see that you are generating type aliases in `ink_lang`. Looks slightly intransparent to me but I don't feel like I currently understand enough to judge this decision.

Robbepop:
Can you adjust the `lang` examples to make use of this?
Also can we still easily test our contracts using the test environment?

examples/lang/erc20/src/lib.rs
ascjones: The AccountId type from `node_runtime` resolves to `sr25519::Public` didn't implement `Debug for `no_std`. I've just added that on my `aj-ink-types` branch: https://github.com/paritytech/substrate/blob/8cf36aff21f6ca85beedc88e6234076f32a50970/core/primitives/src/sr25519.rs#L50

examples/lang/erc20/src/lib.rs
ascjones: Being completely transparent would mean `ink_model::EnvHandler<SrmlEnv<NodeRuntimeTypes>>` here. A tradeoff between transparency and usability?

I suppose though that normally the user shouldn't be seeing `EnvHandler` anyway.

lang/src/gen/build.rs
ascjones: Both aliases used later in codegen. 

https://github.com/paritytech/ink/blob/33616695e6340e4579b939dbb2dfab298fb969d9/lang/src/gen/build.rs#L185

https://github.com/paritytech/ink/blob/33616695e6340e4579b939dbb2dfab298fb969d9/lang/src/gen/build.rs#L364

Alternatively we could generate the full type in each location with `SrmlEnv<#contract.env_types_type>`.

Though as observed in the previous comment the alias will surface to the user if they require the `EnvHandler` in a helper method.

model/src/exec_env.rs
ascjones: It's inlining the impl which used to be in the `env::r#return` function which has been removed: https://github.com/paritytech/ink/pull/108/files#diff-aaf76b409bdca20872661cfad19d09c1L97

types/node_runtime/Cargo.toml
ascjones: The lib is intended just to bring in the env types.

Not sure about this, can discuss further.

model/src/exec_env.rs
Robbepop: So it is no longer possible to call this in `core`? So users need `ink_model` to actually make use of this abstraction layer that abstracts between the real environment and the test environment or is it different now due to the refactoring?

lang/src/gen/build.rs
Robbepop: Okay I see. Maybe a good usability decision although in macro code I tend to always write out all names to make the generated code more robust in case of user inserted aliases or binding. Not sure if this code can be controlled in that way though.

examples/lang/erc20/src/lib.rs
Robbepop: > I suppose though that normally the user shouldn't be seeing EnvHandler anyway.

Yeah I would hope so but atm we cannot prevent this in all cases unfortunately.

lang/src/gen/build.rs
ascjones: Replace type alias with explicit types in https://github.com/paritytech/ink/pull/108/commits/48732edf535f27825aed62d31a1d0eec4eafc998

examples/lang/erc20/src/lib.rs
ascjones: Type alias removed for transparency: https://github.com/paritytech/ink/pull/108/commits/48732edf535f27825aed62d31a1d0eec4eafc998

model/src/exec_env.rs
ascjones: You're right the test-env abstraction is broken now. I'm currently looking into this.

core/src/env/test_env.rs
```diff
@@ -30,12 +29,13 @@ use core::cell::{
...
 };
-use std::convert::TryFrom;
+use parity_codec::{Decode, Encode};
+use std::marker::PhantomData;
 
 /// A wrapper for the generic bytearray used for data in contract events.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct EventData {
-    topics: Vec<env::Hash>,
+    topics: Vec<Vec<u8>>,
```
Robbepop: Couldn't we use `Hash` again if we made `EventData` generic as well?

core/src/env/test_env.rs
```diff
@@ -121,19 +121,19 @@ pub struct TestEnvData {
     /// # Note
     ///
     /// The current address can be adjusted by `TestEnvData::set_address`.
-    address: srml::AccountId,
+    address: Vec<u8>,
```
Robbepop: Same here.

core/src/env/test_env.rs
```diff
@@ -471,52 +455,38 @@ impl TestEnv {
...
-        TEST_ENV_DATA.with(|test_env| test_env.borrow().load(key))
-    }
-}
-
 macro_rules! impl_env_getters_for_test_env {
     ( $( ($fn_name:ident, $ret_name:ty) ),* ) => {
         $(
             fn $fn_name() -> $ret_name {
-                TEST_ENV_DATA.with(|test_env| test_env.borrow().$fn_name())
+                TEST_ENV_DATA.with(|test_env| Decode::decode(&mut &test_env.borrow().$fn_name()[..]).unwrap())
```
Robbepop: Please replace `unwrap` with `expect` here.

core/src/env/traits.rs
```diff
@@ -18,12 +18,13 @@ use crate::{
...
     storage::Key,
 };
+use core::hash::Hash;
 use parity_codec::Codec;
 
 /// The environmental types usable by contracts defined with pDSL.
 pub trait EnvTypes {
     /// The type of an address.
-    type AccountId: Codec + PartialEq + Eq;
+    type AccountId: Codec + PartialEq + Eq + Hash + Copy;
```
Robbepop: Why does it need to be `Hash` and `Copy`? We need to take care to not be more restrictive than Substrate.

examples/lang/flipper/src/lib.rs
```diff
@@ -17,13 +17,15 @@
...
 use ink_core::{
-    env::println,
+    env::DefaultSrmlTypes,
     memory::format,
     storage,
 };
 use ink_lang::contract;
 
 contract! {
+    type EnvTypes = DefaultSrmlTypes;
```
Robbepop: Maybe we want to use another syntax here that is less confused with an actual type alias which this isn't behind the scenes.
This is more of a meta information to my understanding.
So maybe we could make use of some attribute-like syntax here.
```rust
#![env = DefaultSrmlTypes]
```
Or something like that.

types/node_runtime/Cargo.toml
Robbepop: Yeah you are right. Let's keep a single responsibility here.

core/src/env/test_env.rs
Robbepop: NVM - saw that there is a translation layer already

core/src/env/traits.rs
ascjones: I agree, these are requirements only for using `AccountId` with a `HashMap` (e.g. ERC20 sample), which is not a strict requirement. Will remove

examples/lang/erc20/src/lib.rs
```diff
@@ -152,12 +152,23 @@ contract! {
 mod tests {
     use super::*;
     use ink_core::env;
-    use std::convert::TryFrom;
+    use parity_codec::Decode;
+
+    /// Create a new AccountId with the given bytes
+    ///
+    /// Required because AccountId is a type alias and cannot be used as a constructor
+    fn new_account_id(data: [u8; 32]) -> AccountId {
```
Robbepop: Maybe in this case trait bounds `TryFrom<&[u8]>` or `From<[u8;32]>` would be useful for `AccountId` and maybe some other types.

Robbepop: We should really not force people to write functions like these for their simple test codes.

examples/lang/erc20/src/lib.rs
```diff
@@ -152,12 +152,23 @@ contract! {
...
+        Decode::decode(&mut data.as_ref()).expect("Valid AccountId bytes")
+    }
+
+    fn set_caller(caller: AccountId) {
```
Robbepop: Maybe we can replace this by another abstraction layer around the test environment.
For example we could introduce a `ConcreteTestEnv` in the following way:

```rust
struct ConcreteTestEnv<T>
where
    T: EnvTypes,
{
    pub fn set_caller(caller: AccountId) {
        env::test::set_caller::<T>(caller)
    }
}
```
And then `ink_lang` could generate something like:
```rust
type TestEnv = ConcreteTestEnv<WhateverTheUserWantedAsTheirEnv>;
```
So they could then write
```rust
TestEnv::set_caller([0x0; 32]);
```
Instead of having to write their own wrappers around all calls.
This encapsulates the environment information into this wrapper type to that a user doesn't have to annotate all test environment calls with the same `EnvTypes` type.

Robbepop: Of course: Feel free to come up with something better!

Robbepop: I would prefer a solution were we could still write `env::test::set_caller` instead of `TestEnv::set_caller`.

model/src/exec_env.rs
ascjones: Sorted

examples/lang/erc20/src/lib.rs
ascjones: Let me know what you think of the solution in https://github.com/paritytech/ink/pull/108/commits/5c2158c33cf54a4e2ec3f8119f6d0d15340bb240

lang/src/gen/build.rs
```diff
@@ -70,7 +70,7 @@ fn codegen_for_contract_env(tokens: &mut TokenStream2, contract: &hir::Contract)
             Moment,
         };
 
-        #[allow(snake_case)] type env = types::ContractEnv;
+        #[allow(snake_case)] type env = types::_ContractEnv;
```
Robbepop: Why?

ascjones: This was a temporary change to fix an issue with conflicting names, shouldn't have been pushed

examples/lang/flipper/src/lib.rs
ascjones: Awaiting outcome of poll on channel

ascjones:
> Can you adjust the lang examples to make use of this?

Done

> Also can we still easily test our contracts using the test environment with the lang layer?

Done

> Can you elaborate what a user of ink! that wants to adept it to their chain types (like Edgeware) needs to adjust in order to make it happen with the changes introduced in this PR?

See main PR comment

core/src/env/mod.rs
```diff
@@ -40,17 +40,29 @@ mod test_env;
...
 pub use traits::*;
 
-/// The environment implementation that is currently being used.
+pub use self::srml::DefaultSrmlTypes;
+
+/// The storage environment implementation that is currently being used.
 ///
 /// This may be either
-/// - `DefaultEnv` for real contract storage
+/// - `SrmlEnv` for real contract storage
```
Robbepop: Please update the docs.

core/src/env/mod.rs
```diff
@@ -40,17 +40,29 @@ mod test_env;
...
 ///   manipulation that may happen on-chain.
 /// - `TestEnv` for emulating a contract environment
 ///   that can be inspected by the user and used
 ///   for testing contracts off-chain.
 #[cfg(not(feature = "test-env"))]
-pub(self) type ContractEnv = self::srml::DefaultSrmlEnv;
+pub type ContractEnvStorage = self::srml::SrmlEnvStorage;
```
Robbepop: Is this correct being `pub` while the other definitions are `pub(self)` ?

core/src/env/test_env.rs
```diff
@@ -17,8 +17,7 @@
 use super::*;
 use crate::{
     env::{
-        self,
-        srml,
+        EnvTypes,
```
Robbepop: Is this already passing the formatting tests?

core/src/env/test_env.rs
```diff
@@ -394,24 +394,26 @@ thread_local! {
 }
 
 /// Test environment for testing SRML contract off-chain.
-pub struct TestEnv;
+pub struct TestEnv<T> {
+    marker: PhantomData<T>,
```
Robbepop: I always prefer setting this to `marker: PhantomData<fn () -> T>` so that the outer type depends less on `T` and make it `Send` and `Sync`.

core/src/env/test_env.rs
```diff
@@ -471,52 +454,39 @@ impl TestEnv {
...
-    }
-}
-
 macro_rules! impl_env_getters_for_test_env {
     ( $( ($fn_name:ident, $ret_name:ty) ),* ) => {
         $(
             fn $fn_name() -> $ret_name {
-                TEST_ENV_DATA.with(|test_env| test_env.borrow().$fn_name())
+                TEST_ENV_DATA.with(|test_env| Decode::decode(&mut &test_env.borrow().$fn_name()[..])
+                    .expect("Encoded environment types are valid"))
```
Robbepop:
```suggestion
                    .expect("environment instances are assumed to be correctly encoded"))
```

core/src/env/traits.rs
```diff
@@ -20,16 +20,30 @@ use crate::{
 };
 use parity_codec::Codec;
 
-/// The environmental types usable by contracts defined with pDSL.
+#[cfg(not(feature = "test-env"))]
+/// The environmental types usable by contracts defined with ink!.
 pub trait EnvTypes {
     /// The type of an address.
-    type AccountId: Codec + PartialEq + Eq;
+    type AccountId: Codec + Clone + PartialEq + Eq;
```
Robbepop: Could you also add a conditional trait bound to `Debug` if `std` is enabled?
I think this could be great for testability using `println` as we sometimes have to.
Or am I misunderstanding that we don't need to add this trait bound for practical uses since the concrete types are already known upon usage?

examples/lang/erc20/Cargo.toml
```diff
@@ -5,9 +5,13 @@ authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
 [dependencies]
-ink_core = { path = "../../../core" }
-ink_model = { path = "../../../model" }
-ink_lang = { path = "../../../lang" }
+# TODO [AJ]: remove specific branches before merging
```
Robbepop: TODO!
Please resolve all TODOs and FIXMEs etc. and tell me when you are done with that.

examples/lang/erc20/src/lib.rs
```diff
@@ -17,17 +17,16 @@
...
-    },
     memory::format,
     storage,
 };
+use ink_model::EnvHandler;
 use ink_lang::contract;
+use ink_types_node_runtime::NodeRuntimeTypes;
 
 contract! {
+    #![env = NodeRuntimeTypes]
```
Robbepop: Just to clarify: We could also write `#![env = ink_types_node_runtime::NodeRuntimeTypes]`, right?

examples/lang/erc20/src/lib.rs
```diff
@@ -68,14 +67,14 @@ contract! {
         /// Returns the total number of tokens in existence.
         pub(external) fn total_supply(&self) -> Balance {
             let total_supply = *self.total_supply;
-            env.println(&format!("Erc20::total_supply = {:?}", total_supply));
+            env::println(&format!("Erc20::total_supply = {:?}", total_supply));
```
Robbepop: Are we able to still use `env.println` ? How about other `env.*` syntax elements?
If this patch introduces a shift here we need to make sure how to handle future cases.

examples/lang/erc20/src/lib.rs
```diff
@@ -130,7 +129,7 @@ contract! {
         }
 
         /// Transfers token from a specified AccountId to another AccountId.
-        fn transfer_impl(&mut self, env: &mut ink_model::EnvHandler, from: AccountId, to: AccountId, value: Balance) -> bool {
+        fn transfer_impl(&mut self, env: &mut EnvHandler<ink_core::env::ContractEnv<NodeRuntimeTypes>>, from: AccountId, to: AccountId, value: Balance) -> bool {
```
Robbepop: Hmm this is really ugly but also a great motivation to hide this ugliness through the macros.

Robbepop: ... in a later PR

examples/lang/erc20/src/lib.rs
```diff
@@ -168,10 +166,10 @@ mod tests {
 
     #[test]
     fn transfer_works() {
-        let alice = AccountId::try_from([0x0; 32]).unwrap();
-        let bob = AccountId::try_from([0x1; 32]).unwrap();
+        let alice: AccountId = From::from([0x0; 32]);
```
Robbepop: Wouldn't it be more readable to just write: `let alice = AccountId::from([0x0; 32]);`
Open to opinions.

examples/lang/erc20/src/lib.rs
```diff
@@ -168,10 +166,10 @@ mod tests {
...
+        let bob: AccountId = From::from([0x1; 32]);
 
-        env::test::set_caller(alice);
+        env::set_caller(alice);
```
Robbepop: So we kinda lost the `test` sub module here.
It arguably isn't that important but what are the chances we can reintroduce it because I kind of like the clear separation between environment and test-environmental functions.

core/src/env/test_env.rs
```diff
@@ -428,38 +430,19 @@ impl TestEnv {
...
+    impl_env_setters_for_test_env!(
+        (set_address, address, T::AccountId),
+        (set_balance, balance, T::Balance),
+        (set_caller, caller, T::AccountId),
+        (set_random_seed, random_seed, T::Hash),
+        (set_now, now, T::Moment)
+    );
 
-    /// Returns an iterator over all emitted events.
     pub fn emitted_events() -> impl IntoIterator<Item = Vec<u8>> {
```
Robbepop: Why not return here the iterator directly?

Robbepop: As far as I know/do, `IntoIterator` is better used for inputs, rather than outputs.

examples/lang/erc20/src/lib.rs
```diff
@@ -218,20 +216,20 @@ mod tests {
...
-        let alice = AccountId::try_from([0x0; 32]).unwrap();
-        let bob = AccountId::try_from([0x1; 32]).unwrap();
+        let alice: AccountId = From::from([0x0; 32]);
+        let bob: AccountId = From::from([0x1; 32]);
 
         // No events to start
-        env::test::set_caller(alice);
-        assert_eq!(env::test::emitted_events().count(), 0);
+        env::set_caller(alice);
+        assert_eq!(env::emitted_events().into_iter().count(), 0);
```
Robbepop: imo `emitted_events` should continue to return an iterator

examples/model/erc20/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2018"
 [dependencies]
 ink_core = { path = "../../../core" }
 ink_model = { path = "../../../model" }
+ink_types_node_runtime = { git = "https://github.com/paritytech/ink-types-node-runtime" }
```
Robbepop: We maybe later want to upload this to crates.io.

examples/model/erc20/src/lib.rs
```diff
@@ -29,6 +29,10 @@ use ink_model::{
     Contract,
     ContractDecl,
 };
+use ink_types_node_runtime::NodeRuntimeTypes;
```
Robbepop: Have you checked if also the model and core examples are still working? I guess core example might no longer work, but that is not because of your PR ...

lang/src/gen/build.rs
```diff
@@ -48,6 +49,31 @@ pub fn generate_code(tokens: &mut TokenStream2, contract: &hir::Contract) {
...
+
+        use types::{
+            AccountId,
+            Balance,
+            Hash,
+            Moment,
+        };
+
+        use ink_core::env::Env as _;
+        #[allow(snake_case)] type env = ink_core::env::ContractEnv<#env_types>;
```
Robbepop: Let's hope that this dirty trick is not too dirty. :D

Robbepop: so we could in theory use `ink_model::ExecHandler<env>` ?
Not that we should!

lang/src/lib.rs
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with ink!.  If not, see <http://www.gnu.org/licenses/>.
 
-#![recursion_limit = "256"]
+#![recursion_limit = "512"]
```
Robbepop: Oh my ... 0.o why was that?

Robbepop: I really hope that dtolnay releases a proc-macro version of `quote` soon.

lang/src/parser.rs
```diff
@@ -59,28 +59,32 @@ impl ast::Item {
 
 impl Parse for ast::Item {
     fn parse(input: ParseStream<'_>) -> Result<Self> {
-        let attrs = syn::Attribute::parse_outer(input)?;
+        let inner_attrs: ast::ItemEnvMeta = input.parse()?;
```
Robbepop: Can't we make use of https://oschwald.github.io/maxminddb-rust/syn/struct.Attribute.html#method.parse_inner ?
`syn::Attribute::parse_inner(input)` would be the perfect fit for us and then we would need to dispatch those attributes into our own. This could also be used later for proper versioning of smart contracts which has already been planned.

Note that if this is too much work for you, then let's leave this to another PR.

Robbepop: Just saw your implementation already goes into a decent direction. So no change needed here.

lang/src/tests/mod.rs
```diff
@@ -240,6 +299,7 @@ fn multiple_states() {
 fn multiple_deploy_handlers() {
     assert_failure(
         quote! {
+            #![env = ink_core::env::DefaultSrmlTypes]
```
Robbepop: Would it maybe be a good idea to make `ink_core::env::DefaultSrmlTypes` be a default to not break too many contracts? Maybe this is a bad idea. Let's discuss this in an issue.

core/src/env/mod.rs
ascjones: You're right it should also be `pub (self)`

core/src/env/traits.rs
ascjones: Yeah we don't need the trait bound at this level, so long as the concrete type is implementing `Debug` it will work.

Note that below is a definition for the `test-env` feature with a `Debug` bound, which is required for test asserts where we don't have the concrete type.

examples/lang/erc20/Cargo.toml
ascjones: Yes will do that now, since we are close to merging.

examples/lang/erc20/src/lib.rs
ascjones: Correct, that part is parsed as a `syn::Type`. e.g. https://github.com/paritytech/ink/pull/108/files#diff-18215f9b77a1fc56d4103de0bf553be0R11

examples/lang/erc20/src/lib.rs
ascjones: Yes you can still do both `env.println` and `env::println`. In this case I was testing that `env::println` still works.

examples/lang/erc20/src/lib.rs
ascjones: Yes definitely! Remember earlier I had a type alias to avoid this but we decided to be explicit here for clarity.

examples/lang/erc20/src/lib.rs
ascjones: Yes I agree. This is a relic of an earlier iteration

lang/src/gen/build.rs
ascjones: I'm not 100% sold on this. I propose we stick with this for this PR. I've opened an issue to address this: #125

examples/model/erc20/src/lib.rs
ascjones: Well, they compile and the tests run - didn't test them on chain...

lang/src/lib.rs
ascjones: Indeed! Can't remember what it was...in any case I've reverted it and it seems to work okay now

examples/lang/erc20/src/lib.rs
ascjones: See also: https://github.com/paritytech/ink/issues/125

examples/lang/erc20/src/lib.rs
ascjones: Yes this is bypassing the `env::test::emitted_events` function. Need to figure this out. #125.

lang/src/tests/mod.rs
ascjones: https://github.com/paritytech/ink/issues/127

core/src/env/test_env.rs
ascjones: Fixed

examples/model/erc20/src/lib.rs
ascjones: Actually turns out this was no longer working, fixed in latest commits

lang/src/gen/build.rs
ascjones: Removed

lang/src/lib.rs
ascjones: Reverted

core/src/env/srml/srml_only/impls.rs
```diff
@@ -77,33 +67,37 @@ where
...
-        if size > 0 {
-            value.resize(size as usize, 0);
-            unsafe {
-                sys::ext_scratch_copy(value.as_mut_ptr() as u32, 0, size);
-            }
-        }
-        value
-    }
+/// The SRML contracts environment.
+#[allow(unused)]
```
Robbepop: Please do a `#[cfg_attr(...)]` here in order to do this conditionally.

examples/lang/erc20/Cargo.toml
```diff
@@ -5,9 +5,12 @@ authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2018"
 
 [dependencies]
-ink_core = { path = "../../../core" }
-ink_model = { path = "../../../model" }
-ink_lang = { path = "../../../lang" }
+ink_core = { git = "https://github.com/paritytech/ink/", package = "ink_core" }
```
Robbepop: We should not depend on `node-runtime-types` via ink! which enforces these GitHub dependencies and instead copy those important examples over to that repo. Generally we want to avoid GitHub dependencies to crates within our own repo.

examples/lang/erc20/src/lib.rs
Robbepop: Clarity definitely wins over ugliness.

core/src/env/srml/srml_only/impls.rs
ascjones: This was redundant so I've removed it

Issue #37, opened 2019-03-20
Fix DynAlloc


Robbepop:
The "new" `DynAlloc` storage allocator is broken.
Whenever using it (e.g. `allocate_using`) the code panics (traps).
The reason for this has yet to be found.

We need to fix it in order to finally deprecate `CellChunkAlloc`.

PR #111, opened 2019-05-31
Attempt to replicate dyn_alloc trapping on deployment

cmichi:
Do not merge!

This is my attempt at replicating the bug from #37. I didn't manage to replicate it, though according to private conversations this should have been sufficient. What I did:

* Modify `flipper` to repeatedly dynamically allocate entries in the vector and deallocate them again (by popping). In total 20.100 items are pushed into the vecs before deallocation begins.
* Modify substrate: deactivate check for `block production took too long`, increase endowment in genesis block hugely, decrease gas/rent/etc. price.
* Deploy contract, create instance and call flipper with generous endowment (`9999999`).

The transaction goes through successfully and I see this output from substrate:
```
...
Runtime: Outer Vec len after pushing: 100
Runtime: Outer Vec len after popping: 0
...
```

### DIFF
--- a/examples/lang/flipper/src/lib.rs
+++ b/examples/lang/flipper/src/lib.rs
@@ -37,6 +37,9 @@ contract! {
     /// alter between `true` and `false` using the `flip` message.
     /// Users can retrieve its current state using the `get` message.
     struct Flipper {
+        alloc: storage::alloc::DynAlloc,
+        // alloc: storage::alloc::BumpAlloc,
+
         /// The current state of our flag.
         value: storage::Value<bool>,
         outer_vec: storage::Vec<storage::Vec<i32>>,
@@ -45,6 +48,7 @@ contract! {
     impl Deploy for Flipper {
         /// Initializes our state to `false` upon deploying our smart contract.
         fn deploy(&mut self) {
+            self.alloc.initialize(());
             self.value.set(false)
         }
     }
@@ -57,19 +61,9 @@ contract! {
 
         /// Push some dynamically allocated entries.
         pub(external) fn push_some(&mut self) {
-            let mut alloc = unsafe {
-                let mut fw_alloc = storage::alloc::BumpAlloc::from_raw_parts(Key([0x0; 32]));
-                let mut dyn_alloc = storage::alloc::DynAlloc::allocate_using(&mut fw_alloc);
-                dyn_alloc.initialize(());
-                dyn_alloc
-            };
-
-            // Uncomment to test with BumpAlloc instead
-            // let mut alloc = unsafe { BumpAlloc::from_raw_parts(Key([0x0; 32])) };
-
-            for i in 0..1000 {
+            for i in 0..2000 {
                 let mut inner_vec = unsafe {
-                    Vec::<i32>::allocate_using(&mut alloc).initialize_into(())
+                    Vec::<i32>::allocate_using(&mut self.alloc).initialize_into(())
                 };
 
                 println(&format!("Inner Vec about to push 3 elts: {:?}", i));
@@ -84,7 +78,7 @@ contract! {
 
         /// Pop some dynamically allocated entries.
         pub(external) fn pop_some(&mut self) {
-            for a in 0..200 {
+            for a in 0..500 {
                 println(&format!("Outer Vec about to pop: {:?}", a));
                 self.outer_vec.pop();
             }
--- a/examples/lang/flipper/src/lib.rs
+++ b/examples/lang/flipper/src/lib.rs
@@ -60,12 +60,12 @@ contract! {
                 dyn_alloc
             };
 
-            for _ in 0..100 {
+            for _ in 0..3010 {
                 let mut inner_vec = unsafe {
                     Vec::<i32>::allocate_using(&mut alloc).initialize_into(())
                 };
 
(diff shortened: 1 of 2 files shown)


examples/lang/flipper/src/lib.rs
```diff
@@ -42,6 +52,32 @@ contract! {
...
         /// Flips the current state of our smart contract.
         pub(external) fn flip(&mut self) {
+            println(&format!("flip"));
+            let mut alloc = unsafe {
+                let mut fw_alloc = storage::alloc::BumpAlloc::from_raw_parts(Key([0x0; 32]));
+                let mut dyn_alloc = storage::alloc::DynAlloc::allocate_using(&mut fw_alloc);
+                dyn_alloc.initialize(());
+                dyn_alloc
+            };
+            let mut alloc = unsafe { BumpAlloc::from_raw_parts(Key([0x0; 32])) };
```
Robbepop: Why are you using the `BumpAlloc` here instead of the `DynAlloc` in order to allocate the `storage::Vec` ?

examples/lang/flipper/src/lib.rs
```diff
@@ -42,6 +52,32 @@ contract! {
...
+
+            for _ in 0..100 {
+                let mut inner_vec = unsafe {
+                    Vec::<i32>::allocate_using(&mut alloc).initialize_into(())
+                };
+
+                for _ in 0..201 {
```
Robbepop: You can reduce this to 3 since it is not what we actually want to test.

examples/lang/flipper/src/lib.rs
```diff
@@ -42,6 +52,32 @@ contract! {
...
+                let mut fw_alloc = storage::alloc::BumpAlloc::from_raw_parts(Key([0x0; 32]));
+                let mut dyn_alloc = storage::alloc::DynAlloc::allocate_using(&mut fw_alloc);
+                dyn_alloc.initialize(());
+                dyn_alloc
+            };
+            let mut alloc = unsafe { BumpAlloc::from_raw_parts(Key([0x0; 32])) };
+
+            for _ in 0..100 {
+                let mut inner_vec = unsafe {
+                    Vec::<i32>::allocate_using(&mut alloc).initialize_into(())
```
Robbepop: This line (if it used `DynAlloc` instead of `BumpAlloc`) is what we are interested in testing >3000 times before deallocating.

examples/lang/flipper/src/lib.rs
cmichi: Accidental commit, was used for debugging.

cmichi:
It works neither for `DynAlloc` nor if using `BumpAlloc` instead. For `DynAlloc` 2307 items can be pushed to the outer vec, for `BumpAlloc` 2314.

I added the testing code (with your comments addressed) as a test to ink and it runs fine (not pushed yet).

So I suspect that it might have something to do with some maximum limit being reached for the wasmi instance in substrate. I have increased every somehow related limit I found (maximum pages, etc.), but haven't had luck so far ‒ the numbers stay the same. The trap is always an unreachable trap in the wasmi instance. My next plans are to bisect the exact place/method where the trap occurs. Do you know by chance if there is something beyond println statements for that (e.g. a stack trace)?

Robbepop:
> It works neither for `DynAlloc` nor if using `BumpAlloc` instead. For `DynAlloc` 2307 items can be pushed to the outer vec, for `BumpAlloc` 2314.
> 
> I added the testing code (with your comments addressed) as a test to ink and it runs fine (not pushed yet).
> 
> So I suspect that it might have something to do with some maximum limit being reached for the wasmi instance in substrate. I have increased every somehow related limit I found (maximum pages, etc.), but haven't had luck so far ‒ the numbers stay the same. The trap is always an unreachable trap in the wasmi instance. My next plans are to bisect the exact place/method where the trap occurs. Do you know by chance if there is something beyond println statements for that (e.g. a stack trace)?

Interesting!
For testing gas usage we could print out the remaining gas which is just another `ext_` call that should already be supported in ink! `lang` layer.
However, I currently do not know of any better way to debug this for limits as we only have the `ext_*` calls available in smart contracts and they do not cover getters for all the limits there are.
Maybe we could introduce some new `ext_*` getters just for the purpose of debugging this particular thing.
It is also interesting that even the super simple `BumpAlloc` for nearly the same amount of allocations fails which also really sounds to me that some limit has been reached.

cmichi:
Good suggestions!

Here's what I get after your proposed changes:

* Pushing 5k items into outer vec (3 items in inner vec) in steps of 1k works for both `BumpAlloc` and `DynAlloc`.
* Attempting to pop in steps of 200 works as well.

Increasing these numbers (popping in larger steps or pushing in larger steps) results in an error during runtime execution at some point (but not because out of gas). It fails at slightly different numbers for both allocators.

Yeah so from my perspective everything seems like the allocators work fine, but it's rather an issue with limitations of the wasmi instance. Question is how we handle this? Do you have an idea for some minimum requirements that we want to fulfill?

Robbepop:
With https://github.com/paritytech/ink/pull/136 being merged we could give this another try and finally deprecate `CellChunkAlloc` if this is found to be working. :)

Robbepop:
I ll close this and you are free to open it again if you want to continue working on this.

Issue #113, opened 2019-06-06
Implementation of HashMap initialization when using eDSL


usd-yamazaki:
If you store a structure in storage::HashMap, you can not use storage::HashMap as a child structure.
The following code will actually cause an error because storage::HashMap::new() does not exist.

```
#![cfg_attr(not(any(test, feature = "test-env")), no_std)]

use ink_core::{
    env::{
        self,
        AccountId,
        Balance,
        Hash,
    },
    storage,
};
use ink_lang::contract;
use parity_codec::{Encode, Decode};

contract! {
    struct Test {
        users: storage::HashMap<AccountId, User>,
    }

    impl Deploy for Test {
        fn deploy(&mut self) {
        }
    }

    impl Test {
        pub(external) fn add(&mut self) {
            let user = User {
                balance: 0,
                items: storage::HashMap::new(),
            };
            self.users.insert(env::caller(), user);
        }
    }
}

#[derive(Encode, Decode)]
struct User {
    balance: Balance,
    items: storage::HashMap<Hash, u128>,
}
```

Robbepop:
What you are trying to achieve is dynamic allocation of storage types.
This is possible in ink!, however requires a different initialization scheme than you are used normally in Rust.

So for example this should work:

```rust
pub(external) fn add(&mut self) {
    let user = User {
        balance: 0,
        items: storage::HashMap::allocate_using(&mut env.dyn_alloc).initialize_into(()),
    };
    self.users.insert(env::caller(), user);
}
```

To be honest this is currently quite some ugly syntax for such a simple thing.
However, we already have some plans on how to make this simpler in the future.
Also please beware, that just yesterday we found a seriour bug in dynamicly allocated instances of storage types that we have to fix - but we are already on it. (@cmichi )

usd-yamazaki:
I did not notice the existence of env.dyn_alloc.
Thank you!

PR #115, opened 2019-06-07
Fix flush

cmichi:
This implements recursive behavior for `flush()`, which is an issue discovered while investigating #37 and also brought up in #113.

I'm a bit unsure for which primitive types we should provide implementations of `flush()`. AFAIK `u8` substrate avoids usage of `u8`, `usize`, `isize` and floating points, due to different behaviors of different architectures.

I would have liked to add better tests (also asserting that the content is correct and not just the number of writes), but it seems that `test_env` is currently a bit limited. The other tests also don't do any "deep inspection". After trying it for a bit I settled on the easier tests for now, but maybe we can keep this in mind for the future (or create a follow up).

### DIFF
--- a/core/src/storage/chunk/sync_chunk/tests.rs
+++ b/core/src/storage/chunk/sync_chunk/tests.rs
@@ -22,14 +22,31 @@ use crate::{
         alloc::{
             AllocateUsing,
             BumpAlloc,
+            DynAlloc,
+            Initialize,
         },
         Flush,
         Key,
     },
     test_utils::run_test,
 };
+use crate::storage;
 
 fn dummy_chunk() -> SyncChunk<u32> {
+    unsafe {
+        let mut alloc = BumpAlloc::from_raw_parts(Key([0x0; 32]));
+        SyncChunk::allocate_using(&mut alloc)
+    }
+}
+
+fn dummy_vec_chunk() -> SyncChunk<storage::Vec<u32>> {
+    unsafe {
+        let mut alloc = BumpAlloc::from_raw_parts(Key([0x0; 32]));
+        SyncChunk::allocate_using(&mut alloc)
+    }
+}
+
+fn dummy_recursive_vec_chunk() -> SyncChunk<storage::Vec<storage::Vec<u32>>> {
     unsafe {
         let mut alloc = BumpAlloc::from_raw_parts(Key([0x0; 32]));
         SyncChunk::allocate_using(&mut alloc)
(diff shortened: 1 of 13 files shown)


core/src/storage/chunk/sync_chunk/tests.rs
```diff
@@ -22,12 +22,15 @@ use crate::{
         alloc::{
             AllocateUsing,
             BumpAlloc,
+            DynAlloc,
```
Robbepop: Are we already using `DynAlloc`?

core/src/storage/collections/stash/impls.rs
```diff
@@ -222,6 +230,14 @@ enum Entry<T> {
     Occupied(T),
 }
 
+impl<T> Flush for Entry<T> where Self: Encode {
+    fn flush_at(&mut self, at: Key) {
+        unsafe {
+            crate::env::store(at, &self.encode()[..]);
```
Robbepop: Maybe we find a way to reduce the code dupe here.

core/src/storage/collections/vec/impls.rs
```diff
@@ -76,12 +78,23 @@ impl<'a, T> Iter<'a, T> {
...
+	Self: Encode,
+	T: parity_codec::Encode,
+	T: Flush,
 {
     fn flush(&mut self) {
         self.len.flush();
         self.cells.flush();
     }
+
+    fn flush_at(&mut self, at: Key) {
```
Robbepop: Under my impression of how this works we only need an impl of `Flush::flush_at` for `SyncChunk` and all fundamental and primitive types, not for types like `storage::Vec` etc since they already simply propagate their flush to their members.

Robbepop: Correct me if I am wrong.

core/src/storage/flush.rs
```diff
@@ -31,7 +31,83 @@
...
+    }
+
+    /// Default implementation which forwards to flush.
+    /// This realizes recursive behavior for e.g. nested vectors.
+    fn flush_at(&mut self, _at: Key) {
+        self.flush();
+    }
+}
+
+impl Flush for i8 where Self: Encode {
```
Robbepop: Why do we need the `Self: Encode` btw?

Robbepop: Also I think you can get rid of lots of code dupe with a decent `macro_rules!` macro. ;)

Robbepop:
I think we can close this in favor of https://github.com/paritytech/ink/pull/136. @cmichi

PR #117, opened 2019-06-11
[chores] Rename pDSL to ink at CONTRIBUTING.md

jckdotim:
This is a simple renaming patch from pDSL to ink (see also #64)


Robbepop:
Thank you!

PR #121, opened 2019-06-13
[lang] Use FnDecl::inputs_with_env. Fixes warnings.

### LINKED ISSUE #119: Unused `ast::FnDecl::inputs_with_env` method
The build warns about the following method being unused: 
https://github.com/paritytech/ink/blob/debb49aa3e7cc111818f913c9c8273686483761b/lang/src/ast.rs#L266-L286

However it looks similar to the following code, which it might have been intended to replace:
https://github.com/paritytech/ink/blob/d803c2e2cb558fd79f1d4bee8cd5a3adb3f59977/lang/src/gen/build.rs#L338-L357

So we should either use the method in place of the above code, or remove it.

ascjones:
Resolves #119


Robbepop:
Just to be sure: Have you checked if examples still work?

ascjones:
> Have you checked if examples still work?

Yes, although I did forget at first.

> Also we should add this to our CI.

Agreed. I'll create an issue.

PR #124, opened 2019-06-18
Implement `ext_dispatch_call`

### LINKED ISSUE #42: Implement infrastructure for smart contract to call back into runtime

ascjones:
Resolves #42. 

- Adds `dispatch_call` function 
- Introduces a new `EnvType` `Call` which must implement `Encode`
- ~Introduces pre-baked `Call` types for srml modules~. *edit*: this part has been moved to https://github.com/paritytech/ink-types-node-runtime/pull/3

Here is an example of dispatching a `Balances::transfer` from a contract:

```
pub(external) fn balance_transfer(&mut self, dest: u32, value: Balance) {
    let transfer_call = calls::Balances::<DefaultSrmlTypes>::transfer(dest, value);
    env.dispatch_call(transfer_call);
}
```
The pre-baked `Call` types are "copies" of the srml module `Call` enums e.g. Balances. Importing the actual types from substrate is possible but it results in a very large wasm blob size.  To ensure compatibility with the equivalent substrate types, tests are required for each `Call` which encode the local type and decode as the substrate type. 

**Note:** This PR only includes a single predefined `Call` type for `Balances`. I think we should consider which additional calls/modules we should add in a follow up PR.

The final piece is the definition of the top level `Call` enum to compose the modules to match the substrate runtime. This can be user defined for custom runtimes by defining the `Call` type in `EnvTypes` e.g.

```
#[derive(Encode, Decode)]
#[cfg_attr(feature = "test-env", derive(Debug, Clone, PartialEq, Eq))]
pub enum Call {
    #[codec(index = "5")]
    Balances(super::calls::Balances<DefaultSrmlTypes>),
}
```

Because of how `parity-scale-codec` encodes enums, chain authors can define only the srml module calls they want to support, so long as the `index` matches up with the the relevant `Call` variant in the target chain's substrate runtime declaration.


lang/src/gen/build.rs
```diff
@@ -57,6 +57,7 @@ fn codegen_for_contract_env(tokens: &mut TokenStream2, contract: &hir::Contract)
             use ink_core::env::{ContractEnv, EnvTypes};
 
             pub type AccountId = <ContractEnv<#env_types> as EnvTypes>::AccountId;
+            pub type AccountIndex = <ContractEnv<#env_types> as EnvTypes>::AccountIndex;
```
Robbepop: What is the difference to AccountId?

ascjones: A lookup of `AccountIndex` -> `AccountId`. e.g. `node-runtime` has a lookup `u32` -> `Public` to resolve an address from an `Index`. See https://github.com/paritytech/substrate/blob/master/srml/balances/src/lib.rs#L417

Robbepop: Ah got it, that's what an `Address` can also refer to instead of an `AccountId`.

.travis.yml
```diff
@@ -33,6 +33,8 @@ install:
   # Print rustc and cargo versions
   - rustc -vV
   - cargo -vV
+  # Install wasm-gc for building substrate libs
+  - cargo install --git https://github.com/alexcrichton/wasm-gc --force
```
Robbepop: I am not experienced with `wasm-gc` etc. but have you considered this?: https://github.com/alexcrichton/wasm-gc#wasm-gc

cli/template/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2018"
 ink_core = { git = "https://github.com/paritytech/ink", package = "ink_core" }
 ink_model = { git = "https://github.com/paritytech/ink", package = "ink_model" }
 ink_lang = { git = "https://github.com/paritytech/ink", package = "ink_lang" }
-parity-codec = { version = "3.3", default-features = false, features = ["derive"] }
+parity-codec = { version = "4.1.1", default-features = false, features = ["derive"] }
```
Robbepop: In general I would simply put `4.1` here. Also this has already been done in the latest set of changes to `master`.

core/Cargo.toml
```diff
@@ -20,7 +20,11 @@ include = ["/Cargo.toml", "src/**/*.rs", "/README.md", "/LICENSE"]
 [dependencies]
 ink_alloc = { path = "../alloc/" }
 ink_utils = { path = "../utils/" }
-parity-codec = { version = "3.2", default-features = false, features = ["derive", "full"] }
+parity-codec = { version = "4.1.1", default-features = false, features = ["derive", "full"] }
+
+[dev-dependencies]
+node-runtime = { git = "https://github.com/paritytech/substrate/", branch="aj-wasm-builder", package = "node-runtime", features = ["std"] }
```
Robbepop: So here we are again ... depending on `node-runtime`. I guess it is only for tests, right? So does that mean that `ink_core` tests now compile forever? If so I would really appreciate this behind a crate feature and as an optional `dev-dependency` to be able to quickly run simple tests that do not rely on the `ext_dispatch_call` functionality.

core/src/env/api.rs
```diff
@@ -64,8 +68,19 @@ pub unsafe fn load(key: Key) -> Option<Vec<u8>> {
...
+    T: Encode,
     E: Env,
 {
     E::r#return(&value.encode()[..])
 }
+
+/// Dispatches a Call into the runtime, for invoking other substrate
+/// modules. Dispatched only after successful contract execution.
+///
+/// The encoded Call MUST be decodable by the target substrate runtime.
```
Robbepop: What happens if it isn't? Can we actually crash the runtime with our smart contracts?

core/src/env/srml/calls.rs
```diff
@@ -0,0 +1,156 @@
...
+use crate::env::EnvTypes;
+use parity_codec::{
+    Decode,
+    Encode,
+    Input,
+    Output,
+};
+use core::convert::TryInto;
+
+#[cfg_attr(feature = "test-env", derive(Debug, Clone, PartialEq, Eq))]
```
Robbepop: Is `test-env` correct here? As already said I think this could even life under its very own crate feature to make the `node-runtime` and `srml_indices` dev-dependencies optional upon testing. Maybe something like `test-dispatch-call` crate feature - use your creativity. Also please tell me what you think about this.

core/src/env/srml/calls.rs
```diff
@@ -0,0 +1,156 @@
...
+pub enum Address<T: EnvTypes> {
+    Id(T::AccountId),
+    Index(T::AccountIndex),
+}
+
```
Robbepop:
```suggestion

/// Returns `b` if `b` is greater than `a` and otherwise `None`.
```

core/src/env/srml/calls.rs
```diff
@@ -0,0 +1,156 @@
...
+fn need_more_than<T: PartialOrd>(a: T, b: T) -> Option<T> {
+    if a < b { Some(b) } else { None }
+}
+
+/// Decode impl copied from
+/// https://github.com/paritytech/substrate/blob/ec62d24c602912f07bbc416711376d9b8e5782c5/srml/indices/src/address.rs#L61
```
Robbepop: Please make use of markdown and properly name the link, a la:

```markdown
Decode implementation copied over from Substrate `Address` that can be found [here](substrate-address).

[substrate-address]: https://github.com/paritytech/substrate/blob/ec62d24c602912f07bbc416711376d9b8e5782c5/srml/indices/src/address.rs#L61
```

Also please notify readers that this implementation must be kept in sync for it to work.
This is a very fragile part of the code base.

core/src/env/srml/calls.rs
```diff
@@ -0,0 +1,156 @@
...
+                need_more_than(0xffffffffu32.into(), Decode::decode(input)?)?
+            ),
+            0xff => Address::Id(Decode::decode(input)?),
+            _ => return None,
+        })
+    }
+}
+
+/// Encode impl copied from
+/// https://github.com/paritytech/substrate/blob/ec62d24c602912f07bbc416711376d9b8e5782c5/srml/indices/src/address.rs#L83
```
Robbepop: Same suggestion as above.

core/src/env/srml/calls.rs
```diff
@@ -0,0 +1,156 @@
...
+            },
+        }
+    }
+}
+
+#[derive(Encode)]
+#[cfg_attr(feature = "test-env", derive(Decode, Debug, Clone, PartialEq, Eq))]
+pub enum Balances<T: EnvTypes> {
+    #[allow(non_camel_case_types)]
+    transfer(Address<T>, #[codec(compact)] T::Balance),
```
Robbepop: Why `non_camel_case_types` ? Please provide rationals here, otherwise I would rather not do this.

core/src/env/srml/srml_only/impls.rs
```diff
@@ -86,10 +86,12 @@ where
     T: EnvTypes,
 {
     type AccountId = <T as EnvTypes>::AccountId;
+    type AccountIndex = <T as EnvTypes>::AccountIndex;
```
Robbepop: Btw.: Why is `AccountId` no longer enough? As far as I know SRML contracts doesn't even really support `AccountIndex` since it would require a 2-phase runtime call in order to be used compared to a `AccountId`.

core/src/env/srml/srml_only/sys.rs
```diff
@@ -72,6 +72,14 @@ extern "C" {
         value_len: u32,
     );
 
+    /// Dispatches a Call into the runtime, for invocation of substrate modules
+    ///
+    /// The data is encoded as T::Call
```
Robbepop: What is `T` here? We probably need other docs for this.

core/src/env/srml/types.rs
```diff
@@ -57,7 +63,7 @@ impl<'a> TryFrom<&'a [u8]> for AccountId {
 }
 
 /// The default SRML balance type.
-pub type Balance = u64;
+pub type Balance = u128;
```
Robbepop: Is this the correct SRML `Balance` type? My latest knowledge was that it is still `u64`. I might be wrong though.

core/src/env/srml/types.rs
```diff
@@ -83,3 +89,55 @@ pub type Moment = u64;
 
 /// The default SRML blocknumber type.
 pub type BlockNumber = u64;
+
+/// The default SRML call type.
+#[derive(Encode)]
+#[cfg_attr(feature = "test-env", derive(Decode, Debug, Clone, PartialEq, Eq))]
+pub enum Call {
+    #[codec(index = "5")]
+    Balances(super::calls::Balances<DefaultSrmlTypes>),
```
Robbepop: It only features `Balances` atm? Is everything else hard coded later? Are custom runtimes supposed to also configure ink! in a way to extend the set of supported dispatch calls here?

core/src/env/srml/types.rs
```diff
@@ -83,3 +89,55 @@ pub type Moment = u64;
...
+}
+
+impl From<super::calls::Balances<DefaultSrmlTypes>> for Call {
```
Robbepop: I think it is simpler to use the `derive_more` crate in such cases.

core/src/env/traits.rs
```diff
@@ -48,8 +55,13 @@ pub trait EnvTypes {
     type Moment: Codec + Clone + PartialEq + Eq + core::fmt::Debug;
     /// The type of block number.
     type BlockNumber: Codec + Clone + PartialEq + Eq + core::fmt::Debug;
+    /// The type of a call into the runtime
+    type Call: Codec + Clone + PartialEq + Eq + core::fmt::Debug;
```
Robbepop: Why does it need `Decode` in `test-env` ? Please add rational.

examples/lang/calls/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+ink_core = { path = "../../../core" }
+ink_model = { path = "../../../model" }
+ink_lang = { path = "../../../lang" }
+parity-codec = { version = "4.1.1", default-features = false, features = ["derive"] }
+
+[lib]
+name = "calls"
+crate-type = ["cdylib"]
+
+[features]
```
Robbepop: Needs to be adjusted to the new `std` default crate feature compatibility.

examples/lang/calls/build.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -e
+
+PROJNAME=calls
+
+# cargo clean
+# rm Cargo.lock
+
+CARGO_INCREMENTAL=0 &&
+cargo build --release --features generate-api-description --target=wasm32-unknown-unknown --verbose
```
Robbepop: Also need adjustments for `std` default feature.

examples/lang/calls/rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2019-05-21
```
Robbepop: We are about to drop `rust-toolchain` files so I think you can also just remove it for this PR.

examples/lang/calls/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+    struct Calls {}
+
+    impl Deploy for Calls {
+        fn deploy(&mut self) {
+        }
+    }
+
+    impl Calls {
+        /// Dispatches a `transfer` call to the Balances srml module
+        pub(external) fn balance_transfer(&mut self, dest: AccountId, value: Balance) {
```
Robbepop:
```suggestion
        pub(external) fn transfer_balance(&mut self, dest: AccountId, value: Balance) {
```

examples/lang/calls/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn dispatches_balances_call() {
+        let mut calls = Calls::deploy_mock();
+        assert_eq!(env::dispatched_calls().into_iter().count(), 0);
+        calls.balance_transfer(1, 10000);
```
Robbepop:
```suggestion
        calls.transfer_balance(1, 10000);
```

examples/lang/calls/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+        assert_eq!(env::dispatched_calls().into_iter().count(), 1);
```
Robbepop: I wish we had better introspection for such cases ... but good to have a test example for this functionality!

model/src/exec_env.rs
```diff
@@ -180,4 +181,8 @@ impl<T: Env> EnvHandler<T> {
     pub fn block_number(&self) -> T::BlockNumber {
         T::block_number()
     }
+
+    pub fn dispatch_call(&self, call: T::Call) {
```
Robbepop:
```suggestion
    pub fn dispatch_call<C>(&self, call: C)
    where
        C: Into<T::Call>,
    {
```

model/src/exec_env.rs
```diff
@@ -180,4 +181,8 @@ impl<T: Env> EnvHandler<T> {
...
+        T::dispatch_raw_call(call.encode().as_slice())
```
Robbepop:
```suggestion
        T::dispatch_raw_call(call.into().encode().as_slice())
```

Robbepop: Also please adjust all uses of `env.dispatch_call` by removing the superflous `.into()` in its params.

core/src/env/srml/types.rs
```diff
@@ -34,17 +34,15 @@ pub enum DefaultSrmlTypes {}
...
         _unconstructable: ()
     }
-    impl Encode for Call {}
+    impl parity_codec::Encode for Call {}
 
     #[cfg(feature = "std")]
-    impl Decode for Call {
-        fn decode<I: Input>(_value: &mut I) -> Option<Self> {
+    impl parity_codec::Decode for Call {
+        fn decode<I: parity_codec::Input>(_value: &mut I) -> Option<Self> {
```
Robbepop: Why do we impl `Decode` for `Call` if we never decode it?

lang/src/tests/noop.rs
```diff
@@ -43,7 +43,6 @@ fn contract_compiles() {
                 use ink_core::env::{ContractEnv, EnvTypes};
 
                 pub type AccountId = <ContractEnv<DefaultSrmlTypes> as EnvTypes>::AccountId;
-                pub type AccountIndex = <ContractEnv<DefaultSrmlTypes> as EnvTypes>::AccountIndex;
```
Robbepop: I guess removed in favor of just using `AccountId` ?

.travis.yml
ascjones: I'm aware of that, but this was a prerequisite explicitly checked by the substrate `wasm-builder`. In any case the `substrate` deps are removed again (for now).

core/Cargo.toml
ascjones: Now moved over to https://github.com/paritytech/ink-types-node-runtime/pull/3. But yes a good idea to have a feature to avoid compiling whole of substrate for these tests each time, and just have it run on CI.

core/src/env/api.rs
ascjones: No just the smart contract execution will fail:

https://github.com/paritytech/substrate/blob/00053422ea9dbdbe7932038c6b58a8c41861ff50/srml/contracts/src/wasm/runtime.rs#L574
https://github.com/paritytech/substrate/blob/5ae532d4f016cfe8874b2e9f76a771b2a5a969dd/srml/contracts/src/exec.rs#L441

core/src/env/srml/calls.rs
ascjones: Agree, this has now moved to https://github.com/paritytech/ink-types-node-runtime/pull/3. Over there the codec tests are the only ones so we need to build the substrate deps everytime anyway.

core/src/env/srml/calls.rs
ascjones: Suggestion implemented in https://github.com/paritytech/ink-types-node-runtime/pull/3

core/src/env/srml/calls.rs
ascjones: Suggestion implemented in https://github.com/paritytech/ink-types-node-runtime/pull/3

core/src/env/srml/calls.rs
ascjones: Suggestion implemented in https://github.com/paritytech/ink-types-node-runtime/pull/3

core/src/env/srml/calls.rs
ascjones: Consistency with the enum type generated by `decl_module!`. So it looks the same as using the `srml-balances` `Call` type directly.

But the name of the variant is not required to be the same (just the position), so I could change it if you prefer.

core/src/env/srml/srml_only/impls.rs
ascjones: I added this because the `srml-balances` module (and other srml modules) expect the `Address` enum as an argument (either Id or Index).

You are correct, and I have removed `AccountIndex` from `EnvTypes`. However as part of https://github.com/paritytech/ink-types-node-runtime/pull/3, you are still able to construct an `Address::Index` to send as part of a `Call` into the runtime.

lang/src/tests/noop.rs
ascjones: AccountIndex now gone

core/src/env/srml/types.rs
ascjones: It's "used" in the test env so is required as a trait bound on `EnvTypes` https://github.com/paritytech/ink/blob/8c6b85235131d262e3a016842d361d6c009f96a8/core/src/env/test_env.rs#L503.

Suppose we could add the `Decode` bound to a generic type parameter `Call` to just this method - what do you think?

examples/lang/calls/src/lib.rs
ascjones: Moved to https://github.com/paritytech/ink-types-node-runtime/pull/3, not implemented though because we agreed in chat that this name was okay for the `Call`

core/src/env/api.rs
Robbepop: But please write that down as comment since it is rather important and non trivial to see.

examples/lang/calls/src/lib.rs
ascjones: Same as above

examples/lang/calls/src/lib.rs
ascjones: Moved to https://github.com/paritytech/ink-types-node-runtime/pull/3. But actually those calls are decoded so you can inspect them - just was too lazy here.

core/src/env/api.rs
ascjones: Comment added see :point_down:

model/src/exec_env.rs
ascjones: Implemented

core/src/env/api.rs
Robbepop: Perfection

model/src/exec_env.rs
ascjones: Done in https://github.com/paritytech/ink-types-node-runtime/pull/3

cli/template/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2018"
 ink_core = { git = "https://github.com/paritytech/ink", package = "ink_core", default-features = false }
 ink_model = { git = "https://github.com/paritytech/ink", package = "ink_model", default-features = false }
 ink_lang = { git = "https://github.com/paritytech/ink", package = "ink_lang", default-features = false }
-parity-codec = { version = "3.3", default-features = false, features = ["derive"] }
+parity-codec = { version = "4.1", default-features = false, features = ["derive"] }
```
Robbepop: What a fail, this should have been done in an earlier PR already. Good you noticed it!

core/src/env/api.rs
```diff
@@ -64,8 +68,20 @@ pub unsafe fn load(key: Key) -> Option<Vec<u8>> {
...
+/// If decoding fails, then the smart contract execution will fail.
+pub fn dispatch_call<T>(call: <T as EnvTypes>::Call)
```
Robbepop: Since this is an interfacing method and `Call` is generally an `enum`, can't we here as well use an `Into<<T as EnvTypes>::Call>` generic input instead?

core/src/env/srml/types.rs
```diff
@@ -31,14 +31,34 @@ use parity_codec::{
 
 /// The SRML fundamental types.
 #[allow(unused)]
+#[cfg_attr(feature = "test-env", derive(Debug, Clone, PartialEq, Eq))]
```
Robbepop: Why only implement those trait when `test-env` feature is enabled? Please add rational.

core/src/env/srml/types.rs
```diff
@@ -31,14 +31,34 @@ use parity_codec::{
...
 pub enum DefaultSrmlTypes {}
 
+/// Private module for default Call type, so it cannot be constructed
+/// For calling into the runtime, a user defined Call type required.
+/// See https://github.com/paritytech/ink-types-node-runtime.
+mod private {
+    #[cfg_attr(feature = "std", derive(Debug, Clone, PartialEq, Eq))]
+    pub struct Call {
```
Robbepop: Can't we just make it `pub enum Call { }` ? Can't be instantiated as well. Or does it cause problems?

core/src/env/srml/types.rs
Robbepop: Then better put it behind `test-env` feature instead of `std` feature and also provide a rational (comment) why it is there.

core/src/env/srml/types.rs
Robbepop: Answer: Yes it is. Resolved.

core/src/env/test_env.rs
```diff
@@ -303,6 +312,10 @@ impl TestEnvData {
             .iter()
             .map(|event_data| event_data.data_as_bytes())
     }
+
```
Robbepop: Missing doc comment!

Robbepop: Also as in `emitted_events` we could provide a `DoubleEndedIterator` here.

core/src/env/test_env.rs
```diff
@@ -477,6 +494,16 @@ where
                 .into_iter()
         })
     }
+
```
Robbepop: Missing doc comment.

core/src/env/traits.rs
```diff
@@ -48,8 +53,13 @@ pub trait EnvTypes {
...
 }
 
+/// Substrate runtime dispatchable call
+pub trait Call<T: EnvTypes>: Encode {}
```
Robbepop: What is this used for? Please add this rational to dev-docs.

core/src/env/traits.rs
```diff
@@ -127,4 +137,7 @@ pub trait Env: EnvTypes {
 
     /// Deposits raw event data through Contracts module.
     fn deposit_raw_event(topics: &[<Self as EnvTypes>::Hash], data: &[u8]);
+
+    /// Dispatches a call into the Runtime
```
Robbepop:
```suggestion
    /// Dispatches a call into the Runtime.
```

examples/lang/calls/Cargo.toml
Robbepop: Still needs adjustment

examples/lang/calls/rust-toolchain
Robbepop: Please delete `rust-toolchain` file.

model/src/exec_env.rs
```diff
@@ -180,4 +181,11 @@ impl<T: Env> EnvHandler<T> {
     pub fn block_number(&self) -> T::BlockNumber {
         T::block_number()
     }
+
```
Robbepop: Missing doc comment.
Would you also be so kind to add a doc-comment to `block_number` above?

examples/lang/calls/Cargo.toml
ascjones: Moved: see https://github.com/paritytech/ink-types-node-runtime/pull/3/files#diff-6f7c540321faf2e34153128ae83caadeR19

core/src/env/traits.rs
ascjones: Removed

core/src/env/srml/types.rs
```diff
@@ -31,14 +31,34 @@ use parity_codec::{
...
 /// The SRML fundamental types.
 #[allow(unused)]
+#[cfg_attr(feature = "test-env", derive(Debug, Clone, PartialEq, Eq))]
 pub enum DefaultSrmlTypes {}
 
+/// Empty enum for default Call type, so it cannot be constructed.
+/// For calling into the runtime, a user defined Call type required.
+/// See https://github.com/paritytech/ink-types-node-runtime.
+///
+/// # Note
```
Robbepop:
```suggestion
/// # Note
///
```

Robbepop: I generally just like having a space there. Personal thing though ...

core/src/env/srml/types.rs
```diff
@@ -31,14 +31,34 @@ use parity_codec::{
...
+/// Some traits are only implemented to satisfy the constraints of the test environment, in order to
+/// keep the code size small.
+#[cfg_attr(feature = "test-env", derive(Debug, Clone, PartialEq, Eq))]
+pub enum Call {}
+impl parity_codec::Encode for Call {}
+
+#[cfg(feature = "test-env")]
+impl parity_codec::Decode for Call {
+    fn decode<I: parity_codec::Input>(_value: &mut I) -> Option<Self> {
+        unimplemented!()
```
Robbepop: This would crash with the message `not yet implemented` but we would never intend to have an implementation for it. So I think it is better to just return `None` and add a comment describing our intent to never implement it.

Issue #95, opened 2019-05-14
How to choose which Substrate version to target?


xlc:
Substrate 1.0 and master are not compatible. What is the plan for ink to handle compatibility issues across different Substrate versions?

Many people are going to use ink with Substrate 1.0 and still some people will want to use against Substrate master.

One example of a contract API breaking change: https://github.com/paritytech/substrate/pull/2563
Not sure if this is the only one given that Substrate have no changelog to read.

bkchr:
CC @Robbepop 

In general I would assume that `ink!` follows master. 1.0 is just a milestone. With 2.0 this will probably settle more, as 2.0 will/should be the basis of Polkadot 1.0.

Robbepop:
This is a good question and we should definitely answer this soon in the official docs.
Since ink! is not a stable, nor a stabily released product it might break from time to time.
We also have not yet implemented proper solutions that would prevent the breakage for future updates to either Substrate or ink!, however, as soon as we stabilize ink! we will definitely do so; or even earlier than that.
Our current recommendation thus is currently to use the current `master` of Substrate together with ink!.

Robbepop:
For a more elaborate answer go [here](https://github.com/paritytech/ink/issues/96#issuecomment-493389831).

xlc:
Does ink aims to support multiple versions of substrate simultaneously? I have no idea how to achieve it but it seems important. Just like how iOS developer will want to ensure his app works with multiple versions of iOS.

Robbepop:
Good question!
We could use some mechanism like this to specify a Substrate version for a concrete ink! smart contract:
https://github.com/paritytech/ink/issues/27
So the smart contract would be specified to work with exactly this version of Substrate.
However, the problem is that it is going to be hard to check the availability of the user's Substrate version.

The only sane thing we can do here is to not break things as soon as ink! stabilizes together with the version specifier that would always specify a certain major version of Substrate.

For example, if you specify `#[version = 1]` in your ink! smart contract you would be depending on Substrate version ^1.0 and the ink! and Substrate teams have to make sure that this is going to work.
This means that your smart contract is compatible with all Substrate versions of the v1.X series, e.g. v1.0, v1.1, etc.

Robbepop:
I think for now it doesn't make sense to start supporting multiple different Substrate versions with ink! since everything in this ecosystem is still in heavy flux. Once Substrate and ink! development starts to stabilize notably we can think about this again.
Until then ink! is always targeting the latest working Substrate release.

PR #129, opened 2019-06-24
Replace input buffer with scratch buffer

ascjones:
Tested with examples/lang/erc20 - works.

Rel: https://github.com/paritytech/substrate/pull/2911

### DIFF
--- a/core/src/env/srml/srml_only/impls.rs
+++ b/core/src/env/srml/srml_only/impls.rs
@@ -113,17 +113,7 @@ where
     T: EnvTypes,
 {
     fn input() -> Vec<u8> {
-        let size = unsafe { sys::ext_input_size() };
-        if size == 0 {
-            Vec::new()
-        } else {
-            let mut buffer = Vec::new();
-            buffer.resize(size as usize, 0);
-            unsafe {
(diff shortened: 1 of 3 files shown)


core/src/env/srml/srml_only/impls.rs
```diff
@@ -113,17 +113,7 @@ where
...
-            Vec::new()
-        } else {
-            let mut buffer = Vec::new();
-            buffer.resize(size as usize, 0);
-            unsafe {
-                sys::ext_input_copy(buffer.as_mut_ptr() as u32, 0, size);
-            }
-            buffer
-        }
+        read_scratch_buffer()
```
Robbepop: I like the simplicity of this change. :+1:

xlc:
does this means ink master no long works with substrate v1.0?

ascjones:
@xlc correct

xlc:
any plan to maintain compatibility with substrate v1.0? substrate master are discouraged to use for development due to frequent breaking changes
Related #95

ascjones:
It's a good point. Just spoke to @Robbepop about this and we will look into implementing versioning when he returns.

PR #131, opened 2019-06-28
[cli] deploy command

### LINKED ISSUE #134: Client command to deploy contracts (put_code)
Deploying a contract involves 2 steps:
1. Upload the code
2. Deploy an instance of the code

This document will address the implementation of 1. If this is successful then adding the next step will be relatively straightforward, the design of which will be addressed in a follow up issue.

## Motivation

- Current workflow to deploy a contract requires the UI, which is great but can be slow for a developer workflow. Requires lots of clicks etc.

- This tool could also be used for automation of deploying contracts, which would be useful for integration tests.

### UX

```
cargo contract deploy [ARGS]
```

#### Args
- **wasm_path:** (optional) Path of compiled wasm contract for upload, defaults to `./target/<name>-pruned.wasm`. Contract crate name extracted from `Cargo.toml`
- **Gas:** (optional) Max gas for deploying contract, defaults to 500000
- **Url:** (optional) Websockets url to a node on the target chain. defaults to `http://localhost:9944`
- **Account Name:** The origin account for deploying the contract. For version 0.1, only built in named accounts from the substrate keyring will be supported. See [signing](#signing)

#### Output
If successful, the `code hash` of the contract is returned. This can then be used to deploy an instance of the contract via the UI, or a future command in this client.

It should also be possible to optionally display the Extrinsic and Block hashes. 

### Implementation

The workflow for submitting the code will be:

- Load the wasm contract code from disk
- Construct an `Extrinsic` with with a `Call:contracts::put_code(gas, code)`
- Sign extrinsic with specified account
- Submit extrinsic via [`author_submitAndWatchExtrinsic`](https://github.com/paritytech/substrate/blob/master/core/rpc/src/author/mod.rs#L70)
- Wait for the extrinsic to be included in a block
- Fetch block with hash via [`chain_getBlock`](https://github.com/paritytech/substrate/blob/master/core/rpc/src/chain/mod.rs#L53)
- Listening to events via [`state_subscribeStorage`](https://github.com/paritytech/substrate/blob/master/core/rpc/src/state/mod.rs#L153), find the event which indicates success (key `contract.CodeStored`), or extrinsic failure. Consider a timeout here in case neither event is received.
- If successful event found, extract the code hash from the event data and return it to the user.

### Substrate Dependency and Types

When interacting with the substrate rpc, we need to have a dependency on substrate crates for the rpc apis, and the messages we send and receive.

This however has some issues that need considering:

- What version of substrate should we depend on?
- Whether and how do we handle multiple versions of substrate?
- Custom runtimes: not all chains will have the same types and module construction. Especially when constructing the `Call` type which needs to be compatible with the target chain. (See: https://github.com/paritytech/ink/pull/108)

## Signing

The initial version will only support the default built in [named accounts](http://localhost:9944). We need to carefully consider the options for key management since it is security critical, and also important to have a clean and flexible UX. Some ideas:

- Copy approach of `subkey` sign, [pass in secret/password on command line](https://github.com/paritytech/substrate/blob/master/subkey/src/main.rs#L175). 
- Use [substrate keystore](https://github.com/paritytech/substrate/blob/master/core/keystore/src/lib.rs#L105) to load key file and prompt for password.
- Allow for generating the unsigned `Extrinsic` to deploy the contract, sign it using some external tool (e.g. subkey), then accept the signed `Extrinsic` for submission. 

## PoC
In order to test some of these ideas there is a PoC available at https://github.com/paritytech/ink/pull/131

ascjones:
Closes #134.

This PR introduces the `deploy` command, which will upload compiled wasm to the chain by calling `put_code` and returning the code hash if successful. Note that this will not instantiate the contract. We should consider the name here, since `deploy` is overloaded in `ink!`, it is the name used for the constructor of the contract called when it is instantiated.

Much of the code here was extracted to https://github.com/paritytech/substrate-subxt, and this PR is waiting for that to be in good shape for use here.

- [x] `put_code`
- [x] extract code hash from event
- [x] cli options
- [x] integration test
- [x] wait for upstream fix to substrate for fetching `RuntimeVersion`: https://github.com/paritytech/substrate/pull/3471, required for adding the version to extrinsic payload in subxt https://github.com/paritytech/substrate-subxt/pull/12
- [x] remove dependency on node_runtime


dvc94ch:
How is this coming along?

ascjones:
@dvc94ch it's been on the back burner, but it was basically working (submitting extrinsic and waiting for it to be included in a block). I got as far as attempting to subscribe to events to extract the code hash which is not working yet. I'll come back to this shortly - but the client pubsub stuff has worked flawlessly so far :+1:

Robbepop:
> I got as far as attempting to subscribe to events to extract the code hash which is not working yet.

Can you elaborate what was not working about it and what your ideas and plans for it are? :)

ascjones:
@Robbepop I think I'm using the wrong hashing for the key for "System Events", might be `twox_128` rather than `blake_256`, need to test and confirm.

Also one blocker I was investigating on the substrate side was https://github.com/paritytech/substrate/issues/3080.

I'll return to this shortly and hopefully get it up and running.

.gitignore
```diff
@@ -21,3 +21,6 @@ Cargo.lock
 
 # Ignore history files.
 **/.history/**
+
+# Ignore rust-toolchain file
+rust-toolchain
```
Robbepop: Why? We should just have no `rust-toolchain` file anymore.

ascjones:
image

Just got this working. Needs a bit of polishing before its ready for review.

dvc94ch:
mmh so this only checks that the extrinsic is included in a block. does it matter if the block is finalized or not?
maybe we should wait for the block to be finalized before returning success, and if it is not finalized try again?

ascjones:
@dvc94ch good point. Could subscribe to `subscribe_finalized_heads` and wait for our block hash. Though this would obviously take a bit longer. 

Alternatively we could be "optimistic" and assume the block will be eventually finalized most of the time. In the event that it isn't, the follow up command (in this case deploying an instance of the contract) would fail. Maybe provide an error message to that effect.

dvc94ch:
I'd like to see this in it's own crate.
Some things I'd recommend to make it more general:

- Add an Error enum, this module shouldn't depend on CommandError.
- Make waiting for block inclusion optional
  for exchanges making a massive amount of transactions, the api should return immediately 
  and the nonce should be passed in externally, probably also the genesis hash. Maybe a 
  builder pattern would be more flexible.
- Add support for waiting for finality
  this is what a traditional wallet might want to do
- While I think it's ok to depend on the core/srml modules we need to loose the dependency on 
  node_runtime

ascjones:
image

cli/Cargo.toml
```diff
@@ -15,21 +15,38 @@ categories = ["cli", "tool"]
...
 parity-wasm = "0.40.2"
 cargo_metadata = "0.8.2"
+runtime_primitives = { git = "https://github.com/paritytech/substrate/", package = "sr-primitives" }
+serde = { version = "1.0", features = ["derive"] }
+srml-contracts = { git = "https://github.com/paritytech/substrate/", package = "srml-contracts" }
+srml-balances = { git = "https://github.com/paritytech/substrate/", package = "srml-balances" }
+srml-system = { git = "https://github.com/paritytech/substrate/", package = "srml-system" }
+srml-indices = { git = "https://github.com/paritytech/substrate/", package = "srml-indices" }
+substrate-primitives = { git = "https://github.com/paritytech/substrate/", package = "substrate-primitives" }
+subxt = { git = "https://github.com/paritytech/substrate-subxt/", package = "substrate-subxt" }
```
Robbepop: Let's get rid of as many direct SRML dependencies as we can.

cli/src/cmd/deploy.rs
```diff
@@ -0,0 +1,172 @@
...
+        ContractsXt,
+    },
+    system::System,
+};
+
+/// Load the wasm blob from the specified path, or defaults to the target contract wasm in the
+/// current project, inferred via the crate metadata.
+fn load_contract_code(path: Option<&PathBuf>) -> Result<Vec<u8>> {
+    let default_wasm_path = build::collect_crate_metadata()
+        .map(|metadata| metadata.dest_wasm())?;
```
Robbepop: might be replacable by `.map(CrateMetadata::dest_wasm)?`. not sure if that's an improvement though.

cli/src/cmd/deploy.rs
```diff
@@ -0,0 +1,172 @@
...
+    let mut data = Vec::new();
+    let mut file = fs::File::open(&contract_wasm_path)
+        .map_err(|e| format!("Failed to open {}: {}", contract_wasm_path.display(), e))?;
+    file.read_to_end(&mut data)?;
+
+    return Ok(data)
+}
+
+/// If the extrinsic was successfully included in a block, attempt to extract the code hash from the
+/// `Contracts::CodeStored` event
```
Robbepop: Normally you would want to write doc-comments as

----
Short description of what it does on a high-level.

Detailed and longer more elaborate description of what it does with mentions of exceptions and potential failures.
----

cli/src/cmd/deploy.rs
```diff
@@ -0,0 +1,172 @@
...
+/// Returns the code hash of the deployed contract if successful.
+///
+/// Optionally supply the contract wasm path, defaults to destination contract file inferred from
+/// Cargo.toml of the current contract project.
+///
+/// Creates an extrinsic with the `Contracts::put_code` Call, submits via RPC, then waits for
+/// the `ContractsEvent::CodeStored` event.
+pub(crate) fn execute_deploy(
+    url: url::Url,
+    surl: &str,
```
Robbepop: `surl` ? is that a thing? :upside_down_face:

cli/src/cmd/deploy.rs
```diff
@@ -0,0 +1,172 @@
...
+mod tests {
+    use assert_matches::assert_matches;
+    use std::{
+        fs,
+        io::Write,
+        path,
+    };
+
+    #[test]
+    #[ignore] // depends on a local substrate node running
```
Robbepop: would it make more sense to ignore this conditionally?

cli/src/cmd/deploy.rs
ascjones: Should be `suri` (secret uri). Copied from `subkey` https://github.com/paritytech/substrate/blob/5a2be8bd56ed5d9dc787fb3b9a76ce10ced4704b/subkey/src/main.rs#L234

cli/src/cmd/deploy.rs
ascjones: You mean e.g. `#[cfg(feature="substrate-node-integration")]`?

Robbepop: yeah for example if that works

cli/src/cmd/deploy.rs
ascjones: Yeah I've change this comment, not sure it is an improvement. Though that method is private and very succinct anyway

cli/Cargo.toml
ascjones: I'm going to move the default `Runtime` definition to `subxt` so we can remove these direct dependencies.

cli/src/cmd/deploy.rs
ascjones: Hmm if I do this it will be run on CI with `test --all-features` (and fail)

PR #133, opened 2019-07-08
Implement ext_call in ink!

taskooh:
I wanna call another contract in a contract.
Is this the right approach to do that?


core/src/env/srml/srml_only/impls.rs
```diff
@@ -29,7 +29,7 @@ use crate::{
 use core::{
     marker::PhantomData,
 };
-use parity_codec::Decode;
+use parity_codec::{Decode,Encode};
```
Robbepop: Please run `cargo fmt` to fix all formatting issues

core/src/env/srml/srml_only/impls.rs
```diff
@@ -145,4 +145,25 @@ where
             )
         }
     }
+
+    fn call(
+        callee : <Self as EnvTypes>::AccountId,
+        gas: u64,
```
Robbepop: This is basically correct, however as far as I remember there is a PR in the pipeline to make the type of `gas` generic. Not sure if it has been implemented for ink! yet.

Robbepop: Actually saw that this is not the case: https://github.com/paritytech/substrate/pull/2944/files#diff-375b14443b202b601fb667ba014f0711R269

core/src/env/traits.rs
```diff
@@ -120,4 +120,6 @@ pub trait Env: EnvTypes {
 
     /// Deposits raw event data through Contracts module.
     fn deposit_raw_event(topics: &[<Self as EnvTypes>::Hash], data: &[u8]);
+
+    fn call(callee : <Self as EnvTypes>::AccountId, gas: u64, value : <Self as EnvTypes>::Balance ,input_data : &[u8])->u32;
```
Robbepop: From the documentation of `ext_call`:

```
// Returns 0 on the successful execution and puts the result data returned
// by the callee into the scratch buffer. Otherwise, returns 1 and clears the scratch
// buffer.
```

While the API that returns a `u32` is correct I think even though this is fairly low-level we maybe want a different return type and value.
For example, we might want to return something like an `Option<ScratchBufferAccess>` to indicate that there was or wasn't a result instead of a `u32`.
Through `ScratchBufferAccess` I mean a thin-wrapper type that simply provides access to the returned data if any. This could be a type that simply wraps calls to `ext_scratch_len` and `ext_scratch_copy` and returns `Option<Vec<u8>>`.

As in simply returning `u32` we should avoid copying the result always from the scratch buffer since the claler might not be interested in the actual result and I am not sure if the compiler is smart enough to optimize these calls over FFI away.

core/src/env/srml/srml_only/impls.rs
taskooh: I added the gas generic type

Robbepop: Have you clicked on the link I provided to you?
The current state of SRML is that even though Gas is meant to be generic we fixed it to u64.
I will ask the SRML side if a generic type for ink! is useful at all given that decision. @pepyakin

core/src/env/srml/srml_only/impls.rs
```diff
@@ -152,20 +152,25 @@ where
         gas: <Self as EnvTypes>::Gas,
         value : <Self as EnvTypes>::Balance,
         input_data : &[u8]
-    ) -> u32 {
+    ) -> Option< Vec<u8> > {
```
Robbepop: Returning a `Vec<u8>` directly is exactly what I warned about since then all success path do a costly read to the scratch buffer and even if the caller doesn't even use it. Also the compiler is probably not powerful enough to actually optimize this call away since it is over the FFI.
What we actually want is a type abstraction that does the trick of reading the scratch buffer.

E.g.:

```rust
struct ReturnData<T> {
     marker: PhantomData<fn () -> T>,
}

impl<T> ReturnData<T>
where
    T: parity_codec::Decode,
{
    fn get() -> T {
        T::decode(&mut &read_scratch_buffer()[..])
    }
}
```

This way we have an API that can only be used to load from the scratch buffer if `Some(_)` is returned. Also it leaves control over loading the actual return data to the caller and also directly decodes it to the underlying return type of the called function. Later we will have to wrap this function in order to build the more high-level alternative that is also type safe with respect to the called function's return type.

core/src/env/srml/srml_only/impls.rs
Robbepop: Talked to @pepyakin and he confirmed that gas is no longer treated as a generic type and ink! should make use of fixed `u64` instead.

taskooh: I misunderstood your intention.
All right, I roll back this change.

core/src/env/srml/srml_only/impls.rs
taskooh: Should I decode the data in this low level api?
To do this, it requires to have another generic type in Env traits.
Also, i'm not sure where to implement this wrapping structure.

core/src/env/srml/srml_only/impls.rs
```diff
@@ -72,6 +72,18 @@ impl EnvStorage for SrmlEnvStorage {
...
 }
 
+/// Wrapped access to return data.
+/// get() Must be used right after ReturnData is returned.
+#[allow(unused)]
+pub struct ReturnData {}
+
+#[allow(unused)]
+impl ReturnData {
+    fn get() -> Vec<u8> {
```
Robbepop: The only thing now missing is making `ReturnData` generic over a `T: Decode` so that `get` returns a `T` directly `fn get() -> T`. It should panic (via `Result::expect`) if the decoding failed. For convenience feel free to also add a `fn get_raw() -> Vec<u8>` that simply does what the current `get` does.

core/src/env/srml/srml_only/impls.rs
```diff
@@ -72,6 +72,18 @@ impl EnvStorage for SrmlEnvStorage {
     }
 }
 
+/// Wrapped access to return data.
+/// get() Must be used right after ReturnData is returned.
+#[allow(unused)]
+pub struct ReturnData {}
```
Robbepop: Make it a tuple struct: `pub struct ReturnData<T>(PhantomData<fn () -> T>);`

core/src/env/srml/srml_only/impls.rs
```diff
@@ -72,6 +72,18 @@ impl EnvStorage for SrmlEnvStorage {
...
+
+#[allow(unused)]
```
Robbepop: why?

core/src/env/srml/srml_only/impls.rs
Robbepop: Also `get` should be public.

core/src/env/srml/srml_only/impls.rs
taskooh: Because sandbox(ext_sandbox_size and ext_sandbox_copy) is overwritten when another message is called.
So if call() is used several times in one contract method, we should use ReturnData::get() before next call() is used.

Robbepop: Yeah this is a problem.
So we have to make sure that this situation does not happen.
Normally users of an API are forced to use whatever they got using the `#[must_use]` attribute.

core/src/env/srml/srml_only/impls.rs
```diff
@@ -73,12 +72,33 @@ impl EnvStorage for SrmlEnvStorage {
...
+pub struct ReturnData<T> {
+    marker: PhantomData<fn() -> T>,
+}
+
+#[allow(unused)]
+impl<T> ReturnData<T>
+where
+    T: Decode,
+{
+    fn get() -> T {
```
Robbepop: Both `get` and `get_raw` should take `self` as their first argument to make them only callable once.
```suggestion
    fn get(self) -> T {
```

core/src/env/srml/srml_only/impls.rs
```diff
@@ -73,12 +72,33 @@ impl EnvStorage for SrmlEnvStorage {
...
+        T::decode(&mut &read_scratch_buffer()[..]).unwrap()
+    }
+
+    fn get_raw() -> Vec<u8> {
```
Robbepop:
```suggestion
    fn get_raw(self) -> Vec<u8> {
```

taskooh:
I implemented ReturnData as trait because implementation of get() differs between srml and test_env
to do this, it was difficult to add generic types to ReturnData itself so i added generic types to the get() function.

taskooh:
I rebased from master. is there something missing?

Robbepop:
Sorry, I forgot to push my latest code review for your PR and was already wondering why you don't react. Let me elaborate a bit more on the idea I stated in a comment above about splitting the `ext_call` API.

---

There currently is a problem with the proposed API that `ext_call` might return data that is to be fetch through the scratch buffer. However, not all calls actually return something. So if you implement an automatic scratch buffer retrieval for potential return data you might end up doing this wasted work.

Also the proposed work around with the additional guard type that you already implemented is not working properly either due to the stated problems with in-between-calls.

The idea to fix this is splitting this API up into two parts:

- `fn ext_invoke(..)`
    Won't return anything but also won't touch the scratch buffer.
    Users should use this for calls where they either won't use the result or where the call won't return data.
- `#[must_use] fn ext_evaluate<T: Decode>(..) -> T`
    Always fetches and decodes the data in the scratch buffer after the call has been made.
    Also guarantees that users actually use the return value.

Robbepop:
Please rebase this on current `master` to fix CI issues.

model/src/exec_env.rs
```diff
@@ -180,4 +181,26 @@ impl<T: Env> EnvHandler<T> {
     pub fn block_number(&self) -> T::BlockNumber {
         T::block_number()
     }
+
+    /// Calls a remote smart contract
```
Robbepop: Also please extend the documentation here since this is a very fragile part of the env API with many open questions. Try asking yourself what you would ask yourself about this and try to cover those questions in the docs. Same for the other API.

lang/src/encode_input_data.rs
```diff
@@ -0,0 +1,78 @@
+use ink_core::memory::vec::Vec;
+use ink_utils;
+use parity_codec::Encode;
+
+pub trait EncodeSafe{
```
Robbepop: What is the point?

core/src/env/traits.rs
```diff
@@ -127,4 +130,22 @@ pub trait Env: EnvTypes {
...
+    );
+
+    /// Calls a remote smart contract and return access to scratch buffer with return data
+    #[must_use]
+    fn call_evaluate<T: Decode>(
+        callee: <Self as EnvTypes>::AccountId,
+        gas: u64,
+        value: <Self as EnvTypes>::Balance,
+        input_data: &[u8],
+    ) -> Option<T>;
```
Robbepop: For later extensibility I think it is better if we return `Result<T, CallError>` here instead where `CallError` is for now just a `struct CallError` that can be extended later on.

core/src/env/traits.rs
```diff
@@ -127,4 +130,22 @@ pub trait Env: EnvTypes {
...
     fn deposit_raw_event(topics: &[<Self as EnvTypes>::Hash], data: &[u8]);
+
+
+    /// Calls a remote smart contract and return access to scratch buffer without return data
+    fn call_invoke(
+        callee: <Self as EnvTypes>::AccountId,
+        gas: u64,
+        value: <Self as EnvTypes>::Balance,
+        input_data: &[u8]
+    );
```
Robbepop: For `call_invoke` we want to return `Result<(), CallError>` since the call to another smart contract might be errorneous. `CallError` is simply a `struct CallError;` that we can extend later on.

core/src/env/test_env.rs
```diff
@@ -399,6 +399,17 @@ impl TestEnvData {
...
+
+    #[allow(unused_variables)]
+    pub fn call<T: Decode>(
+        &mut self,
+        callee: &[u8],
+        gas: u64,
+        value: &[u8],
+        input_data: &[u8],
+    ) -> Option<T> {
+        unimplemented!();
```
Robbepop: This should simply record the call by pushing a `struct CallData { callee, gas, value, input_data }` into a vector of those call data. See how it is done for events.

core/src/env/srml/srml_only/impls.rs
```diff
@@ -151,4 +147,50 @@ where
...
+            );
+        }
+    }
+
+    fn call_evaluate<U: Decode>(
+        callee: <Self as EnvTypes>::AccountId,
+        gas: u64,
+        value: <Self as EnvTypes>::Balance,
+        input_data: &[u8],
+    ) -> Option<U> {
```
Robbepop: As explained below we want to return `Result<T, CallError>` here.

core/src/env/srml/srml_only/impls.rs
```diff
@@ -151,4 +147,50 @@ where
             )
         }
     }
+    fn call_invoke(
+        callee: <Self as EnvTypes>::AccountId,
+        gas: u64,
+        value: <Self as EnvTypes>::Balance,
+        input_data: &[u8],
+    ) {
```
Robbepop: As explained below we want to return `Result<(), CallError>` here.

Robbepop:
Hey how is it going with this PR?
Can you please rebase on to master so we can actually merge it as soon as you fixed what I have already mentioned to you?
It would be very nice if we could finally pull this over the finishing line.

lang/src/encode_input_data.rs
taskooh: Since parity_codec::Encode includes generic type, cannot make trait object of "Encode". So i temporary made another trait to use in this code, which is not using generic types.

taskooh:
@Robbepop Sorry for being late, i was busy for my job.
I realize that test_env is quite not done.

core/src/env/traits.rs
```diff
@@ -136,4 +143,21 @@ pub trait Env: EnvTypes {
...
+    /// Calls a remote smart contract without returning data
+    fn call_invoke(
+        callee: <Self as EnvTypes>::AccountId,
+        gas: u64,
+        value: <Self as EnvTypes>::Balance,
+        input_data: &[u8]
+    ) -> Result<(), Self::CallError>;
+
+    /// Calls a remote smart contract and return encoded data
+    #[must_use]
```
Robbepop: Since we are using `Result` as return type we can remove this `#[must_use]` again since `Result` is `#[must_use]` on a type level already.

core/src/env/srml/srml_only/impls.rs
```diff
@@ -107,6 +112,8 @@ impl<T> Env for SrmlEnv<T>
 where
     T: EnvTypes,
 {
+    type CallError = CallError;
```
Robbepop: We do not want to introduce an associated type for `CallError` at this point.

core/src/env/srml/srml_only/impls.rs
```diff
@@ -69,6 +72,8 @@ impl EnvStorage for SrmlEnvStorage {
     }
 }
 
+pub struct CallError;
```
Robbepop: Needs documentation

Robbepop:
```suggestion
/// Errors encountered by calling a remote contract.
///
/// # Note
///
/// This is currently just a placeholder for potential future error codes.
pub struct CallError;
```

core/src/env/test_env.rs
```diff
@@ -47,6 +47,13 @@ impl EventData {
     }
 }
 
+struct CallData {
```
Robbepop:
```suggestion
/// Emulates the data given to remote smart contract call instructions.
struct CallData {
```

core/src/env/test_env.rs
```diff
@@ -179,6 +186,8 @@ pub struct TestEnvData {
     gas_left: Vec<u8>,
     /// The total transferred value.
     value_transferred: Vec<u8>,
+    /// The calls.
```
Robbepop:
```suggestion
    /// Invoked calls to remote contracts during test execution.
```

core/src/env/test_env.rs
```diff
@@ -47,6 +47,13 @@ impl EventData {
     }
 }
 
+struct CallData {
+    calee: Vec<u8>,
```
Robbepop:
```suggestion
    callee: Vec<u8>,
```

core/src/env/test_env.rs
```diff
@@ -317,6 +328,23 @@ impl TestEnvData {
     pub fn dispatched_calls(&self) -> impl DoubleEndedIterator<Item = &[u8]> {
         self.dispatched_calls.iter().map(Vec::as_slice)
     }
+
+    /// Appends new call to the end of the bytearray.
```
Robbepop:
```suggestion
    /// Records a new call.
```

core/src/env/test_env.rs
```diff
@@ -417,6 +445,18 @@ impl TestEnvData {
...
+
+    pub fn call<T: Decode>(
+        &mut self,
+        callee: &[u8],
+        gas: u64,
+        value: &[u8],
+        input_data: &[u8],
+    ) -> Result<T,CallError> {
+        add_call(callee, gas, value, input_data);
+        unimplemented!();
```
Robbepop: remove this

Robbepop:
```suggestion
```

Robbepop: I can see why you left it there. The returning of this is a bit more involved.
What I would recommend doing is the following:
Create another field in the test environment data (`Vec<u8>`) that is settable by the user and that is actually taken upon calling `call` to decode into the expected `T`. So upon testing a smart contract using the test environment a tester can completely control it and say what it expects from the outside.

core/src/env/test_env.rs
```diff
@@ -567,6 +607,31 @@ where
     fn dispatch_raw_call(data: &[u8]) {
         TEST_ENV_DATA.with(|test_env| test_env.borrow_mut().dispatch_call(data))
     }
+
+    fn call_invoke(
+        callee: T::AccountId,
+        gas: u64,
+        value: T::Balance,
+        input_data: &[u8],
+    ) {
```
Robbepop: should have same return type (`Result`) as emulated functions

core/src/env/test_env.rs
```diff
@@ -567,6 +607,31 @@ where
...
+            .with(|test_env| test_env.borrow_mut().call(callee, gas, value, input_data));
+        return Result<(),CallError{}>;
+    }
+
+    fn call_evaluate<U : Decode>(
+        callee: T::AccountId,
+        gas: u64,
+        value: T::Balance,
+        input_data: &[u8],
+    ) -> Option<U> {
```
Robbepop: Should have same return type as emulated function (`Result`)

lang/src/encode_input_data.rs
Robbepop: It is not idiomatic to do that and confusing. Please find another way to test this functionality. You are free to make limited use of macro_rules! macros in order to provide the functionality of your tests if required.

Robbepop:
@taskooh we are about to implement a more elaborate version of this feature.

For this we want to base our work on your PR, however, this requires the last remaining work done by you, otherwise we cannot merge it. We do not want to throw away all the work done by you so far. ;)
It would be nice if you could fix the remaining problems so that we can move on this whole feature.

core/src/env/srml/srml_only/impls.rs
taskooh: Do you have any idea where to implement CallError struct?
I want to use same struct in core/src/env/* and in model/src/exec_env.rs

Robbepop: Since `ink_model` depends on `ink_core` just define it in `ink_core` in `ink_core::env` and use it from `ink_model`.

core/src/env/srml/srml_only/impls.rs
```diff
@@ -153,4 +165,56 @@ where
...
+                gas,
+                value.as_ptr() as u32,
+                value.len() as u32,
+                input_data.as_ptr() as u32,
+                input_data.len() as u32,
+            );
+            if success == 0 {
+                Ok(())
+            } else {
+                Err(CallError {})
```
Robbepop:
```suggestion
                Err(CallError)
```

core/src/env/srml/srml_only/impls.rs
```diff
@@ -153,4 +165,56 @@ where
...
+                gas,
+                value.as_ptr() as u32,
+                value.len() as u32,
+                input_data.as_ptr() as u32,
+                input_data.len() as u32,
+            );
+            if success == 0 {
+                Ok(U::decode(&mut &read_scratch_buffer()[..]).unwrap())
+            } else {
+                Err(CallError {})
```
Robbepop:
```suggestion
                Err(CallError)
```

core/src/env/test_env.rs
```diff
@@ -567,6 +601,34 @@ where
...
+        callee: T::AccountId,
+        gas: u64,
+        value: T::Balance,
+        input_data: &[u8],
+    ) -> Result<(), CallError> {
+        let callee = &(callee.encode())[..];
+        let value = &(value.encode())[..];
+        let res = TEST_ENV_DATA
+            .with(|test_env| test_env.borrow_mut().call(callee, gas, value, input_data));
+        match res {
```
Robbepop: you can replace this `match` expression with a simple combinator: `.map(|_res| ())`

lang/src/encode_input_data.rs
```diff
@@ -0,0 +1,82 @@
+use ink_core::memory::vec::Vec;
```
Robbepop: please just remove this entire file. it is confusing and doesn't even test the calling infrastructure. we will find suitable tests for the call infrastructure at a later point.

core/src/env/test_env.rs
taskooh: I got it!
Also, i need to check error of `call`. Should i add a field of error flag?
Now it can catch only the decoding error, but cant catch error while executing `call`.

core/src/env/test_env.rs
```diff
@@ -47,6 +50,15 @@ impl EventData {
     }
 }
 
+/// Emulates the data given to remote smart contract call instructions.
+#[allow(unused)]
```
Robbepop: why would you need `#[allow(unused)]`?

core/src/env/test_env.rs
```diff
@@ -567,6 +609,35 @@ where
...
+        callee: T::AccountId,
+        gas: u64,
+        value: T::Balance,
+        input_data: &[u8],
+    ) -> Result<U, CallError> {
+        let callee = &(callee.encode())[..];
+        let value = &(value.encode())[..];
+        let return_data = TEST_ENV_DATA
+            .with(|test_env| test_env.borrow_mut().call(callee, gas, value, input_data));
+        match U::decode(&mut &return_data[..]) {
```
Robbepop: can be substituted with `.map(|bytes| U::decode(&mut &bytes[..]).map_err(CallError)`

taskooh: Ok, i'll try!

core/src/env/test_env.rs
taskooh: Errors occurred:
```
error: field is never used: `callee`
  --> core/src/env/test_env.rs:55:5
   |
55 |     callee: Vec<u8>,
   |     ^^^^^^^^^^^^^^^
   |
note: lint level defined here
  --> core/src/lib.rs:42:2
   |
42 |     unused,
   |     ^^^^^^
   = note: #[deny(dead_code)] implied by #[deny(unused)]

error: field is never used: `gas`
  --> core/src/env/test_env.rs:56:5
   |
56 |     gas: u64,
   |     ^^^^^^^^

error: field is never used: `value`
  --> core/src/env/test_env.rs:57:5
   |
57 |     value: Vec<u8>,
   |     ^^^^^^^^^^^^^^

error: field is never used: `input_data`
  --> core/src/env/test_env.rs:58:5
   |
58 |     input_data: Vec<u8>,
   |     ^^^^^^^^^^^^^^^^^^^

error: aborting due to 4 previous errors
```

taskooh:
@Robbepop @ascjones Is the `return_data` in
https://github.com/paritytech/ink/pull/133/conflicts
same as `call_return` which I implemented?

core/src/env/srml/srml_only/impls.rs
```diff
@@ -153,4 +157,56 @@ where
...
+
+    fn call_invoke(
+        callee: <Self as EnvTypes>::AccountId,
+        gas: u64,
+        value: <Self as EnvTypes>::Balance,
+        input_data: &[u8],
+    ) -> Result<(), CallError> {
+        let callee = callee.encode();
+        let value = value.encode();
+        unsafe {
```
Robbepop: I would appreciate to keep the bounds of `unsafe` code as low as possible.
For example in this case you could go with:
```rust
let result = unsafe {
    sys::ext_call(
        callee.as_ptr() as u32,
        callee.len() as u32,
        gas,
        value.as_ptr() as u32,
        value.len() as u32,
        input_data.as_ptr() as u32,
        input_data.len() as u32,
    )
};
if result != 0 {
    return Err(CallError)
}
Ok(())
```

core/src/env/srml/srml_only/impls.rs
```diff
@@ -153,4 +157,56 @@ where
...
+
+    fn call_evaluate<U: Decode>(
+        callee: <Self as EnvTypes>::AccountId,
+        gas: u64,
+        value: <Self as EnvTypes>::Balance,
+        input_data: &[u8],
+    ) -> Result<U, CallError> {
+        let callee = callee.encode();
+        let value = value.encode();
+        unsafe {
```
Robbepop: Same here as in `call_invoke` with `unsafe`

core/src/env/srml/srml_only/impls.rs
ascjones: I would go further and suggest that at the `core` layer we have just a `call` method which contains this unsafe block. This is more consistent with the mostly 1-1 mapping to `sys` functions at this layer. 

Then we could move the `call_invoke` and `call_evaluate` up to `exec_env` in the model layer, which would call into this.

Robbepop: I am fine with both approaches. Maybe you are right that it is a better fit in `ExecEnv` though.

Robbepop:
Even though there are some outstanding things to do, we (@ascjones and me) decided that we are going to merge this PR and refine it quickly to base our enhancements on top of it. I hope this is to the interest of you @taskooh . Thank you for your work!

PR #143, opened 2019-07-15
Extract ink_core/memory into its own sub-crate

### LINKED ISSUE #110: Extract ink_core/memory into its own sub-crate
We really need a further abstraction layer to handle memory types.

Right now memory types that can be found in Rust's alloc crate (that is stable since recently) life in `ink_core` but for example we shouldn't depend on `ink_core` from within the upcoming `ink_abi`.

So we want to extract `ink_core/memory` sub module out of `ink_core` into its own crate that we can from then on depend on.
Proposed sub-crate name: `ink_memory`

ascjones:
Resolves #110.

For now I've reexported `ink_memory` as `memory` from `ink_core`, so all existing imports are valid. Undecided about whether to directly import `ink_memory` for usages in `core` and `model`.


Robbepop:
According to our new guidelines every crate requires a README that (shortly) describes what it is about.

ascjones:
So not just a symlink to the root README like in the other crates?

Robbepop:
> So not just a symlink to the root README like in the other crates?

That's a good question.
I think it would be best if we provide a small description with rationals for the subcrate's existence for all the minor sub crates and then a link to the major README. What do you think?

ascjones:
As an experiment, I'm generating the README with [cargo readme](https://github.com/livioribeiro/cargo-readme). It will use the crate module level docs and the template to generate the README.

If we like this idea we can use it for other crates and use a shared template, and perhaps some automation to generate the READMEs.

ascjones:
@Robbepop does that seem a reasonable solution for the README?

Robbepop:
I really like the direction of separating this functionality out into its own sub-crate, however, I think we can do even better than simply moving the bits over. This could potentially act as a `std` to `no_std` + `alloc` compatibility layer instead of just an `alloc` shim.

So all `std` pieces would always be loaded from within this sub-crate.
For the purpose of usability we should also provide a prelude that we can simply import.
Also maybe the name of this compatibility layer should be different from `ink_memory`.

What about `ink_std`?  - That's may be a bit confusing due to `ink_core` existence though ...
Maybe later we can also split apart `ink_core` into `ink_env` and `ink_storage`.
Also we might want to add an `ink_traits` for bridging between the different sub-crates. At this point I should probably just open a new issue ...

ascjones:
See #157

PR #149, opened 2019-07-24
Make CI happy again

Robbepop:
Current CI problem with this PR:

```
Running "kcov" "--exclude-pattern=/home/travis/.cargo" "--verify" "/home/travis/build/paritytech/ink/target/cov/ink_lang-0700eacf7505e32b" "/home/travis/build/paritytech/ink/target/debug/ink_lang-0700eacf7505e32b"
/home/travis/build/paritytech/ink/target/debug/ink_lang-0700eacf7505e32b: error while loading shared libraries: libtest-2d1bd552fc52b268.so: cannot open shared object file: No such file or directory
```


Robbepop:
> LGTM, notwithstanding the CI error which appears to be unrelated

I am not sure. I know this CI error from the past but I forgot how I fixed it back then. Let's keep this PR open for ideas. Or try solving the remaining CI error with another PR?

ascjones:
We could merge on the basis that this PR means the CI gets a *lot* further than master: https://travis-ci.org/paritytech/ink.

On second thoughts the title of the PR is not satisfied so let's fix it properly before merging.

Robbepop:
Hmm I think you are right. It is already a huge benefit to have restored at least the majority of the ink! CI pipeline compared to the current `master`.

Issue #123, opened 2019-06-13
Build/test examples in CI


ascjones:
To prevent examples breaking, and as an extra test of changes: we should build and run tests as part of the CI for all the examples.

Rel #88 #103 #121 #116

ascjones:
Superseded by #159

PR #151, opened 2019-07-26
 [core] implement empty Flush for default AccountId and Hash types

ascjones:
Fixes erc20 example

### DIFF
--- a/core/src/env/srml/types.rs
+++ b/core/src/env/srml/types.rs
@@ -14,20 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with ink!.  If not, see <http://www.gnu.org/licenses/>.
 
-use core::{
-    array::TryFromSliceError,
-    convert::TryFrom,
-};
+use core::{array::TryFromSliceError, convert::TryFrom};
 
-use crate::{
-    env::EnvTypes,
-    storage::Flush,
-    impl_empty_flush_for,
-};
-use parity_codec::{
-    Decode,
-    Encode,
-};
+use crate::{env::EnvTypes, impl_empty_flush_for, storage::Flush};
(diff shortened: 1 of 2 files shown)


core/src/env/srml/types.rs
```diff
@@ -83,3 +77,5 @@ pub type Moment = u64;
 
 /// The default SRML blocknumber type.
 pub type BlockNumber = u64;
+
+impl_empty_flush_for!(AccountId, Hash);
```
Robbepop: I don't know if this is an overkill but okay ... :D

Robbepop: Also we might want to add additional constraints to the `EnvTypes` so that they are guaranteed to implement `Flush`. This might become a problem for `node-runtime` though ...

Robbepop: I am currently thinking about using `#![feature(optin_builtin_traits)]` which could greatly help here.

ascjones:
> We should really really add some of our examples to the CI for testing purposes.

Yeah: https://github.com/paritytech/ink/issues/123

core/src/env/srml/types.rs
ascjones: > Also we might want to add additional constraints to the `EnvTypes` so that they are guaranteed to implement `Flush`. This might become a problem for `node-runtime` though ...

Yeah I though about this too - would certainly catch the compilation errors earlier.

> I don't know if this is an overkill but okay ... :D

Perhaps, I thought since we have done this for all primitive types for the universality

Robbepop:
I ll merge this now even without the mentioned requested changes since it is important. Also we can do the other changes in another PR.

PR #155, opened 2019-07-29
[ci] Improve Travis CI scripts

Robbepop:
This fixes the used rust compiler channel to `nightly-2019-07-19` which is assumed to be working properly (at least on my machine). Also made normal `nightly` channel OK for failure. So it is not required to be working, unlike the `nightly-2019-07-19` channel. But with this in mind we can always check if the current nightly works.

Also this makes checks independent from each other. So all checks are always run.
On top of that it features a new CI summary that looks like the following upon success of failure:

**Success:**
```
CI Summary
==========

compile (all-features):        ok
compile (no-default-features): ok
formatting:                    ok
clippy (all-features):         ok
clippy (no-default-features):  ok
test (all-features):           ok
build Wasm:                    ok
coverage analysis:             ok
All checks have passed!
```

**Failure:**
```
CI Summary
==========

compile (all-features):        ERROR
compile (no-default-features): ok
formatting:                    ERROR
clippy (all-features):         ERROR
clippy (no-default-features):  ok
test (all-features):           ERROR
build Wasm:                    ok
coverage analysis:             ok
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,12 +61,95 @@ before_script:
 
 script:
 - |
-  cargo check --verbose --all --all-features &&
-  cargo check --verbose --all --no-default-features &&
-  cargo fmt --verbose --all -- --check &&
-  cargo clippy --verbose --all --all-features -- -D warnings &&
-  cargo clippy --verbose --all --no-default-features -- -D warnings &&
-  cargo test --verbose --all --all-features &&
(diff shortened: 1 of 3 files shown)


Robbepop:
> Looks like you are doing CI while you're doing CI.

Haha, that's exactly how I felt! :D

PR #156, opened 2019-07-30
sync parity-codec's version in template, land events example and lang flipper examples

flyq:
sync parity-codec's version in `template`, `land events example` and `lang flipper example` to 4.1, so that `cargo contract new xxx` can get right dependencies in `Cargo.toml`.


flyq:
It's my pleasure!

PR #170, opened 2019-08-15
Sync with SRML contracts for call/create status codes

Robbepop:
This is a fairly quick synchronization of SRML contracts recent refactoring that added support for `u32` return codes of `call` and `create` instructions of Wasm smart contracts in order to propagate error and success stati.

A follow-up refactoring of `ink_model` might make sense in order to do some long standing clean ups with the current architecture.

More information about the changes in SRML contracts can be found here: https://github.com/paritytech/substrate/pull/3320/commits

### DIFF
--- a/model/src/contract.rs
+++ b/model/src/contract.rs
@@ -288,7 +288,7 @@ where
 
 /// A return status code for `deploy` and `dispatch` calls back to the SRML contracts module.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
-pub struct RetCode(u32);
+pub struct RetCode(u8);
 
 impl RetCode {
     /// Indicates a successful execution.
@@ -303,7 +303,7 @@ impl RetCode {
 
     /// Returns the internal `u32` value.
     pub fn to_u32(self) -> u32 {
-        self.0
+        self.0 as u32
     }
 }
 
--- a/model/src/contract.rs
+++ b/model/src/contract.rs
@@ -477,8 +477,8 @@ where
         // Consumes the contract since nothing should be done afterwards.
         use ink_core::storage::alloc::Initialize as _;
         self.env.initialize(());
-        let deploy_params = DeployArgs::decode(&mut &input[..])
-            .map_err(|_err| RetCode::failure())?;
+        let deploy_params =
+            DeployArgs::decode(&mut &input[..]).map_err(|_err| RetCode::failure())?;
         (self.deployer.deploy_fn)(&mut self.env, deploy_params);
         self.env.state.flush();
         Ok(())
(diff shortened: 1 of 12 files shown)


model/src/contract.rs
```diff
@@ -286,13 +286,43 @@ where
     }
 }
 
+/// A return status code for `deploy` and `dispatch` calls back to the SRML contracts module.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct RetCode(u32);
```
ascjones: Why not make this a `u8` to enforce the max `255`? We're calling `to_u32()` in the lang layer anyway.

model/src/contract.rs
```diff
@@ -425,25 +470,28 @@ where
...
+    fn deploy_with(&mut self, input: &[u8]) -> Result<(), RetCode> {
         // Deploys the contract state.
         //
         // Should be performed exactly once during contract lifetime.
         // Consumes the contract since nothing should be done afterwards.
         use ink_core::storage::alloc::Initialize as _;
         self.env.initialize(());
-        let deploy_params = DeployArgs::decode(&mut &input[..]).unwrap();
+        let deploy_params = DeployArgs::decode(&mut &input[..])
+            .map_err(|_err| RetCode::failure())?;
```
ascjones: Do we plan to use the error codes 1-254 so we can debug where the error originates? 

Also can't believe that `unwrap` has been there all along!

model/src/contract.rs
```diff
@@ -474,6 +522,7 @@ where
 
     fn deploy(&mut self, input: Self::DeployArgs) {
         self.deploy_with(&input.encode()[..])
+            .expect("`deploy` failed to execute properly")
```
ascjones: The `TestableContract` could also return the error codes so we could, well, test them.

model/src/contract.rs
Robbepop: good point actually!

model/src/contract.rs
Robbepop: The `unwrap` should have been an `expect` but it didn't even matter because SRML contracts didn't provide any way to introspect the `expect` message anyway. Now we have better utilities.
We have yet to find out how we are going to handle failures in ink!.

model/src/contract.rs
Robbepop: Yeah but this is the wrong layer in my opinion because normally you do not test the `ink_model` layer and if it behaves properly upon incorrect input but instead you want to test your smart contract with valid input. Also this is guarded by `ink_lang`.

model/src/contract.rs
```diff
@@ -286,13 +286,43 @@ where
...
-    fn deploy(self);
+    ///
+    /// # Note
+    ///
+    /// Only the least-significant 8 bits of the `u32` return value may
+    /// be used to indicate success or an error state.
+    /// Zero (`0`) represents a successful execution, (`255`) means invalid
+    /// execution (e.g. trap) and any value in between represents a non-
+    /// specified invalid execution.
+    /// All other bits in the `u32` must be 0.
```
ascjones: This comment is redundant now that `RetCode` enforces it. Perhaps could move it (and the duplicate on deploy) to the `RetCode` type?

Robbepop:
> boom

:taco:

ascjones:
> Waiting for CI to finish.

:skull:

PR #171, opened 2019-08-16
[core] ext_call PR followup and refinements

Robbepop:
Quick followup PR for #133 .

### DIFF
--- a/core/src/env/srml/srml_only/impls.rs
+++ b/core/src/env/srml/srml_only/impls.rs
@@ -173,8 +173,8 @@ where
     ) -> Result<(), CallError> {
         let callee = callee.encode();
         let value = value.encode();
-        unsafe {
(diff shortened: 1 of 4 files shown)


core/src/env/srml/srml_only/impls.rs
```diff
@@ -199,21 +198,20 @@ where
...
                 input_data.as_ptr() as u32,
                 input_data.len() as u32,
-            );
-            if success == 0 {
-                U::decode(&mut &read_scratch_buffer()[..]).map_err(|_| CallError)
-            } else {
-                Err(CallError)
-            }
+            )
+        };
```
ascjones: I've made a quick PR into this branch with this block of code extracted to a method #172. Because only the last 3 lines of each of these methods are interesting.

PR #176, opened 2019-08-20
[core] Fix memory vec

riusricardo:
- Fix compilation error
- Update build scripts


core/src/env/srml/srml_only/calls.rs
```diff
@@ -14,16 +14,19 @@
 // You should have received a copy of the GNU General Public License
 // along with ink!.  If not, see <http://www.gnu.org/licenses/>.
 
+// We need this to fix a rustfmt issue. https://github.com/rust-lang/rustfmt/issues/3750
+#[rustfmt::skip]
+
```
Robbepop:
```suggestion
```

PR #177, opened 2019-08-22
[cli] handle error when contract package already exists, friendly error messages

ascjones:
- If a directory with a `Cargo.toml` already exists, then it fails with a friendly error message.
- If the directory exists then the template is unpacked to the existing directory
  - If any other file with the same name exists it will fail
- Error implements `Display` and this will be printed via `eprintln!`


cli/src/cmd/new.rs
```diff
@@ -69,6 +75,13 @@ fn initialize_for_lang(name: &str) -> Result<()> {
                     fs::create_dir_all(&p)?;
                 }
             }
+            if outpath.exists() {
```
DemiMarie-parity: This is racy.  Instead, the file should be opened with the `O_EXCL|O_NOFOLLOW` flags on \*nix.

cli/src/cmd/new.rs
```diff
@@ -113,4 +126,17 @@ mod tests {
             r#"Err(CommandError { kind: Other("Contract names cannot contain hyphens") })"#
         )
     }
+
+    #[test]
```
DemiMarie-parity: More tests needed.

ascjones:
@Robbepop good to merge this now?

cli/src/cmd/error.rs
```diff
@@ -30,12 +30,34 @@ pub enum CommandErrorKind {
...
     kind: CommandErrorKind,
 }
 
+impl std::fmt::Display for CommandError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.kind)
+    }
+}
+
 impl From<IoError> for CommandError {
```
Robbepop: We could replace impls like these by usage of: https://crates.io/crates/derive_more

cli/src/cmd/error.rs
```diff
@@ -30,12 +30,34 @@ pub enum CommandErrorKind {
     Other(String),
 }
 
+impl std::fmt::Display for CommandErrorKind {
```
Robbepop: Could be replace by usage of: https://jeltef.github.io/derive_more/derive_more/display.html

cli/src/cmd/error.rs
```diff
@@ -60,6 +82,14 @@ impl From<&str> for CommandError {
     }
 }
 
+impl From<String> for CommandError {
```
Robbepop: Maybe you can unify `From<&str>` and `From<String>` by implementing for `T: Into<String>`.

cli/src/cmd/new.rs
```diff
@@ -84,10 +104,10 @@ fn initialize_for_lang(name: &str) -> Result<()> {
         }
     }
 
-    Ok(())
+    Ok(format!("Created contract {}", name))
```
Robbepop: Why is returning a string needed here?

cli/src/cmd/error.rs
ascjones: I would agree, but to do so I would need to move `CommandErrorKind` variants up to the `CommandError` enum itself so they can be derived. Happy to do so if you agree.

cli/src/cmd/error.rs
ascjones: This doesn't work because some of the other errors e.g. `IoError` already implement `Into<String>` which causes conflicting implementations with `derive_more::From`

cli/src/cmd/new.rs
ascjones: So we can print a success message to the stdout: https://github.com/paritytech/ink/pull/177/files#diff-4c438f60247f494e2aee67fcce9e11e0R105

ascjones:
Existing reviews seem to have been auto dismissed when I pushed for some reason.

cli/src/cmd/error.rs
```diff
@@ -20,80 +20,21 @@ use std::{
 };
 use zip::result::ZipError;
 
-/// The kinds of command errors.
-#[derive(Debug)]
-pub enum CommandErrorKind {
+/// An error that can be encountered while executing commands.
+#[derive(Debug, derive_more::From, derive_more::Display)]
+pub enum CommandError {
```
Robbepop: aww much better <3

PR #186, opened 2019-09-05
Deprecate CellChunkAlloc for DynAlloc

cmichi:
Should close https://github.com/paritytech/ink/issues/37.

The repo tests run fine now, I also did a test-deploy of [a modified contract which does a lot of dynamic allocations](https://github.com/paritytech/ink/pull/111/files#diff-a8079a97a67414c6d1c9f30cca5f6ac9R63-R77).

Is there anything else we should test?


cmichi:
Hmm, it's strange that the CI fails. I don't see why, the only thing that somehow stood out to me is this: https://travis-ci.org/paritytech/ink/jobs/580102467#L1171-L1177 (though I'm not sure why it would output that diff).

PR #189, opened 2019-09-18
ink! model 2.0

Robbepop:
- Adds support for contracts having multiple constructors
- Introduce `constructors!` macro that works similar to `messages!`
- Change syntax in `messages!` macro to make it aware of `mut` messages
- Make `env` available through `self` where `Self`is the `Storage` struct
- Rename `state!` to `storage!`
- Rewrite the whole `ink_model` to improve the dispatcher


Robbepop:
Closed because work has been taken over to `ink_lang2`.
Please do _not_ delete the associated branch since this PR contains potentially interesting work for the future.

PR #190, opened 2019-09-18
ink abi v0.2

Robbepop:
- Support multiple constructors
- Docs and minor improvements
- Update `derive_more` to version `0.15`
- Update `syn`, `quote` and `proc_macro` dependencies to `1.0`
- Add `DisplayName` feature to leak type names non-transparently.
    - E.g. without `DisplayName` usages of type aliases are never communicated and the ABI just provide information about the aliased type but not about the type alias.
- Adjust `ink_lang` to generate code for the `ink_abi` changes
    - It does not yet make use of the new display name functionality


abi/src/specs.rs
```diff
@@ -76,11 +80,12 @@ pub enum Valid {}
 /// The message builder is not ready to finalize construction.
 pub enum Invalid {}
 
+/// A builder for contracts.
 pub struct ContractSpecBuilder<S = Invalid> {
     /// The name of the to-be-constructed contract specification.
     name: <MetaForm as Form>::String,
-    /// The deploy handler of the to-be-constructed contract specification.
-    deploy: Option<DeploySpec>,
+    /// The constructors of the to-be-constructed constract specification.
```
ascjones:
```suggestion
    /// The constructors of the to-be-constructed contract specification.
```

lang/src/hir.rs
```diff
@@ -244,10 +244,12 @@ impl Contract {
                     }
                     Some(self_ty) => {
                         match self_ty {
-                            ast::FnArg::SelfValue(_) | ast::FnArg::Captured(_) => bail!(
+                            ast::FnArg::SelfValue(_) | ast::FnArg::Captured(_) => {
+                                bail!(
```
ascjones: Indentation looks wrong here

Robbepop: i think that's what the latest `rustfmt` changes did - will look into it though

PR #192, opened 2019-09-19
Implement the build command

pepyakin:


examples/lang/delegator/build-all.sh
```diff
@@ -1,6 +0,0 @@
-#!/bin/bash
```
Robbepop: Hmm ... do we have a proper solution for this?

Robbepop: We could simply leave the `build-all.sh` script but invoking `cargo contract build` instead of the local `build.sh` scripts.

README.md
```diff
@@ -41,12 +41,17 @@ Use the scripts provided under `scripts` directory in order to run checks on eit
...
-For running the example smart contracts found under `examples` you also need to install some further tools:
+For building the example smart contracts found under `examples` you will need to have `cargo-contract` installed.
 
-- `wabt` (WebAssembly Binary Toolkit)
-- `wasm-utils` (Parity's Wasm Utilities)
+```
+cargo install cargo-contract
+```
+
+Execute the following command in shell while in an example's directory:
```
Robbepop: Between `cargo install cargo-contract` and `cargo contract build` we should maybe also include a section for `cargo contract new` since that is the logical step in between.

Robbepop:
btw.: this is amazing! :tada:

README.md
pepyakin: While I understand and agree with you in general, I think this is not appropriate place for this kind of documentation: this particular piece is only about running examples.

pepyakin:
Not sure what's up with CI

Robbepop:
> Not sure what's up with CI
CI reports
```
thread 'cmd::new::tests::dont_overwrite_existing_files_not_in_cargo_project' panicked at 'assertion failed: `(left == right)`

  left: `"Ok(\"Created contract dont_overwrite_existing_files\")"`,

 right: `"Err(Other(\"New contract file dont_overwrite_existing_files/build.sh already exists\"))"`', cli/src/cmd/new.rs:157:9
```

pepyakin:
This PR doesn't strip custom sections, so binaries produced by it are huge.

Robbepop:
> This PR doesn't strip custom sections, so binaries produced by it are huge.

Will you work on this?

cli/src/cmd/build.rs
```diff
@@ -153,6 +154,16 @@ fn ensure_maximum_memory_pages(
     Ok(())
 }
 
+/// Strips all custom sections.
+///
+/// Presently all custom sections are not required so they can be stripped safely.
+fn strip_custom_sections(module: &mut Module) {
```
Robbepop: haha nice! looks simple

riusricardo: This is much better now, but it still doubles the size of master's build. However, it has a reasonable size.

pepyakin:
@riusricardo I don't expect the difference. Can you publish the two binaries so that I can compare them?

riusricardo:
@pepyakin https://github.com/riusricardo/flipper-contract

pepyakin:
Yeah, sure. As I said, this is something that I didn't expect and should be fixable.

pepyakin:
Yep, turns out that parity-wasm treats as the name section specially so adding another case solved the issue. Not I am getting 23kb for each contract.

riusricardo:
Nice! I'm also getting 23 KB on ser-build branch.

PR #197, opened 2019-10-01
[cli] fix `new` command tests, pass in optional target dir

ascjones:
Both the `new` and `build` commands now accept an optional target dir, which is used by the tests to pass in the temporary directory.

This fixes the broken tests and prevents the test contract files being created in the working dir.


Robbepop:
Does this also fix the `rustfmt` issues? The CI is either not yet updated or stating that this PR still has formatting errors but I think the CI is simply not up to date yet.

Robbepop:
Let's wait for CI to pass.

cli/src/cmd/build.rs
```diff
@@ -88,14 +92,31 @@ pub fn collect_crate_metadata() -> Result<CrateMetadata> {
...
+        .unwrap_or_default()
+        .contains("-nightly");
+
+    if !is_nightly_default {
+        cmd.arg("+nightly");
+    }
+
+    let output = cmd
         .args(&[
-            "+nightly",
```
Robbepop: I am still undecided about this because for CI it makes total sense but not for all people that are generally working with stable Rust.

ascjones:
> Minor grumbles but will approve anyway because we really need this merged finally.

Can we merge it finally then? I'm inclined to agree about the `+nightly` stuff - but don't have the time right now to figure out a better solution to get it to work on CI.

PR #201, opened 2019-10-05
Erc721 example

riusricardo:
- Erc721 Example

### DIFF
--- a/examples/lang/erc721/src/lib.rs
+++ b/examples/lang/erc721/src/lib.rs
@@ -21,11 +21,11 @@ use ink_core::{
     memory::format,
     storage,
 };
+pub use core::result::Result;
 use ink_lang::contract;
 use ink_model;
 
 pub type EnvHandler = ink_model::EnvHandler<ink_core::env::ContractEnv<DefaultSrmlTypes>>;
-pub use core::result::Result;
 pub type TokenId = u32;
 pub type Counter = u32;
 
--- a/examples/lang/erc721/src/lib.rs
+++ b/examples/lang/erc721/src/lib.rs
@@ -219,7 +219,7 @@ contract! {
 
         fn burn(&mut self, env: &EnvHandler, from: &AccountId, id: &TokenId)-> Result<(), u32> {
             let caller = env.caller();
-            if !self.exists(id){
+            if !self.exists(id) {
                  return Err(Error::TokenNotFound.to_u32())
             };
             if self.owner_of(id) != caller && *from != AccountId::from([0x0; 32]) {
--- a/examples/lang/erc721/src/lib.rs
+++ b/examples/lang/erc721/src/lib.rs
@@ -25,7 +25,7 @@ use ink_lang::contract;
 use ink_model;
 
 pub type EnvHandler = ink_model::EnvHandler<ink_core::env::ContractEnv<DefaultSrmlTypes>>;
-pub type Result<T, E> = core::result::Result<T, E>;
+pub use core::result::Result;
 pub type TokenId = u32;
 pub type Counter = u32;
 
--- a/examples/lang/erc721/src/lib.rs
+++ b/examples/lang/erc721/src/lib.rs
@@ -1,0 +1,230 @@
+// Copyright 2018-2019 Parity Technologies (UK) Ltd.
+// This file is part of ink!.
+//
+// ink! is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// ink! is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with ink!.  If not, see <http://www.gnu.org/licenses/>.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use ink_core::{
+    env::DefaultSrmlTypes,
+    memory::format,
+    storage,
+};
+use ink_lang::contract;
+use ink_model::EnvHandler;
+use ink_core::storage::alloc::Initialize as _;
+use ink_core::storage::alloc::AllocateUsing as _;
+use std::ops::Index;
+
+pub type Result<T, E> = core::result::Result<T, E>;
+
+contract! {
+    #![env = DefaultSrmlTypes]
+
+    // Event deposited when a token transfer occurs
+    event Transfer {
+        from: AccountId,
+        to: AccountId,
+        token_id: u32,
+    }
+
+    // Event deposited when an approval occurs
+    event Approval {
+        owner: AccountId,
+        to: AccountId,
+        token_id: u32,
+    }
+
+    /// The storage items for a typical ERC721 token implementation.
+    struct Erc721 {
+        /// The total supply.
+        token_owner: storage::HashMap<u32, AccountId>,
+        token_approvals: storage::HashMap<u32, AccountId>,
+        owned_tokens: storage::HashMap<AccountId, storage::Vec<u32>>,
+        all_tokens: storage::Vec<u32>,
+        name: storage::Vec<u8>,
+        symbol: storage::Vec<u8>,
+
+    }
+
+    impl Deploy for Erc721 {
+        fn deploy(&mut self, name: u8, symbol: u8) {
+            self.name.push(name);
+            self.symbol.push(symbol);
+        }
+    }
+
+    impl Erc721 {
+
+        pub(external) fn get_total_supply(&self) -> u32 {
+            let total_supply = self.all_tokens.len();
+            env.println(&format!("Erc20::total_supply = {:?}", total_supply));
+            total_supply
+        }
+        
+        pub(external) fn get_balance(&self, owner: AccountId) -> u32 {
+            let balance = self.balance_of(&owner);
+            env.println(&format!("Erc721::balance_of(owner = {:?}) = {:?}", owner, balance));
+            balance
+        }
+
+        pub(external) fn get_owner(&self, token_id: u32) -> AccountId {
+            let owner = self.owner_of(&token_id);
+            env.println(&format!("Erc721::owner_of(token = {:?}) = {:?}", token_id, owner));
+            owner
+        }
+
+    }
+
+    impl Erc721 {
+        
+        fn approve(&mut self, env: &mut EnvHandler<ink_core::env::ContractEnv<DefaultSrmlTypes>>, to: &AccountId, token_id: &u32) -> Result<(), &'static str> {
+            let caller = env.caller();
+
+            if caller == self.owner_of(token_id){
+
+                self.token_approvals.insert(*token_id, caller);
+
+                env.emit(Approval {
+                    owner: caller,
+                    to: *to,
+                    token_id: *token_id,
+                });
+
+                Ok(())
+
+            } else{
+                Err("not owner")
+            }
+        }
+
+        fn transfer_from(&mut self, env: &mut EnvHandler<ink_core::env::ContractEnv<DefaultSrmlTypes>>, from: &AccountId, to: &AccountId, token_id: &u32) -> Result<(), &'static str> {        
+            let caller = env.caller();
+
+            if self.approved_or_owner(&caller, token_id){
+
+                self.clear_approval(from, token_id)?;
+                self.remove_token_from(from, token_id)?;
+                self.add_token_to(env, to, token_id)?;
+
+                env.emit(Transfer {
+                    from: *from,
+                    to: *to,
+                    token_id: *token_id,
+                });
+
+                Ok(())
+
+            } else{
+                Err("not approved")
+            }
+            
+        }
+
+        fn clear_approval(&mut self, from: &AccountId, token_id: &u32) -> Result<(), &'static str> {
+            
+            if *from == self.owner_of(token_id){
+                match self.token_approvals.remove(token_id) {
+                    Some(_res) => Ok(()),
+                    None => Err("cannot remove token approval")
+                }
+            } else{
+                Err("not owner")
+            }
+        }
+
+        fn remove_token_from(&mut self, from: &AccountId, token_id: &u32) -> Result<(), &'static str> {
+
+            if self.owner_of(token_id) != *from {
+                return Err("not owner")
+            };
+
+            if !self.owned_tokens.contains_key(from){
+                return Err("tokens not found");
+            };
+
+            let tokens = self.owned_tokens
+            .get_mut(from)
+            .expect("cannot get tokens");
+
+            match tokens.swap_remove(*tokens.index(*token_id)) {
+                Some(_res) => Ok(()),
+                None => Err("cannot remove token")
+            }
+        }
+
+        fn add_token_to(&mut self, env: &mut EnvHandler<ink_core::env::ContractEnv<DefaultSrmlTypes>>, to: &AccountId, token_id: &u32) -> Result<(), &'static str> {
+
+            if self.owner_of(token_id) != AccountId::from([0x0; 32]){
+                return Err("already assigned")
+            };
+
+            if !self.owned_tokens.contains_key(to){
+
+                let new_vec = unsafe {storage::Vec::<u32>::allocate_using(&mut env.dyn_alloc)}.initialize_into(());
+                
+                self.owned_tokens
+                    .insert(*to, new_vec)
+                    .ok_or("cannot add tokens")?;
+            };
+
+            let tokens = self.owned_tokens
+                .get_mut(to)
+                .expect("cannot find tokens");
+
+            tokens.push(*token_id);
+
+            Ok(())
+        }
+
+        fn balance_of(&self, of: &AccountId, ) -> u32 {
+            
+            let balance = match self.owned_tokens.get(of) {
+                Some(num) => num.len(),
+                None => 0u32,
+            };
+            balance
+        }
+
+        fn owner_of(&self, token_id: &u32, ) -> AccountId {
+            *self.token_owner.get(token_id).unwrap_or(&AccountId::from([0x0; 32]))
+        }
+
+        fn approved_for(&self, token_id: &u32) -> AccountId {
+            *self.token_approvals.get(token_id).unwrap_or(&AccountId::from([0x0; 32]))
+        }
+
+        fn approved_or_owner(&self, owner: &AccountId, token_id: &u32) -> bool {
+            *owner == self.owner_of(token_id) ||  *owner == self.approved_for(token_id)
+        }
+    }
+}
+
+#[cfg(all(test, feature = "test-env"))]
+mod tests {
+    use super::*;
+    use ink_core::env;
+    type Types = ink_core::env::DefaultSrmlTypes;
+
+    #[test]
+    fn deployment_works() {
+        let alice = AccountId::from([0x0; 32]);
+        env::test::set_caller::<Types>(alice);
+
+        let erc721 = Erc721::deploy_mock(1,2);
+        assert_eq!(erc721.get_total_supply(), 0);
+        assert_eq!(erc721.get_balance(alice), 0);
+        assert_eq!(erc721.get_owner(1), AccountId::from([0x0; 32]));
+    }
+}
--- a/examples/lang/erc721/src/lib.rs
+++ b/examples/lang/erc721/src/lib.rs
@@ -36,11 +36,11 @@ struct Counter {
 }
 
 impl Counter {
-    pub fn add(&mut self, by: u32) {
+    pub fn increase(&mut self) {
         self.value += 1;
     }
 
-    pub fn sub(&mut self, by: u32) {
+    pub fn decrease(&mut self) {
         self.value -= 1;
     }
 
@@ -168,7 +168,7 @@ contract! {
             let count = self.owned_tokens_count
                 .get_mut(from)
                 .ok_or("cannot get account count")?;
-            count.sub(1u32);
+            count.decrease();
             Ok(())
         }
 
@@ -184,7 +184,7 @@ contract! {
             let count = self.owned_tokens_count
                 .get_mut(to)
                 .ok_or("cannot get account count")?;
-            count.add(1u32);
+            count.increase();
             Ok(())
             } else{
                 Err("token has owner")
--- a/examples/lang/erc721/src/lib.rs
+++ b/examples/lang/erc721/src/lib.rs
@@ -186,7 +186,7 @@ contract! {
         }
 
         fn add_token_to(&mut self, to: &AccountId, id: &TokenId) -> Result<(), u32> {
-            if self.exists(id){
+            if self.exists(id) {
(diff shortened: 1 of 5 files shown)


examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+
+use ink_core::{
+    env::DefaultSrmlTypes,
+    memory::format,
+    storage,
+};
+use ink_lang::contract;
+use ink_model::EnvHandler;
+use ink_core::storage::alloc::Initialize as _;
+use ink_core::storage::alloc::AllocateUsing as _;
```
Robbepop: Please use `cargo fmt` (or `rustfmt` directly) to format the code.
This makes it much more readable, also for code reviews.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+use std::ops::Index;
```
Robbepop: Never use `std` in a contract implementation.
Also use the `core` crate or `ink_core::memory` if you want to access entities in the `alloc` crate but also prefer not to use `alloc` directly.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+pub type Result<T, E> = core::result::Result<T, E>;
+
+contract! {
+    #![env = DefaultSrmlTypes]
+
+    // Event deposited when a token transfer occurs
+    event Transfer {
+        from: AccountId,
+        to: AccountId,
+        token_id: u32,
```
Robbepop: I think it would be better to create a type alias `pub type TokenId = u32;` instead of using `u32` directly.

Robbepop: Also then you can simply name the field `id` because it is clear what kind of id this refers to.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /// The storage items for a typical ERC721 token implementation.
+    struct Erc721 {
+        /// The total supply.
+        token_owner: storage::HashMap<u32, AccountId>,
+        token_approvals: storage::HashMap<u32, AccountId>,
+        owned_tokens: storage::HashMap<AccountId, storage::Vec<u32>>,
```
Robbepop: Following the Open Zepelin implementation of Erc721 they are using a simple counter on the user side of things. So instead of `owned_tokens: storage::HashMap<AccountId, storage::Vec<u32>>` they simple store `owned_token_count: storage::HashMap<AccountId, u32>;`. Maybe we should stick to their implementation as a reference and also we would no longer need dynamic allocations.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+        all_tokens: storage::Vec<u32>,
+        name: storage::Vec<u8>,
+        symbol: storage::Vec<u8>,
```
Robbepop: What is symbol used for?

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+        token_id: u32,
+    }
+
+    /// The storage items for a typical ERC721 token implementation.
+    struct Erc721 {
+        /// The total supply.
+        token_owner: storage::HashMap<u32, AccountId>,
+        token_approvals: storage::HashMap<u32, AccountId>,
+        owned_tokens: storage::HashMap<AccountId, storage::Vec<u32>>,
+        all_tokens: storage::Vec<u32>,
```
Robbepop: I don't think we need this.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+        name: storage::Vec<u8>,
+        symbol: storage::Vec<u8>,
+
+    }
+
+    impl Deploy for Erc721 {
+        fn deploy(&mut self, name: u8, symbol: u8) {
```
Robbepop: `name` should be `ink_core::memory::vec::Vec<u8>` or short `Vec<u8>` (note the missing `storage`!) And I still don't get what `symbol` is used for but the same probably applies for it as well. We could in theory even make use of `String` but then again some people dislike it being used so we don't.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+            self.name.push(name);
+            self.symbol.push(symbol);
+        }
+    }
+
+    impl Erc721 {
+
+        pub(external) fn get_total_supply(&self) -> u32 {
```
Robbepop: Is this part of the "official" Erc721 standard? I am missing it here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/IERC721.sol Maybe I am looking wrong.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+                env.emit(Approval {
+                    owner: caller,
+                    to: *to,
+                    token_id: *token_id,
+                });
+
+                Ok(())
+
+            } else{
+                Err("not owner")
```
Robbepop: I would prefer us sticking as closely as possible to the actual Erc721 standard. We can later always do it better since we are using Rust but for now we should try to stick to the given standard.

examples/lang/erc721/src/lib.rs
Robbepop: https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/token/ERC721

examples/lang/erc721/src/lib.rs
riusricardo: https://github.com/eduadiez/DAppNodeNFT/blob/master/contracts/ERC721.sol

Robbepop: Then this implementation seems to me to be a superset of the Erc721 standard.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        self.value -= 1;
+    }
+
+    pub fn get(&self) -> u32 {
+        *self.value
+    }
+}
+
+impl Flush for Counter {
+	fn flush(&mut self) {}
```
Robbepop: you need to actually flush any storage entity in the `Flush` implementation.
so `storage::Value<u32>` should be flushed for a correct implementation.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+use scale::{Encode, Decode};
+
+pub type Result<T, E> = core::result::Result<T, E>;
+pub type TokenId = u32;
+
+#[derive(Encode, Decode)]
+#[cfg_attr(feature = "ink-generate-abi", derive(Metadata))]
+struct Counter {
+    value: storage::Value<u32>,
+}
```
Robbepop: This also requires an `ink_core::storage::alloc::AllocateUsing` implementation.

Robbepop: However, I think it isn't necessary to use `storage::Value<u32>` within `Counter` but instead you could simply use `u32` and if you require to store `Counter` in the contract's storage you can still store it as `storage::Value<Counter>`.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        pub(external) fn get_owner(&self, id: u32) -> AccountId {
+            let owner = self.owner_of(&id);
+            env.println(&format!("Erc721::owner_of(token = {:?}) = {:?}", id, owner));
+            owner
+        }
+    }
+
+    impl Erc721 {
+        fn approve(&mut self, env: &mut EnvHandler<ink_core::env::ContractEnv<DefaultSrmlTypes>>, to: &AccountId, id: &TokenId) -> Result<(), &'static str> {
```
Robbepop: Please create a type alias `type EnvHandler = EnvHandler<ink_core::env::ContractEnv<DefaultSrmlTypes>>` or similar to end this abomination.
I am so glad we are getting rid of this in the upcoming `ink_lang2` codegen.

examples/lang/erc721/src/lib.rs
riusricardo: I did a few changes as you suggested. Do I still need to implement the flush if it is not anymore a `storage::Value<u32>`?
Flush is implemented at `storage::Value`

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+        fn mint(&mut self, env: &EnvHandler, to: &AccountId, id: &TokenId) -> Result<(), &'static str> {
+            if *to == AccountId::from([0x0; 32]){
+                return Err("cannot mint to account zero")
+            };
+            if self.exists(id){
+                return Err("token already minted")
+            };
+
+            self.token_owner
+                .insert(*id, *to)
```
riusricardo: @Robbepop any idea why it fails to insert a token? It can do the next steps and insert a counter but the token fails.

riusricardo: I can't figure out the reason but if I comment the next line `.ok_or("cannot insert token")?;` it just works

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+use ink_core::{
+    env::DefaultSrmlTypes,
+    memory::format,
+    storage,
+};
+use ink_lang::contract;
+use ink_model;
+
+pub type EnvHandler = ink_model::EnvHandler<ink_core::env::ContractEnv<DefaultSrmlTypes>>;
+pub type Result<T, E> = core::result::Result<T, E>;
```
gavofyork: why not just:

```suggestion
pub use core::result::Result;
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+pub type ErrNo = u32;
+
+const NOT_OWNER: ErrNo = 1;
+const INSERT_ERROR: ErrNo = 2;
+const TOKEN_EXISTS: ErrNo = 3;
+const NOT_APPROVED: ErrNo = 4;
+const TOKEN_NOT_FOUND: ErrNo = 5;
+const CAN_NOT_GET_COUNTER: ErrNo = 6;
+const ACCOUNT_ZERO_NOT_ALLOWED: ErrNo = 7;
+const CAN_NOT_REMOVE: ErrNo = 8;
```
Robbepop: Please use a Rust `enum` for error kinds. You can implement a `to_u32` method on it if you want to use raw `u32` values instead upon returning to SRML contracts.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+            env.emit(Transfer {
+                from: *from,
+                to: *to,
+                id: *id,
+            });
+            Ok(())
+        }
+
+        fn clear_approval(&mut self, id: &TokenId) -> Result<(), ErrNo> {
+            if !self.token_approvals.contains_key(id){
```
Robbepop: Please run `cargo fmt` on the file to fix formatting issues.

examples/lang/erc721/src/lib.rs
```diff
@@ -16,28 +16,44 @@
...
-const TOKEN_NOT_FOUND: ErrNo = 5;
-const CAN_NOT_GET_COUNTER: ErrNo = 6;
-const ACCOUNT_ZERO_NOT_ALLOWED: ErrNo = 7;
-const CAN_NOT_REMOVE: ErrNo = 8;
+#[cfg(feature = "ink-generate-abi")]
+use type_metadata::Metadata;
+
+#[derive(Encode, Decode, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "ink-generate-abi", derive(Metadata))]
+pub enum ErrNo {
```
Robbepop: can we simply call this `Error`?

examples/lang/erc721/src/lib.rs
```diff
@@ -76,11 +76,14 @@ contract! {
     struct Erc721 {
         token_owner: storage::HashMap<TokenId, AccountId>,
         token_approvals: storage::HashMap<TokenId, AccountId>,
-        owned_tokens_count: storage::HashMap<AccountId, Counter>
+        owned_tokens_count: storage::HashMap<AccountId, Counter>,
+        symbol: storage::Value<u8>,
```
Robbepop: Why do we need symbol? I though we discussed this before and concluded that we won't need it for a basic Erc721 implementation, didn't we? Also as far as I see it this symbol serves as UI element so we won't want to push for that.

riusricardo: Removed. This was a fix for the ABI error.

riusricardo: It turns out that it needs to initialize a variable on deploy or it breaks the ABI generation.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+    CanNotRemove,
+    CanNotGetCounter,
+    AccountZeroNotAllowed,
+}
+
+impl From<Error> for u32 {
+    fn from(item: Error) -> Self {
+        item as u32
+    }
+}
```
Robbepop: I would generally not implement such `From` impls for primitive types.
Maybe I am a bit too over-restrictive here: What I would do instead is to simply provide a method on `Error`: `fn to_u32(self) -> u32` and also I would derive `Copy` (and `Clone`) for `Error` because we can.

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+};
+
+pub type EnvHandler = ink_model::EnvHandler<ink_core::env::ContractEnv<DefaultSrmlTypes>>;
+pub type TokenId = u32;
+pub type Counter = u32;
+
+#[cfg(feature = "ink-generate-abi")]
+use type_metadata::Metadata;
+
+#[derive(Encode, Decode, Debug, PartialEq, Eq)]
```
Robbepop: add derives for `Copy` and `Clone` because we can

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                owner: caller,
+                to: *to,
+                id: *id,
+            });
+            Ok(())
+        }
+
+        fn transfer_from(&mut self, env: &EnvHandler, from: &AccountId, to: &AccountId, id: &TokenId) -> Result<(), Error> {
+            let caller = env.caller();
+            if !self.exists(id){
```
Robbepop:
```suggestion
            if !self.exists(id) {
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                return Err(Error::TokenNotFound);
```
Robbepop:
```suggestion
                return Err(Error::TokenNotFound)
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+            env.emit(Transfer {
+                from: *from,
+                to: *to,
+                id: *id,
+            });
+            Ok(())
+        }
+
+        fn clear_approval(&mut self, id: &TokenId) -> Result<(), Error> {
+            if !self.token_approvals.contains_key(id){
```
Robbepop:
```suggestion
            if !self.token_approvals.contains_key(id) {
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                return Ok(());
```
Robbepop:
```suggestion
                return Ok(())
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+            };
+
+            match self.token_approvals.remove(id) {
+                Some(_res) => Ok(()),
+                None => Err(Error::CanNotRemove)
+            }
+        }
+
+        fn remove_token_from(&mut self, from: &AccountId, id: &TokenId) -> Result<(), Error> {
+            if !self.exists(id){
```
Robbepop:
```suggestion
            if !self.exists(id) {
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                return Err(Error::TokenNotFound);
```
Robbepop:
```suggestion
                return Err(Error::TokenNotFound)
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+
+            self.decrease_counter_of(from)?;
+            self.token_owner
+                .remove(id)
+                .ok_or(Error::CanNotRemove)?;
+            Ok(())
+        }
+
+        fn add_token_to(&mut self, to: &AccountId, id: &TokenId) -> Result<(), Error> {
+            if self.exists(id){
```
Robbepop:
```suggestion
            if self.exists(id) {
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                 return Err(Error::TokenExists)
+            };
+
+            self.increase_counter_of(to)?;
+            if !self.token_owner.insert(*id, *to).is_none() {
+                return Err(Error::CanNotInsert)
+            };
```
Robbepop:
```suggestion
            }
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+            Ok(())
+        }
+
+        fn mint(&mut self, env: &EnvHandler, to: &AccountId, id: &TokenId) -> Result<(), Error> {
+            if *to == AccountId::from([0x0; 32]){
```
Robbepop:
```suggestion
            if *to == AccountId::from([0x0; 32]) {
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                return Err(Error::AccountZeroNotAllowed)
+            };
+            if self.exists(id){
```
Robbepop:
```suggestion
            if self.exists(id) {
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                return Err(Error::TokenExists)
+            };
+
+            if !self.token_owner.insert(*id, *to).is_none() {
+                return Err(Error::CanNotInsert)
+            };
```
Robbepop:
```suggestion
            }
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+            Ok(())
+        }
+
+        fn mint(&mut self, env: &EnvHandler, to: &AccountId, id: &TokenId) -> Result<(), Error> {
+            if *to == AccountId::from([0x0; 32]){
+                return Err(Error::AccountZeroNotAllowed)
+            };
+            if self.exists(id){
+                return Err(Error::TokenExists)
+            };
```
Robbepop:
```suggestion
            }
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                from: AccountId::from([0x0; 32]),
+                to: *to,
+                id: *id,
+            });
+            Ok(())
+        }
+
+        fn burn(&mut self, env: &EnvHandler, from: &AccountId, id: &TokenId)-> Result<(), Error> {
+            let caller = env.caller();
+            if !self.exists(id){
```
Robbepop:
```suggestion
            if !self.exists(id) {
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+            Ok(())
+        }
+
+        fn increase_counter_of(&mut self, of: &AccountId) -> Result<(), Error> {
+            if self.balance_of(of) > 0 {
+                let count = self.owned_tokens_count
+                    .get_mut(of)
+                    .ok_or(Error::CanNotGetCounter)?;
+                *count += 1;
+                return Ok(());
```
Robbepop:
```suggestion
                return Ok(())
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+            } else{
```
Robbepop:
```suggestion
            } else {
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                match self.owned_tokens_count.insert(*of, 1) {
+                    Some(_) => return Err(Error::CanNotInsert),
```
Robbepop:
```suggestion
                    Some(_) => Err(Error::CanNotInsert),
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                    None => return Ok(())
```
Robbepop:
```suggestion
                    None => Ok(()),
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                };
```
Robbepop:
```suggestion
                }
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+            };
```
Robbepop:
```suggestion
            }
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+                .ok_or(Error::CanNotGetCounter)?;
+            *count -= 1;
+            Ok(())
+        }
+
+        fn balance_of(&self, of: &AccountId) -> u32 {
+            *self.owned_tokens_count.get(of).unwrap_or(&0u32)
+        }
+
+        fn owner_of(&self, id: &TokenId, ) -> AccountId {
```
Robbepop:
```suggestion
        fn owner_of(&self, id: &TokenId) -> AccountId {
```

examples/lang/erc721/src/lib.rs
Robbepop:
```suggestion
#[derive(Copy, Clone, Encode, Decode, Debug, PartialEq, Eq)]
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+impl Error {
+    fn to_u32(self) -> u32 {
+        self as u32
+    }
+}
+
+contract! {
+    #![env = DefaultSrmlTypes]
+
+    // Event deposited when a token transfer occurs
```
Robbepop:
```suggestion
    /// Event deposited when a token transfer occurs
```

examples/lang/erc721/src/lib.rs
```diff
@@ -0,0 +1,453 @@
...
+    event Transfer {
+        from: AccountId,
+        to: AccountId,
+        id: TokenId,
+    }
+
+    // Event deposited when an approval occurs
```
Robbepop:
```suggestion
    /// Event deposited when an approval occurs
```

Issue #15, opened 2019-01-15
Implement storage::BinaryHeap


Robbepop:
The pDSL is currently missing a heap data structure that enables to store ranked entities.

## Motivation

Since sorting is very compute intense (O(n*log(n)) it is not a good idea to sort even medium sized sets of data in a smart contract. Instead one should refer to using a data structure such as the binary heap.

## Design Goals

The design should mirror Rust's `BinaryHeap` and should in that sense also be a MAX-heap.
However, since smart contracts are very dependent on the best possible performance one might think about having a MIN- and a MAX- heap. This issue serves as meeting point for discussions around this topic.

## Proposed API

To mirror Rust's `BinaryHeap` as closely as possible we want the following API:

```rust
impl<T> BinaryHeap<T> {
    /// Returns the length of the heap.
    fn len(&self) -> u32;

    /// Returns `true` if the heap is empty.
    fn is_empty(&self) -> bool;

    /// Returns the greatest item if not empty.
    fn peek(&self) -> Option<&T>;

    /// Mutates the greatest item if not empty and returns a reference to the result.
    fn mutate_peek(&mut self) -> Option<&T>;

    /// Removes the greatest item from the heap if not empty.
    fn pop(&mut self) -> Option<T>;

    /// Pushes an item onto the heap.
    fn push(&mut self, val: T);

    /// Returns an iterator over all items of the heap.
    fn iter(&self) -> impl Iterator<Item = &T>;
```

Robbepop:
With the `Flush` inplace we can get even closer to the original Std interface.

```rust
impl<T> BinaryHeap<T> {
    /// Returns the length of the heap.
    fn len(&self) -> u32;

    /// Returns `true` if the heap is empty.
    fn is_empty(&self) -> bool;

    /// Returns the greatest item if not empty.
    fn peek(&self) -> Option<&T>;

    /// Mutates the greatest item if not empty and returns a reference to the result.
    fn peek_mut(&mut self) -> Option<&mut T>;

    /// Removes the greatest item from the heap if not empty.
    fn pop(&mut self) -> Option<T>;

    /// Pushes an item onto the heap.
    fn push(&mut self, val: T);

    /// Returns an iterator over all items of the heap.
    fn iter(&self) -> impl Iterator<Item = &T>;
}
```

An enhancement of `Stash` could use an efficient version of a priority queue (binary heap) as well as some storage allocator implementations.

PR #206, opened 2019-10-11
Implement Heap storage primitive

cmichi:
Closes https://github.com/paritytech/ink/issues/15.

I named it `Heap` instead of `BinaryHeap`, since it's implemented as a ternay heap and though it doesn't make a difference to the API it might be confusing otherwise.


cmichi:
The PR has been updated and is ready for review.

For the moment I left `AccessWraper` in the `BinaryHeap` module, it's quite general though and could easily be moved to e.g. `storage/chunk/` as a general abstraction.

core/src/storage/collections/binary_heap/access_wrapper.rs
```diff
@@ -0,0 +1,117 @@
...
+};
+
+// Number of values stored in each entry of the `SyncChunk`.
+const COUNT: u32 = 3;
+
+#[derive(Copy, Clone, Debug, Encode, Decode)]
+pub struct Group<T> ([Option<T>; COUNT as usize]);
+
+#[derive(Debug, Encode, Decode)]
+pub struct AccessWrapper<T> (SyncChunk<Group<T>>);
```
Robbepop: Could we maybe rename this to something less generic? Maybe `DuplexSyncChunk`.
Also in ink! we generally prefer speaking field names with normal structs towards tuple-structs usage.
However, I think this use case is okay since it is really just a wrapper.

core/src/storage/collections/binary_heap/access_wrapper.rs
```diff
@@ -0,0 +1,117 @@
...
+
+impl<T> Flush for AccessWrapper<T>
+where
+    T: Flush,
+    SyncChunk<Group<T>>: Flush,
```
Robbepop: I think only `SyncChunk<Group<T>>: Flush` is necessary here.

core/src/storage/collections/binary_heap/access_wrapper.rs
```diff
@@ -0,0 +1,117 @@
...
+{
+    fn flush(&mut self) {
+        self.0.flush();
+    }
+}
+
+impl<T> AccessWrapper<T>
+where
+    T: scale::Encode + scale::Decode + Copy + Clone,
```
Robbepop: So this whole thing only works for `Copy + Clone` types? That's rather restrictive in my opinion.
In general we even want complex storage types to be stored within any other storage data structure.

core/src/storage/collections/binary_heap/impls.rs
```diff
@@ -0,0 +1,461 @@
...
+/// The heap depends on `Ord` and is a max-heap by default. In order to
+/// make it a min-heap implement the `Ord` trait explicitly on the type
+/// which is stored in the heap.
+///
+/// Provides `O(log(n))` push and pop operations.
+#[derive(Debug)]
+#[cfg_attr(feature = "ink-generate-abi", derive(Metadata))]
+pub struct BinaryHeap<T> {
+    /// Stores densely packed general heap information.
+    header: storage::Value<BinaryHeapHeader>,
```
Robbepop: Do we need the header here for just one element or is it expected to have more header elements in the future?

core/src/storage/collections/binary_heap/impls.rs
```diff
@@ -0,0 +1,461 @@
...
+pub struct Values<'a, T> {
+    /// The underlying iterator.
+    iter: Iter<'a, T>,
+}
+
+impl<'a, T> Values<'a, T> {
+    /// Creates a new iterator for the given storage heap.
+    pub(crate) fn new(heap: &'a BinaryHeap<T>) -> Self
+    where
+        T: scale::Codec + Ord + Copy,
```
Robbepop: Why `Copy` ?

core/src/storage/collections/binary_heap/impls.rs
```diff
@@ -0,0 +1,461 @@
...
+    {
+        Self { iter: heap.iter() }
+    }
+}
+
+impl<T> Flush for BinaryHeap<T>
+where
+    T: Ord + Encode + Flush + Copy,
```
Robbepop: We generally should get rid of all `Copy` restrictions.

core/src/storage/collections/binary_heap/impls.rs
cmichi: Good point, I removed it now and replaced it with `len` (I don't expect more fields).

core/src/storage/collections/binary_heap/impls.rs
```diff
@@ -0,0 +1,448 @@
...
+/// which is stored in the heap.
+///
+/// Provides `O(log(n))` push and pop operations.
+#[derive(Debug)]
+#[cfg_attr(feature = "ink-generate-abi", derive(Metadata))]
+pub struct BinaryHeap<T> {
+    /// The number of nodes stored in the heap.
+    len: storage::Value<u32>,
+    /// The nodes of the heap.
+    entries: DuplexSyncChunk<T>,
```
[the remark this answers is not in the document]
cmichi: Yeah, so the layout is currently not ideal. This was just the lowest hanging fruit ‒ taking the "array" and splitting it into groups to save together without further optimization of grouping.

For [this binary max heap](https://en.wikipedia.org/wiki/File:Max-Heap.svg) the layout is currently as follows (three nodes in each group):
```
Group 0: [100, 19, 36]
Group 1: [17, 3, 25]
Group 2: [1, 2, 7]
```
(I can also add a test and comment for this once we know which direction we're going.)

Hmm, maybe it would be an idea to always keep the root in it's own group 0 and then batch pairs of two nodes together. But really, it's all just premature optimization. Do you have any thoughts?

(Btw, the `BTreeMap` algorithm will do this optimization for reads a lot better, since that data structure inherently is suited for minimizing reads.)

core/src/storage/collections/binary_heap/duplex_sync_chunk.rs
```diff
@@ -18,6 +18,18 @@
 //! of values in one cell (instead of the `SyncChunk` behavior of storing
 //! one value per cell). The intention is to reduce expensive fetch
 //! operations from storage.
+//!
+//! **NOTE** This wrapper is geared explicitly towards a binary tree
+//! structure -- the value which is stored at index `0` (the root) will
+//! always be stored in its own group with no other values in it. The
+//! intention is to be able to store child nodes paired together
+//! in a group.
```
Robbepop: "... because for query operations you have to access both elements anyways. This allows to skip one expensive read for every accessed pair."

^--- The above is the rational why we actually do this.

core/src/storage/collections/binary_heap/tests.rs
```diff
@@ -0,0 +1,347 @@
...
+    fn cmp(&self, other: &V) -> Ordering {
+        other.0.cmp(&self.0)
+    }
+}
+
+impl PartialOrd for V {
+    fn partial_cmp(&self, other: &V) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
Robbepop: Why no `#[ðerive(PartialOrd, Ord)]` ?

core/src/storage/collections/binary_heap/impls.rs
```diff
@@ -0,0 +1,448 @@
...
+    where
+        T: scale::Codec + Ord,
+    {
+        Self { iter: heap.iter() }
+    }
+}
+
+impl<T> Flush for BinaryHeap<T>
+where
+    T: Ord + Encode + Flush,
```
Robbepop: Why `Ord` + `Encode` ?

core/src/storage/collections/binary_heap/impls.rs
```diff
@@ -0,0 +1,448 @@
...
+    DuplexSyncChunk<T>: Flush,
+{
+    fn flush(&mut self) {
+        self.len.flush();
+        self.entries.flush();
+    }
+}
+
+#[cfg(feature = "ink-generate-abi")]
+impl<T: Ord> HasLayout for BinaryHeap<T>
```
Robbepop: Why is `Ord` required here?

core/src/storage/collections/binary_heap/impls.rs
```diff
@@ -0,0 +1,448 @@
...
+                LayoutField::of("entries", &self.entries),
+            ],
+        )
+        .into()
+    }
+}
+
+impl<'a, T> Iterator for Values<'a, T>
+where
+    T: Codec + Ord,
```
Robbepop: Please remove all trait bounds that are locally not necessary.

core/src/storage/collections/binary_heap/duplex_sync_chunk.rs
```diff
@@ -0,0 +1,194 @@
...
+};
+use scale::{
+    Decode,
+    Encode,
+};
+
+// Number of values stored in each entry of the `SyncChunk`.
+// Note that the first group (at index `0`) will only ever
+// contain one value.
+const COUNT: u32 = 2;
```
Robbepop: To make the implementation more generic and more usable with respect to other components that might need other values for `COUNT` we could make `Group` generic over array-like instances `[(); N]` where `N` reflects `COUNT` and by offering a trait `Count { const N: usize; }` that is auto implemented for `[(); 0..=32]` we could go a long way I guess. But then again I am not sure if Rust's current const-eval capabilities are strong enough for such a use-case. We would need to research. Also I think this is better suited in a follow-up PR. :)

core/src/storage/collections/binary_heap/impls.rs
cmichi: :+1: I removed some other bounds, these ones are necessary though because the `Iterator` impl needs access to e.g. `heap.get()` which requires `Ord` when comparing nodes against each other and `Codec` when interacting with the `DuplexSyncChunk` impl (which requires `Codec`).

core/src/storage/collections/binary_heap/duplex_sync_chunk.rs
cmichi: That's a really good point, I was already wondering how we could make it generic, while still having the `COUNT` determined at compile time. I'll try to take that into account when reusing this module for `BTreeMap`.

PR #209, opened 2019-10-14
[cli] add generate-abi command

ascjones:
~~cargo contract build --abi will execute the abi generation post build.~~

~~Alternatively we could have this as a separate command - but I decided to add it as a flag to `build`, since a developer might want to regenerate the abi as part of the build if the contract interface has changed.~~

**Edit** changed it to a separate command: `cargo contract generate-abi`

### DIFF
--- a/cli/src/cmd/abi.rs
+++ b/cli/src/cmd/abi.rs
@@ -31,7 +31,7 @@ pub(crate) fn execute_generate_abi() -> Result<String> {
         "--package",
         "abi-gen",
         "--verbose",
-    ])?;
+    ], None)?;
 
     let metadata = MetadataCommand::new().exec()?;
     let mut abi_path = metadata.target_directory.clone();
--- a/cli/src/cmd/abi.rs
+++ b/cli/src/cmd/abi.rs
@@ -18,20 +18,19 @@ use crate::cmd::{
     Result,
 };
 use cargo_metadata::MetadataCommand;
-
-
+use std::path::PathBuf;
 
 /// Executes build of the smart-contract which produces a wasm binary that is ready for deploying.
 ///
 /// It does so by invoking build by cargo and then post processing the final binary.
-pub(crate) fn execute_generate_abi() -> Result<String> {
+pub(crate) fn execute_generate_abi(dir: Option<&PathBuf>) -> Result<String> {
     println!(" Generating abi");
 
     super::exec_cargo("run", &[
         "--package",
         "abi-gen",
         "--verbose",
-    ], None)?;
+    ], dir)?;
 
     let metadata = MetadataCommand::new().exec()?;
     let mut abi_path = metadata.target_directory.clone();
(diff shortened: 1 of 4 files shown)


cli/src/main.rs
```diff
@@ -87,7 +87,11 @@ enum Command {
     },
     /// Builds the smart contract.
     #[structopt(name = "build")]
-    Build {},
+    Build {
+        /// Generate the contract abi artifacts
+        #[structopt(long = "abi")]
+        generate_abi: bool,
+    },
```
cmichi: This is the output I get for `--help`:
```
$ cargo contract build --help
…
USAGE:
    cargo contract build

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information
```
Could you make the `--abi` flag show up here?

cli/src/cmd/build.rs
```diff
@@ -88,20 +88,10 @@ pub fn collect_crate_metadata() -> Result<CrateMetadata> {
...
 
-/// Invokes `cargo build` in the current directory.
-///
-/// Currently it assumes that user wants to use `+nightly`.
-fn build_cargo_project() -> Result<()> {
-    // We also assume that the user uses +nightly.
+fn exec_cargo(command: &str, args: &[&'static str]) -> Result<()> {
     let output = Command::new("cargo")
-        .args(&[
-            "+nightly",
```
riusricardo: I'm guessing that removing +nightly was intentional.
Should we now use `cargo +nightly contract build --abi` as the correct build command?

ascjones: Actually that wasn't intentional, I've added it back. Not sure whether running `cargo +nightly contract build --abi` would mean the invoked `cargo` commands also receive the flag.

The handling of this is improved in https://github.com/paritytech/ink/pull/197, which I hope will be merged first.

riusricardo: If you run `cargo +nightly contract build --abi` it compiles correctly. I don't use nightly as default and it only builds if I use the `+nightly` flag. The final question will be if we hide it as part of the build command or it should be handled explicitly?

Robbepop:
> 
> 
> `cargo contract build --abi` will execute the abi generation post build.
> 
> Alternatively we could have this as a separate command - but I decided to add it as a flag to `build`, since a developer might want to regenerate the abi as part of the build if the contract interface has changed.

Please implement this as seperate command `generate-abi` since the rational here does not apply because the crate has to be compiled with different crate features anyway. Also people should not be confused by some technical details that ABI generation actually requires building the contract. Strictly speaking this is an implementation detail that we should hide.

ascjones:
Note that test failures are fixed in separate PR #197

Robbepop:
It would be nice if this command had a test.

ascjones:
> All in all LGTM but misses at least one small test.

Agree. I'll add a test once #197 is merged.

Robbepop:
> 
> 
> > All in all LGTM but misses at least one small test.
> 
> Agree. I'll add a test once #197 is merged.

merged #197

cli/src/cmd/abi.rs
```diff
@@ -0,0 +1,70 @@
...
+/// Executes build of the smart-contract which produces a wasm binary that is ready for deploying.
+///
+/// It does so by invoking build by cargo and then post processing the final binary.
+pub(crate) fn execute_generate_abi(dir: Option<&PathBuf>) -> Result<String> {
+    println!(" Generating abi");
+
+    super::exec_cargo("run", &[
+        "--package",
+        "abi-gen",
+        "--verbose",
```
Robbepop: we forgot to add `--no-default-features` here. otherwise i takes `std` feature which causes to output the off-chain test environment instead of the SRML environment. this isn't a big problem but could be improved. also maybe we want to provide `--release` since I have the feeling that `--release` is faster counterintuitively.

ascjones: https://github.com/paritytech/ink/pull/234

PR #211, opened 2019-10-15
Fix format and clippy

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

### DIFF
--- a/abi/src/specs.rs
+++ b/abi/src/specs.rs
@@ -15,7 +15,10 @@
 // along with ink!.  If not, see <http://www.gnu.org/licenses/>.
 
 use core::marker::PhantomData;
-use serde::Serialize;
+use serde::{
+    Serialize,
+    Serializer,
+};
 use type_metadata::{
     form::{
         CompactForm,
@@ -190,7 +193,8 @@ pub struct ConstructorSpec<F: Form = MetaForm> {
     /// The name of the message.
     name: F::String,
     /// The selector hash of the message.
-    selector: u32,
+    #[serde(serialize_with = "serialize_selector")]
+    selector: [u8; 4],
     /// The parameters of the deploy handler.
     args: Vec<MessageParamSpec<F>>,
     /// The deploy handler documentation.
@@ -234,7 +238,7 @@ impl ConstructorSpec {
         ConstructorSpecBuilder {
             spec: Self {
                 name,
-                selector: 0,
+                selector: [0u8; 4],
                 args: Vec::new(),
                 docs: Vec::new(),
             },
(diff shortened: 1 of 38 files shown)


abi/src/specs.rs
```diff
@@ -169,7 +169,7 @@ impl ContractSpecBuilder<Valid> {
 
 impl ContractSpec {
     /// Creates a new contract specification.
-    pub fn new(name: <MetaForm as Form>::String) -> ContractSpecBuilder {
+    pub fn builder(name: <MetaForm as Form>::String) -> ContractSpecBuilder {
```
Robbepop: I am against naming this `builder` even though this would fix a `clippy` lint.
However, we should sometimes question certain clippy lints and also we could come up with better names.
For example `builder` is not suitable but maybe `build` was, but then again people could be confused since we can also `build` (aka compile) smart contracts.

core/src/env2/call.rs
```diff
@@ -59,11 +59,13 @@ impl<'a> From<&'a [u8]> for Selector {
...
+    #[allow(clippy::should_implement_trait)]
     pub fn from_str(name: &str) -> Self {
         From::from(name.as_bytes())
     }
 
     /// Returns the underlying bytes of the selector.
+    #[allow(clippy::trivially_copy_pass_by_ref)]
     pub fn to_bytes(&self) -> [u8; 4] {
         self.bytes
     }
```
Robbepop: We won't implement the associated traits here because traits cannot be `const fn` which is a major pita for the `Selector` type which is going to be used in `const fn` contexts.

core/src/storage/alloc/bump_alloc.rs
```diff
@@ -39,7 +39,7 @@ pub struct BumpAlloc {
 impl BumpAlloc {
     /// Creates a new forward allocator for the given raw parts.
     ///
-    /// # Note
+    /// # Safety
     ///
     /// Do not use this directly!
     /// This is meant to be used by pDSL internals only.
```
Robbepop: Highly outdated docs. You are right that we should have used the `Safety` section here, however, the safety section is normally used to communicate what to take care and look for if you are about to call it so we should eventually do that here as well.

abi/src/specs.rs
Robbepop: Let's allow to name this method `new` for now please.

Robbepop:
@koushiro i would really like to merge this once the grumbles have been fixed.

Is this going anywhere soonish(TM)?

abi/src/specs.rs
```diff
@@ -356,6 +356,7 @@ impl MessageSpec {
 /// Some of the fields are guarded by a type-state pattern to
 /// fail at compile-time instead of at run-time. This is useful
 /// to better debug code-gen macros.
+#[allow(clippy::type_complexity)]
 pub struct MessageSpecBuilder<Selector, Mutates, Returns> {
     spec: MessageSpec,
     marker: PhantomData<fn() -> (Selector, Mutates, Returns)>,
```
Robbepop: Clippy really thinks this type is too complex?! Well, ... maybe because of the 3 type params.

PR #217, opened 2019-10-25
Fix CI cargo workspace issue

TriplEight:


ascjones:
I was playing with [`cargo hack`](https://github.com/taiki-e/cargo-hack) and seems like it would solve the problem of building all packages in the workspace e.g. `cargo hack check --all`

Robbepop:
> I was playing with [`cargo hack`](https://github.com/taiki-e/cargo-hack) and seems like it would solve the problem of building all packages in the workspace e.g. `cargo hack check --all`

I do not like the idea to depend on something 3rd-party like `cargo-hack` in order to work around bugs of nightly cargo. But I agree that the current situation is a total mess.

Robbepop:
> I was playing with [`cargo hack`](https://github.com/taiki-e/cargo-hack) and seems like it would solve the problem of building all packages in the workspace e.g. `cargo hack check --all`

I do not like the idea to depend on something 3rd-party like `cargo-hack` in order to work around bugs of nightly cargo. But I agree that the current situation is a total mess.

Robbepop:
Since this solution with `--package X` doesn't work at all and since I have already another PR open that tackles the same underlying problem I am considering closing this issue. @TriplEight ?

Robbepop:
Closed because outdated. Also no proper solution. Other PR fixed similar problem. Closed.

PR #225, opened 2019-11-09
improve some diagnostics for ink! lang

Robbepop:


Robbepop:
> LGTM!
> 
> Btw: What is the minimal set of ink!-annotated elements for a valid smart contract? Is it required to have at least one constructor? What about the rest (storage, messages, …) ‒ are they optional?

At the moment every valid ink! contract is required to have at least one `#[ink(constructor)]` and at least one `#[ink(message)]` as well as exactly one `#[ink(storage)]` but that's it.

cmichi:
> At the moment every valid ink! contract is required to have at least one `#[ink(constructor)]` and at least one `#[ink(message)]` as well as exactly one `#[ink(storage)]` but that's it.

Ok, correct me if I'm wrong, but I think at the moment we don't have a test for a missing `#[ink(storage)]`?

Robbepop:
> > At the moment every valid ink! contract is required to have at least one `#[ink(constructor)]` and at least one `#[ink(message)]` as well as exactly one `#[ink(storage)]` but that's it.
> 
> Ok, correct me if I'm wrong, but I think at the moment we don't have a test for a missing `#[ink(storage)]`?

Hmm, ... just tested for this scenario and you seem to be right. I thought I implemented this check but also a test for this is missing currently. So if we actually remove the `#[ink(storage)]` from a storage struct at the moment we receive a pretty arcane and confusing error message:

```
 --> $DIR/14-missing-storage-struct.rs:5:1
  |
5 | #[ink::contract(version = "0.1.0")]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: there must be at least 2 conflicting storages; qed
```

Thank you for pointing that out!

PR #227, opened 2019-11-10
ink! 2.0 follow-up: Remove the old ink! 1.0

Robbepop:
This PR removes all the deprecated ink! 1.0 components and sub crates and modules that are no longer in use after the transition towards ink! 2.0.

## Blocked-by

- [x] Updated ink! tutorial & workshop
- [x] Blog post about ink! 2.0
- [x] Blog post for transitioning from ink! 1.0 to ink! 2.0

## Tasks

- [x] Remove old `ink_lang` 1.0 syntax & codegen
- [x] Remove old `ink_model` 1.0 dispatching layer
- [ ] Remove old `ink_core::env` 1.0 environment handler
    Doing so exposed some problems with `RawCell` and `RawChunk` that are still directly accessing the deprecated `ink_core::env::api` methods. These calls should generally be replaced. Also we might want to remove `RawCell`, `TypedCell`, `RawChunk` and `TypedChunk` since their abstraction purpose isn't as useful as we thought when we created them.
    - [ ] Some parts of these transition problems can be fixed by https://github.com/paritytech/ink/pulls. So this PR is partly blocked by it.
- [x] Remove old ink! 1.0 example contracts


riusricardo:
Is this PR still blocked?

Robbepop:
> 
> 
> Is this PR still blocked?

The PR is blocked for 2 reasons:

1. We want to make sure that ink! 2.0 supports all features ink! 1.0 supported which is currently not the case due to limitations of the off-chain test environment.
2. There are some technical aspects that we have to solve before we can get rid of ink! 1.0.

Robbepop:
Superseeded by https://github.com/paritytech/ink/pull/312. Closed.

PR #229, opened 2019-11-11
Update to derive_more v0.99

Robbepop:
Simply updates all dependencies from within ink! to `derive_more` crate to the newest 0.99 release.

The bigger advantage of having updated `derive_more` dependency is that the old version was the last dependency that was still depending on the non-`1.0` syn, quote and proc_macro2 crates.

### DIFF
--- a/core/src/env2/error.rs
+++ b/core/src/env2/error.rs
@@ -40,13 +40,13 @@ pub enum Error {
     Call(CallError),
     Create(CreateError),
     Codec(scale::Error),
-    #[from(ignore)] InvalidStorageKey,
-    #[from(ignore)] InvalidStorageRead,
-    #[from(ignore)] InvalidContractCall,
-    #[from(ignore)] InvalidContractCallReturn,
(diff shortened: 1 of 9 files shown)


core/src/env2/error.rs
```diff
@@ -40,13 +40,13 @@ pub enum Error {
...
-    InvalidContractInstantiation,
-    InvalidContractInstantiationReturn,
-    InvalidRandomSeed,
+    #[from(ignore)] InvalidStorageKey,
+    #[from(ignore)] InvalidStorageRead,
+    #[from(ignore)] InvalidContractCall,
+    #[from(ignore)] InvalidContractCallReturn,
+    #[from(ignore)] InvalidContractInstantiation,
+    #[from(ignore)] InvalidContractInstantiationReturn,
+    #[from(ignore)] InvalidRandomSeed,
```
Robbepop: Note: Already opened [an issue](https://github.com/JelteF/derive_more/issues/105) for this abomination.

PR #230, opened 2019-11-11
2.0 Syntax Release Notes

shawntabrizi:


RELEASES.md
```diff
@@ -0,0 +1,250 @@
...
+		self.env().emit_event(Transferred {
+			from: None,
+			to: Some(caller),
+			amount: initial_supply,
+		});
+	}
+//-snip-
+```
+
+You can name them whatever you want.
```
cmichi: The prefix `__ink` for functions is reserved, if a constructor is named that way the contract won't compile.

RELEASES.md
```diff
@@ -0,0 +1,250 @@
...
+
+EnvHandler is no longer exposed to the user and instead the environment is now always accessed via self.env() is all ink! constructors, messages and private methods.
+
+It is possible to enable the dynamic environment that allows for dynamic allocations by specifying dynamic_allocations = true in the parameters of the ink! header. This is disabled by default.
+
+Testing contracts off-chain is done by cargo test and users can simply use the standard routines of creating unit test modules within the ink! module itself (#[cfg(test)] mod tests { use super::*; #[test] fn my_test() { ... } }). Test instances of contracts can be created with MyContract::my_constructor(a, b) if the contract has an #[ink(storage)] struct MyContract { ... } and an #[ink(constructor)] fn my_constructor(a: A, b: B) { ... }. Messages can simply be called on the returned instance as if MyContract::my_constructor returns a Self instance.
+
+The off-chain test environment has lost a bit of power compared to the old ink! lang. It currently is no longer possible to query and set special test data about the environment but these restirctions are going to be lifted in the future.
```
cmichi:
```suggestion
The off-chain test environment has lost a bit of power compared to the old ink! lang. It currently is no longer possible to query and set special test data about the environment but these restrictions are going to be lifted in the future.
```

cmichi:
Thanks for writing this, Shawn! Can we add something like this somewhere?

> Every valid ink! contract is required to have at least one `#[ink(constructor)]`, at least one `#[ink(message)]` and exactly one `#[ink(storage)]` attribute.

shawntabrizi:
One idea was to share something like this:
https://github.com/paritytech/ink/pull/230/commits/41899f2f0eafc33ce38e4ff73cd9bcfb246162af

Which is a diff of a contract before and after. We could add comments inline to help explain what has changed.

The other idea is this MD file:

https://github.com/paritytech/ink/blob/fa0c833bd1fb096b4552655615a2e0183e104ea2/RELEASES.md

Which is very much still in progress, but the table is a little large to be rendered.

Last idea I have not executed on is just everything in a single column with clear headers.

RELEASES.md
```diff
@@ -0,0 +1,343 @@
...
+}
+```
+
+Test instances of contracts can be created with something like:
+
+```rust
+let contract = MyContract::my_constructor(a, b);
+```
+
+Messages can simply be called on the returned instance as if MyContract::my_constructor returns a
```
cmichi:
```suggestion
Messages can simply be called on the returned instance as if `MyContract::my_constructor` returns a
```

RELEASES.md
```diff
@@ -0,0 +1,414 @@
...
+
+Follow the instructions below to understand how to migrate your ink! 1.0 contracts to this new ink!
+2.0 syntax.
+
+## Declaring a Contract
+
+The fundamental change with the new ink! syntax is how we declare a new contract.
+
+We used to wrap the whole ink! contract into a `contract!` macro. At that point, all syntax within
+the macro could be custom, and mistakenly we used that to define custom syntax that made our code
```
cmichi: Instead of "mistakenly" I would say something like "in our first iteration". I mean, we tried that approach intentionally and it turned out that there are a number of issues with it.

RELEASES.md
```diff
@@ -0,0 +1,414 @@
...
+delegation and ink! messages can easily call other ink! messages.
+
+Given another ink! contract like `mod Accumulator { ... }`, we can call any of its functions just
+like we would expect:
+
+```rust
+use accumulator::Accumulator;
+let result = Accumulator::my_function()
+```
+
```
cmichi:
```suggestion

For more details (e.g. on how deployment and instantiation) behave take a look at our delegator example.

```

RELEASES.md
```diff
@@ -0,0 +1,414 @@
...
+Test instances of contracts can be created with something like:
+
+```rust
+let contract = MyContract::my_constructor(a, b);
+```
+
+Messages can simply be called on the returned instance as if `MyContract::my_constructor` returns a
+Self instance.
```
cmichi:
```suggestion
`Self` instance.
```

RELEASES.md
shawntabrizi: Agreed and updated.

RELEASES.md
```diff
@@ -0,0 +1,414 @@
...
+
+## ink!-less Implementations
+
+It is also possible to annotate an entire impl blocks with:
+
+```rust
+#[ink(impl)]
+impl Contract {
+    fn internal_function(&self) {
+        self.env().emit_event(Hi);
```
Robbepop: It might be confusing to have `self.env().emit_event(Hi);` here

RELEASES.md
```diff
@@ -0,0 +1,414 @@
...
+```
+
+</td>
+</tr>
+</table>
+
+## Contract Tests
+
+Testing contracts off-chain is done by `cargo test` and users can simply use the standard routines
+of creating unit test modules within the ink! module itself:
```
Robbepop: Users are free to declare their unit-tests wherever they want. Maybe we shall make that more clear at this point.

RELEASES.md
```diff
@@ -0,0 +1,414 @@
...
+    .create()
+    .expect("failed at instantiating the accumulator contract");
+```
+
+</td>
+<td>
+
+```rust
+let accumulator = Accumulator::new(init_value)
+    .value(total_balance / 4)
```
Robbepop: Maybe you want to show that setting `.gas_limit(..)` is possible as well.

RELEASES.md
```diff
@@ -0,0 +1,414 @@
...
+
+It is now possible to call ink! messages and ink! constructors. So ink! constructors allow
+delegation and ink! messages can easily call other ink! messages.
+
+Given another ink! contract like `mod Accumulator { ... }`, we can call any of its functions just
+like we would expect:
+
+```rust
+use accumulator::Accumulator;
+let result = Accumulator::my_function()
```
Robbepop: This is not really correct syntax unless `my_function` was a `my_constructor` but then again this is only possible in the new ink! 2.0.

RELEASES.md
```diff
@@ -0,0 +1,414 @@
...
+#[ink::contract(version = "0.1.0", types = MyCustomTypes)]
+```
+
+By default, we use `DefaultSrmlTypes`, so you don't need to define anything unless you plan to use
+custom types.
+
+### Dynamic Allocation
+
+It is possible to enable the dynamic environment that allows for dynamic allocations by specifying
+`dynamic_allocations = true` in the parameters of the ink! header. This is disabled by default.
```
Robbepop: We should make it clear that dynamic environment is still under research and not yet stable. Will possibly change in the future.

RELEASES.md
```diff
@@ -0,0 +1,414 @@
...
+contract! {
+    ...
+}
+```
+
+</td>
+<td>
+
+```rust
+#[ink::contract(version = "0.1.0")]
```
Robbepop: Technically it is `ink_lang2::contract` but since we import `ink_lang2` using `use ink_lang2 as ink;` above this works.

shawntabrizi: Will it always be `ink_lang2`? Or will you rename it once you refactor syntax 1 out?

RELEASES.md
shawntabrizi: I changed the line to "within the ink! project."

RELEASES.md
shawntabrizi: It is technically not wrong though right? There could be a defined event `Hi` elsewhere?

Robbepop: Yeah but I think this `Hi` is rather confusing to beginners.

RELEASES.md
Robbepop: We might want to rename it, yes. I was even thinking about introducing a shim-crate `ink` that covers all other crates so that contract authors no longer have to include X different crates but just this single one instead.

Robbepop: Names are seldomly final in ink! :P

shawntabrizi: Well... dunno if you want me to act on the change or not :P

Robbepop: Okay sorry, I think we should include the `use` statement to make it clear what happened on this line. E.g.:
```
use ink_lang2 as ink;

#[ink::contract(version = "0.1.0")]
mod foo {
    ...
}
```

RELEASES.md
```diff
@@ -0,0 +1,455 @@
...
+
+## ink!-less Implementations
+
+It is also possible to annotate an entire impl blocks with:
+
+```rust
+#[ink(impl)]
+impl Contract {
+    fn internal_function(&self) {
+        self.env().emit_event(Hi);
```
shawntabrizi:
```suggestion
        self.env().emit_event(EventName);
```

RELEASES.md
shawntabrizi: Updated to `EventName`

RELEASES.md
Robbepop: Thanks you. That's a bit better.

RELEASES.md
shawntabrizi:
```suggestion
use ink_lang2 as ink;

#[ink::contract(version = "0.1.0")]
```

shawntabrizi: Done :)

README.md
```diff
@@ -46,10 +46,10 @@ Use the scripts provided under `scripts` directory in order to run checks on eit
 For building the example smart contracts found under `examples` you will need to have `cargo-contract` installed.
 
 ```
-cargo install --git https://github.com/paritytech/ink cargo-contract
+cargo install --git https://github.com/paritytech/ink cargo-contract --force
 ```
 
-Add `--force` option to update to the most recent `cargo-contract` version.
+We use the `--force` option to ensure you are updated to the most recent `cargo-contract` version.
```
Robbepop:
```suggestion
We use the `--force` to ensure you are updated to the most recent `cargo-contract` version.
```

Robbepop: Please :D

PR #231, opened 2019-11-11
Update ERC20 to match specification

shawntabrizi:
Some names and functions don't match specification.

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-20.md


shawntabrizi:
@Robbepop It is just a rename and compiles, so I think you can merge whenever.

Robbepop:
> @Robbepop It is just a rename and compiles, so I think you can merge whenever.

Okay then I trust you with this and merge.

PR #236, opened 2019-11-13
Add ./examples README & basic build instructions

Stefie:


README.md
```diff
@@ -51,15 +51,30 @@ cargo install --git https://github.com/paritytech/ink cargo-contract
 
 Add `--force` option to update to the most recent `cargo-contract` version.
 
-Please visit [the documentation](https://substrate.dev/substrate-contracts-workshop/#/0/building-your-contract)
-about building contracts and generating metadata.
+### Build example contract and generate the contracts ABI
 
-### Hello, World! - The Flipper
+To build a single example and generate the contracts wasm file, navigate to the root of the example smart contract and run:
```
Robbepop:
```suggestion
To build a single example and generate the contracts Wasm file, navigate to the root of the example smart contract and run:
```

README.md
```diff
@@ -51,15 +51,30 @@ cargo install --git https://github.com/paritytech/ink cargo-contract
...
+
+`cargo contract build`
+
+To generate the Application Binary Interface (ABI) of the contract, run the following command:
```
Robbepop:
```suggestion
To generate the Application Binary Interface (ABI), or metadata, of the contract, run the following command:
```

README.md
```diff
@@ -51,15 +51,30 @@ cargo install --git https://github.com/paritytech/ink cargo-contract
...
+
+`cargo contract generate-abi`
+
+You should now have a optimized `<contract-name>.wasm` file and a `abi.json` file in the `./target` folder of the contract.
```
Robbepop:
```suggestion
You should now have an optimized `<contract-name>.wasm` file and an `abi.json` file in the `target` folder of the contract.
```

examples/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+```
+cargo install --git https://github.com/paritytech/ink cargo-contract
+```
+
+Add `--force` option to update to the most recent `cargo-contract` version.
+
+## Build example contract and generate the contracts ABI
+
+To build a single example and generate the contracts wasm file, navigate to the root of the smart contract and run the following command:
```
Robbepop:
```suggestion
To build a single example and generate the contracts Wasm file, navigate to the root of the smart contract and run the following command:
```

examples/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+`cargo contract build`
+
+To generate the Application Binary Interface (ABI) of the contract, run the following command:
```
Robbepop:
```suggestion
To generate the Application Binary Interface (ABI), or metadata, of the contract, run the following command:
```

examples/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+`cargo contract generate-abi`
+
+You should now have a optimized `<contract-name>.wasm` file and a `abi.json` file in the `./target` folder of your contract.
```
Robbepop:
```suggestion
You should now have an optimized `<contract-name>.wasm` file and an `abi.json` file in the `target` folder of your contract.
```

README.md
Stefie: I think there are too many commas here, a native speaker should check the punctuation. // cc @shawntabrizi @ascjones

shawntabrizi: @Stefie Robin expressed to be that the term "ABI" is quite a poor description of the file. In general we should refer to it as the contract metadata.

If I were to write the sentence with both I would say:

> To generate the contract metadata (a.k.a. the contract ABI), run the following command:

In this case, I do not write out ABI, cause either they know it already, or they don't care about "a.k.a."

Stefie: Hmm, ok. I copied the wording from the workshop https://substrate.dev/substrate-contracts-workshop/#/0/building-your-contract, though. It should be changed there, too, then.

But I don't fully agree because `ABI` is kind of a fixed term that is used everywhere in the Polkadot-JS repos (https://github.com/search?q=org%3Apolkadot-js+abi&type=Code) whereas `metadata` refers to the substrate metadata exposed via RPC.

I think it's more confusing to use two terms for the same thing in different places (metadata & abi) than to just stick to one term (abi) everywhwere.

Stefie: And besides the fact that it's already an established term in the ecosystem, it actually does make sense...

> In computer software, an application binary interface (ABI) is an interface between two binary program modules; often, one of these modules is a library or operating system facility, and the other is a program that is being run by a user.

> An ABI defines how data structures or computational routines are accessed in machine code, which is a low-level, hardware-dependent format; in contrast, an API defines this access in source code, which is a relatively high-level, hardware-independent, often human-readable format. A common aspect of an ABI is the calling convention, which determines how data is provided as input to or read as output from computational routines

Stefie:
> [#236 (comment)](https://github.com/paritytech/ink/pull/236#discussion_r345774607)
> 
> Follow-up to this comment: This is not correct. [...] We have been using the wrong names for too long time and that's why created even more confusion. I would like to somehow find out of this mess that we have created but not sure how and if we still can.

Ok, I got your point! I actually don't have strong feelings about what term will be used in the end, I only have strong feelings about 2 things:
1. The change shouldn't start in an ./example README file, but in the ink! code itself. The README should just reflect what's implemented in the code and the generated file is literally called `abi.json`. At the moment the term `abi` is used [everywhere in ink!](https://github.com/paritytech/ink/search?p=1&q=abi&unscoped_q=abi), the workshop, the docs  & the polkadot.js API/ecosystem, so it makes sense to explain it in the examples README, until that's not the case anymore.
2. If there's a serious approach to change the term `abi` to `metadata`, it should happen everywhere in the substrate smart contract ecosystem, not only in ink!, because otherwise it could/ will be quite confusing for users.

Robbepop:
>     2\. If there's a serious approach to change the term `abi` to `metadata`, it should happen everywhere in the substrate smart contract ecosystem, not only in ink!, because otherwise it could/ will be quite confusing for users.

Isn't it already called metadata in the Substrate ecosystem?
In general I agree with you but let's just start here.

README.md
Robbepop: https://github.com/paritytech/ink/pull/239

examples/README.md
Stefie: I changed it to the wording suggested by @shawntabrizi for now.

examples/README.md
```diff
@@ -0,0 +1,33 @@
+# ink! code examples
+
+This folder contains a set of example contracts for ink!.
+
+Have a look at the different examples to better understand how to use ink! to build your own Substrate smart contracts.
+
+Besides teaching you how to write smart contracts, this folder is also meant for testing purposes. If you want to contribute to Parity ink!, please run the `check-examples.sh` script in the `./scripts` folder before submitting your PR.
```
Robbepop:
```suggestion
Besides teaching you how to write smart contracts, this folder is also meant for testing purposes. If you want to contribute to Parity ink!, please run the `check-workspace.sh` and `check-examples.sh` script in the `./scripts` folder before submitting your PR.
```

examples/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Preparation
+
+For building the example smart contracts found in this folder you will need to have `cargo-contract` installed.
+
+```
+cargo install --git https://github.com/paritytech/ink cargo-contract
+```
+
+Add `--force` option to update to the most recent `cargo-contract` version.
```
Robbepop:
```suggestion
Use the `--force` to update to the most recent `cargo-contract` version.
```

examples/README.md
Stefie: `check-workspace.sh` doesn't have anything to do with this ./examples folder, that's why I didn't mention it here. `check-examples.sh` on the other hand is using it.

It is already mentioned in the general README , that both files in the ./scripts should be run before committing a PR.

Robbepop: That's true but I was wondering on the whole section as a whole because we shouldn't regard it as standard practice that people start hacking on our examples and file PRs but this paragraph gave me the idea that this might be intended dev behaviour. So we might even want to consider removing this paragraph. I do not mean that devs should not file PRs for obvious improvements to our examples though.

README.md
```diff
@@ -43,7 +43,7 @@ For building the example smart contracts found under `examples` you will need to
 cargo install --git https://github.com/paritytech/ink cargo-contract --force
 ```
 
-We use the `--force` to ensure you are updated to the most recent `cargo-contract` version.
+Use the `--force` option to ensure you are updated to the most recent `cargo-contract` version.
```
Robbepop: Thanks :smiley:

examples/README.md
Stefie: Removed.

examples/README.md
```diff
@@ -0,0 +1,31 @@
...
+This folder contains a set of example contracts for ink!.
+
+Have a look at the different examples to better understand how to use ink! to build your own Substrate smart contracts.
+
+## Preparation
+
+For building the example smart contracts found in this folder you will need to have `cargo-contract` installed.
+
+```
+cargo install --git https://github.com/paritytech/ink cargo-contract
```
shawntabrizi:
```suggestion
cargo install --force --git https://github.com/paritytech/ink cargo-contract
```

examples/README.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+Use the `--force` option to update to the most recent `cargo-contract` version.
```
shawntabrizi:
```suggestion
We use the `--force` to update to the most recent `cargo-contract` version.
```

Stefie: Is that some kind of Star Wars reference?

README.md
```diff
@@ -49,17 +43,32 @@ For building the example smart contracts found under `examples` you will need to
...
```
Robbepop:
```suggestion
Use the `--force` to ensure you are updated to the most recent `cargo-contract` version.
```

PR #238, opened 2019-11-13
Update SCALE codec from 1.0 -> 1.1

Robbepop:


lang2/macro/Cargo.toml
```diff
@@ -16,12 +16,12 @@ include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
...
-scale = { package = "parity-scale-codec", version = "1.0", default-features = false, features = ["derive"] }
+scale = { package = "parity-scale-codec", version = "1.1", default-features = false, features = ["derive"] }
 quote = "1"
-syn = { version = "1", features = ["parsing", "full", "extra-traits"] }
-proc-macro2 = "1"
+syn = { version = "1.0", features = ["parsing", "full", "extra-traits"] }
+proc-macro2 = "1.0"
 heck = "0.3"
-itertools = { version = "0.8", default-features = false }
+itertools = { version = "0.8.1", default-features = false }
```
ascjones: Why add the patch version here?

Robbepop: Good catch! However, I like sticking to newest versions and also it has had no update since Dec 2018. Also there are performance improvements like this: https://github.com/rust-itertools/itertools/commit/521938f5e35fc376b80165bf1530d798e7464989
(We are actually using this function afair.)

PR #255, opened 2019-11-20
Fix erc20 transferring bug

### LINKED ISSUE #254: erc20 token example transferring bug
image

Transferring to yourself `x` balance will actually add `x` balance!
```rust
            let to_balance = self.balance_of_or_zero(&to);  // <- A
            self.balances.insert(from.clone(), from_balance - value); // <-B
            self.balances.insert(to.clone(), to_balance + value); // <-C
```
Should be B A C

kigawas:
Fix #254


kigawas:
How come it can `increase coverage by 51.77%`? :open_mouth:

Robbepop:
The 26% coverage reported previously was a bug.

PR #256, opened 2019-11-20
Add core/derive crate to add derive macro for some traits

Robbepop:
Adds `derive` macros to the `ink_core` crate for some of its traits:

Namely

- `ink_core::storage::Flush`
- `ink_core::storage::alloc::AllocateUsing`
- ~~`ink_core::storage::alloc::Initialize`~~

`ink_core::storage::alloc::Initialize` was found to be too specific to have a proper derive.


core/derive/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+
+pub(crate) fn allocate_using_derive(mut s: synstructure::Structure) -> TokenStream2 {
+    // We cannot implement AllocateUsing on enums because we cannot specify
+    // which variant we are going to use.
+    if let syn::Data::Enum(ref _enum_data) = s.ast().data {
+        panic!("cannot derive AllocateUsing for enums")
+    }
+    s.bind_with(|_| synstructure::BindStyle::Move);
+    s.add_bounds(synstructure::AddBounds::Fields);
+    let body = s.variants()[0].construct(|field, _| {
```
ascjones: Would be good to have a comment here that structs have a single "variant". Or even assert that `s.variants().len() == 0` - just to make it clear what's going on here.

core/derive/tests/compile_tests.rs
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[test]
+fn compile_tests() {
+    let t = trybuild::TestCases::new();
+    t.pass("tests/ui/pass/flush.rs");
+    t.pass("tests/ui/pass/allocate_using.rs");
+    // t.compile_fail("tests/ui/fail/01-constructor-returns.rs");
```
ascjones: Still required?

Robbepop: No but I leave it as placeholder since I wanted to add some failure tests.

core/derive/src/lib.rs
Robbepop: Single variant is already guaranteed by the design of `synstructure` and the fact that I added a check to disallow `enum` (that is the only way to have multiple variants). The design of `synstructure` treats everything as `enum` which is brilliant but confusing at first sight. I will add some docs that explains this.

Robbepop:
> LGTM.
> 
> Though not sure I understand the exercising of the trait impls in the compile tests - are they invoked at all?

I just improved `AllocateUsing` tests and will do the same for `Flush` tests now.

PR #264, opened 2019-11-20
Fix selector encoding

cmichi:
By encoding as array of strings instead of a stringified array.

Before:
```
"[\"0x8C",\"0x97",\"0xDB",\"0x39"]"
```

After:
```
["0x8C", "0x97", "0xDB", "0x39"] 
```

Needs to be communicated to polkadot-js before merge.


Robbepop:
Will merge as soon as Polkadot JS side is ready.

Robbepop:
Superseeded by https://github.com/paritytech/ink/pull/296.

PR #269, opened 2019-11-22
Use Flush derives and generally update the Flush trait

Robbepop:
Update the `Flush` trait and use derive `Flush` proc. macros in `ink_core` and elsewhere.

Blocked by: https://github.com/paritytech/ink/pull/256


core/src/lib.rs
```diff
@@ -58,3 +58,6 @@ pub mod env;
 pub mod env2;
 pub mod memory;
 pub mod storage;
+
+// Needed for derive macros of `core/derive` sub crate.
+pub(crate) use crate as ink_core;
```
ascjones: Very clever!

Robbepop:
> LGTM. Why did we have all those empty impls of `fn flush` in the first place?

This is the question I have asked myself for a long time.

ascjones:
CI failing

Robbepop:
> CI failing

Fixed

Issue #245, opened 2019-11-17
Refactor dynamic allocator


Robbepop:
The current `Flush` and dynamic storage has problems in that dynamically allocated storage entities cannot be properly deallocated from storage because of a missing global accessor to the dynamic storage allocator as well as a missing accessor to the flusher.

We have to fix both problems while maintaining the efficiency gains through the new `ink_core::env2`.

## Global Instance

The only solution I currently see is that we introduce a globally accesible object that handles dynamic storage allocations and deallocations. This is a major move from the internal `DynEnv` approach that we have right now that tightly encapsulates the dynamic environment which enables us to perfectly reuse our internal buffers to avoid many (re-)allocations on the heap. However, this system does not allow access from anywhere and requires to have references from anywhere in order to allow for generic and ubiquitous operations such as deallocation (`Drop`) and flushing (through `Flush` trait).

## Buffer Pool

In order to maintain efficiency a solution might be to introduce a global buffer pool that is accessible from anywhere through shared references (`&self`) and enables buffer reusing. It is basically a pool of buffers that allow for intermediate buffer usages from anywhere to avoid heap allocations.

A possible interface for this could be:

```rust
/// Returns a reference to a shared buffer.
/// Shared buffers are generally a `Vec<u8>` under the hood and allow to store intermediate results.
/// When such a `SharedBuffer` is dropped (through its `Drop`) impl it is returned back to the `BufferPool` for later reuse.
fn get_buffer(&self) -> SharedBuffer;
```

The major problem is what the exact lifetime of such a returned `SharedBuffer` shall be optimally. We don't want them to be stored anywhere for an extended period of time since they are generally meant to be used as intermediate buffers. So we should attach some lifetime to them in the way how `RefCell` does that through `RefMut` etc.

Robbepop:
The https://github.com/paritytech/ink/pull/311 introduced a new dynamic storage allocator that actually works. The only thing left is a follow-up to introduce and use a so-called `PushOnDrop` abstractions but the rest already works as expected for the new `storage2` module. Closed.

PR #270, opened 2019-11-22
Initial implementation of the global buffer arena

Robbepop:
Implements the buffer pool mentioned in this issue: https://github.com/paritytech/ink/issues/245

The implementation of the `BufferArena` shall solve the problem of throttling heap memory allocations due to heavy usage in the layer between SRML contracts and ink!. Most often buffers are short-lived since they are just used for encoding and decoding intermediate values. The `BufferArena` acts as a globally accessible heap allocation cache.

Provides a single API:

- `fn get_buffer(&self) -> BufferRef`

Where `BufferRef` implements all traits required for ink! and SRML contracts interaction.
When such a `BufferRef` goes out of scope its internal buffer is added back to the global `BufferArena` instance.

The implementation tries to be as efficient as possible while maintaining checks to identify malicious smart contracts that accumulate many cached buffers.

This small module will be the backbone of the future dynamic allocator and maybe also of a reimplementation of the environmental accessor.


core/src/lib.rs
```diff
@@ -41,6 +41,7 @@
     trivial_numeric_casts,
     unused_extern_crates
 )]
+#![feature(cell_update)]
```
pepyakin: Is using this feature really worth it? While the code looks prettier indeed I think we can define `Cell::update` locally, right?

Robbepop: will remove it

core/src/env2/buffer_arena.rs
```diff
@@ -0,0 +1,358 @@
...
+        /// CRITICAL NOTE
+        /// =============
+        ///
+        /// The wrapped `BufferArena` type itself is __NOT__ `Sync` since it is using
+        /// `Cell` and `RefCell` internally instead of the thread-safe alternatives.
+        /// However, since Wasm smart contracts are guaranteed to operated single
+        /// threaded we can allow for this unsafe `Sync` implementation to allow
+        /// for having the global static `BUFFER_ARENA` variable and as long as we
+        /// are only operating single threaded this shouldn't be unsafe.
+        // #[cfg(not(feature = "std"))]
```
cmichi: Did you leave this comment in here accidentally?

core/src/env2/buffer_arena.rs
```diff
@@ -0,0 +1,358 @@
...
+            /// The wrapped buffer arena.
+            arena: BufferArena,
+        }
+
+        /// CRITICAL NOTE
+        /// =============
+        ///
+        /// The wrapped `BufferArena` type itself is __NOT__ `Sync` since it is using
+        /// `Cell` and `RefCell` internally instead of the thread-safe alternatives.
+        /// However, since Wasm smart contracts are guaranteed to operated single
```
cmichi:
```suggestion
        /// However, since Wasm smart contracts are guaranteed to operate single
```

core/src/env2/buffer_arena.rs
```diff
@@ -0,0 +1,358 @@
...
+    /// required to use them as environment buffer.
+    ///
+    /// - [`Reset`]: Allows resetting the buffer. This clears all the elements,
+    ///            however, it retains the memory allocation.
+    /// - [`EnlargeTo`]: Safely enlarges the buffer to the required minimum size
+    ///                  if it isn't already large enough.
+    /// - [`core::convert::AsRef`]`<[u8]>`: Returns a shared view into the byte buffer.
+    /// - [`core::convert::AsMut`]`<[u8]>`: Returns an exclusive view into the byte buffer.
+    pub fn get_buffer(&self) -> BufferRef {
+        self.in_use.set(self.in_use.get() + 1);
```
cmichi:
```suggestion
        self.in_use.set(self.in_use() + 1);
```

core/src/env2/buffer_arena.rs
```diff
@@ -0,0 +1,358 @@
...
+    Cell,
+    RefCell,
+};
+
+/// The maximum amount of used byte buffers at the same time.
+///
+/// Since the whole point behind this byte buffer arena is to cache
+/// allocated heap memory the number of concurrent byte buffers that are
+/// in use at the same time should be kept small.
+const IN_USE_LIMIT: usize = 1000;
```
cmichi: So the limit is defined by the number of instances, but what is the byte limit per instance? I wonder if it makes more sense to define the arena limit as a number of bytes, since this would provide a better idea of the upper byte boundary.

Does the number have any reasoning behind it or is it more a rough value of thumb for now?

Robbepop: The limit is arbitrary.
However imposing it makes sense since there is one way how you could abuse this global buffer arena and it is when you end up with many many different buffers, so when `in_use` is a high number.
I expect normal use cases to spawn at most 5-10 buffers so 1000 is already pretty intense.

A regulation on the total bytes in use could be introduced but it should definitely be possible to have for example one giant buffer if that's what has been needed.

PR #284, opened 2019-11-27
Add BTreeMap storage primitive

cmichi:
- [x] Provide API for insert operations
- [x] Provide entry API
- [x] Provide API for removing elements
- [x] More doc comments
- [x] More tests


core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1132 @@
...
+    cmp::Ord,
+    ptr,
+};
+use super::search::{self, SearchResult};
+
+/// Each node in the tree has 2 * B children.
+pub(crate) const B: usize = 6;
+
+/// Number of elements which can be stored in one node of the tree.
+pub const CAPACITY: usize = 2 * B - 1;
```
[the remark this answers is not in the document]
cmichi: I added a comment to clarify it.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1132 @@
...
+#[cfg(feature = "ink-generate-abi")]
+use type_metadata::Metadata;
+use core::{
+    cmp::Ord,
+    ptr,
+};
+use super::search::{self, SearchResult};
+
+/// Each node in the tree has 2 * B children.
+pub(crate) const B: usize = 6;
```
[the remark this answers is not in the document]
cmichi: For the moment I just took the same parameters as [the Rust stdlib implementation](https://github.com/rust-lang/rust/blob/master/src/liballoc/collections/btree/node.rs#L42-L44). So yeah kinda arbitrarily, but I don't have a better idea until we can do benchmarks.

One thing that I just thought about: If we change these consts at any point this will require to have all existing `BTreeMap`s which are stored on-chain to be migrated. An alternative would be to store these params within the `BTreeMapHeader`. Hmm, but come to think of it this brings a lot of disadvantages, like us having to test and support all kinds of parameters and be endlessly downwards-compatible. Yeah, so from my POV we should stay with consts.

core/src/storage/collections/btree_map/search.rs
```diff
@@ -0,0 +1,132 @@
...
+            KVHandle::new(node_index, idx)
+        )
+    }
+}
+
+/// Conducts a linear search for `key` in the elements contained in a node.
+pub fn search_linear<K, V, Q>(
+    node: &Node<K, V>,
+    key: &Q
+) -> (u32, bool)
```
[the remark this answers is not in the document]
cmichi: So the thing is I also need to return the `u32` here (which is the last position examined), so unfortunately `Option<u32>` wouldn't help us here. I added a comment to clarify the meaning of the return value.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+        slice.len() - idx - 1,
+    );
+    ptr::write(slice.get_unchecked_mut(idx), val);
+}
+
+/// Extracts the element at `idx` from `slice` while shifting all subsequent items to
+/// the left by one.
+///
+/// Returns the extracted element.
+unsafe fn slice_remove<T>(slice: &mut [Option<T>], idx: usize) -> Option<T> {
```
Robbepop: Isn't this what `Vec::remove(&mut self, index: usize)` does?
https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+/// Inserts `val` at `idx` into `slice` while shifting all subsequent items to
+/// the right by one. The last element of the slice will fall out.
+unsafe fn slice_insert<T>(slice: &mut [T], idx: usize, val: T) {
+    ptr::copy(
+        slice.as_ptr().add(idx),
+        slice.as_mut_ptr().add(idx + 1),
+        slice.len() - idx - 1,
+    );
+    ptr::write(slice.get_unchecked_mut(idx), val);
+}
```
Robbepop: Isn't this what `Vec::insert(&mut self, index: usize, elem: T)` does?
https://doc.rust-lang.org/std/vec/struct.Vec.html#method.insert

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+    /// assert_eq!(map["poneyland"], 12);
+    /// if let Entry::Occupied(o) = map.entry("poneyland") {
+    ///     *o.into_mut() += 10;
+    /// }
+    /// assert_eq!(map["poneyland"], 22);
+    /// ```
+    pub fn into_mut(self) -> &'a mut V {
+        self.into_kv_mut()
+            .expect("every occupied entry always has a key/value pair; qed")
+            .1
```
Robbepop: I'd love to see named structs instead of tuples or tuple structs in such cases since something like `.1` is not very readable tbh.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+    ///
+    ///     // We can use the same Entry multiple times.
+    ///     *o.get_mut() += 2;
+    /// }
+    /// assert_eq!(map["poneyland"], 24);
+    /// ```
+    pub fn get_mut(&mut self) -> &mut V {
+        self.kv_mut()
+            .expect("every occupied entry always has a key/value pair; qed")
+            .1
```
Robbepop: I'd love to see named structs instead of tuples or tuple structs in such cases since something like `.1` is not very readable tbh.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+    ///
+    /// if let Entry::Occupied(o) = map.entry("poneyland") {
+    ///     assert_eq!(o.get(), &12);
+    /// }
+    /// ```
+    pub fn get(&self) -> &V {
+        self.tree
+            .get_kv(self.handle)
+            .expect("every occupied entry always has a key/value pair; qed")
+            .1
```
Robbepop: I'd love to see named structs instead of tuples or tuple structs in such cases since something like `.1` is not very readable tbh.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+    ///
+    /// let mut map: BTreeMap<&str, usize> = new_btree_map();
+    /// map.entry("poneyland").or_insert(12);
+    /// assert_eq!(map.entry("poneyland").key(), &"poneyland");
+    /// ```
+    pub fn key(&self) -> &K {
+        self.tree
+            .get_kv(self.handle)
+            .expect("every occupied entry always has a key/value pair; qed")
+            .0
```
Robbepop: I'd love to see named structs instead of tuples or tuple structs in such cases since something like `.1` is not very readable tbh.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+    /// the right one.
+    Split(KVHandle, K, V, NodeHandle),
+}
+
+enum UnderflowResult {
+    AtRoot,
+    EmptyParent(NodeHandle),
+    Merged(NodeHandle),
+    Stole(NodeHandle),
+}
```
Robbepop: Missing docs.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+    fn from(handle: KVHandle) -> NodeHandle {
+        NodeHandle::new(handle.node)
+    }
+}
+
+impl From<u32> for NodeHandle {
+    fn from(index: u32) -> NodeHandle {
+        NodeHandle::new(index)
+    }
+}
```
Robbepop: Sure you want this? Because a trait implementation is always public. But for improved type safety by using index spaces guided by index types it makes sense to keep implementations like these as private as possible.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+
+impl NodeHandle {
+    pub fn new(node: u32) -> Self {
+        Self { node }
+    }
+
+    pub fn node(&self) -> u32 {
+        self.node
+    }
+}
```
Robbepop: Docs missing.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+        }
+    }
+
+    /// Returns the number of elements stored in the tree.
+    pub fn len(&self) -> usize {
+        self.len as usize
+    }
+
+    /// Returns the number of edges this node has.
+    pub fn edges(&self) -> usize {
```
Robbepop: I'd rename this to `len_edges`. I thought this was returning an iterator over the edges instead tbh.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+    pub parent_idx: Option<u32>,
+
+    /// The array storing the keys for a node.
+    pub keys: [Option<K>; CAPACITY],
+
+    /// The array storing the values for a node.
+    pub vals: [Option<V>; CAPACITY],
+
+    /// The pointers to the children of this node.
+    pub edges: [Option<u32>; 2 * B],
```
Robbepop: I'd replace `2 * B` with an `EDGES` constant and we should state that `CAPACITY + 1 == EDGES` must always be true.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+/// which children nodes can be linked. Each child has a link to its parent.
+///
+/// Each node is stored as one storage entity. This reduces storage access,
+/// since with each fetch the entire content of a node (all its elements, etc.)
+/// are fetched.
+#[derive(PartialEq, Eq, Encode, Decode)]
+#[cfg_attr(feature = "ink-generate-abi", derive(Metadata))]
+pub struct Node<K, V> {
+    /// This node's parent node.
+    pub parent: Option<u32>,
```
Robbepop: Shouldn't this better be an `Option<NodeHandle>`?

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+
+    /// This node's index into the parent node's `edges` array.
+    /// If, for example, `parent_idx = Some(2)` this refers to the
+    /// second position in the `edges` array of its parent node.
+    pub parent_idx: Option<u32>,
```
Robbepop: Shouldn't this better be an `Option<KVHandle>`?

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+/// for performance reasons so that they all reside in the same
+/// storage entity. This allows implementations to perform less reads
+/// and writes to the underlying contract storage.
+#[derive(Encode, Decode)]
+#[cfg_attr(feature = "ink-generate-abi", derive(Metadata))]
+pub(super) struct BTreeMapHeader {
+    /// The latest vacant index.
+    next_vacant: Option<u32>,
+    /// The index of the root node.
+    root: Option<u32>,
```
Robbepop: This should probably be `Option<NodeHandle>`.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+///
+/// Separation of these fields into a sub structure has been made
+/// for performance reasons so that they all reside in the same
+/// storage entity. This allows implementations to perform less reads
+/// and writes to the underlying contract storage.
+#[derive(Encode, Decode)]
+#[cfg_attr(feature = "ink-generate-abi", derive(Metadata))]
+pub(super) struct BTreeMapHeader {
+    /// The latest vacant index.
+    next_vacant: Option<u32>,
```
Robbepop: This should probably be `Option<NodeHandle>`.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+    fn correct_all_childrens_parent_links(&mut self, handle: &NodeHandle) {
+        let node = self.get_node(handle).expect("node must exist");
+        let len = node.len();
+        self.correct_childrens_parent_links(handle, 0, len + 1);
+    }
+
+    fn has_children(&mut self, handle: &NodeHandle) -> bool {
+        let node = self.get_node(handle).expect("node must exist");
+        node.edges() > 0
+    }
```
Robbepop: I know these are just internal utility and helper routines, however, I would still appreciate proper doc comments here.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1872 @@
...
+};
+use scale::{
+    Codec,
+    Decode,
+    Encode,
+};
+#[cfg(feature = "ink-generate-abi")]
+use type_metadata::Metadata;
+
+/// Each node in the tree has 2 * B children.
```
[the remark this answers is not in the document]
cmichi: I chose the number solely based on the reason that the Rust standard library implementation [uses the same number](https://github.com/rust-lang/rust/blob/master/src/liballoc/collections/btree/node.rs#L42). Since we don't have empirical data yet I took this number as a first starting point.

But yeah, I agree, there needs to be a logical argument for the number of children. Once people use the data structure it will be hard to migrate to a new number.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+/// and writes to the underlying contract storage.
+#[derive(Encode, Decode)]
+#[cfg_attr(feature = "ink-generate-abi", derive(Metadata))]
+struct BTreeMapHeader {
+    /// The latest vacant node index.
+    next_vacant: Option<NodeHandle>,
+    /// The latest vacant pair index.
+    next_vacant_pair: Option<KVStoragePointer>,
+    /// The index of the root node.
+    root: Option<NodeHandle>,
```
Robbepop: Is this not always 0? If not, when? why?

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+/// for performance reasons so that they all reside in the same
+/// storage entity. This allows implementations to perform less reads
+/// and writes to the underlying contract storage.
+#[derive(Encode, Decode)]
+#[cfg_attr(feature = "ink-generate-abi", derive(Metadata))]
+struct BTreeMapHeader {
+    /// The latest vacant node index.
+    next_vacant: Option<NodeHandle>,
+    /// The latest vacant pair index.
+    next_vacant_pair: Option<KVStoragePointer>,
```
Robbepop: Emulating a `storage::Stash` here?

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+    pub(super) fn get_node(&self, handle: NodeHandle) -> Option<&Node<K, V>> {
+        let entry = self.entries.get(handle.node())?;
+        match entry {
+            InternalEntry::Occupied(occupied) => Some(occupied),
+            InternalEntry::Vacant(_) => None,
+        }
+    }
+
+    /// Returns the keys stored in this node.
+    pub(super) fn keys_in_node(&self, node: NodeHandle) -> [Option<&K>; CAPACITY] {
```
Robbepop: Wouldn't it be better if we instead return `&[K]` here?
Or is it possible that the resulting array has holes in it? E.g. `[Some, Some, None, Some]`

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+            (Split(left, pair_ptr, right), idx) => {
+                ins_pair_ptr = pair_ptr;
+                ins_edge = right;
+                out_storage_pointer = idx;
+                self.ascend(left.node())
+            }
+        };
+
+        loop {
+            match cur_parent {
```
Robbepop: Please add a comment providing a rational that explains why this loop always breaks.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+                    .expect("handle must be valid");
+                (hole, old_pair_ptr, nl)
+            }
+        };
+
+        let mut handle = small_leaf.node();
+        while new_len < CAPACITY / 2 {
+            match self.handle_underfull_node(handle) {
+                UnderflowResult::AtRoot => break,
+                UnderflowResult::EmptyParent(_) => unreachable!(),
```
Robbepop: Please add a comment in the `unreachable!` providing a rational that states exactly why it is unreachable.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+        let left_node = self
+            .get_node_mut(left_child)
+            .expect("left child must exist");
+        let left_len = left_node.len();
+
+        debug_assert!(left_len + right_len < CAPACITY);
+
+        left_node.set_pair(left_len, removed_pair_ptr);
+        let src = &right_pairs[..right_len];
+        let range = left_len + 1..left_len + 1 + right_len;
```
Robbepop: Personally I'd prefer some parentheses for this.

Robbepop:
```suggestion
        let range = (left_len + 1)..(left_len + 1 + right_len);
```

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+
+                Some(new_root)
+            }
+        };
+
+        let node = self.get_node_mut(handle).expect("node must exist");
+        node.set_len(node.len() - 1);
+
+        (pair_ptr, edge)
+    }
```
Robbepop: Lot's of `... must exist` assertions in this function. Sometimes this indicates minor structuring smells.

core/src/storage/collections/btree_map/node.rs
```diff
@@ -0,0 +1,439 @@
...
+    pub(super) pairs: [Option<KVStoragePointer>; CAPACITY],
+
+    /// The pointers to the children of this node.
+    pub(super) edges: [Option<NodeHandle>; EDGES],
+
+    /// Number of elements stored in this node.
+    len: u32,
+
+    /// Marker for compile-time checking of correct key/value pair types.
+    marker: PhantomData<(K, V)>,
```
Robbepop:
```suggestion
    marker: PhantomData<fn() -> (K, V)>,
```

core/src/storage/collections/btree_map/tests.rs
```diff
@@ -0,0 +1,442 @@
...
+        // when
+        // when the element is removed again the node count has
+        // to go back to one since the nodes should have been merged.
+        assert_eq!(map.remove(&one_more), Some(one_more * 10));
+        assert_eq!(map.len(), 11);
+        assert_eq!(map.node_count(), 1);
+        assert!(map.get_node(NodeHandle::new(0)).is_some());
+        Ok(())
+    })
+}
```
Robbepop: Nice!

core/src/storage/collections/btree_map/node.rs
cmichi: Where do you see the advantage of the `fn() -> ` here?

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+/// ```no_compile
+/// const B: usize = 2;
+/// const CAPACITY: usize = 2 * B - 1;
+/// keys  = [    a,    b,    c    ];
+/// edges = [ 1,    2,    3,    4 ];
+/// ```
+pub(super) const CAPACITY: usize = 2 * B - 1;
+
+/// Pointer to the index where a key/value pair is stored in `kv_pairs`.
+pub(super) type KVStoragePointer = u32;
```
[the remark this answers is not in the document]
cmichi: Good point, I renamed it.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+
+/// The node type, either a `Leaf` (a node without children) or `Internal`
+/// (a node with children). This distinction makes it easier to handle e.g.
+/// cases where nodes are merged or implementing recursion.
+pub(super) enum HandleType {
+    Leaf,
+    Internal,
```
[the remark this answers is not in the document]
cmichi: I agree, `Branch` seems a term which is a lot more speaking. I originally took the naming from the Rust implementation. I renamed it now everywhere. `Internal` was also confusing since our implementation also has an `InternalEntry` which is something completely different.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+/// `BTreeMap` standard library implementation. The Rust implementation
+/// is in-memory, whereas this implementation uses the ink! storage
+/// primitives (`SyncChunk`, etc.).
+#[cfg_attr(feature = "ink-generate-abi", derive(Metadata))]
+pub struct BTreeMap<K, V> {
+    /// Stores densely packed general BTreeMap information.
+    header: storage::Value<BTreeMapHeader>,
+
+    /// Contains the nodes of the tree.
+    entries: SyncChunk<InternalEntry<K, V>>,
```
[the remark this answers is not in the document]
cmichi: I renamed `entries` to `nodes`, but your other suggestion doesn't work out since `InternalEntry<K, V>` encapsulates our recycling storage mechanism:
```
enum InternalEntry<K, V> {
    /// A vacant entry pointing to the next vacant index.
    Vacant(Option<NodeHandle>),
    /// An occupied entry contains a tree node with its elements.
    Occupied(Node<K, V>),
}
```
But as you see there is already a `Node<K, V>` struct which encapsulates what you most probably wanted to get at.

core/src/storage/collections/btree_map/impls.rs
cmichi: The root index can change. For example, if an element is inserted it is inserted into a leaf; if the leaf is already full this results in changes to the tree structure. These changes are propagated upwards. This might result in an overfull root node. The algorithm handles this by creating a new root node with the old root as a child.

core/src/storage/collections/btree_map/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+//! keys  = [    a,    b,    c    ];
+//! vals  = [    a,    b,    c    ];
+//! edges = [ 1,    2,    3,    4 ];
+//! ```
+//!
+//! Here the left child of element `a` would be the node with the index `1`, its
+//! right child the node with index `2`.
+//!
+//! This concept of multiple elements stored in one node suits our needs well,
+//! since expensive storage fetches are reduced.
```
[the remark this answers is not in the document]
cmichi: One node = one storage cell at the moment, since `nodes: SyncChunk<InternalEntry<K, V>>` with `InternalEntry::Occupied(Node<K, V>)`.

core/src/storage/collections/btree_map/impls.rs
```diff
@@ -0,0 +1,1963 @@
...
+
+    /// The key/value pairs stored in the tree.
+    /// We don't store this in `entries` for performance reasons. Instead each
+    /// node of the tree merely stores a pointer to the storage index where
+    /// the key/value pair is store.
+    /// This way it's more effective to balance the tree, since we only ever
+    /// have to move the `u32` index, instead of the whole `K`/`V` (which
+    /// would imply decoding, encoding, etc.).
+    kv_pairs: SyncChunk<InternalKVEntry<K, V>>,
```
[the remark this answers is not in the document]
cmichi: > Also you could rename `InternalKVEntry<K, V>` to simply `Entry<K, V>`.

There is already a `pub enum Entry<'a, K, V>` exported by `BTreeMap`, since I need to expose `Entry` for the entries API publicly.

core/src/storage/collections/btree_map/impls.rs
cmichi: I would leave it at `[Option<&K>]` since this method is used by the node search stuff, which e.g. knows that search can be stopped once a `None` element is reached.

core/src/storage/collections/btree_map/impls.rs
cmichi: Kind of, that's where I took it from :-).

core/src/storage/collections/btree_map/impls.rs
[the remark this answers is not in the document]
cmichi: So from what I understand (e.g. from [this](http://cglab.ca/~abeinges/blah/rust-btree-case/) extensive description of an earlier implementation) the `6` also doesn't have that much of a reasoning behind it, but rather seemed like a good value for the linear search conducted within a node when searching for a key while also keeping the algorithm cache-friendly.

I think for our case the best thing would be to have some proper benchmarks. Though I can suppose the results will vary widely depending on what you put into the `BTreeMap`.

core/src/storage/collections/btree_map/impls.rs
cmichi: Yeah I agree. It's unfortunate because we need to fetch the edge, child and node behind a handle and each of those could potentially fail ‒ though it shouldn't if the implementation is sound.

cmichi:
@Robbepop I have implemented most requested changes and to the ones where I have something different to say I replied to your comment. So yeah from my perspective the PR is a good first version to iterate over once we have better benchmarking tools.

Could you please take another look?

PR #287, opened 2019-11-29
Add Erc721 example with ink2.0 syntax

XatMassacrE:


examples/lang2/erc721/Cargo.toml
```diff
@@ -0,0 +1,58 @@
+[package]
+name = "erc721"
+version = "0.1.0"
+authors = ["XatMassacrE"]
```
Robbepop: I am not sure if we are going to accept example authors to be different from the default. Have to ask ...

examples/lang2/erc721/Cargo.toml
```diff
@@ -0,0 +1,58 @@
...
+edition = "2018"
+
+[dependencies]
+ink_abi = { path = "../../../abi", default-features = false, features = ["derive"], optional = true }
+ink_core = { path = "../../../core", default-features = false }
+ink_lang2 = { path = "../../../lang2", default-features = false }
```
Robbepop: Please make use of the new `ink_prelude` crate as has been done in other parts of ink! and other examples.

examples/lang2/erc721/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![feature(proc_macro_hygiene)]
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use ink_lang2 as ink;
+use ink_core::{
+    memory::format,
```
Robbepop: Please update to current master and use `ink_prelude` instead.

examples/lang2/erc721/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+    #[ink(storage)]
+    struct Erc721 {
+        /// Mapping from token to owner.
+        token_owner: storage::HashMap<Hash, AccountId>,
+        /// Mapping from token to approvals users.
+        token_approvals: storage::HashMap<Hash, AccountId>,
+        /// Mapping from owner to number of owned token.
+        owned_tokens_count: storage::HashMap<AccountId, u128>,
+        /// Mapping from owner to operator approvals
+        operator_approvals: storage::HashMap<AccountId, storage::HashMap<AccountId, bool>>,
```
Robbepop: It is recommended to use
```rust
operator_approvals: storage::HashMap<(AccountId, AccountId), bool>
```
instead to avoid having to use dynamic storage.
Dynamic storage allocations are a thing but are currently unstable and not fleshed out.

examples/lang2/erc721/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+        owner: AccountId,
+        #[ink(topic)]
+        operator: AccountId,
+        approved: bool
+    }
+
+    impl Erc721 {
+        #[ink(constructor)]
+        fn new(&mut self) {
+            self.none_account.set(AccountId::from([0x0; 32]));
```
Robbepop: In other example contracts we do not recommend having this none account that stems from Solidity.

examples/lang2/erc721/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+
+        /// ===========================
+        /// NOT REQUIRED, JUST FOR TEST
+        /// ===========================
+        #[ink(message)]
+        fn mint(&mut self, id: Hash) -> bool {
+            let caller = self.env().caller();
+            let owner = self.owner_of_or_none(id);
+            if owner != *self.none_account {
+                self.env().println(&format!("Erc721::mint::token is exist"));
```
Robbepop: Generally we don't want to put `println` into example contract codes.
Users should only make use of `println` for debugging purposes while developing.

examples/lang2/erc721/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+            let owner = self.owner_of_or_none(id);
+            self.env().println(&format!("Erc721::owner_of(token = {:?}) = {:?}", id, owner));
+            owner
+        }
+
+        /// The approved address for this token, or the none address if there is none
+        #[ink(message)]
+        fn get_approved(&self, id: Hash) -> AccountId {
+            let account = self.approved_of_or_none(id);
+            account
```
Robbepop: Can be simplified to just `self.approved_of_or_none(id)`

Robbepop: Also instead of none account you could use `Option<AccountId>` as return type which would be more idiomatic Rust.

examples/lang2/erc721/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+                    unsafe {
+                        use ink_core::storage::alloc::AllocateUsing;
+                        use ink_core::storage::alloc::Initialize;
+                        use ink_core::storage::alloc::BumpAlloc;
+                        use ink_core::storage::Key;
+                        let mut alloc = BumpAlloc::from_raw_parts(Key([0x0; 32]));
+                        let mut default_hashmap: storage::HashMap<AccountId, bool> = storage::HashMap::allocate_using(&mut alloc).initialize_into(());
+                        default_hashmap.insert(to, approved);
+                        self.operator_approvals.insert(caller, default_hashmap);
+                    }
```
Robbepop: Basically this whole thing is screwed but also can be easily solved by just using `storage::HashMap<(AcocuntId, AccountId), bool>` as described above.

XatMassacrE:
@Robbepop Hi, i have update most of the problems you came up.

examples/lang2/erc721/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    struct Erc721 {
+        /// Mapping from token to owner.
+        token_owner: storage::HashMap<Hash, AccountId>,
+        /// Mapping from token to approvals users.
+        token_approvals: storage::HashMap<Hash, AccountId>,
+        /// Mapping from owner to number of owned token.
+        owned_tokens_count: storage::HashMap<AccountId, u128>,
+        /// Mapping from owner to operator approvals
+        operator_approvals: storage::HashMap<(AccountId, AccountId), bool>
+    }
```
Robbepop: Due to being compiled for `wasm32` which is a 32-bit architecture we can simplify certain parts of this. For example you could use a `struct TokenId(u32)` instead of `Hash` for the token ID which is just 32 bit instead of 256 bit large structure. That should help in reducing gas costs.
Also `owned_tokens_count` could therefore be a mapping from `AccountId -> u32` instead of `u128`.

XatMassacrE:
@Robbepop i have fixed the two questions.

Robbepop:
> Sorry for not responding for longer time.
> This looks good overall now.
> I am willing to merge if it has been adjusted to the most recent ink! changes.
> Also CI needs to be adressed.
> 
> Furthermore you could add some tests since the new off-chain test environment regained its full powers.

No answer + superseeded by https://github.com/paritytech/ink/pull/337. Closed.

PR #291, opened 2019-12-03
Bring Debug trait bound back for env2::EnvTypes under test-env

liuchengxu:


Robbepop:
Please tell us our use case for this change.

liuchengxu:
I was trying to upgrade https://github.com/paritytech/ink-types-node-runtime to ink2.0 and this will resolve the first error when doing the migration.

屏幕快照 2019-12-03 下午10 32 44

Robbepop:
> 
> 
> I was trying to upgrade https://github.com/paritytech/ink-types-node-runtime to ink2.0 and this will resolve the first error when doing the migration.
> 
> 屏幕快照 2019-12-03 下午10 32 44

Why does it need to be debug formatted in here anyways?

liuchengxu:
I have not digged into it. But `Deubg` trait exists in `env` under feature `test-env`, so I think it would also be useful in `env2`.

Robbepop:
> 
> 
> I have not digged into it. But `Deubg` trait exists in `env` under feature `test-env`, so I think it would also be useful in `env2`.

Generally you should only ever use `Debug` types in test code and test code is always concrete, meaning it never operates on generics that have to fullfil some trait bounds. So even if we have no such `Debug` trait bound you can still `Debug` print your types in the contracts test code if they support doing that since you are operating on concrete types that the compiler can deduce.

The `ink_core::env` is generally deprecated and shouldn't be taken care of.

liuchengxu:
`Debug` trait bound is required in https://github.com/paritytech/ink-types-node-runtime/blob/a377b14338914df339f432c5c41ef3ebee008738/src/calls.rs#L27-L29 . I still think it's used for the `test-env`. Any thoughts? @ascjones

ascjones:
> Generally you should only ever use Debug types in test code and test code is always concrete, 

That's true but IIRC there was a requirement for `Debug` in the `TestEnv<T>` itself (which doesn't know the concrete types). I can't remember what for. 

Perhaps with the new `test-env` impl it is no longer required. What happens if we remove that `derive(Debug)` from `Address` - will it compile?

liuchengxu:
Well, it does compile with all the related `Debug` removed.

ascjones:
Cool so we probably don't need this change then. Have you ported the example + tests in `ink-types-node-runtime`?

liuchengxu:
Closing then.

@ascjones Just made it compile. The tests of `ink-types-node-runtime` are still broken. Some apis used in the tests are seemingly removed and not sure if they have replacements at the moment. I'll prepare a PR later and then we could move the discussion to `ink-types-node-runtime`.

PR #293, opened 2019-12-03
Add DomainNameService example contract

Robbepop:
ink! 2.0 port for the contract presented in https://medium.com/@chainx_org/secure-and-decentralized-polkadot-domain-name-system-e06c35c2a48d.


examples/lang2/dns/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+        old_owner: Option<AccountId>,
+        #[ink(topic)]
+        new_owner: AccountId,
+    }
+
+    impl DomainNameService {
+        /// Creates a new domain name service contract.
+        #[ink(constructor)]
+        fn new(&mut self) {
+            self.default_address.set(AccountId::from([0x00; 32]));
```
cmichi: Dumb question, but couldn't we use `Default::default()`? Seems like it would simplify the code and be more readable as well.

examples/lang2/dns/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+        #[ink(message)]
+        fn is_name_exist(&self, name: Hash) -> bool {
+            self.name_exists(name)
+        }
+
+        fn name_exists(&self, name: Hash) -> bool {
+            if self.name_to_owner.get(&name).is_some() {
+                return true
+            }
+            false
```
cmichi: Could be simplified to just `self.name_to_owner.get(&name).is_some()` as the function body (GitHub sadly can't do multi-line suggestions yet).

examples/lang2/dns/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use ink_core::storage;
+use ink_lang2 as ink;
+
+#[ink::contract(version = "0.1.0")]
+mod dns {
+    #[ink(storage)]
+    struct DomainNameService {
+        /// A hashmap to store all name to addresses mapping.
```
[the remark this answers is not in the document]
cmichi: 1-n vs. 1-1 mapping (we have the latter, since the hashmap contains the owner-address mapping)

examples/lang2/dns/src/lib.rs
Robbepop: I wish I could undo some commits. :D

Robbepop:
> In the spirit of encouraging an economical handling of storage I suggest to add a
> 
> ```
> #[ink(message)]
> fn unregister(&mut self, name: Hash) -> bool { … }
> ```
> 
> Since we want to encourage people testing their contracts it would be nice to have some tests. I think it also makes it visible how easy it is to use the idiomatic Rust way of testing with ink!. Wdyt?
> 
> Besides these nitpicks the code looks really clean! +1

Now that we have a proper off-chain testing framework we could actually add tests. The linked block post even already provides plenty of tests we could simply reuse.

examples/dns/.ink/abi_gen/main.rs
```diff
@@ -0,0 +1,7 @@
+fn main() -> Result<(), std::io::Error> {
+    let abi = <contract::DomainNameService as ink_lang2::GenerateAbi>::generate_abi();
```
Robbepop:
```suggestion
    let abi = <contract::DomainNameService as ink_lang::GenerateAbi>::generate_abi();
```

examples/dns/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+    struct DomainNameService {
+        /// A hashmap to store all name to addresses mapping.
+        name_to_address: storage::HashMap<Hash, AccountId>,
+        /// A hashmap to store all name to owners mapping.
+        name_to_owner: storage::HashMap<Hash, AccountId>,
+        /// The default address.
+        default_address: storage::Value<AccountId>,
+    }
+
+    /// Errors that can occure upon calling this contract.
```
cmichi:
```suggestion
    /// Errors that can occur upon calling this contract.
```

PR #296, opened 2019-12-10
New metadata format

### LINKED ISSUE #294: Contract metadata format
@jacogr is not happy with the current contract metadata format - it requires a hand-rolled parser which is difficult to maintain. 

I will add some proposed metadata formats to this issue once I come up with them. Also we will most likely need to make changes in https://github.com/type-metadata/type-metadata. In the meantime I will collect some raw thoughts from Riot chat here:

------------------
This brings me to my next point, after looking at all the examples - I would appreciate it if we can actually move the ABI JSON also to something standard that aligns with the way we actually do things in Substrate. As it stands Keith had to roll a parser by hand since he couldn't use any of the "define the structure, and it parse" code that we use everywhere.

This means that when we eventually get it in metadata, all that work is completely wasted an has to be re-done. (Instead of having well-defined structures that can parse byte streams, JSON, etc.)

This means -

- no Rust-like namings, stick to JSON standard, e.g. `return_type:` ... va `returnType:` ...
- build up proper structures, so they can be parsed as such, e.g. `range.offset`: ... vs `range: { offset: ... }`
- this is not a `[u8; 4]` output, `["0x5E","0xBD","0x88","0xD6"]`, this is `0x5edb88d6`
Everything being non-aligned-with-Substrate in this world is just creating untold pain. With the simple things above in place, you can just feed the createType a definition and magically full-parsed objects appear. Currently there are special types defined, a hand-rolled parser for each type of field, special conversions for e.g. the selectors, etc. etc.

----------------
I am telling you that the ABI JSON as it stands,

- encodes data incorrectly
- has like "range.offset", "range.len" next to each other (instead of enums)
- cannot be parsed with anything that we use to parse the normal SCALE . and RPC stuff
So I want a proper format that I can define in terms of structures (enums, structs) that I can just . parse "magiclally".

ascjones:
Closes #294.

- Use externally tagged enums e.g. instead of `"range.offset"` use `"range": { "offset" }`
- Convert field names to `camelCase` instead of `snake_case`
- Selector bytes encoded as hex string e.g. instead of `"[\"0x07\",\"0x5B\",\"0xCD\",\"0x15\"]"` use `"0x075bcd15"`
- Convert from `type-metadata` to https://github.com/paritytech/scale-info

The new type registry serialization depends on https://github.com/paritytech/scale-info/pull/3, so only once that is merged will you get the full new format.

e.g. Incrementer:

```
{
  "registry": {
    "strings": [
      "Storage",
      "flipper",
      "__ink_private",
      "__ink_storage",
      "value",
      "Value",
      "ink_core",
      "storage",
      "cell",
      "SyncCell",
      "sync_cell",
      "Key",
      "ink_primitives",
      "key",
      "new",
      "init_value",
      "bool",
      "default",
      "flip",
      "get"
    ],
    "types": [
      {
        "composite": {
          "name": 1,
          "namespace": [
            2,
            2,
            3,
            4
          ],
          "fields": [
            {
              "name": 5,
              "type": 2
            }
          ]
        }
      },
      {
        "composite": {
          "name": 6,
          "namespace": [
            7,
            8,
            5
          ],
          "params": [
            3
          ],
          "fields": [
            {
              "name": 9,
              "type": 4
            }
          ]
        }
      },
      {
        "primitive": "bool"
      },
      {
        "composite": {
          "name": 10,
          "namespace": [
            7,
            8,
            9,
            11
          ],
          "fields": [
            {
              "name": 9,
              "type": 5
            }
          ]
        }
      },
      {
        "composite": {
          "name": 12,
          "namespace": [
            13,
            14
          ],
          "fields": [
            {
              "type": 6
            }
          ]
        }
      },
      {
        "array": {
          "len": 32,
          "type": 7
        }
      },
      {
        "primitive": "u8"
      }
    ]
  },
  "storage": {
    "struct": {
      "type": 1,
      "fields": [
        {
          "name": 5,
          "layout": {
            "struct": {
              "type": 2,
              "fields": [
                {
                  "name": 9,
                  "layout": {
                    "range": {
                      "offset": "0x0000000000000000000000000000000000000000000000000000000000000000",
                      "len": 1,
                      "elemType": 3
                    }
                  }
                }
              ]
            }
          }
        }
      ]
    }
  },
  "contract": {
    "name": 2,
    "constructors": [
      {
        "name": 15,
        "selector": "0x5ebd88d6",
        "args": [
          {
            "name": 16,
            "type": {
              "id": 3,
              "displayName": [
                17
              ]
            }
          }
        ],
        "docs": []
      },
      {
        "name": 18,
        "selector": "0x0222ff18",
        "args": [],
        "docs": []
      }
    ],
    "messages": [
      {
        "name": 19,
        "selector": "0x8c97db39",
        "mutates": true,
        "args": [],
        "returnType": null,
        "docs": []
      },
      {
        "name": 20,
        "selector": "0x25444afe",
        "mutates": false,
        "args": [],
        "returnType": {
          "id": 3,
          "displayName": [
            17
          ]
        },
        "docs": []
      }
    ],
    "events": [],
    "docs": []
  }
}
```


Robbepop:
Superseedes https://github.com/paritytech/ink/pull/264

jacogr:
Is it possible to sneak in a version number into the `InkProject` struct? That would be beneficial going forward as well.

Robbepop:
> Is it possible to sneak in a version number into the `InkProject` struct? That would be beneficial going forward as well.

I was talking to Sean Young who implements the Solang Solidity to Wasm compiler and we want to share our metadata format for the contracts.
So for this purpose I think it would also be beneficial to include some kind of "language" property into the `InkProject` struct that is simply `ink` for ink! contracts for a start.

jacogr:
@Robbepop It makes sense, the additional metadata is quite easy to add as a sub-structure, and is certainly very beneficial as we expand.

ascjones:
Should this versioning/language field be done as a separate PR?

Robbepop:
> Should this versioning/language field be done as a separate PR?

Yes, also can you write an issue for that so we can track and discuss there what needs to be added exactly?

jacogr:
Goint through, think this is not correctly named/consistent -

```
"constructors": [
      {
        "name": 36,
        "selector": "0x5ebd88d6",
        "args": [
          {
            "name": 37,
            "type": {
              "ty": 4,
              "displayName": [
                38
              ]
            }
          }
        ],
        "docs": []
      },
```

`ty` above

ascjones:
So should `ty` be `type` then? So it would be `type`.`type`

Robbepop:
> So should `ty` be `type` then? So it would be `type`.`type`

Well `ty` is an ID, so just call it `id`.

ascjones:
> > Should this versioning/language field be done as a separate PR?
> 
> Yes, also can you write an issue for that so we can track and discuss there what needs to be added exactly?

https://github.com/paritytech/ink/issues/299

ascjones:
> Let's merge this as soon as Polkadot JS gave their acknowledgement that their JSON parser can trivially parse the output of this PR.

@jacogr are we good to go on your end?

jacogr:
@ascjones No, I'm not good, atm juggling as well as I can (so did what I could in "off-time"), sadly there are higher priorities API-wise.

ascjones:
> PS: Still need an indication as to by the `Vec<u8>` definition due to the internals as `{ elems: 'Vec<u8>` }` is there.

Not sure exactly what you're referring to here. Is that from a type def - can you point me at it in context?

Robbepop:
> > PS: Still need an indication as to by the `Vec<u8>` definition due to the internals as `{ elems: 'Vec<u8>` }` is there.
> 
> Not sure exactly what you're referring to here. Is that from a type def - can you point me at it in context?

I think this is mainly about some bad misnamings in `type-metadata` crate.

jacogr:
@ascjones So basically in the example you gave above, the following types (parsed), are being exposed -

```
name: `Vec<u8>'
type:  { elems: 'Vec<u8>' }
```

(the type in position 14)

This is used by `ink_core::env2::env_access::mutable::EnvAccessMut` (position 12) which in turn is used by `ink_core::env2::env_access::immutable::EnvAccess` (position 9)

Robbepop:
> 
> 
> @ascjones So basically in the example you gave above, the following types (parsed), are being exposed -
> 
> ```
> name: `Vec<u8>'
> type:  { elems: 'Vec<u8>' }
> ```
> 
> (the type in position 14)
> 
> This is used by `ink_core::env2::env_access::mutable::EnvAccessMut` (position 12) which in turn is used by `ink_core::env2::env_access::immutable::EnvAccess` (position 9)
> 
> So 2 things there -
> 
>     * not convinced the `EnvAccessMut` & `EnvAccess` is needed to be exposed, since they are not used by calls
> 
>     * the `Vec<u8> === { elems: Vec<u8> }` is just weird. The inner elems actually refer to a primitive/correct type definition, not sure why the outer is called `Vec<u8>` as well, also without a namespace in this case.

The reason for the second bullet point is that we have plans to use the metadata information for more contexts than just Polkadot JS UI where we are not only interested in how the types exactly encode and decode using SCALE codec but also in how they are structured. But just plans so far.

jacogr:
I'm happy with that, but the `Vec<u8> = { elems: Vec<u8> }` is just beyond weird. We certainly don't have a `struct Vec<u8> { elems: Vec<u8> }` defined somewhere in the code.

Robbepop:
> I'm happy with that, but the `Vec<u8> = { elems: Vec<u8> }` is just beyond weird. We certainly don't have a `struct Vec<u8> { elems: Vec<u8> }` defined somewhere in the code.
> 
> So if it was namespaced, could "get it" (would still like to see the definition though), however it is not.

Yeah this is what I mean by confusion. We should have called the inner `Vec<u8>` more like `dynarray<u8>` or so in contract to Rust's `array`. So, yes. A bit unfortunate naming but we could fix that.

Robbepop:
Anything against this change from the Polkadot JS side of things? You'd need to implement this. I would like to merge this soon but your implementation has to be done first. @jacogr @kwingram25

jacogr:
The `Vec` thing is still a blocker atm. It destroys the registry by overriding the base `Vec<u8>`. (And not going to hack around it like we did previously, hacks come at a cost)

jacogr:
Is it possible to supply a new incrementer ABI after https://github.com/paritytech/ink/pull/312? (It looks like that may solve the Vec issue... from my reading. At the very least, it helps having latest.)

Robbepop:
> Is it possible to supply a new incrementer ABI after #312? (It looks like that may solve the Vec issue... from my reading. At the very least, it helps having latest.)

PR #312 did not change contract metadata at all. However, there are other PRs in the pipeline that will.

jacogr:
Ok, was hoping with the EnvAccess (which is the root of the issues), the problematic Vec re-assign would go the way of the do-do.

Robbepop:
> Ok, was hoping with the EnvAccess (which is the root of the issues), the problematic Vec re-assign would go the way of the do-do.

Ah if that's what you mean - yes there should no longer be an `EnvAccess` in the metadata.

ascjones:
bors try

ascjones:
Now updated to use `scale-info` with new format. e.g. Flipper:

```
{
  "registry": {
    "strings": [
      "Storage",
      "flipper",
      "__ink_private",
      "__ink_storage",
      "value",
      "Value",
      "ink_core",
      "storage",
      "cell",
      "SyncCell",
      "sync_cell",
      "Key",
      "ink_primitives",
      "key",
      "new",
      "init_value",
      "bool",
      "default",
      "flip",
      "get"
    ],
    "types": [
      {
        "composite": {
          "name": 1,
          "namespace": [
            2,
            2,
            3,
            4
          ],
          "fields": [
            {
              "name": 5,
              "type": 2
            }
          ]
        }
      },
      {
        "composite": {
          "name": 6,
          "namespace": [
            7,
            8,
            5
          ],
          "params": [
            3
          ],
          "fields": [
            {
              "name": 9,
              "type": 4
            }
          ]
        }
      },
      {
        "primitive": "bool"
      },
      {
        "composite": {
          "name": 10,
          "namespace": [
            7,
            8,
            9,
            11
          ],
          "fields": [
            {
              "name": 9,
              "type": 5
            }
          ]
        }
      },
      {
        "composite": {
          "name": 12,
          "namespace": [
            13,
            14
          ],
          "fields": [
            {
              "type": 6
            }
          ]
        }
      },
      {
        "array": {
          "len": 32,
          "type": 7
        }
      },
      {
        "primitive": "u8"
      }
    ]
  },
  "storage": {
    "struct": {
      "type": 1,
      "fields": [
        {
          "name": 5,
          "layout": {
            "struct": {
              "type": 2,
              "fields": [
                {
                  "name": 9,
                  "layout": {
                    "range": {
                      "offset": "0x0000000000000000000000000000000000000000000000000000000000000000",
                      "len": 1,
                      "elemType": 3
                    }
                  }
                }
              ]
            }
          }
        }
      ]
    }
  },
  "contract": {
    "name": 2,
    "constructors": [
      {
        "name": 15,
        "selector": "0x5ebd88d6",
        "args": [
          {
            "name": 16,
            "type": {
              "id": 3,
              "displayName": [
                17
              ]
            }
          }
        ],
        "docs": []
      },
      {
        "name": 18,
        "selector": "0x0222ff18",
        "args": [],
        "docs": []
      }
    ],
    "messages": [
      {
        "name": 19,
        "selector": "0x8c97db39",
        "mutates": true,
        "args": [],
        "returnType": null,
        "docs": []
      },
      {
        "name": 20,
        "selector": "0x25444afe",
        "mutates": false,
        "args": [],
        "returnType": {
          "id": 3,
          "displayName": [
            17
          ]
        },
        "docs": []
      }
    ],
    "events": [],
    "docs": []
  }
}
```

abi/src/layout.rs
```diff
@@ -176,13 +173,12 @@ impl IntoCompact for LayoutField {
...
     /// The single key for cells or the starting key address for chunks.
-    #[serde(rename = "range.offset", serialize_with = "serialize_key")]
+    #[serde(serialize_with = "serialize_key")]
     offset: LayoutKey,
     /// The amount of associated key addresses starting from the offset key.
-    #[serde(rename = "range.len")]
     len: u32,
     /// The element type stored under the associated keys.
-    #[serde(rename = "range.elem_type")]
+    #[serde(rename = "elemType")]
```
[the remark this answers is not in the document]
ascjones: camelCase for direct mapping to javascript types. See original PR https://github.com/type-metadata/type-metadata/pull/36

examples/flipper/.ink/abi_gen/metadata.json
```diff
@@ -0,0 +1,192 @@
+{
```
Robbepop, line 1: This whole file should be removed.

examples/flipper/something/.rustc_info.json
```diff
@@ -0,0 +1 @@
+{"rustc_fingerprint":2833492635811596006,"outputs":{"1164083562126845933":["rustc 1.45.0-nightly (a74d1862d 2020-05-14)\nbinary: rustc\ncommit-hash: a74d1862d4d87a56244958416fd05976c58ca1a8\ncommit-date: 2020-05-14\nhost: x86_64-unknown-linux-gnu\nrelease: 1.45.0-nightly\nLLVM version: 9.0\n",""],"16929990295604814582":["___\n",""],"4476964694761187371":["___\nlib___.rlib\nlib___.so\nlib___.so\nlib___.a\nlib___.so\n/home/andrew/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu\ndebug_assertions\nproc_macro\ntarget_arch=\"x86_64\"\ntarget_endian=\"little\"\ntarget_env=\"gnu\"\ntarget_family=\"unix\"\ntarget_feature=\"fxsr\"\ntarget_feature=\"mmx\"\ntarget_feature=\"sse\"\ntarget_feature=\"sse2\"\ntarget_has_atomic=\"16\"\ntarget_has_atomic=\"32\"\ntarget_has_atomic=\"64\"\ntarget_has_atomic=\"8\"\ntarget_has_atomic=\"ptr\"\ntarget_has_atomic_load_store=\"16\"\ntarget_has_atomic_load_store=\"32\"\ntarget_has_atomic_load_store=\"64\"\ntarget_has_atomic_load_store=\"8\"\ntarget_has_atomic_load_store=\"ptr\"\ntarget_os=\"linux\"\ntarget_pointer_width=\"64\"\ntarget_thread_local\ntarget_vendor=\"unknown\"\nunix\n",""]},"successes":{}}
```
Robbepop, line 1: This file should be removed.

ascjones:
I know the current storage module is being replaced, but I have fixed the compilation errors anyway

ascjones:
Latest incrementer `metadata.json`:
```
{
  "registry": {
    "strings": [
      "incrementer",
      "__ink_private",
      "__ink_storage",
      "Storage",
      "value",
      "ink_core",
      "storage",
      "Value",
      "cell",
      "sync_cell",
      "SyncCell",
      "ink_primitives",
      "key",
      "Key",
      "new",
      "init_value",
      "i32",
      "default",
      "inc",
      "by",
      "get"
    ],
    "types": [
      {
        "path": [
          1,
          1,
          2,
          3,
          4
        ],
        "composite": {
          "fields": [
            {
              "name": 5,
              "type": 2
            }
          ]
        }
      },
      {
        "path": [
          6,
          7,
          5,
          8
        ],
        "params": [
          3
        ],
        "composite": {
          "fields": [
            {
              "name": 9,
              "type": 4
            }
          ]
        }
      },
      {
        "primitive": "i32"
      },
      {
        "path": [
          6,
          7,
          9,
          10,
          11
        ],
        "composite": {
          "fields": [
            {
              "name": 9,
              "type": 5
            }
          ]
        }
      },
      {
        "path": [
          12,
          13,
          14
        ],
        "composite": {
          "fields": [
            {
              "type": 6
            }
          ]
        }
      },
      {
        "array": {
          "len": 32,
          "type": 7
        }
      },
      {
        "primitive": "u8"
      }
    ]
  },
  "storage": {
    "struct": {
      "type": 1,
      "fields": [
        {
          "name": 5,
          "layout": {
            "struct": {
              "type": 2,
              "fields": [
                {
                  "name": 9,
                  "layout": {
                    "range": {
                      "offset": "0x0000000000000000000000000000000000000000000000000000000000000000",
                      "len": 1,
                      "elemType": 3
                    }
                  }
                }
              ]
            }
          }
        }
      ]
    }
  },
  "contract": {
    "name": 1,
    "constructors": [
      {
        "name": 15,
        "selector": "0x5ebd88d6",
        "args": [
          {
            "name": 16,
            "type": {
              "id": 3,
              "displayName": [
                17
              ]
            }
          }
        ],
        "docs": []
      },
      {
        "name": 18,
        "selector": "0x0222ff18",
        "args": [],
        "docs": []
      }
    ],
    "messages": [
      {
        "name": 19,
        "selector": "0xe7d0590f",
        "mutates": true,
        "args": [
          {
            "name": 20,
            "type": {
              "id": 3,
              "displayName": [
                17
              ]
            }
          }
        ],
        "returnType": null,
        "docs": []
      },
      {
        "name": 21,
        "selector": "0x25444afe",
        "mutates": false,
        "args": [],
        "returnType": {
          "id": 3,
          "displayName": [
            17
          ]
        },
        "docs": []
      }
    ],
    "events": [],
    "docs": []
  }
}
```

abi/src/specs.rs
```diff
@@ -830,36 +833,5 @@ fn serialize_selector<S>(s: &[u8; 4], serializer: S) -> Result<S::Ok, S::Error>
...
-
-        // when
-        let json = serde_json::to_string(&cs.into_compact(&mut registry)).unwrap();
-
-        // then
-        assert_eq!(
-            json,
-            r#"{"name":1,"selector":"[\"0x07\",\"0x5B\",\"0xCD\",\"0x15\"]","args":[],"docs":[]}"#
-        );
-    }
```
Robbepop, line 864: Was this test moved or removed?

Issue #297, opened 2019-12-11
Constraining `invoke_runtime` type


ascjones:
In `env2` the function `dispatch_call` was renamed `invoke_runtime`. Previously the type of the call was constrained to `EnvTypes::Call` to ensure that only valid calls could be dispatched:

https://github.com/paritytech/ink/blob/8165b244cec50d6dbf83a147c13c6d863154caf3/model/src/exec_env.rs#L195-L200

The new `invoke_runtime` allows any type which implements `Encode`:

https://github.com/paritytech/ink/blob/b7ca7525563e576ebef41801174c413d43166dc1/core/src/env2/srml/impls.rs#L352-L360

Should we consider replicating the old api whereby the type of the `Call` is restricted? This will guarantee that the `Call` is decodable by the runtime (assuming the types are compatible). 

This was discovered by @liuchengxu: https://github.com/paritytech/ink-types-node-runtime/pull/7.

jaybutera:
I second this, as someone new to the code base it would also make it clear what type needs to be passed in.

Which I assume is now `CallData` https://github.com/paritytech/ink/blob/master/core/src/env2/call/utils.rs#L74

Robbepop:
> I second this, as someone new to the code base it would also make it clear what type needs to be passed in.
> 
> Which I assume is now `CallData` https://github.com/paritytech/ink/blob/master/core/src/env2/call/utils.rs#L74

The `CallData` type is not correct in this context.
We currently do not have a `Call` type in the `EnvTypes` of `env2` in ink! so we'd have to recreate that one.

Robbepop:
This is partially implemented in https://github.com/paritytech/ink/pull/312 and can be closed once the PR has been merged.

PR #302, opened 2019-12-24
[core] Constrain invoke_runtime parameter

jaybutera:
As discussed in #297. Use the same constraint `Into<<Self as EnvTypes>::Call>` as in the first [env/api.rs#L84](https://github.com/paritytech/ink/blob/master/core/src/env/api.rs#L84).

This is also a prerequisite to exposing the `invoke_runtime` method as will be done from https://github.com/paritytech/ink-types-node-runtime/pull/7

### DIFF
--- a/core/src/env2/srml/impls.rs
+++ b/core/src/env2/srml/impls.rs
@@ -349,13 +349,13 @@ where
         ext::deposit_event(topics, data);
     }
 
-    fn invoke_runtime<O, V>(buffer: &mut O, call_data: &V)
+    fn invoke_runtime<O, V>(buffer: &mut O, call_data: V)
     where
         O: scale::Output + AsRef<[u8]> + Reset,
-        V: scale::Encode,
+        V: Into<<Self as EnvTypes>::Call>,
     {
         buffer.reset();
-        call_data.encode_to(buffer);
+        call_data.into().encode_to(buffer);
         ext::dispatch_call(buffer.as_ref());
     }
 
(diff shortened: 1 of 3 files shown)


core/src/env2/srml/impls.rs
```diff
@@ -349,13 +349,13 @@ where
         ext::deposit_event(topics, data);
     }
 
-    fn invoke_runtime<O, V>(buffer: &mut O, call_data: &V)
+    fn invoke_runtime<O, V>(buffer: &mut O, call_data: V)
```
Robbepop: Since we are just encoding the `call_data` parameter which takes a `&self` we do not need to take over ownership of anything. So please change it back to just `&<Self as EnvTypes>::Call`. So we can remove the `V` generic parameter as far as I can see.
Besides this looks good.

jaybutera: Sure thing. I just followed the old convention - figured having it implement Into might fit a general case I hadn't considered.

core/src/env2/srml/impls.rs
```diff
@@ -349,13 +349,12 @@ where
         ext::deposit_event(topics, data);
     }
 
-    fn invoke_runtime<O, V>(buffer: &mut O, call_data: V)
+    fn invoke_runtime<O, V>(buffer: &mut O, call_data: &<Self as EnvTypes>::Call)
```
Robbepop: `V` still here

jaybutera:
@Robbepop If this looks good can we push it through? Just trying to finish for my own use :smile:

PR #310, opened 2020-01-13
Docs Generation

TriplEight:
Generates docs on merges, schedules (if any), tags.
Can add `only: changes: "*.rs"` if needed.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -79,7 +79,7 @@ publish-docs:
     - git fetch
     # - git checkout --track origin/tst-pages
     - git checkout -B origin/tst-pages
-    - git add -f ${CARGO_TARGET_DIR}/docs/*
+    - git add -f ${CARGO_TARGET_DIR}/doc/*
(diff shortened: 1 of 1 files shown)


Robbepop:
> Can add `only: changes: "*.rs"` if needed.

What does this do?

.gitlab-ci.yml
```diff
@@ -195,3 +196,28 @@ examples-generate-metadata:
...
+
+#### stage:                        publish
+
+publish-docs:
+  stage:                           publish
+  <<:                              *docker-env
+  only:
+    - master
+    - schedules
+    - tags
```
Robbepop: Not sure we need this outside merged to `master` tbh

TriplEight:
> What does this do? 

pipeline will only launch if `*.rs` (or any other type, or a dir if you want) files were changed.

.gitlab-ci.yml
TriplEight: - `schedules` can be a nightly job on `master`, we just didn't need it before.
- `tags` will do a documentation update when you release.
Don't you need any of those?

PR #311, opened 2020-01-14
Implement storage (revision 2) module

Robbepop:
# New Contract Initialization & Finalization

## Technical ToDo List

- [ ] Add unit tests for `SpreadLayout` & `PackedLayout` impls of Rust primitives.
- [x] Normalize trait bounds for some storage abstractions.
- [x] Add `expect` for some iterator impls of storage collections.
- [x] Decide about usage of `storage::Pack` in `storage2::stash::Entry` internally.
    - We decided to use `storage::Pack` in the internals of data structures and disallow types `T` that cannot be put into such a `storage::Pack`. Users can opt-into using `storage2::Box` to be able to pack otherwise non-package types. 
- [x] Add `storage2::Bitvec`: required for dynamic storage allocator.
- [x] Dynamic storage allocator.
- [x] ~Make `storage2::Stash` generic over a symbol type.~
    - We opted not to do this for the sake of keeping generated Wasm file sizes low since this could explode in the face of monomorphization.
- [x] Implement missing non-storage traits for `storage2::HashMap`.
- [x] Write unit tests for `storage2::collections::HashMap`.
- [x] Write more unit tests for dynamic storage allocator.
- [x] Write more unit tests for `storage2::collections::Bitvec`.
- [x] Write `SpreadLayout` unit tests for storage collections:
    - [x] `storage::Vec`
    - [x] `storage::SmallVec`
    - [x] `storage::Stash`
    - [x] `storage::HashMap`
    - [x] `storage::Box`
    - [x] `storage::Bitvec`
- [x] Add unit tests for some storage abstractions:
    - [x] `storage::Pack`
    - [x] `storage::Memory`
    - [x] `storage::Box`
- [x] Write unit tests for `lazy` abstractions.
    - [x] `LazyCell`
    - [x] `LazyArray`
    - [x] `LazyIndexMap`
    - [x] `LazyHashMap`
- [x] Make storage collections work on `PackedLayout` elements
    - [x] `Vec`
    - [x] `SmallVec`
    - [x] `Stash` - still uses `Pack<Entry<T>>` internally instead of `Entry<T>`
    - [x] `HashMap` - still uses `Pack<ValueEntry<T>>` internally instead of `ValueEntry<T>`
    - [x] `Bitvec` - still uses `Pack<Bits256>` instead of `Bits256` internally
- [x] Implement `Drop` for `storage` data structures to clean up their associated storage.
    - [x] `Box`
    - [x] `Vec`
    - [x] `Stash`
    - [x] `SmallVec`
    - [x] `HashMap`
- [x] Improve interfaces so that it is possible to avoid the extraneous storage read in `clear_packed_at` implementations of some lazy data structures.
    - [x] `storage2::LazyArray`
    - [x] `storage2::LazyIndexMap`
    - [x] `storage2::LazyHashMap`
    - [x] Implement the trait setting for all built-in types.

## Follow-up Work Items

- [ ] Add `PushOnDrop` abstraction
    - [ ] Make use of `PushOnDrop` for dynamic storage allocator.
    - Rough implementation
        ```rust
        pub struct PushOnDrop<T>
        where
            T: SpreadLayout,
        {
            at: Key,
            what: T,
        }

        impl<T> Drop for PushOnDrop<T>
        where
            T: SpreadLayout,
        {
            fn drop(&mut self) {
                let prevent_drop = ManuallyDrop::new(self.what);
                push_spread_root::<T>(&*prevent_drop, &self.at);
            }
        }
        ```
- [ ] Move `storage2::collections::boxed` into `storage2::alloc`.
- [x] Add `storage2::Memory<T>` to specify entities that are never synchronized with the contract storage. This is useful for intermediate state between messages.
- [x] Implement `Debug` impls that can properly handle the `UnsafeCell` field and displays the cache below.
- [x] Improve API of `lazy::Entry`, e.g. some methods should be `&self` instead of `&mut self` etc.
- [x] Add prefix bytes to `storage::HashMap::to_offset_key` key calculation.
- [x] Turn the data structure underlying to the `DynamicAllocator` for storage into a "generic" data structure usable by other ink! abstractions.
    - [x] Find a proper name for it.
        - [x] `BitStash` maybe?
- [ ] Make `SpreadLayout::pull_spread` and utility functions such as `pull_spread_root` return `Result` instead of panicking.
- [ ] Improve efficiency of finding the first free slot in the dynamic storage allocator. We can do this by using the information in the full mask of a 256-bit block instead of iterating over the `u8` masks which should be much more efficient.
- [ ] Try to remove the code duplication in `LazyIndexMap` and `LazyHashMap`.
- [ ] Implement more operations on storage collections:
    - [ ] `storage::Vec`
        - [ ] `clear(&mut self)`: Empties the storage vector. More efficient than doing it manually. Might be useful for small storage vector instances.
        - [ ] `set(&mut self, index: Index, new_value: T)`: Sets the value of an indexed elements without loading and returning the old value which is more efficient. There currently is no alternative to that.
- [ ] Implement `storage::HashMap` entry API
- [ ] Improve ergonomics for storage trait implementations.
    - [ ] Forward as packed should be simple to implement. (e.g. always packed)
- [ ] Add `storage2::Stash::take_drop`.
    - [ ] Make use of `storage2::Stash::take_drop` in `storage2::HashMap::remove`
- [ ] Smooth edges of initial dynamic storage allocator implementation.
- [x] Write proc. macros for deriving the new storage traits:
    - [x] `SpreadLayout`
    - [x] `PackedLayout`
    - Implemented in follow-up PR: https://github.com/paritytech/ink/pull/399
- [ ] Adjust `ink_lang`
    - [ ] Introduce new `#[ink(constructor)]` syntax and semantics to adjust to the changes.
    - [ ] Adjust test cases for new `#[ink(constructor)]` syntax and semantics
- [ ] Adjust example contracts for the changes.
- [ ] **Bonus:** Refactor `storage::Key` to be based on `u128` instead of `[u8; 32]`.
    - We do this because we hope that we can reduce some computational overhead by this in key offset calculations that have to be done always upon contract initialization or finalization.
- [ ] Remove old `storage` module and replace all usage by the new `storage2` module.
- [ ] Implement missing high-level `storage` data structures:
    - [ ] `VecDeque`: Storage data structure with an API similar to Rust's `VecDeque`.
    - [ ] `BTreeMap`: Re-implementation of old `storage::BTreeMap`.
    - [ ] `BinaryHeap`: Re-implementation of old `storage::BinaryHeap`.
- [ ] Implement utility `storage` abstraction `Trie` that uses storage subtries to pack storage entities.
    - Blocked by the bulk removal feature of storage subtries in Substrate. Read more about this here: https://github.com/paritytech/substrate/issues/5215

## Why?

Currently ink! has some major problems around contract initialization (contract instantiation), contract setup upon a call and finalization (contract tear-down after instantiation or call).

Contract instantiation and the setup phase upon calling a contract are treated the same even though they do completely different tasks and also have different assumptions about the state of the contract storage.

- The contract instantiation has to assume that the contract storage is basically empty.
- The setup phase (upon calling a contract) can actually assume a more or less properly setup contract storage and directly work on that.

Due to this major problem of unifying two concepts that should not have been unified many abstractions in ink! have developed in an incorrect way leading to bad UX and non optimal performance and memory usage characteristics.

## New Design

This PR changes the way how contract instantiation and contract call setup phase are handled and adjusts the contract finalization.

So far `#[ink(constructor)]` messages always had to be `&mut self` and return nothing as if they were constructors of known object oriented languages. However, Rust constructors normally return `Self` and don't have a receiver. This design separated ink! from Rust quite a bit and made it awkward for Rustaceans to "unlearn" constructors.

The new design will treat ink! constructors just like Rust constructors. This has the consequence that you can no longer access the environment via `self` because there is no longer a `self` receiver in an `#[ink(constructor)]`. For that purpose we introduce another universal way of accessing the environment from within `#[ink(constructor)]` and `#[ink(message)]` functions via `Self::env()`.

#### OLD

```rust
#[ink(constructor)]
fn new(&mut self, initial_supply: Balance) {
    self.total_supply.set(initial_supply);
    self.balances.insert(caller, initial_supply);
    self.env().emit_event(Transfer { from: None, to: Some(caller), value: initial_supply })
```

#### NEW

```rust
#[ink(constructor)]
fn new(initial_supply: Balance) -> Self {
    let caller = Self::env().caller();
    let mut balances = storage::BTreeMap::new();
    balances.insert(caller, initial_supply);
    Self::env().emit_event(Transfer { from: None, to: Some(caller), value: initial_supply });
    Self {
        total_supply: storage::Lazy::new(initial_supply),
        balances,
        allowances: storage::BTreeMap::new(),
    }
}
```

## New Storage Primitives

So far ink! storage has been mainly managed by the following primitives:

- `storage::SyncCell`: Low-level primitive to load and store a single cached cell of type `T`.
- `storage::SyncChunk`: Low-level primitive to load and store a chunk of cached and cells of type `T`.
- `storage::Value`: High-level primitive to manage a single `T`.
- `storage::Vec`: High-level primitive to manage a vector of `T`.
And other storage collections similar to `storage::Vec` that manage a whole set of storage entities each living in their own cell.

All of these primitives mainly have a single or multiple `ink_primitives::Key` instances to maintain a mapping to some concrete contract storage entries all the time. This was designed to allow to `flush` the contract storage at any point during contract execution which could be a valuable counter measurement against re-entrancy attacks upon calling another contract. However, this whole mechanic of enforcing all storage types to always have direct pointers into their mapped storage region has significant drawbacks. For once we cannot support Rust primitive types since they themselves cannot hold such a key.

The new storage abstraction data structures would no longer have their respective contract storage region encoded into themselves. Instead they rely on contract initialization and finalization routines to provide them with the correct mapping. So those mappings are now managed by the routines instead of by the data structures.

Furthermore this means that some already existing storage abstractions need adjustments, others might even be removed.

### New Storage Types

The new storage types introduced by this PR design are the following:

| Name | Purpose |
|:-:|:-|
| `storage::Lazy<T>` | Allows to lazily load upon contract calls. |
| `storage::Pack<T>` | Packs `T` into using just a single cell instead of distributing to as many cells as possible. |
| `storage::Box<T>` | Stores an indirection to a dynamically allocated `T` on the storage. |
| `storage::LazyMap<K, V>` | Low-level primitive and a generalization of the concept of a chunk by mapping from `K` instead of an index. |

Removed or deprecated storage types are:

| Name | Reason |
|:-:|:-|
| `storage::SynCell<T>` | Replaced and generalized mainly by `storage::Lazy<T>` |
| `storage::SynChunk<T>` | Replaced and generalized mainly by `storage::LazyMap<T>` |
| `storage::Value<T>` | Used to represent a lazily loaded `T`, either use `T` or `storage::Lazy<T>` |

The other `storage` collections will remain mostly the same but are having adjusted interfaces to meet the new requirements. E.g. they will have proper constructors, e.g. `storage::Vec::new()` and implement the new storage traits.

### Example Use Cases

1. You simply want to store and load a single `i32`.
    - You simply use `i32` as type. It won't be lazily loaded but that wasn't a requirement.

2. You want to have a lazily loadable `i64` value.
    - You use `storage::Lazy<i64>`. It will load lazily but besides that behave the same as a `i64`.

Note that `storage` collections that manage multiple elements are always lazy.
Lazy loading storage primitives necessarily need to have a key somehow to lazily load from when needed.

### Lazy

The structure of `storage::Lazy<T>` is roughly this:

```rust
enum Lazy<T> {
    Lazy(Key), // Has a single key when initialized as lazy.
    Loaded(T),  // Has the T when initialized by value or when already loaded.
}
```

However, due to the fact that we want to access the `T` of a lazy by reference and that the loading of a `Lazy<T>` should require a `&self` receiver instead of a `&mut self` receiver we need a few low-level adjustments to let that happen. Fore more details the reader should look into the source of the PR.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -21,8 +21,6 @@
 //! emulator for simple off-chain testing.
 
 #![cfg_attr(not(feature = "std"), no_std)]
-#![allow(unused_unsafe)]
-// #![deny(unsafe_op_in_unsafe_fn)]
 #![deny(
     missing_docs,
     bad_style,
--- a/core/src/storage2/alloc/allocator.rs
+++ b/core/src/storage2/alloc/allocator.rs
@@ -34,7 +34,7 @@ type Index32 = u8;
 /// The dynamic allocator.
 ///
 /// Manages dynamic storage allocations in a very efficient and economic way.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct DynamicAllocator {
     /// Counter for set bits in a 256-bit chunk of the `free` list.
     ///
@@ -229,7 +229,7 @@ impl DynamicAllocator {
 }
 
 /// Stores the number of set bits for each 256-bits block in a compact `u8`.
-#[derive(Debug, scale::Encode, scale::Decode)]
+#[derive(Debug, PartialEq, Eq, scale::Encode, scale::Decode)]
 struct CountFree {
     /// Set bits per 256-bit chunk.
     counts: [u8; 32],
@@ -289,7 +289,7 @@ impl ::core::ops::IndexMut<u8> for CountFree {
     }
 }
 
-#[derive(Debug, Copy, Clone, scale::Encode, scale::Decode)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, scale::Encode, scale::Decode)]
 pub struct FullMask(u32);
 
 impl Default for FullMask {
--- a/core/src/storage2/alloc/allocator.rs
+++ b/core/src/storage2/alloc/allocator.rs
@@ -52,6 +52,7 @@ pub struct DynamicAllocator {
 
 impl SpreadLayout for CountFree {
     const FOOTPRINT: u64 = 1;
+    const REQUIRES_DEEP_CLEAN_UP: bool = false;
 
     fn pull_spread(ptr: &mut KeyPtr) -> Self {
         forward_pull_packed::<Self>(ptr)
--- a/core/src/storage2/alloc/allocator.rs
+++ b/core/src/storage2/alloc/allocator.rs
@@ -232,36 +232,6 @@ impl Default for CountFree {
     }
 }
 
-impl CountFree {
-    /// Returns an iterator yielding shared references to the set-bit counts.
-    pub fn iter(&self) -> core::slice::Iter<u8> {
-        self.counts.iter()
-    }
-
-    /// Returns an iterator yielding exclusive references to the set-bit counts.
-    pub fn iter_mut(&mut self) -> core::slice::IterMut<u8> {
-        self.counts.iter_mut()
-    }
-}
-
-impl<'a> IntoIterator for &'a CountFree {
-    type Item = &'a u8;
-    type IntoIter = core::slice::Iter<'a, u8>;
-
-    fn into_iter(self) -> Self::IntoIter {
-        self.iter()
-    }
-}
-
-impl<'a> IntoIterator for &'a mut CountFree {
-    type Item = &'a mut u8;
-    type IntoIter = core::slice::IterMut<'a, u8>;
-
-    fn into_iter(self) -> Self::IntoIter {
-        self.iter_mut()
-    }
-}
-
 impl ::core::ops::Index<u8> for CountFree {
     type Output = u8;
 
--- a/core/src/storage2/alloc/allocator.rs
+++ b/core/src/storage2/alloc/allocator.rs
@@ -104,19 +104,6 @@ impl DynamicAllocator {
         }
     }
 
-    /// Resets the dynamic allocator.
-    ///
-    /// # Note
-    ///
-    /// This method is only needed for testing when tools such as `miri` run all
-    /// tests on the same thread in sequence so every test has to reset the
-    /// statically allocated instances like the storage allocator before running.
-    #[cfg(test)]
-    pub fn reset(&mut self) {
-        self.counts = StorageVec::new();
-        self.free = StorageBitvec::new();
-    }
-
     /// Returns the bit position of the first 256-bit chunk with zero bits
     /// in the `free` list.
     ///
--- a/core/src/storage2/alloc/allocator.rs
+++ b/core/src/storage2/alloc/allocator.rs
@@ -23,7 +23,6 @@ use crate::storage2::{
         PackedLayout,
         SpreadLayout,
     },
-    Pack,
     Vec as StorageVec,
 };
 use ink_primitives::Key;
@@ -46,7 +45,7 @@ pub struct DynamicAllocator {
     /// In theory it is possible to search up to 8192 storage cells for free
     /// slots with a single contract storage look-up. By iterating over the 32
     /// `SetBits32` instances of a single instance.
-    counts: StorageVec<Pack<CountFree>>,
+    counts: StorageVec<CountFree>,
     /// Stores a bit for every allocated or free storage cell.
     free: StorageBitvec,
 }
@@ -74,7 +73,7 @@ impl PackedLayout for CountFree {
 }
 
 impl SpreadLayout for DynamicAllocator {
-    const FOOTPRINT: u64 = <StorageVec<Pack<CountFree>> as SpreadLayout>::FOOTPRINT
+    const FOOTPRINT: u64 = <StorageVec<CountFree> as SpreadLayout>::FOOTPRINT
         + <StorageBitvec as SpreadLayout>::FOOTPRINT;
 
     fn pull_spread(ptr: &mut KeyPtr) -> Self {
@@ -127,7 +126,7 @@ impl DynamicAllocator {
     fn position_first_zero(&mut self) -> Option<u64> {
         // Iterate over the `counts` list of a dynamic allocator.
         // The counts list consists of packs of 32 counts per element.
-        for (n, counts) in self.counts.iter_mut().map(Pack::as_inner_mut).enumerate() {
+        for (n, counts) in self.counts.iter_mut().enumerate() {
             if let Some(i) = counts.position_first_zero() {
                 let n = n as u64;
                 let i = i as u64;
@@ -187,7 +186,7 @@ impl DynamicAllocator {
                 // We need to push another counts element.
                 let mut counter = CountFree::default();
                 counter[0_u8] = 1;
-                self.counts.push(Pack::from(counter));
+                self.counts.push(counter);
             }
             // Return the new slot.
             DynamicAllocation(self.free.len() - 1)
--- a/core/src/storage2/alloc/allocator.rs
+++ b/core/src/storage2/alloc/allocator.rs
@@ -14,171 +14,47 @@
 
 use super::DynamicAllocation;
 use crate::storage2::{
-    collections::Bitvec as StorageBitvec,
+    collections::BitStash,
     traits::{
-        forward_clear_packed,
-        forward_pull_packed,
-        forward_push_packed,
         KeyPtr,
-        PackedLayout,
         SpreadLayout,
     },
-    Vec as StorageVec,
 };
-use ink_primitives::Key;
-
-/// An index into one of the 32 `SetBit32` entries.
-type Index32 = u8;
 
 /// The dynamic allocator.
 ///
 /// Manages dynamic storage allocations in a very efficient and economic way.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, Default, PartialEq, Eq)]
 pub struct DynamicAllocator {
-    /// Counter for set bits in a 256-bit chunk of the `free` list.
-    ///
-    /// For every 256-bit chunk stored in `free` stores a `u8` that counts
-    /// the number of set bits in the 256-bit chunk. This information is used
-    /// to compact the information in `free` to make a `first fit` linear
-    /// search for a new free storage slot more scalable.
-    ///
-    /// In theory it is possible to search up to 8192 storage cells for free
-    /// slots with a single contract storage look-up. By iterating over the 32
-    /// `SetBits32` instances of a single instance.
-    counts: StorageVec<CountFree>,
-    /// Stores a bit for every allocated or free storage cell.
-    free: StorageBitvec,
-}
-
-impl SpreadLayout for CountFree {
-    const FOOTPRINT: u64 = 1;
-    const REQUIRES_DEEP_CLEAN_UP: bool = false;
-
-    fn pull_spread(ptr: &mut KeyPtr) -> Self {
-        forward_pull_packed::<Self>(ptr)
-    }
-
-    fn push_spread(&self, ptr: &mut KeyPtr) {
-        forward_push_packed::<Self>(self, ptr)
-    }
-
-    fn clear_spread(&self, ptr: &mut KeyPtr) {
-        forward_clear_packed::<Self>(self, ptr)
-    }
-}
-
-impl PackedLayout for CountFree {
-    fn pull_packed(&mut self, _at: &Key) {}
-    fn push_packed(&self, _at: &Key) {}
-    fn clear_packed(&self, _at: &Key) {}
+    allocations: BitStash,
 }
 
 impl SpreadLayout for DynamicAllocator {
-    const FOOTPRINT: u64 = <StorageVec<CountFree> as SpreadLayout>::FOOTPRINT
-        + <StorageBitvec as SpreadLayout>::FOOTPRINT;
+    const FOOTPRINT: u64 = <BitStash as SpreadLayout>::FOOTPRINT;
 
     fn pull_spread(ptr: &mut KeyPtr) -> Self {
         Self {
-            counts: SpreadLayout::pull_spread(ptr),
-            free: SpreadLayout::pull_spread(ptr),
+            allocations: SpreadLayout::pull_spread(ptr),
         }
     }
 
     fn push_spread(&self, ptr: &mut KeyPtr) {
-        SpreadLayout::push_spread(&self.counts, ptr);
-        SpreadLayout::push_spread(&self.free, ptr);
+        SpreadLayout::push_spread(&self.allocations, ptr);
     }
 
     fn clear_spread(&self, ptr: &mut KeyPtr) {
-        SpreadLayout::clear_spread(&self.counts, ptr);
-        SpreadLayout::clear_spread(&self.free, ptr);
+        SpreadLayout::clear_spread(&self.allocations, ptr);
     }
 }
 
 impl DynamicAllocator {
-    /// Creates a new dynamic storage allocator.
-    pub fn new() -> Self {
-        Self {
-            counts: StorageVec::new(),
-            free: StorageBitvec::new(),
-        }
-    }
-
-    /// Returns the bit position of the first 256-bit chunk with zero bits
-    /// in the `free` list.
-    ///
-    /// Returns the bit position of the first bit in the 256-bit chunk and not
-    /// the chunk position since that's what [`Bitvec::get_chunk`] expects.
-    ///
-    /// Also directly increases the count of the first found free bit chunk.
-    fn position_first_zero(&mut self) -> Option<u64> {
-        // Iterate over the `counts` list of a dynamic allocator.
-        // The counts list consists of packs of 32 counts per element.
-        for (n, counts) in self.counts.iter_mut().enumerate() {
-            if let Some(i) = counts.position_first_zero() {
-                let n = n as u64;
-                let i = i as u64;
-                return Some(n * (32 * 256) + i * 256)
-            }
-        }
-        None
-    }
-
-    /// Returns the number of required counts elements.
-    fn required_counts(&self) -> u32 {
-        let capacity = self.free.capacity();
-        if capacity == 0 {
-            return 0
-        }
-        1 + ((capacity - 1) / (32 * 256)) as u32
-    }
-
     /// Returns a new dynamic storage allocation.
     ///
     /// # Panics
     ///
     /// If the dynamic allocator ran out of free dynamic allocations.
     pub fn alloc(&mut self) -> DynamicAllocation {
-        if let Some(index) = self.position_first_zero() {
-            if index == self.free.len() as u64 {
-                self.free.push(true);
-                return DynamicAllocation(self.free.len() - 1)
-            }
-            let mut bits256 = self
-                .free
-                .get_chunk_mut(index as u32)
-                .expect("must exist if indices have been found");
-            if let Some(first_zero) = bits256.position_first_zero() {
-                bits256
-                    .get_mut(first_zero)
-                    .expect("first zero is invalid")
-                    .set();
-                DynamicAllocation(index as u32 + first_zero as u32)
-            } else {
-                // We found a free storage slot but it isn't within the valid
-                // bounds of the free list but points to its end. So we simply
-                // append another 1 bit (`true`) to the free list and return
-                // a new dynamic allocation pointing to it. No need to push to
-                // the counts list in this case.
-                self.free.push(true);
-                DynamicAllocation(self.free.len() - 1)
-            }
-        } else {
-            // We found no free dynamic storage slot:
-            // Check if we already have allocated too many (2^32) dynamic
-            // storage allocations and panic if that's the case.
-            // Otherwise allocate a new pack of 256-bits in the free list
-            // and mirror it in the counts list.
-            self.free.push(true);
-            if self.counts.len() < self.required_counts() {
-                // We need to push another counts element.
-                let mut counter = CountFree::default();
-                counter[0_u8] = 1;
-                self.counts.push(counter);
-            }
-            // Return the new slot.
-            DynamicAllocation(self.free.len() - 1)
-        }
+        DynamicAllocation(self.allocations.put())
     }
 
     /// Frees the given dynamic storage allocation.
@@ -193,135 +69,15 @@ impl DynamicAllocator {
     /// in the `free` list.
     pub fn free(&mut self, allocation: DynamicAllocation) {
         let index = allocation.get();
-        let mut access = self.free.get_mut(index).expect("index is out of bounds");
-        if !access.get() {
-            // If the dynamic allocation bit is not set this indicates that
-            // the dynamic storage location has already been cleaned up and
-            // this might be a double free.
+        if !self
+            .allocations
+            .take(index)
+            .expect("invalid dynamic storage allocation")
+        {
             panic!(
                 "encountered double free of dynamic storage: at index {}",
                 index
-            );
+            )
         }
-        // Set to `0` (false) which means that this slot is available again.
-        access.reset();
-        // Update the counts list.
-        let counts_id = index / (256 * 32);
-        let byte_id = ((index / 256) % 32) as u8;
-        self.counts
-            .get_mut(counts_id)
-            .expect("invalid counts ID")
-            .dec(byte_id);
-    }
-}
-
-/// Stores the number of set bits for each 256-bits block in a compact `u8`.
-#[derive(Debug, PartialEq, Eq, scale::Encode, scale::Decode)]
-struct CountFree {
-    /// Set bits per 256-bit chunk.
-    counts: [u8; 32],
-    /// Since with `u8` can only count up to 255 but there might be the need
-    /// to count up to 256 bits for 256-bit chunks we need to store one extra
-    /// bit per counter to determine filled chunks.
-    full: FullMask,
-}
-
-impl Default for CountFree {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
-impl ::core::ops::Index<u8> for CountFree {
-    type Output = u8;
-
-    fn index(&self, index: u8) -> &Self::Output {
-        &self.counts[index as usize]
-    }
-}
-
-impl ::core::ops::IndexMut<u8> for CountFree {
-    fn index_mut(&mut self, index: u8) -> &mut Self::Output {
-        &mut self.counts[index as usize]
-    }
-}
-
-#[derive(Debug, Copy, Clone, PartialEq, Eq, scale::Encode, scale::Decode)]
-pub struct FullMask(u32);
-
-impl Default for FullMask {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
-impl FullMask {
-    /// Creates a new full mask with every flag set to `false`.
-    pub fn new() -> Self {
-        Self(0)
-    }
-
-    /// Returns `true` if the 256-bit chunk at the given index is full.
-    pub fn is_full(self, index: u8) -> bool {
-        assert!(index < 32);
-        (self.0 >> (31 - index as u32)) & 0x01 == 1
-    }
-
-    /// Sets the flag for the 256-bit chunk at the given index to `full`.
-    pub fn set_full(&mut self, index: u8) {
-        self.0 |= 1_u32 << (31 - index as u32);
-    }
-
-    /// Resets the flag for the 256-bit chunk at the given index to not `full`.
-    pub fn reset_full(&mut self, index: u8) {
-        self.0 &= !(1_u32 << (31 - index as u32));
-    }
-}
-
-impl CountFree {
-    /// Creates a new 32-entity set bit counter initialized with zeros.
-    pub fn new() -> Self {
-        Self {
-            counts: Default::default(),
-            full: Default::default(),
-        }
-    }
-
-    /// Returns the position of the first free `u8` in the free counts.
-    ///
-    /// Returns `None` if all counts are `0xFF`.
-    pub fn position_first_zero(&mut self) -> Option<u8> {
-        for (i, count) in self.counts.iter_mut().enumerate() {
-            if !self.full.is_full(i as u8) {
-                if *count == !0 {
-                    self.full.set_full(i as u8);
-                } else {
-                    *count += 1;
-                }
-                return Some(i as u8)
-            }
-        }
-        None
-    }
-
-    /// Increases the number of set bits for the given index.
-    ///
-    /// Returns the new number of set bits.
-    ///
-    /// # Panics
-    ///
-    /// - If the given index is out of bounds.
-    /// - If the increment would cause an overflow.
-    pub fn dec(&mut self, index: Index32) -> u8 {
-        assert!(index < 32, "index is out of bounds");
-        if self.full.is_full(index) {
-            self.full.reset_full(index);
-        } else {
-            let new_value = self.counts[index as usize]
-                .checked_sub(1)
-                .expect("set bits decrement overflowed");
-            self.counts[index as usize] = new_value;
-        }
-        self.counts[index as usize]
     }
 }
--- a/core/src/storage2/alloc/init.rs
+++ b/core/src/storage2/alloc/init.rs
@@ -77,7 +77,9 @@ cfg_if! {
         pub fn initialize_for(phase: ContractPhase) {
             let instance = unsafe { &mut GLOBAL_INSTANCE };
             // We do not allow reinitialization for Wasm targets for performance reasons.
-            assert!(!instance.is_initialized());
+            if let DynamicAllocatorState::Initialized(_) = instance {
+                panic!("cannot reinitialize dynamic storage allocator instance in Wasm");
+            }
             *instance = phase.into();
         }
 
@@ -91,13 +93,13 @@ cfg_if! {
                 DynamicAllocatorState::UninitDeploy => {
                     let mut allocator = DynamicAllocator::new();
                     let result = f(&mut allocator);
-                    *instance = allocator;
+                    *instance = DynamicAllocatorState::Initialized(allocator);
                     result
                 }
                 DynamicAllocatorState::UninitCall => {
                     let mut allocator = pull_spread_root::<DynamicAllocator>(&Key(DYNAMIC_ALLOCATOR_KEY_OFFSET));
                     let result = f(&mut allocator);
-                    *instance = allocator;
+                    *instance = DynamicAllocatorState::Initialized(allocator);
                     result
                 }
                 DynamicAllocatorState::Initialized(ref mut allocator) => {
--- a/core/src/storage2/alloc/init.rs
+++ b/core/src/storage2/alloc/init.rs
@@ -1,0 +1,155 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::DynamicAllocator;
+use cfg_if::cfg_if;
+use crate::storage2::traits::pull_spread_root;
+use ink_primitives::Key;
+
+/// The default dynamic allocator key offset.
+///
+/// This is where the dynamic allocator is stored on the contract storage.
+const DYNAMIC_ALLOCATOR_KEY_OFFSET: [u8; 32] = [0xFE; 32];
+
+/// The phase in which a contract execution can be.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum ContractPhase {
+    /// Initializes the global dynamic storage allocator from scratch.
+    ///
+    /// Upon initialization it will be created from scratch as if the
+    /// contract has been deployed for the first time.
+    Deploy,
+    /// Initializes the global dynamic storage allocator from storage.
+    ///
+    /// Upon initialization the dynamic storage allocator will be pulled
+    /// from the contract storage with the assumption that a former
+    /// contract deployment has already taken place in the past.
+    Call,
+}
+
+/// The state of the dynamic allocator global instance.
+#[derive(Debug)]
+enum DynamicAllocatorState {
+    /// The global instance has not yet been initialized.
+    ///
+    /// Upon initialization it will be created from scratch as if the
+    /// contract has been deployed for the first time.
+    UninitDeploy,
+    /// The global instance has not yet been initialized.
+    ///
+    /// Upon initialization it will be pulled from the contract storage
+    /// with the assumption that a former contract deployment has already
+    /// taken place in the past.
+    UninitCall,
+    /// The global instance has already been initialized successfully.
+    Initialized(DynamicAllocator),
+}
+
+impl From<ContractPhase> for DynamicAllocatorState {
+    fn from(phase: ContractPhase) -> Self {
+        match phase {
+            ContractPhase::Deploy => DynamicAllocatorState::UninitDeploy,
+            ContractPhase::Call => DynamicAllocatorState::UninitCall,
+        }
+    }
+}
+
+cfg_if! {
+    if #[cfg(all(not(feature = "std"), target_arch = "wasm32"))] {
+        // Procedures for the Wasm compilation:
+
+        /// The global instance for the dynamic storage allocator.
+        static mut GLOBAL_INSTANCE: DynamicAllocatorState = DynamicAllocatorState::UninitDeploy;
+
+        /// Commands the (re-)initialization of the global instance for the dynamic
+        /// storage allocator.
+        pub fn initialize_for(phase: ContractPhase) {
+            let instance = unsafe { &mut GLOBAL_INSTANCE };
+            // We do not allow reinitialization for Wasm targets for performance reasons.
+            assert!(!instance.is_initialized());
+            *instance = phase.into();
+        }
+
+        /// Runs the given closure on the global instance for the dynamic storage allocator.
+        pub fn on_call<F, R>(f: F) -> R
+        where
+            F: FnOnce(&mut DynamicAllocator) -> R,
+        {
+            let instance = unsafe { &mut GLOBAL_INSTANCE };
+            match instance {
+                DynamicAllocatorState::UninitDeploy => {
+                    let mut allocator = DynamicAllocator::new();
+                    let result = f(&mut allocator);
+                    *instance = allocator;
+                    result
+                }
+                DynamicAllocatorState::UninitCall => {
+                    let mut allocator = pull_spread_root::<DynamicAllocator>(&Key(DYNAMIC_ALLOCATOR_KEY_OFFSET));
+                    let result = f(&mut allocator);
+                    *instance = allocator;
+                    result
+                }
+                DynamicAllocatorState::Initialized(ref mut allocator) => {
+                    f(allocator)
+                }
+            }
+        }
+
+    } else if #[cfg(feature = "std")] {
+        // Procedures for the off-chain environment and testing compilation:
+
+        use ::core::cell::RefCell;
+        thread_local!(
+            /// The global instance for the dynamic storage allocator.
+            static GLOBAL_INSTANCE: RefCell<DynamicAllocatorState> = RefCell::new(DynamicAllocatorState::UninitDeploy);
+        );
+
+        /// Commands the (re-)initialization of the global instance for the dynamic
+        /// storage allocator.
+        pub fn initialize_for(phase: ContractPhase) {
+            GLOBAL_INSTANCE.with(|instance| instance.replace_with(|_| phase.into()));
+        }
+
+        /// Runs the given closure on the global instance for the dynamic storage allocator.
+        pub fn on_call<F, R>(f: F) -> R
+        where
+            F: FnOnce(&mut DynamicAllocator) -> R,
+        {
+            GLOBAL_INSTANCE.with(|instance| {
+                match &mut *instance.borrow_mut() {
+                    instance @ DynamicAllocatorState::UninitDeploy => {
+                        let mut allocator = DynamicAllocator::new();
+                        let result = f(&mut allocator);
+                        *instance = DynamicAllocatorState::Initialized(allocator);
+                        result
+                    }
+                    instance @ DynamicAllocatorState::UninitCall => {
+                        let mut allocator = pull_spread_root::<DynamicAllocator>(&Key(DYNAMIC_ALLOCATOR_KEY_OFFSET));
+                        let result = f(&mut allocator);
+                        *instance = DynamicAllocatorState::Initialized(allocator);
+                        result
+                    }
+                    DynamicAllocatorState::Initialized(instance) => {
+                        f(instance)
+                    }
+                }
+            })
+        }
+
+    } else {
+        compile_error! {
+            "ink! only support compilation as `std` or `no_std` + `wasm32-unknown`"
+        }
+    }
+}
--- a/core/src/storage2/alloc/init.rs
+++ b/core/src/storage2/alloc/init.rs
@@ -121,7 +121,9 @@ cfg_if! {
         /// Commands the (re-)initialization of the global instance for the dynamic
         /// storage allocator.
         pub fn initialize_for(phase: ContractPhase) {
-            GLOBAL_INSTANCE.with(|instance| instance.replace_with(|_| phase.into()));
+            GLOBAL_INSTANCE.with(|instance| {
+                core::mem::forget(instance.replace_with(|_| phase.into()))
+            });
         }
 
         /// Runs the given closure on the global instance for the dynamic storage allocator.
--- a/core/src/storage2/alloc/init.rs
+++ b/core/src/storage2/alloc/init.rs
@@ -122,7 +122,7 @@ cfg_if! {
         /// storage allocator.
         pub fn initialize_for(phase: ContractPhase) {
             GLOBAL_INSTANCE.with(|instance| {
-                core::mem::forget(instance.replace_with(|_| phase.into()))
+                instance.replace_with(|_| phase.into())
             });
         }
 
--- a/core/src/storage2/alloc/init.rs
+++ b/core/src/storage2/alloc/init.rs
@@ -92,7 +92,7 @@ cfg_if! {
             let instance = unsafe { &mut GLOBAL_INSTANCE };
             match instance {
                 DynamicAllocatorState::UninitDeploy => {
-                    let mut allocator = DynamicAllocator::new();
+                    let mut allocator = DynamicAllocator::default();
                     let result = f(&mut allocator);
                     *instance = DynamicAllocatorState::Initialized(allocator);
                     result
--- a/core/src/storage2/alloc/init.rs
+++ b/core/src/storage2/alloc/init.rs
@@ -13,8 +13,8 @@
 // limitations under the License.
 
 use super::DynamicAllocator;
-use cfg_if::cfg_if;
 use crate::storage2::traits::pull_spread_root;
+use cfg_if::cfg_if;
 use ink_primitives::Key;
 
 /// The default dynamic allocator key offset.
--- a/core/src/storage2/alloc/init.rs
+++ b/core/src/storage2/alloc/init.rs
@@ -134,7 +134,7 @@ cfg_if! {
             GLOBAL_INSTANCE.with(|instance| {
                 match &mut *instance.borrow_mut() {
                     instance @ DynamicAllocatorState::UninitDeploy => {
-                        let mut allocator = DynamicAllocator::new();
+                        let mut allocator = DynamicAllocator::default();
                         let result = f(&mut allocator);
                         *instance = DynamicAllocatorState::Initialized(allocator);
                         result
--- a/core/src/storage2/alloc/init.rs
+++ b/core/src/storage2/alloc/init.rs
@@ -40,6 +40,7 @@ pub enum ContractPhase {
 
 /// The state of the dynamic allocator global instance.
 #[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
 enum DynamicAllocatorState {
     /// The global instance has not yet been initialized.
     ///
--- a/core/src/storage2/collections/bitstash/mod.rs
+++ b/core/src/storage2/collections/bitstash/mod.rs
@@ -1,0 +1,174 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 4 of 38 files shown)


core/src/storage/lazy_chunk.rs
```diff
@@ -0,0 +1,393 @@
...
+    fn for_cached_entries<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce(&mut EntryMap<T>) -> R,
+    {
+        // SAFETY: We operate on an exclusive reference on `BTreeMap` within the
+        //         given closure while our method receiver is only a shared reference.
+        //         However, due to encapsulation of the exclusive reference within
+        //         the given closure we cannot leak the exclusive reference outside
+        //         of the closure. So the below action is safe in this regard.
+        f(unsafe { &mut *self.cached_entries.get() })
```
0x7CFE: Could this code be simultaneously called from several places? Like threads or async functions. How do we protect against such scenario?

Robbepop: All ink! code is always executed in a single thread. So no.

core/src/storage/lazy.rs
```diff
@@ -0,0 +1,344 @@
...
+    /// Performs the given closure on the mutable lazy kind.
+    ///
+    /// # Note
+    ///
+    /// Actions on the mutable lazy kind are performed within the closure
+    /// to not leak exclusive references to it to the outside. This is important
+    /// since the `for_kind` method itself operates only on `&self`.
+    fn for_kind<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce(&mut LazyKind<T>) -> R,
```
rphmeier: An explicit lifetime annotation would be a good idea here, to ensure it's not setting the lifetime equal to `&'_ self`, which would allow escape.

```suggestion
        F: for<'a> FnOnce(&'a mut LazyKind<T>) -> R,
```

core/src/storage/lazy_chunk.rs
```diff
@@ -0,0 +1,393 @@
...
+    /// Performs the given closure on the mutable cached entries.
+    ///
+    /// # Note
+    ///
+    /// Actions on the mutable lazy entries are performed within the closure
+    /// to not leak exclusive references to it to the outside. This is important
+    /// since the `for_entries` method itself operates only on `&self`.
+    fn for_cached_entries<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce(&mut EntryMap<T>) -> R,
```
rphmeier:
```suggestion
        F: for<'a> FnOnce(&'a mut EntryMap<T>) -> R,
```

core/src/storage/lazy_chunk.rs
```diff
@@ -0,0 +1,393 @@
...
+                let mut ptr = KeyPtr::from(offset);
+                Push::push(&**entry, &mut ptr);
+            }
+        });
+    }
+}
+
+impl<T> LazyChunk<T>
+where
+    T: Unpin + scale::Decode,
```
tomaka: You don't need `Unpin` because you never prove to `T` that you have pinned it. You never create a `Pin<&mut T>` or `Pin<&T>` or anything similar anywhere.

core/src/storage/lazy_chunk.rs
tomaka: I'm really not sure that this is actually safe.
I don't know why, but it's very suspicious that `Cell` doesn't provide a API that provides a closure taking a mutable reference to its content, and indeed it seems that it's wrong to do so: https://github.com/rust-lang/rust/issues/50186#issuecomment-411169568

tomaka: It is very likely however that this isn't safe only because it's possible to do something very stupid that no-one would reasonably do.

rphmeier: But, this function should itself be `unsafe` because it is not safe to do 

```
let x = chunk.get(index);
chunk.for_cached_entries(|map| map.remove(&index));
// *x points to dropped data now.
```

Robbepop: I think I will simply remove the function since it provides no safe interface.

Robbepop: > But, this function should itself be `unsafe` because it is not safe to do
> 
> ```
> let x = chunk.get(index);
> chunk.for_cached_entries(|map| map.remove(&index));
> // *x points to dropped data now.
> ```

Wow ... how could I not see this immediately. O.O

Robbepop: The helper function is completely broke: https://github.com/paritytech/ink/pull/311#discussion_r375966554
I will just remove it. Thanks for the pointers.

core/src/storage/lazy_chunk.rs
Robbepop: @tomaka Actually Rust compiler thinks that I need it:

```
error[E0596]: cannot borrow data in a dereference of `core::pin::Pin<env::types::_::_parity_scale_codec::alloc::boxed::Box<storage::lazy_chunk::Entry<T>>>` as mutable
   --> core/src/storage/lazy_chunk.rs:248:50
    |
248 |             BTreeMapEntry::Occupied(occupied) => &mut **occupied.into_mut(),
    |                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot borrow as mutable
    |
    = help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `core::pin::Pin<env::types::_::_parity_scale_codec::alloc::boxed::Box<storage::lazy_chunk::Entry<T>>>`

error[E0596]: cannot borrow data in a dereference of `core::pin::Pin<env::types::_::_parity_scale_codec::alloc::boxed::Box<storage::lazy_chunk::Entry<T>>>` as mutable
   --> core/src/storage/lazy_chunk.rs:256:17
    |
256 | /                 &mut **vacant.insert(Box::pin(Entry {
257 | |                     value,
258 | |                     mutated: core::cell::Cell::new(false),
259 | |                 }))
    | |___________________^ cannot borrow as mutable
    |
    = help: trait `DerefMut` is required to modify through a dereference, but it is not implemented for `core::pin::Pin<env::types::_::_parity_scale_codec::alloc::boxed::Box<storage::lazy_chunk::Entry<T>>>`
```

core/src/storage/lazy.rs
```diff
@@ -0,0 +1,344 @@
...
+    fn for_kind<F, R>(&self, f: F) -> R
+    where
+        F: for<'a> FnOnce(&'a mut LazyKind<T>) -> R,
+    {
+        // SAFETY: We operate on an exclusive reference on `LazyKind` within the
+        //         given closure while our method receiver is only a shared reference.
+        //         However, due to encapsulation of the exclusive reference within
+        //         the given closure we cannot leak the exclusive reference outside
+        //         of the closure. So the below action is safe in this regard.
+        f(unsafe { &mut *self.kind.get() })
```
pepyakin: Is this an adequate proof though? I think you can still get a mutable reference, for example:

```rust
lazy.for_kind(|a| lazy.for_kind(|b| {}));
```

core/src/storage/lazy_chunk.rs
tomaka: `impl<T> Unpin for Entry<T> {}`

Or just don't use `Pin<Box<>>` but just `Box<>`.

core/src/storage/lazy.rs
Robbepop: Yep, completely broken ... removed the internal utility function.

core/src/storage/traits/push.rs
```diff
@@ -0,0 +1,304 @@
...
+                // is used even though we have tried to prevent this at compile
+                // time.
+                if <T as StorageSize>::SIZE > 32 {
+                    return
+                }
+                // In theory we'd need to clear all subfields of the `Option<T>`
+                // which are technically `<T as StorageSize>::SIZE` many.
+                // However, this could be a performance hazard so we should
+                // really not do that. We should add a check to disable this
+                // implementation for `StorageSize` of more than 32. (TODO)
```
ascjones: TODO in future or as part of this PR?

core/src/storage/lazy/lazy_chunk.rs
```diff
@@ -0,0 +1,384 @@
...
+            let offset: Key = key + (index as u64 * elem_size);
+            let mut ptr = KeyPtr::from(offset);
+            PushForward::push_forward(&**entry, &mut ptr);
+        }
+    }
+}
+
+impl<T> LazyChunk<T>
+where
+    T: scale::Decode,
```
ascjones: `StorageSize + PullForward`?

core/src/storage/lazy/lazy_chunk.rs
```diff
@@ -0,0 +1,384 @@
...
+        //         `BTreeMap` will never invalidate those references.
+        //         By returning a raw pointer we enforce an `unsafe` block at
+        //         the caller site to underline that guarantees are given by the
+        //         caller.
+        let cached_entries = unsafe { &mut *self.cached_entries.get() };
+        use ink_prelude::collections::btree_map::Entry as BTreeMapEntry;
+        match cached_entries.entry(index) {
+            BTreeMapEntry::Occupied(occupied) => &mut **occupied.into_mut(),
+            BTreeMapEntry::Vacant(vacant) => {
+                let value = match crate::env::get_contract_storage::<T>(key + index) {
```
ascjones: `pull_forward`

core/src/storage/lazy/lazy_chunk.rs
```diff
@@ -0,0 +1,384 @@
...
+    ///
+    /// # Panics
+    ///
+    /// - If the lazy chunk is in an invalid state that forbids interaction.
+    /// - If the lazy chunk is not in a state that allows lazy loading.
+    fn lazily_load_mut(&mut self, index: Index) -> &mut Entry<T> {
+        // SAFETY: Dereferencing the raw-pointer here is safe since we
+        //         encapsulated this whole call with a `&mut self` receiver.
+        //         Also `Entry` instances are all pinned, so mutating the
+        //         `BTreeMap` that stores tham won't invalidate references
```
ascjones:
```suggestion
        //         `BTreeMap` that stores them won't invalidate references
```

core/src/storage/lazy/lazy_chunk.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// Returns the freshly loaded or already loaded entry of the value.
+    ///
+    /// # Panics
+    ///
+    /// - If the lazy chunk is in an invalid state that forbids interaction.
+    /// - If the lazy chunk is not in a state that allows lazy loading.
+    fn lazily_load_mut(&mut self, index: Index) -> &mut Entry<T> {
+        // SAFETY: Dereferencing the raw-pointer here is safe since we
+        //         encapsulated this whole call with a `&mut self` receiver.
+        //         Also `Entry` instances are all pinned, so mutating the
```
ascjones: Pinned how? The type is now `BTreeMap<Index, Box<Entry<T>>>`

core/src/storage/lazy/lazy_chunk.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// - Use [`LazyChunk::put`]`(None)` in order to remove an element.
+    /// - Prefer this method over [`LazyChunk::put_get`] if you are not interested
+    ///   in the old value of the same cell index.
+    ///
+    /// # Panics
+    ///
+    /// - If the lazy chunk is in an invalid state that forbids interaction.
+    /// - If the decoding of the old element at the given index failed.
+    pub fn put(&mut self, _index: Index, _new_value: Option<T>) {
+        todo!()
```
ascjones: I guess we want to implement this

core/src/storage/traits/push.rs
Robbepop: That's a good question. I hope I can find a proper solution before we merge this PR.

core/src/storage/lazy/lazy_chunk.rs
Robbepop: I am currently working that out. Not yet done.

core/src/storage/lazy/lazy_chunk.rs
Robbepop: They _were_ pinned but now only the `Box` and the handling around it makes sure that elements are not moved.

core/src/storage/lazy/mod.rs
```diff
@@ -0,0 +1,248 @@
+// Copyright 2018-2019 Parity Technologies (UK) Ltd.
```
cmichi:
```suggestion
// Copyright 2018-2020 Parity Technologies (UK) Ltd.
```

core/src/storage/lazy/lazy_chunk.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+impl<T> PullForward for LazyChunk<T>
+where
+    T: StorageSize,
+{
+    fn pull_forward(ptr: &mut KeyPtr) -> Self {
+        Self {
+            key: Some(ptr.next_for::<Self>()),
+            cached_entries: UnsafeCell::new(BTreeMap::new()),
```
cmichi: Nice to see it used :+1:

core/src/storage/lazy/lazy_chunk.rs
```diff
@@ -0,0 +1,384 @@
...
+/// A lazy storage chunk that spans over a whole chunk of storage cells.
+///
+/// # Note
+///
+/// This is mainly used as low-level storage primitives by other high-level
+/// storage primitives in order to manage the contract storage for a whole
+/// chunk of storage cells.
+///
+/// A chunk of storage cells is a contiguous range of 2^32 storage cells.
+#[derive(Debug, Default)]
```
cmichi:
```suggestion
#[derive(Default)]
```

cmichi: There are a number of other `Debug` derives in the PR, I haven't annotated them in the review, but you might want to do a full text search.

core/src/storage/traits/pull.rs
```diff
@@ -0,0 +1,274 @@
...
+    T: PullAt,
+{
+    fn pull_at(at: Key) -> Self {
+        Self::new(<T as PullAt>::pull_at(at))
+    }
+}
+
+const _: () = {
+    use ink_prelude::collections::{
+        BTreeSet,
```
cmichi: I can't seem to find the `BTreeSet` module? Did you commit it?

core/src/storage/traits/push.rs
```diff
@@ -0,0 +1,304 @@
...
+macro_rules! impl_push_for_array {
+    ( $($len:literal),* $(,)? ) => {
+        $(
+            impl<T> PushForward for [T; $len]
+            where
+                Self: ArrayLenLessEquals32,
+                T: PushForward,
+            {
+                fn push_forward(&self, ptr: &mut KeyPtr) {
+                    // TODO: Insert check to assert that the array length is
```
cmichi: ToDo for this PR or follow-up?

Robbepop: The PR is not yet done so all TODOs will be worked on before merge.

core/src/storage/traits/pull.rs
Robbepop: This is the Rust `alloc` crate `BTreeSet` not ink!'s `storage::BTreeMap`

core/src/storage/lazy/lazy_chunk.rs
Robbepop: I have to think about the usefulness of those `Debug` derives in the context of our new off-chain environment. Maybe it actually makes sense to have it since we can assume dead code elimination for `no_std` (e.g. Wasm) builds.

core/src/storage2/lazy/lazy_array.rs
```diff
@@ -0,0 +1,444 @@
...
+    ///
+    /// If any of the given indices is out of bounds.
+    pub fn swap(&mut self, a: Index, b: Index) {
+        if a == b {
+            // Bail out early if both indices are the same.
+            return
+        }
+        assert!(a < Self::capacity(), "a is out of bounds");
+        assert!(b < Self::capacity(), "b is out of bounds");
+        let (loaded_a, loaded_b) =
```
Robbepop, line 426: `cargo miri test` reports undefined behaviour at this line of code.
I have already checked it and am pretty confident about it since it is very similar to the `swap` implementation of `storage2::Vec` and that one doesn't trigger undefined behavior according to `miri`.
It would be very nice if reviewers could be extra careful about this code to spot potential pitfalls.

The `miri` error is the following:

```
error: Undefined Behavior: trying to reborrow for Unique, but parent tag <2220142> does not have an appropriate item in the borrow stack
   --> core/src/storage2/lazy/lazy_array.rs:426:14
    |
426 |         let (loaded_a, loaded_b) =
    |              ^^^^^^^^ trying to reborrow for Unique, but parent tag <2220142> does not have an appropriate item in the borrow stack
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
    = note: inside `storage2::lazy::lazy_array::LazyArray::<u8, typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UTerm, typenum::bit::B1>, typenum::bit::B0>, typenum::bit::B0>>::swap` at core/src/storage2/lazy/lazy_array.rs:426:14
note: inside `storage2::collections::smallvec::SmallVec::<u8, typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UTerm, typenum::bit::B1>, typenum::bit::B0>, typenum::bit::B0>>::swap` at core/src/storage2/collections/smallvec/mod.rs:247:9
   --> core/src/storage2/collections/smallvec/mod.rs:247:9
    |
247 |         self.elems.swap(a, b)
    |         ^^^^^^^^^^^^^^^^^^^^^
note: inside `storage2::collections::smallvec::tests::swap_works` at core/src/storage2/collections/smallvec/tests.rs:273:5
   --> core/src/storage2/collections/smallvec/tests.rs:273:5
    |
273 |     vec.swap(0, 1);
    |     ^^^^^^^^^^^^^^
note: inside closure at core/src/storage2/collections/smallvec/tests.rs:262:1
   --> core/src/storage2/collections/smallvec/tests.rs:262:1
    |
262 | / fn swap_works() {
263 | |     let elems = [b'A', b'B', b'C', b'D'];
264 | |     let mut vec = vec_from_slice(&elems);
265 | |
...   |
280 | |     assert_eq_slice(&vec, &[b'C', b'A', b'D', b'B']);
281 | | }
    | |_^
    = note: inside `<[closure@core/src/storage2/collections/smallvec/tests.rs:262:1: 281:2] as std::ops::FnOnce<()>>::call_once - shim` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/ops/function.rs:232:5
    = note: inside `<fn() as std::ops::FnOnce<()>>::call_once - shim(fn())` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/ops/function.rs:232:5
    = note: inside `test::__rust_begin_short_backtrace::<fn()>` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libtest/lib.rs:517:5
    = note: inside closure at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libtest/lib.rs:508:30
    = note: inside `<[closure@DefId(38:631 ~ test[140d]::run_test[0]::{{closure}}[2]) 0:fn()] as std::ops::FnOnce<()>>::call_once - shim(vtable)` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/ops/function.rs:232:5
    = note: inside `<std::boxed::Box<dyn std::ops::FnOnce() + std::marker::Send> as std::ops::FnOnce<()>>::call_once` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/liballoc/boxed.rs:1008:9
    = note: inside `<std::panic::AssertUnwindSafe<std::boxed::Box<dyn std::ops::FnOnce() + std::marker::Send>> as std::ops::FnOnce<()>>::call_once` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/panic.rs:318:9
    = note: inside `std::panicking::r#try::do_call::<std::panic::AssertUnwindSafe<std::boxed::Box<dyn std::ops::FnOnce() + std::marker::Send>>, ()>` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/panicking.rs:331:40
    = note: inside `std::panicking::r#try::<(), std::panic::AssertUnwindSafe<std::boxed::Box<dyn std::ops::FnOnce() + std::marker::Send>>>` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/panicking.rs:274:15
    = note: inside `std::panic::catch_unwind::<std::panic::AssertUnwindSafe<std::boxed::Box<dyn std::ops::FnOnce() + std::marker::Send>>, ()>` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/panic.rs:394:14
    = note: inside `test::run_test_in_process` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libtest/lib.rs:541:18
    = note: inside closure at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libtest/lib.rs:450:39
    = note: inside `test::run_test::run_test_inner` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libtest/lib.rs:475:13
    = note: inside `test::run_test` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libtest/lib.rs:505:28
    = note: inside `test::run_tests::<[closure@DefId(38:230 ~ test[140d]::console[0]::run_tests_console[0]::{{closure}}[2]) 0:&mut test::console::ConsoleTestState, 1:&mut std::boxed::Box<dyn test::formatters::OutputFormatter>]>` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libtest/lib.rs:284:13
    = note: inside `test::run_tests_console` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libtest/console.rs:280:5
    = note: inside `test::test_main` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libtest/lib.rs:120:15
    = note: inside `test::test_main_static` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libtest/lib.rs:139:5
    = note: inside `main`
    = note: inside closure at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/rt.rs:67:34
    = note: inside closure at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/rt.rs:52:73
    = note: inside `std::sys_common::backtrace::__rust_begin_short_backtrace::<[closure@DefId(1:6025 ~ std[1995]::rt[0]::lang_start_internal[0]::{{closure}}[0]::{{closure}}[0]) 0:&dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe], i32>` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/sys_common/backtrace.rs:130:5
    = note: inside closure at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/rt.rs:52:13
    = note: inside `std::panicking::r#try::do_call::<[closure@DefId(1:6024 ~ std[1995]::rt[0]::lang_start_internal[0]::{{closure}}[0]) 0:&&dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe], i32>` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/panicking.rs:331:40
    = note: inside `std::panicking::r#try::<i32, [closure@DefId(1:6024 ~ std[1995]::rt[0]::lang_start_internal[0]::{{closure}}[0]) 0:&&dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe]>` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/panicking.rs:274:15
    = note: inside `std::panic::catch_unwind::<[closure@DefId(1:6024 ~ std[1995]::rt[0]::lang_start_internal[0]::{{closure}}[0]) 0:&&dyn std::ops::Fn() -> i32 + std::marker::Sync + std::panic::RefUnwindSafe], i32>` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/panic.rs:394:14
    = note: inside `std::rt::lang_start_internal` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/rt.rs:51:25
    = note: inside `std::rt::lang_start::<()>` at /home/robrob/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/rt.rs:67:5
    = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to previous error
```

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -0,0 +1,421 @@
...
+                    // We need to load the entity before we remove its associated contract storage
+                    // because it requires a deep clean-up which propagates clearing to its fields,
+                    // for example in the case of `T` being a `storage::Box`.
+                    clear_spread_root_opt::<T>(entry.value().into(), key)
+                } else {
+                    // The type does not require deep clean-up so we can simply clean-up
+                    // its associated storage cell and be done without having to load it first.
+                    let footprint = <T as SpreadLayout>::FOOTPRINT;
+                    assert!(
+                        footprint <= 32,
```
ascjones, line 122: Similar to fragment in `optspec::clear_associated_storage_cells` except that is `footprint < 32` is that intentional? Also should this limit be documented for the `FOOTPRINT` trait constant?

core/src/storage2/traits/impls/prims.rs
```diff
@@ -0,0 +1,235 @@
...
+            }
+            None => {
+                true.push_spread(ptr);
+            }
+        }
+    }
+
+    fn clear_spread(&self, ptr: &mut KeyPtr) {
+        if let Some(value) = self {
+            ptr.advance_by(1);
```
ascjones, line 81: As mentioned in IM, should we be clearing the bool prefix here

core/src/storage2/traits/impls/prims.rs
```diff
@@ -0,0 +1,235 @@
...
+            }
+            Err(error) => {
+                false.push_spread(ptr);
+                <E as SpreadLayout>::push_spread(error, ptr);
+            }
+        }
+    }
+
+    fn clear_spread(&self, ptr: &mut KeyPtr) {
+        ptr.advance_by(1);
```
ascjones, line 155: Should clear bool flag here as in Option<T>?

core/src/storage2/lazy/lazy_cell.rs
Robbepop: good found! I wanted to refactor clean up of storage anyways so I now just did. Resolved.

core/src/lib.rs
```diff
@@ -21,8 +21,12 @@
 //! emulator for simple off-chain testing.
 
 #![cfg_attr(not(feature = "std"), no_std)]
+#![allow(unused_unsafe)]
+// #![deny(unsafe_op_in_unsafe_fn)]
```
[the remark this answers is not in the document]
Robbepop: I added `#![allow(unused_unsafe)]` to prepare for the incoming `#![deny(unsafe_op_in_unsafe_fn)]`, however you are right that this was a bad idea.

core/src/storage2/traits/impls/prims.rs
```diff
@@ -0,0 +1,239 @@
...
+where
+    T: SpreadLayout,
+{
+    const FOOTPRINT: u64 = 1 + <T as SpreadLayout>::FOOTPRINT;
+    const REQUIRES_DEEP_CLEAN_UP: bool = <T as SpreadLayout>::REQUIRES_DEEP_CLEAN_UP;
+
+    fn push_spread(&self, ptr: &mut KeyPtr) {
+        match self {
+            Some(value) => {
+                false.push_spread(ptr);
```
ascjones, line 70: should be `true`

core/src/storage2/traits/impls/prims.rs
```diff
@@ -0,0 +1,239 @@
...
+                <T as SpreadLayout>::push_spread(value, ptr)
+            }
+            None => {
+                true.push_spread(ptr);
```
ascjones, line 74: should be `false`

core/src/storage2/traits/impls/prims.rs
```diff
@@ -0,0 +1,239 @@
...
+            // We do not really need the reference to `self.is_some()`
+            // in order to clean-up the `bool` value from the storage.
+            // However the API is demanding a reference so we give it one.
+            <bool as SpreadLayout>::clear_spread(&self.is_some(), ptr);
+            <T as SpreadLayout>::clear_spread(value, ptr)
+        }
+    }
+
+    fn pull_spread(ptr: &mut KeyPtr) -> Self {
+        if <bool as SpreadLayout>::pull_spread(ptr) {
```
ascjones, line 90: See `push_spread` for true/false which appear to be the wrong way around

core/src/storage2/traits/impls/prims.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    fn clear_spread(&self, ptr: &mut KeyPtr) {
+        if let Some(value) = self {
+            // We do not really need the reference to `self.is_some()`
+            // in order to clean-up the `bool` value from the storage.
+            // However the API is demanding a reference so we give it one.
+            <bool as SpreadLayout>::clear_spread(&self.is_some(), ptr);
+            <T as SpreadLayout>::clear_spread(value, ptr)
+        }
```
ascjones, line 86: Still need to clear the flag if it is `None`

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -0,0 +1,403 @@
...
+
+    /// Returns an exclusive reference to the entry.
+    fn load_entry_mut(&mut self) -> &mut Entry<T> {
+        // SAFETY: We load the entry either from cache of from contract storage.
+        //
+        //         This is safe because we are just returning a shared reference
+        //         from within a `&self` method. This also cannot change the
+        //         loaded value and thus cannot change the `mutate` flag of the
+        //         entry. Aliases using this method are safe since ink! is
+        //         single-threaded.
```
ascjones, line 261: This SAFETY comment is inaccurate - copied from `load_entry`

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -0,0 +1,403 @@
...
+    ///
+    /// # Panics
+    ///
+    /// If decoding the loaded value to `T` failed.
+    #[must_use]
+    pub fn get(&self) -> Option<&T> {
+        self.load_entry().value().into()
+    }
+
+    /// Returns a shared reference to the value.
```
ascjones, line 281:
```suggestion
    /// Returns an exclusive reference to the value.
```

core/src/storage2/lazy/entry.rs
```diff
@@ -0,0 +1,277 @@
...
+};
+use ink_prelude::vec::Vec;
+use ink_primitives::Key;
+
+/// The entry of a single cached value of a lazy storage data structure.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Entry<T> {
+    /// The value or `None` if the value has been removed.
+    value: Option<T>,
+    /// This is `true` if the `value` is dirty and needs to be synchronized
```
ascjones, line 44: Comment needs to be updated for `EntryState` enum

core/src/storage2/lazy/entry.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// Pulls the entity from the underlying associated storage as packed representation.
+    ///
+    /// # Note
+    ///
+    /// Mainly used by lazy storage abstractions that only allow operating on
+    /// packed storage entities such as [`LazyIndexMap`] or [`LazyArray`].
+    pub fn pull_packed_root(root_key: &Key) -> Self {
+        let mut entry =
+            Self::new(pull_packed_root_opt::<T>(root_key), EntryState::Preserved);
+        PackedLayout::pull_packed(&mut entry, root_key);
```
ascjones, line 194: I think this is redundant, since the inner `T` will have `pull_packed` called via `pull_packed_root_opt` if it has a value. Certainly the other two methods don't call their respective methods. This would make the `impl<T> PackedLayout for Entry<T>` redundant too but might want to keep it for completeness.

core/src/storage2/traits/impls/prims.rs
Robbepop: i start to feel that we should generally use `u8` here instead of booleans. It is also more general and can be extended to work on enums in general that way.

core/src/storage2/lazy/lazy_imap.rs
```diff
@@ -0,0 +1,733 @@
...
+        }
+    }
+}
+
+impl<V> LazyIndexMap<V>
+where
+    V: PackedLayout,
+{
+    /// Returns an offset key for the given index.
+    pub fn key_at(&self, index: Index) -> Option<Key> {
```
ascjones, line 265: Is it worth refactoring out the duplication with `LazyHashHap`? The main difference being this method

core/src/storage2/lazy/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+
+impl<T> Lazy<T>
+where
+    T: SpreadLayout,
+{
+    /// Returns a shared reference to the lazily loaded value.
+    ///
+    /// # Note
+    ///
+    /// This loads the value from the contract storage if this did not happed before.
```
ascjones, line 117:
```suggestion
    /// This loads the value from the contract storage if this did not happen before.
```

core/src/storage2/lazy/lazy_imap.rs
Robbepop: There are actually many more differences but in theory you are right. I actually even started with a shared code base and just separated their implementations because it became a bit too complex. However, we can think about some follow-up PRs that take on this code dupe in isolation.

core/src/storage2/collections/boxed/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+impl<T> Box<T>
+where
+    T: SpreadLayout,
+{
+    /// Returns a shared reference to the boxed value.
+    ///
+    /// # Note
+    ///
+    /// This loads the value from the pointed to contract storage
+    /// if this did not happed before.
```
ascjones, line 96:
```suggestion
    /// if this did not happen before.
```

core/src/storage2/collections/boxed/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub fn get(boxed: &Self) -> &T {
+        &boxed.value
+    }
+
+    /// Returns an exclusive reference to the boxed value.
+    ///
+    /// # Note
+    ///
+    /// This loads the value from the pointed to contract storage
+    /// if this did not happed before.
```
ascjones, line 111:
```suggestion
    /// if this did not happen before.
```

core/src/storage2/alloc/init.rs
```diff
@@ -0,0 +1,160 @@
...
+            let instance = unsafe { &mut GLOBAL_INSTANCE };
+            match instance {
+                DynamicAllocatorState::UninitDeploy => {
+                    let mut allocator = DynamicAllocator::default();
+                    let result = f(&mut allocator);
+                    *instance = DynamicAllocatorState::Initialized(allocator);
+                    result
+                }
+                DynamicAllocatorState::UninitCall => {
+                    let mut allocator = pull_spread_root::<DynamicAllocator>(&Key(DYNAMIC_ALLOCATOR_KEY_OFFSET));
```
ascjones, line 101: Where's the equivalent `push_spread_root` for the DynamicAllocator?

core/src/storage2/alloc/init.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        /// Commands the (re-)initialization of the global instance for the dynamic
+        /// storage allocator.
+        pub fn initialize_for(phase: ContractPhase) {
+            let instance = unsafe { &mut GLOBAL_INSTANCE };
+            // We do not allow reinitialization for Wasm targets for performance reasons.
+            if let DynamicAllocatorState::Initialized(_) = instance {
+                panic!("cannot reinitialize dynamic storage allocator instance in Wasm");
+            }
+            *instance = phase.into();
```
ascjones, line 84: Under what circumstances would we go from `UninitCall` back to `UninitDeploy`?

core/src/storage2/alloc/init.rs
Robbepop: in its implied `Drop` implementation and also this has to be guided by `ink_lang` later.

core/src/storage2/alloc/init.rs
Robbepop: Never in on-chain mode. In off-chain mode we can do this if we execute tests in a single thread which happens on for example `miri`.

core/src/storage2/alloc/init.rs
ascjones: I can't find the `Drop` implementation which does this - just trying to understand the lifecycle of the allocator

core/src/storage2/collections/bitvec/iter.rs
```diff
@@ -0,0 +1,310 @@
...
+            if let Some(ref mut front_iter) = self.front_iter {
+                if let front @ Some(_) = front_iter.next() {
+                    self.remaining -= 1;
+                    return front
+                }
+            }
+            match self.bits256_iter.next() {
+                None => {
+                    if let Some(back) = self.back_iter.as_mut()?.next() {
+                        self.remaining -= 1;
```
ascjones, line 151: So this is exercised in concert with `next_back()`, I think a test for this would be good. The existing ones just go forwards and backwards so wouldn't hit this or from the other direction in `DoubleEndedIterator`

core/src/storage2/collections/stash/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+    ///
+    /// Returns the stash index that the element was put into.
+    pub fn put(&mut self, new_value: T) -> Index {
+        let new_entry = Some(Entry::Occupied(new_value));
+        let new_index = if let Some(index) = self.last_vacant_index() {
+            // Put the new element to the most recent vacant index if not all entries are occupied.
+            let old_entry = self
+                .entries
+                .put_get(index, new_entry)
+                .expect("a `next_vacant` index must point to an occupied cell");
```
ascjones, line 362: expect message appears out of sync. Perhaps you mean `last_vacant_index()` rather than `next_vacant`

core/src/storage2/collections/stash/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+            let vacant_entry = match old_entry {
+                Entry::Vacant(vacant_entry) => vacant_entry,
+                Entry::Occupied(_) => {
+                    unreachable!("next_vacant must point to a vacant entry")
```
ascjones, line 366: Ditto ^^

core/src/storage2/collections/stash/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+            // different look-ups to update them.
+            if let Some(entry) = self
+                .entries
+                .get_mut(prev)
+                .map(Entry::try_to_vacant_mut)
+                .expect("`prev` must point to an existing entry at this point")
+            {
+                entry.next = at;
+            }
+            if let Some(entry) = self
```
ascjones, line 452: It looks like `next`and `prev` should always be be vacant, because they are one of: `root_vacant.prev`, `root_vacant.next`, `index` (from `last_vacant_index`), or `at` which we have just assigned as vacant. 

If I'm right we could replace the `if let Some(entry)` here and above with another `expect`.

core/src/storage2/collections/stash/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+            return None
+        }
+        // At this point we know that the entry is occupied with a value.
+        let new_vacant_entry = Entry::Vacant(VacantEntry { prev, next });
+        let taken_entry = core::mem::replace(entry_mut, new_vacant_entry);
+        // Update links from and to neighbouring vacant entries.
+        if prev == next {
+            // Previous and next are the same so we can update the vacant
+            // neighbour with a single look-up.
+            if let Some(entry) = self
```
ascjones, line 432: Potentially replace with expect, see :point_down:

core/src/storage2/collections/stash/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+                .map(Entry::try_to_vacant_mut)
+                .expect("`next` must point to an existing entry at this point")
+            {
+                entry.prev = at;
+                entry.next = at;
+            }
+        } else {
+            // Previous and next vacant entries are different and thus need
+            // different look-ups to update them.
+            if let Some(entry) = self
```
ascjones, line 444: Potentially replace with expect, see :point_down:

core/src/storage2/collections/stash/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+    /// a callback function that is called for every moved entry
+    /// with a shared reference to the entries value and the old as well
+    /// as the new index.
+    ///
+    /// # Note
+    ///
+    /// - If `max_iterations` is `Some` concrete value it is used in order to
+    ///   bound the number of iterations and won't try to defrag until the stash
+    ///   is optimally compacted.
+    /// - Users are adviced to call this method using `Some` concrete
```
ascjones, line 494:
```suggestion
    /// - Users are advised to call this method using `Some` concrete
```

core/src/storage2/collections/stash/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+                    // the length and other header information.
+                    let vacant_index = self
+                        .last_vacant_index()
+                        .expect("it has been asserted that there are vacant entries");
+                    callback(index, vacant_index, &value);
+                    let new_entry = Some(Entry::Occupied(value));
+                    let old_entry = self
+                        .entries
+                        .put_get(vacant_index, new_entry)
+                        .expect("a `next_vacant` index must point to an occupied cell");
```
ascjones, line 535: Expect message appears out of date perhaps `next_index` refers to `last_vacant_index()`

core/src/storage2/collections/stash/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+                    let vacant_entry = match old_entry {
+                        Entry::Vacant(vacant_entry) => vacant_entry,
+                        Entry::Occupied(_) => {
+                            unreachable!("next_vacant must point to a vacant entry")
```
ascjones, line 539: Ditto :point_up:

core/src/storage2/alloc/init.rs
Robbepop: The allocator doesn't require an explicit `Drop` implementation and so doesn't its internal `storage::BitStash` but still their `Drop` impls forward implicitely to the `Drop` impls of their fields. So what gets dropped in the end is a `storage::Bitvec` and a `storage::Vec`.

Robbepop: The life cycle of the dynamic storage allocator is different in `std` and `Wasm` compilation for reasons of making it better testable in `std` (off-chain) environments. In Wasm compilation we know there is only one thread and we construct a dynamic storage allocator within a `ManuallyDrop` at the start but before that we tell it how to construct:
- Either for `construct` or for `call`. Upon `contsruct` we create a brand new dynamic storage allocator because we cannot rely on having one on the storage already. And upon `call` we simply pull the one that has to be there from the `construct`.

core/src/storage2/collections/bitstash/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// Counter for set bits in a 256-bit chunk of the `free` list.
+    ///
+    /// For every 256-bit chunk stored in `free` stores a `u8` that counts
+    /// the number of set bits in the 256-bit chunk. This information is used
+    /// to compact the information in `free` to make a `first fit` linear
+    /// search for a new free storage slot more scalable.
+    ///
+    /// In theory it is possible to search up to 8192 storage cells for free
+    /// slots with a single contract storage look-up. By iterating over the 32
+    /// `SetBits32` instances of a single instance.
```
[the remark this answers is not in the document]
Robbepop: Many refactorings have taken place. ^^

core/src/storage2/alloc/init.rs
ascjones: But `Drop` in these cases will clear, not push - where does the `push` happen?

core/src/storage2/collections/bitstash/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+                // append another 1 bit (`true`) to the free list and return
+                // a new dynamic allocation pointing to it. No need to push to
+                // the counts list in this case.
+                self.free.push(true);
+                self.free.len() - 1
+            }
+        } else {
+            // We found no free dynamic storage slot:
+            // Check if we already have allocated too many (2^32) dynamic
+            // storage allocations and panic if that's the case.
```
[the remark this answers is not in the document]
Robbepop: Yep need to adjust some comments. Thanks.

ascjones:
Also I would suggest another pass from the `unsafe` gurus for just the `unsafe` blocks

core/src/storage2/collections/hashmap/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+            .unwrap_or(false)
+    }
+
+    /// Defragments storage used by the storage hash map.
+    ///
+    /// Returns the number of storage cells freed this way.
+    ///
+    /// # Note
+    ///
+    /// This frees storage that is hold but not necessary for the hash map to hold.
```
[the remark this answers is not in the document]
Robbepop: Not sure I want to include technical comments of internals into the API docs. But will add this comment to the code-docs inside the method. Thanks.

core/src/storage2/collections/stash/mod.rs
Robbepop: to be honest I am not sure about those changes anymore. it has been a while since I touched this code. :sweat_smile:

core/src/storage2/collections/stash/mod.rs
Robbepop: done

core/src/storage2/collections/stash/mod.rs
Robbepop: done

core/src/storage2/collections/stash/mod.rs
Robbepop: yep, done!

core/src/storage2/alloc/init.rs
Robbepop: The `push` will be managed by the `ink_lang` side.

core/src/storage2/collections/bitvec/iter.rs
Robbepop: I have seen you already added a test case for this. so this can be resolved i guess.

core/src/storage2/alloc/allocator.rs
```diff
@@ -0,0 +1,83 @@
...
+        SpreadLayout::clear_spread(&self.allocations, ptr);
+    }
+}
+
+impl DynamicAllocator {
+    /// Returns a new dynamic storage allocation.
+    ///
+    /// # Panics
+    ///
+    /// If the dynamic allocator ran out of free dynamic allocations.
```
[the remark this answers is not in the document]
Robbepop: added the check.

core/src/storage2/collections/hashmap/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+    pub fn contains_key<Q>(&self, key: &Q) -> bool
+    where
+        K: Borrow<Q>,
+        Q: Ord + PartialEq<K> + Eq + scale::Encode + ToOwned<Owned = K>,
+    {
+        self.values
+            .get(key)
+            .map(|entry| entry.key_index)
+            .and_then(|key_index| {
+                self.keys.get(key_index).map(|stored_key| key == stored_key)
```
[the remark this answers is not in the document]
Robbepop: greatly simplified the method. thanks.

core/src/storage2/alloc/init.rs
Robbepop: We will add a `PushOnDrop` abstraction in a follow-up PR that fixes this problem. This abstraction will also be useful for the static storage of a smart contract.

core/src/storage2/collections/hashmap/mod.rs
```diff
@@ -291,13 +291,12 @@ where
...
-        self.values
-            .get(key)
-            .map(|entry| entry.key_index)
-            .and_then(|key_index| {
-                self.keys.get(key_index).map(|stored_key| key == stored_key)
-            })
-            .unwrap_or(false)
+        // We do not check if the given key is equal to the queried key which is
+        // what normally a hash map implementation does because we do not resolve
+        // or prevent collissions in this hash map implementation at any level.
```
ascjones, line 296:
```suggestion
        // or prevent collisions in this hash map implementation at any level.
```

core/src/storage2/collections/hashmap/mod.rs
```diff
@@ -291,13 +291,12 @@ where
...
+        // Having a collission is very unlikely to ever happen though since we
```
ascjones, line 297:
```suggestion
        // Having a collision is virtually impossible since we
```

core/src/storage2/collections/hashmap/tests.rs
```diff
@@ -23,7 +23,7 @@ use crate::{
 use ink_primitives::Key;
 
 #[test]
-fn new_vec_works() {
+fn new_works() {
     // `StorageHashMap::new`
     let vec = <StorageHashMap<u8, i32>>::new();
```
ascjones, line 28:
```suggestion
    let hmap = <StorageHashMap<u8, i32>>::new();
```

and usages

Swader:
Quick question here since it's merged but also mentioned in the v3.0 docs issue - is this going in already, or will it be part of the 3.0 release only?

Robbepop:
> 
> 
> Quick question here since it's merged but also mentioned in the v3.0 docs issue - is this going in already, or will it be part of the 3.0 release only?

Yes, this PR resembles parts of the upcoming ink! 3.0.

Issue #301, opened 2019-12-24
Add `invoke_runtime` to EnvAccessMut wrapper


jaybutera:
#### Problem
`invoke_runtime` is a function of the Env trait [env2/traits.rs#L150](https://github.com/paritytech/ink/blob/master/core/src/env2/traits.rs#L150)

But is not implemented by its wrapper trait, `EnvAccessMut` [env2/env_access/mutable.rs#L156](https://github.com/paritytech/ink/blob/master/core/src/env2/env_access/mutable.rs#L156) that is accessible from an ink contract.

#### Solution
In order to call this method from a contract, the `invoke_runtime` definition should be added to the`EnvAccessMut` trait.

#### More Context
Other methods like `invoke_contract` in `Env` are implemented in both traits. On the same note, the env2 `CallBuilder` code [specifically mentions](https://github.com/paritytech/ink/blob/master/core/src/env2/call/builder.rs#L115) its use for calling other contracts, but doesn't mention invoking the runtime. It is somewhat unclear what type should be passed into `invoke_runtime` since the only constraint is that the input parameter implements `Encode`. Maybe this can be constrained to an `EnvTypes::Call` in the `invoke_runtime` definition like it is in [env/api.rs#L84](https://github.com/paritytech/ink/blob/master/core/src/env/api.rs#L84)
[Related issue] #297

jaybutera:
Ok it looks like this feature has already been implemented elsewhere but not yet submitted in a PR https://github.com/paritytech/ink-types-node-runtime/pull/7

Robbepop:
This is partially implemented in https://github.com/paritytech/ink/pull/312 and can be closed once the PR has been merged.

Issue #266, opened 2019-11-21
Rename address env property to account_id


kigawas:
```rust
    impl_get_property_for! {
        /// Returns the address of the caller of the executed contract.
        fn caller<Caller>() -> T::AccountId;
        /// Returns the transferred balance for the contract execution.
        fn transferred_balance<TransferredBalance>() -> T::Balance;
        /// Returns the current price for gas.
        fn gas_price<GasPrice>() -> T::Balance;
        /// Returns the amount of gas left for the contract execution.
        fn gas_left<GasLeft>() -> T::Balance;
        /// Returns the current block time in milliseconds.
        fn now_in_ms<NowInMs>() -> T::Moment;
        /// Returns the address of the executed contract.
        /* rename? */ fn address<Address>() -> T::AccountId;
        /// Returns the balance of the executed contract.
        fn balance<Balance>() -> T::Balance;
        /// Returns the current rent allowance for the executed contract.
        fn rent_allowance<RentAllowance>() -> T::Balance;
        /// Returns the current block number.
        fn block_number<BlockNumber>() -> T::BlockNumber;
        /// Returns the minimum balance of the executed contract.
        fn minimum_balance<MinimumBalance>() -> T::Balance;
    }

```

#66

Robbepop:
I think that's a good idea. I also thought about this but didn't do it to match the old interface.

kigawas:
Maybe just add a new function and label the old as deprecated

Robbepop:
This is implemented in #312 and can be closed once the PR has been merged.

Issue #233, opened 2019-11-12
Implement simple API for test environment


riusricardo:
ink! 1.0 had a predefined way of setting the caller on the test env.
`env::test::set_caller::<Types>(AccountId::from([0x0; 32]));`

Robbepop:
The current off-chain test environment of `ink_core::env2` is not capable of setting simple environmental values, e.g. the caller of the next transaction. This should be implemented quickly so that users can implement more elaborate off-chain tests again.

Robbepop:
This is implemented in #312 and can be closed once the PR has been merged.

Issue #220, opened 2019-11-03
ink_lang2 follow-up: Remove old ink_core::env, ink_lang and ink_model


Robbepop:
This is a follow-up on the https://github.com/paritytech/ink/pull/191 PR.
Also before we can work on this we have to ensure that we have a proper transition guide up and running for users of `ink_lang`.

The transition guide needs to teach users of `ink_lang` what has changed in comparison to `ink_lang2` and it shall guide users on what they can and should do with the new engine. Also it should provide help to them to easily take the task of transitioning to the new `ink_lang2`.

Note that this task is _NOT_ about renaming `ink_lang2` to `ink_lang`. It will simply be called `ink_lang2` for some time until we come up with a proper naming scheme for it.

## Sub Tasks

- [ ] Remove `ink_core::env` and relink everything that directly depends on it to `ink_core::env2`
- [ ] Remove `ink_model` and everything that directly depends on it
- [ ] Remove `ink_lang`
- [ ] Remove all `ink_lang` examples

Robbepop:
This is implemented in #312 and can be closed once the PR has been merged.

Muhammad-Altabba:
The samples at Substrate Contracts Workshops are not updated yet (https://substrate.dev/substrate-contracts-workshop/)

PR #312, opened 2020-01-18
Implementation of ink! env (revision 3)

Robbepop:
#194 Re-implementation of `env` module (revision 3)

## Issues that can be closed upon merging this:

- [ ] https://github.com/paritytech/ink/issues/301
- [ ] https://github.com/paritytech/ink/issues/297
- [ ] https://github.com/paritytech/ink/issues/233
- [ ] https://github.com/paritytech/ink/issues/220
- [ ] https://github.com/paritytech/ink/issues/266

## ToDos

- [x] Finalize implementation of on-chain `env` (revision 3)
- [x] FInalize implementation of off-chain `env` (revision 3)
    - [x] Querying of all contract, chain spec or execution context properties.
    - [x] Read and write from and to contract storage.
    - [x] Printing to contract execution console and put them into records so that they can be queried.
    - [x] Calling runtime functions and registering runtime call handler.
    - [x] Reading from runtime storage and setting it.
    - [x] Push & pop execution contexts.
    - [x] Getting and setting random block hashes. (`ext_random`)
    - [x] Emitting of events stored into records that can be queried.
    - [x] Support for advancing blocks.
    - [x] Initialization routines.
- [x] Adjust `ink_lang2` for `env` (revision 3)
    - [x] ~Add dynamic allocator instance.~ (Follow-up PR)
    - [x] Add utility wrappers in `ink_lang2` crate that replace missing `EnvAccess` types.
    - [x] Adjust the macro generated output of `ink_lang2_macro` to use `env` (revision 3).
- [x] Remove all usage of `env` (revision 1) and `env` (revision 2) from ink! codebase
    - [x] Remove all usage of `env` (revision 1)
    - [x] Remove all usage of `env` (revision 2)
- [x] Remove `env` (revision 1) and `env` (revision 2)
    - [x] Remove `env` (revision 1)
    - [x] Remove `env` (revision 2)
- [x] Put `env` (revision 3) as `ink_env` into its own ink! crate.
- [x] Remove `ink_model`
    - [x] Remove `ink_model` ink! crate
    - [x] Remove `ink_model` layer example contracts
- [x] Remove `ink_lang` (revision 1)
    - [x] Remove `ink_lang` (revision 1) ink! crate
    - [x] Remove `ink_lang` (revision 1) example contracts
- [x] Remove `ink_core`-only example contracts
- [x] Return `Option<Result<T>>` for `get_contract_storage`
- [x] ~Move `ink_core::storage` into `ink_storage` crate~ (Follow-up PR)
- [x] ~Remove `test-env` crate feature~ (Follow-up PR)

## Follow-up PRs

- [ ] Add dynamic allocator instance.
- [ ] Move `ink_core::storage` into `ink_storage` crate
- [ ] Move `ink_core::env` into `ink_env` crate
- [ ] Remove `test-env` crate feature and replace it in parts with `ink-test` crate feature

## Crate Dependency Graph

Anticipated dependency graph after planned follow-up PRs have been merged.

2020-02-01-012659_1346x692_scrot

## Description

- This is a great simplification of `env2` while still providing its efficiency guarantees.
- The `test-env` feature is going to be removed.
- The only compilation configurations that are going to be allowed are either `no_std` + `wasm32-unknown-unknown` or `std` (e.g. for off-chain testing). The environmental interface is generally the same for both configurations. However, in the case of `#[cfg(test)]` there is going to be another `test` sub module to allow for low-level off-chain testing environment manipulations.

The interface has a typed and an untyped subset of APIs.
- An example of a typed interface is `fn caller` since it returns a `T::AccountId` where `T: EnvTypes`.
- An example of an untyped interface is `fn get_contract_storage<T>` since it doesn't have any dependencies on a specific `EnvTypes` provider.
- As in `env2` all operations between the smart contract and its environment are still implicitly cached and buffered to avoid dynamic allocations. We could even reduce this to no dynamic allocations at all in this implementation by using a static buffer instead.
- Whereas `env` (revision 2) took over the job to also provide dynamic allocators this is no longer true for `env` (revision 3) and will instead be handled in its own sub module (e.g. `dyn_alloc`).

The `env` revision 3 will replace all usage of `env` (revision 1) and `env` (revision 2).

### The new off-chain test API

2020-01-28-003043_1662x1007_scrot

### The contract calling and instantiation API

2020-01-23-150207_1666x868_scrot

### The general environment API

2020-01-23-150154_871x1205_scrot

## Contract Restoration

For the first time we also provide high-level APIs for contract restoration.

2020-01-18-133809_1161x1184_scrot

## Runtime Emulation

Set certain runtime storage entries in the emulated off-chain runtime to allow querying them by the contract.

2020-01-27-163857_903x240_scrot

Initialize off-chain runtime dispatch handlers that emulate your runtime behaviour from the perspective of an invoking contract.

2020-01-27-163848_846x284_scrot


core/src/env3/api.rs
```diff
@@ -0,0 +1,442 @@
+//! The public raw interface towards the host Wasm engine.
```
cmichi: License is missing.

core/src/env3/call/builder.rs
```diff
@@ -0,0 +1,190 @@
...
+        Selector,
+    },
+    EnvTypes,
+    Result,
+};
+
+/// Represents a return type.
+///
+/// Used as a marker type to differentiate at compile-time between invoke and evaluate.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
cmichi: There are a number of `Debug` derives in the PR. Are they required?

core/src/env3/call/create.rs
```diff
@@ -0,0 +1,201 @@
...
+            state: Default::default(),
+        }
+    }
+}
+
+impl<T, C, Seal, CodeHash> CreateBuilder<T, C, Seal, CodeHash>
+where
+    T: EnvTypes,
+{
+    /// Sets the maximumly allowed gas costs for the call.
```
cmichi:
```suggestion
    /// Sets the maximum allowed gas costs for the call.
```

core/src/env3/call/builder.rs
Robbepop: I think they don't matter because dead code elimination will throw them out again if not needed.

core/src/env3/api.rs
Robbepop: thank you!

core/src/env3/api.rs
```diff
@@ -0,0 +1,450 @@
...
+    })
+}
+
+/// Returns the value stored under the given key in the contract's storage.
+///
+/// # Errors
+///
+/// - If the key's entry is empty
+/// - If the decoding of the typed value failed
+pub fn get_contract_storage<R>(key: Key) -> Result<R>
```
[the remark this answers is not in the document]
Robbepop: I did not yet come back to this refactoring.

core/src/env3/engine/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+};
+use cfg_if::cfg_if;
+
+pub trait OnInstance: Env + TypedEnv {
+    fn on_instance<F, R>(f: F) -> R
+    where
+        F: FnOnce(&mut Self) -> R;
+}
+
+cfg_if! {
```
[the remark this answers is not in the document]
Robbepop: That's super bad because I wanted to use this neat macro at some other places as well. Hmm. Maybe file a bug report for IntelliJ + Rust plugin? I am not sure we should guide our design by IDE bugs tbh but this sounds really weird though ...

core/src/env3/api.rs
```diff
@@ -0,0 +1,450 @@
...
+{
+    <EnvInstance as OnInstance>::on_instance(|instance| TypedEnv::gas_left::<T>(instance))
+}
+
+/// Returns the current block time in milliseconds.
+///
+/// # Errors
+///
+/// If the returned value cannot be properly decoded.
+pub fn now_in_ms<T>() -> Result<T::TimeStamp>
```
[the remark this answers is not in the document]
Robbepop: thanks I will consider renaming it or maybe provide some wrapper type

core/src/env3/api.rs
```diff
@@ -0,0 +1,450 @@
...
+/// Invokes a contract message.
+///
+/// # Errors
+///
+/// - If the called contract does not exist.
+/// - If the called contract is a tombstone.
+/// - If arguments passed to the called contract message are invalid.
+/// - If the called contract execution has trapped.
+/// - If the called contract ran out of gas upon execution.
+/// - If given too few endowment.
```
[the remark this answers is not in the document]
Robbepop: As far as I am aware you can also transfer value for contract calls.

core/src/env3/call/builder.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub fn callee(&self) -> &E::AccountId {
+        &self.callee
+    }
+
+    /// The gas limit for the contract instantiation.
+    pub fn gas_limit(&self) -> u64 {
+        self.gas_limit
+    }
+    /// The endowment for the instantiated contract.
+    pub fn endowment(&self) -> &E::Balance {
```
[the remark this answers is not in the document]
Robbepop: I think `transferred_value` is better here

core/src/env3/engine/mod.rs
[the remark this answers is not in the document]
ascjones: Seems to be a known issue: https://github.com/intellij-rust/intellij-rust/issues/4627

> I am not sure we should guide our design by IDE bugs tbh but this sounds really weird though

I agree. I can just maintain a local patch which "expands" this macro.

How do other setups handle this? I can always use VS code or vim or whatever for developing `ink!`. I at least want `go to definition` to work.

core/src/env3/api.rs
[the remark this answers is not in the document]
ascjones: The point is that it is listed under `Errors` - will it error if there is "too few endowment"?

core/src/env3/engine/mod.rs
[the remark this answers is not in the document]
Robbepop: I am using VS Code with the `rust-analyzer` extension and it works pretty well for the ink! codebase. The `go to definition` works well, too besides many other neat things. I am pretty happy with `rust-analyzer` already.

core/src/env3/api.rs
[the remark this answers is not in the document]
Robbepop: How about `block_timestamp`?

lang/src/env_access.rs
```diff
@@ -0,0 +1,359 @@
...
+    ///
+    /// # Params
+    ///
+    /// - `account_id`: Account ID of the to-be-restored contract.
+    /// - `code_hash`: Code hash of the to-be-restored contract.
+    /// - `rent_allowance`: Rent allowance of the restored contract
+    ///                     upon successful restoration.
+    /// - `filtered_keys`: Storage keys that will be removed for the tombstone hash
+    ///                    match calculation that decide whether the original contract
+    ///                    storage and the storage of the restorer contract is equal.
```
ascjones: Sentence is hard to read, perhaps something like: `Storage keys to be excluded when calculating the tombstone hash, which is used to decide whether the original contract and the to-be-restored contract have matching storage.`

lang/src/env_access.rs
```diff
@@ -0,0 +1,359 @@
...
+    ///
+    /// # Usage
+    ///
+    /// A smart contract that has too few funds to pay for its storage fees
```
ascjones:
```suggestion
    /// A smart contract that has insufficient funds to pay for its storage fees
```

core/src/env/types.rs
```diff
@@ -0,0 +1,336 @@
...
+    type Hash = Hash;
+    type Timestamp = Timestamp;
+    type BlockNumber = BlockNumber;
+    type Call = Call;
+}
+
+/// The default balance type.
+pub type Balance = u128;
+
+/// The default time stamp type.
```
ascjones:
```suggestion
/// The default timestamp type.
```

core/src/env/types.rs
```diff
@@ -0,0 +1,336 @@
...
+pub type Timestamp = u64;
+
+/// The default block number type.
+pub type BlockNumber = u64;
+
+/// This call type guarantees to never be constructed.
+///
+/// This has the effect that users of the default SRML types are
```
ascjones:
```suggestion
/// This has the effect that users of the default env types are
```

core/src/env/types.rs
```diff
@@ -0,0 +1,336 @@
...
+/// For more info visit: https://github.com/paritytech/ink-types-node-runtime
+#[derive(Debug)]
+pub enum Call {}
+
+impl Encode for Call {
+    fn encode(&self) -> Vec<u8> {
+        // The implementation enforces at runtime that `Encode` is not called
+        // for the default SRML `Call` type but for performance reasons this check
+        // is removed for the on-chain (release mode) version.
+        debug_assert!(false, "cannot encode default SRML `Call` type");
```
ascjones:
```suggestion
        debug_assert!(false, "cannot encode default `Call` type");
```

core/src/env/types.rs
```diff
@@ -0,0 +1,336 @@
...
+        Vec::new()
+    }
+}
+
+impl scale::Decode for Call {
+    fn decode<I: scale::Input>(_value: &mut I) -> Result<Self, scale::Error> {
+        // This implementation is only to satisfy the Decode constraint in the
+        // test environment. Since Call cannot be constructed then just return
+        // None, but this should never be called.
+        Err("The default SRML `Call` type cannot be used for runtime calls".into())
```
ascjones:
```suggestion
        Err("The default `Call` type cannot be used for runtime calls".into())
```

core/src/env/api.rs
```diff
@@ -12,127 +12,439 @@
...
+/// If the returned value cannot be properly decoded.
+pub fn gas_left<T>() -> Result<T::Balance>
+where
+    T: EnvTypes,
+{
+    <EnvInstance as OnInstance>::on_instance(|instance| TypedEnv::gas_left::<T>(instance))
 }
 
-/// Returns the given data back to the caller.
+/// Returns the current block time in milliseconds.
```
ascjones:
```suggestion
/// Returns the current block timestamp.
```

core/src/env/api.rs
```diff
@@ -12,127 +12,439 @@
...
+}
+
+/// Returns the account ID of the executed contract.
 ///
 /// # Note
 ///
-/// This operation must be the last operation performed by a called
-/// smart contract before it returns the execution back to its caller.
-pub fn return_data<T, E>(data: T)
+/// This method was formerly known as `address`.
```
ascjones: :prince: :guitar:

core/src/env/api.rs
```diff
@@ -12,127 +12,439 @@
...
+///
+/// # Errors
+///
+/// - If the code hash is invalid.
+/// - If the arguments passed to the instantiation process are invalid.
+/// - If the instantiation process traps.
+/// - If the instantiation process runs out of gas.
+/// - If given too few endowment.
+/// - If the returned account ID failed to decode properly.
+pub fn create_contract<T, C>(params: &InstantiateParams<T, C>) -> Result<T::AccountId>
```
ascjones:
```suggestion
pub fn instantiate_contract<T, C>(params: &InstantiateParams<T, C>) -> Result<T::AccountId>
```

core/src/env/api.rs
```diff
@@ -12,127 +12,439 @@
...
+///
+/// # Params
+///
+/// - `account_id`: Account ID of the to-be-restored contract.
+/// - `code_hash`: Code hash of the to-be-restored contract.
+/// - `rent_allowance`: Rent allowance of the restored contract
+///                     upon successful restoration.
+/// - `filtered_keys`: Storage keys that will be removed for the tombstone hash
+///                    match calculation that decide whether the original contract
+///                    storage and the storage of the restorer contract are equal.
```
ascjones: See suggestions in another copy of this comment. How can we make sure these comments don't fall out of sync? Perhaps just have the comment in one place and then link to it?

core/src/env/backend.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    /// Instantiates another contract.
+    ///
+    /// # Errors
+    ///
+    /// - If the instantiation process traps.
+    /// - If the code hash is invalid.
+    /// - If given too few endowment.
+    /// - If the instantiation process runs out of gas.
+    fn create_contract<T, C>(
```
ascjones:
```suggestion
    fn instantiate_contract<T, C>(
```

core/src/env/engine/on_chain/ext.rs
```diff
@@ -67,38 +156,65 @@ pub fn deposit_event(topics: &[u8], data: &[u8]) {
...
 
-pub fn get_storage(key: &[u8]) -> RetCode {
-    unsafe { sys::ext_get_storage(key.as_ptr() as u32) }.into()
+pub fn clear_storage(key: &[u8]) {
+    unsafe { sys::ext_set_storage(key.as_ptr() as u32, 0, 0, 0) }
 }
 
-pub fn get_runtime_storage(key: &[u8]) -> RetCode {
-    unsafe { sys::ext_get_runtime_storage(key.as_ptr() as u32, key.len() as u32) }.into()
+pub fn get_storage(key: &[u8]) -> Result<()> {
```
ascjones: Missing comments for these pub fns

core/src/env/api.rs
Robbepop: ? :sweat_smile:

ascjones: https://www.google.com/search?client=ubuntu&hs=VkJ&channel=fs&q=Why+is+Prince+called+the+Artist+Formerly+Known+as+Prince%3F&sa=X&ved=2ahUKEwj8vtP1tLXnAhWVonEKHQtBA_IQzmd6BAgREAw&biw=1720&bih=1273

core/src/env/api.rs
Robbepop: I also thought about this problem. I think linking of doc comments is a good idea here. :+1:

core/src/env/engine/on_chain/impls.rs
```diff
@@ -0,0 +1,334 @@
...
+    ) -> Result<()>
+    where
+        T: EnvTypes,
+    {
+        // Reset the contract-side buffer to append onto clean slate.
+        self.reset_buffer();
+        // Append the encoded `call_data`, `endowment` and `call_data`
+        // in order and remember their encoded regions within the buffer.
+        let callee = self.append_encode_into_buffer(call_params.callee());
+        let endowment = self.append_encode_into_buffer(call_params.transferred_value());
```
ascjones:
```suggestion
        let transferred_value = self.append_encode_into_buffer(call_params.transferred_value());
```

primitives/Cargo.toml
```diff
@@ -1,31 +1,33 @@
...
-    "std",
-    "ink_core/test-env",
-]
+default = ["std"]
 std = [
-    "ink_core/std",
     "ink_prelude/std",
     "scale/std",
+    "type-metadata",
+    "type-metadata/std",
```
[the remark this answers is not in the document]
Robbepop: The first enables the optional `type-metadata` dependency, the second enables the `type-metadata` `std` crate feature. However, we could simply enable that by default tbh.

core/src/env/engine/on_chain/impls.rs
```diff
@@ -0,0 +1,334 @@
...
+    ) -> Result<R>
+    where
+        T: EnvTypes,
+        R: scale::Decode,
+    {
+        self.invoke_contract_impl(call_params)?;
+        self.decode_scratch_buffer().map_err(Into::into)
+    }
+
+    fn create_contract<T, C>(
```
ascjones:
```suggestion
    fn instantiate_contract<T, C>(
```

core/src/env/api.rs
Robbepop: xD

core/src/env/engine/on_chain/ext.rs
Robbepop: I think it is okay for them to be missing doc comments since they are not really public to the outside of the on-chain module. Also they are all just thin wrappers around their corresponding `ext_*` functions which do have extensive docs.

lang/src/env_access.rs
Robbepop: yeah much better, thanks!

Robbepop:
> taco

Well deserved taco. xD

PR #318, opened 2020-02-06
Docs generating

TriplEight:
not super happy with all of this wizardry, but I didn't find a good way to config Jeckyll the way it would treat Rustdoc normally.
That's why the design decision was to take Jekyll `_config.yml` with theme, `README.md` from the current branch and `target/doc` dir to the root and remove everything else.
https://paritytech.github.io/ink/

Alternatively, could change links `https://paritytech.github.io/ink/ink_abi` and `https://paritytech.github.io/ink/ink_core` in `README.md` to `https://paritytech.github.io/ink/target/doc/ink_abi` and `https://paritytech.github.io/ink/target/doc/ink_core` respectively. But decided that anyway it's better to remove everything else to make it unavailable from GH pages.


TriplEight:
Now you can add
```
documentation = "https://paritytech.github.io/ink/something"
homepage = "https://paritytech.github.io/ink/"
```
to manifests.

Robbepop:
> Now you can add
> 
> ```
> documentation = "https://paritytech.github.io/ink/something"
> homepage = "https://paritytech.github.io/ink/"
> ```
> 
> to manifests.

That's very nice! Thank you. :)

PR #319, opened 2020-02-06
Support quickcheck based property testing

cmichi:
I'm open for other approaches to solve this.

The problem we have with the new env rev 3 is that quickcheck tests don't run, since quickcheck runs multiple tests in one thread whereas our testing environment expects one test per thread (because of  `thread_local!`).

With multiple tests being subsequently run in the same thread by quickcheck we get an `AlreadyInitialized` assertion error.


core/src/env/engine/mod.rs
```diff
@@ -17,11 +17,19 @@ use crate::env::backend::{
     TypedEnv,
 };
 use cfg_if::cfg_if;
+use core::cell::RefCell;
+use std::thread::LocalKey;
```
cmichi: The std lib import is probably an issue, can you think of another way to get to `LocalKey`?

core/Cargo.toml
```diff
@@ -52,3 +57,4 @@ ink-generate-abi = [
     "type-metadata",
     "std",
 ]
+ink-fuzz = []
```
Robbepop: `ink-fuzz` should imply `std` feature.

core/src/env/engine/mod.rs
```diff
@@ -17,11 +17,21 @@ use crate::env::backend::{
...
 pub trait OnInstance: Env + TypedEnv {
+    const INSTANCE: LocalKey<RefCell<EnvInstance>>;
+
     fn on_instance<F, R>(f: F) -> R
     where
         F: FnOnce(&mut Self) -> R;
+
+    fn on_uninitialized_instance<F, R>(f: F) -> R
+    where
+        F: FnOnce(&mut Self) -> R;
```
Robbepop: Please try to find a way without modifying this core trait.
Also I don't know what the difference between `on_instance` and `on_uninitialized_instance` should be because `on_instance` has nothing to do with its initialization.

core/src/env/engine/off_chain/mod.rs
```diff
@@ -224,15 +224,27 @@ impl EnvInstance {
...
+
+    fn on_uninitialized_instance<F, R>(f: F) -> R
+        where
+            F: FnOnce(&mut Self) -> R,
+    {
+        Self::INSTANCE.with(|instance| {
+            let inst = EnvInstance::uninitialized();
+            instance.replace(inst);
+            f(&mut instance.borrow_mut())
+        })
```
Robbepop: If you want to initialize the off-chain environment in a specific way it is better to add another generalized initialization routine to `env::test` api.

cmichi:
@Robbepop Thanks for the comments! I implemented it in a bit of a different way now. Could you take another look please?

core/src/env/engine/off_chain/test_api.rs
```diff
@@ -299,6 +299,25 @@ where
...
 
+/// Initializes the whole off-chain environment.
+/// Uses a fresh off-chain environment instance!
+///
+/// # Note
+///
+/// - Initializes the off-chain environment with default values that fit most
+/// uses cases.
+/// - The off-chain environment _must_ be initialized before use.
+pub fn recreate_and_initialize_as_default<T>() -> Result<()>
```
Robbepop: So this whole thing is basically resetting the off-chain environment instance to default values? Could we call this procedure `reset_to_defaults` then? :) Also we could potentially be more efficient by introducing some more robust API that either initializes or resets with default values.

core/src/env/engine/off_chain/test_api.rs
```diff
@@ -312,6 +331,22 @@ where
     f(default_accounts)
 }
 
+/// Runs the given closure test function with the default configuration
+/// for the off-chain environment.
+///
+/// Doesn't reuse an off-chain environment which might already exist in
+/// this thread, but instead uses a new off-chain environment instance.
+pub fn run_multiple_tests_in_thread<T, F>(f: F) -> Result<()>
```
Robbepop: I currently am failing to see how this function has anything to do with running a closure in multiple threads.

Robbepop:
> With multiple tests being subsequently run in the same thread by quickcheck we get an AlreadyInitialized assertion error.

If tests run consecutively this should have been fixed in my upcoming PR for `storage2` at least. We could port this fix potentially to `storage` (revision 1) if needed.

core/src/storage2/collections/hashmap/tests.rs
```diff
@@ -22,6 +22,63 @@ use crate::{
 };
 use ink_primitives::Key;
 
+#[cfg(feature = "ink-fuzz")]
```
Robbepop, line 25: I think it is better if we move all `ink-fuzz` dependent fuzz tests into their own file so that we can have only one `#[cfg(feature = "ink-fuzz)]` and do not spill those conditional compilation flags around the entire code base. Also I'd like to rename `ink-fuzz` into `ink-fuzz-tests`.

cmichi:
Just updated the PR and it can be reviewed again.

I was able to remove the bulk of the original work, which was originally necessary because of the issue with offchain environments not being reset and quickcheck running in single thread mode. This has resolved itself through other changes in the meantime.

So this PR now "just" contains some fuzz tests for `storage2/HashMap` :-).

In order to execute all tests including fuzz tests use `cargo test --features ink-fuzz`. In order to just execute the ones from this PR use `cargo test --features ink-fuzz -- randomized_`. Imho a dedicated prefix for fuzz tests is a good practice (we could also use `fuzz_` instead of `randomized_`, I'm open to that).

cmichi:
Updated! Execute all fuzz tests via `cargo test --features ink-fuzz-tests -- fuzz_tests`.

core/src/storage2/collections/hashmap/fuzz_tests.rs
```diff
@@ -0,0 +1,121 @@
...
+    let mut cnt_inserts = 0;
+    let mut previous_even_x = None;
+    let inserts_each = inserts_each as i32;
+
+    xs.into_iter().for_each(|x| {
+        if x % 2 == 0 {
+            // On even numbers we insert
+            for key in x..x + inserts_each {
+                let val = key * 10;
+                if let None = map.insert(key, val) {
```
Robbepop, line 41:
```suggestion
                if map.insert(key, val).is_none() {
```

core/src/storage2/collections/hashmap/fuzz_tests.rs
```diff
@@ -0,0 +1,121 @@
...
+                    assert_eq!(map.get(&key), Some(&val));
+                    cnt_inserts += 1;
+                }
+                assert_eq!(map.len(), cnt_inserts);
+            }
+            if let None = previous_even_x {
```
Robbepop, line 47:
```suggestion
            if previous_even_x.is_none() {
```

core/src/storage2/collections/hashmap/fuzz_tests.rs
```diff
@@ -0,0 +1,121 @@
...
+                previous_even_x = Some(x);
+            }
+        } else if x % 2 == 1 && previous_even_x.is_some() {
```
Robbepop, line 50:
```suggestion
        } else if previous_even_x.is_some() {
```

core/src/storage2/collections/hashmap/fuzz_tests.rs
```diff
@@ -0,0 +1,121 @@
...
+    env::test::run_test::<env::DefaultEnvTypes, _>(|_| {
+        // given
+        let xs: Vec<i32> = xs.into_iter().unique().collect();
+        let xth = xth.max(1);
+        let mut map = <StorageHashMap<i32, i32>>::new();
+        let mut len = map.len();
+
+        // when
+        // 1) insert all
+        xs.iter().for_each(|i| {
```
Robbepop, line 95: Why not use a normal `for` loop here instead of the `for_each`?

core/src/storage2/collections/hashmap/fuzz_tests.rs
```diff
@@ -0,0 +1,121 @@
...
+            assert_eq!(map.insert(*i, i * 10), None);
+            len += 1;
+            assert_eq!(map.len(), len);
+        });
+
+        // 2) remove every `xth` element of `xs` from the map
+        xs.iter().enumerate().for_each(|(x, i)| {
```
Robbepop, line 102: `for_each` usage is normally discouraged. Simply use a normal `for` loop instead.

core/src/storage2/collections/hashmap/fuzz_tests.rs
```diff
@@ -0,0 +1,121 @@
...
+            if x % xth == 0 {
+                assert_eq!(map.take(&i), Some(i * 10));
+                len -= 1;
+            }
+            assert_eq!(map.len(), len);
+        });
+
+        // then
+        // everything else must still be get-able
+        xs.iter().enumerate().for_each(|(x, i)| {
```
Robbepop, line 112: Same here.

cmichi:
>Can you provide us with overall results of your fuzz tests? Have they found bugs?

I used it heavily while implementing the `BTreeMap` and found a number of bugs there. Those were mostly edge cases which I hadn't considered for unit tests beforehand, I then added tests for those cases. The fuzz tests for `BTreeMap` look very similar to the ones for this data structure and I can provide a follow-up PR at some point (maybe also for other data structures, it's always similar, a pattern of insert/removes ops). I would wait until we have agreed on a fuzzing library though.

The fuzz tests were left out of the original `BTreeMap` PR since that PR already introduced a lot of new stuff, instead I had created this separate PR for the already existing `HashMap` at that time.

>Also have you looked into proptest?

As discussed in DM I have created a follow-up issue to evaluate and migrate to a better fuzzing library (#464).

core/src/storage2/collections/hashmap/fuzz_tests.rs
```diff
@@ -0,0 +1,127 @@
...
+    env::test::run_test::<env::DefaultEnvTypes, _>(|_| {
+        // given
+        let xs: Vec<i32> = xs.into_iter().unique().collect();
+        let xth = xth.max(1);
+        let mut map = <StorageHashMap<i32, i32>>::new();
+        let mut len = map.len();
+
+        // when
+        // 1) insert all
+        xs.iter().for_each(|i| {
```
Robbepop, line 95: still `for_each` used instead of normal loop

PR #320, opened 2020-02-13
[examples] fix up examples for new `cargo-contract` with `xbuild`

ascjones:
Make the examples work with the updated `cargo-contract` which uses `xbuild` for optimum Wasm binary sizes: https://github.com/paritytech/cargo-contract/pull/33

- Remove `.cargo/config` files (hardcoded linker args are now passed in directly by `cargo-contract`)
- Add `+nightly` to Delegator build script (`cargo-contract` no longer automatically uses the nightly toolchain)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,7 +13,7 @@ stages:
 
 variables:
   GIT_STRATEGY:                    fetch
-  GIT_DEPTH:                       "3"
+  GIT_DEPTH:                       3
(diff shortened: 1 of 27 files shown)


ascjones:
>
    Xargo.toml
    Special Cargo.toml fields
    .cargo/config file

`Xargo.toml` is now gone, the `.cargo.config` file is gone, as well as any `xbuild` config in the `Cargo.toml`. Further improvements now would be removing the need for the `[profile.release]` section. This will happen in a follow up PR: https://github.com/paritytech/cargo-contract/issues/40

Robbepop:
I think we can ignore CI failure since it seems to have nothing to do with this PR.

PR #324, opened 2020-02-19
Workaround a clippy bug

TriplEight:
ineffective as a lasersight on a shotgun; but makes clippy work

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -123,7 +123,6 @@ clippy-std:
   needs:
     - check-std
   script:
-    - rustup default nightly-2020-02-01
     - for crate in ${ALL_CRATES}; do
         cargo +nightly-2020-02-01 clippy --verbose --all-features --manifest-path ${crate}/Cargo.toml -- -D warnings;
       done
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -123,8 +123,9 @@ clippy-std:
   needs:
     - check-std
   script:
+    - rustup default nightly-2020-02-01
```
Robbepop: Please remove these lines. Not needed and just confusing.

TriplEight: Default toolchain in the docker image is `nightly-2020-02-17`.
If remove these overrides, it will use clippy from that toolchain.
Moreover, this workaround is not forever here.

Robbepop: No it won't since we specified the `nightly-2020-02-01` toolchain below with `+nightly-2020-02-01` at the `clippy` invokation.

TriplEight: ah of course, thanks

TriplEight:
Now you can merge it @Robbepop , the test will work after you'll merge the https://github.com/paritytech/ink/pull/323 as well

PR #327, opened 2020-02-26
Add plain multisig wallet example

athei:
This implements a plain multisig wallet with configurable owners and confirmation threshold. It is modeled after the popular [gnosis multisig wallet](https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWallet.sol). It allows executing arbitrary calls using the wallet origin to other contracts after receiving enough confirmations of the contract `owners`.

**Note**
The [polkadot-js/apps UI](https://polkadot.js.org/apps/) cannot be used to create the contract because of this issue: polkadot-js/api#1549. However, you need a dedicated DAPP for this contract anyways to have a good UX.

# Differences from the gnosis wallet

This contract stays close to the gnosis wallet. However, I took the liberty of changing some parts of the contract where I saw room for improvement. 

## State cleanup
The gnosis contract does not clean up historic state whatsoever. This mainly concerns the map that stores the confirmations for each owner and transaction. This seems reasonable when there is no state rent and the cleanup is expensive in computation.  My implementation does this somewhat expensive pruning in each call that obsoletes the state. So every call cleans up after itself.

## Execution is always a separate call
In the gnosis contract a confirmation triggers the execution when the confirmation threshold is reached. I opted for a more simpler approach where the execution must be triggered separately. This avoids cluttering the return type of the confirmation that otherwise had to return information about the possible execution.

## Avoid heavy `ensure_confirmed` computation
The gnosis contract needs to filter through the confirmations when determining if a transaction is confirmed. We avoid this singularity by adding a `confirmation_count` storage item that is kept synchronous with the current confirmations for each transaction. The trade off is the added complexity of doing this.

# TODO
This is a draft for a reason. There are some things left todo.

- [x] Documentation
- [x] State clearly that this contract is an example and not audited or ready for production
- [x] Fix `cargo contract generate-metadata`
- [x] Emit Events
- [x] Switch from `HashMap` to `BTreeMap`. ~Blocked by #334~
- [x] Tests

### DIFF
--- a/examples/multisig_plain/.cargo/config
+++ b/examples/multisig_plain/.cargo/config
@@ -1,4 +1,0 @@
-[target.wasm32-unknown-unknown]
-rustflags = [
-	"-C", "link-args=-z stack-size=65536 --import-memory"
-]
--- a/examples/multisig_plain/.cargo/config
+++ b/examples/multisig_plain/.cargo/config
@@ -1,0 +1,4 @@
+[target.wasm32-unknown-unknown]
+rustflags = [
+	"-C", "link-args=-z stack-size=65536 --import-memory"
+]
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -184,7 +184,11 @@ mod multisig_plain {
                 .map_err(|_| ())
         }
 
-        fn confirm_by_caller(&mut self, confirmer: AccountId, transaction: TransactionId) {
+        fn confirm_by_caller(
+            &mut self,
+            confirmer: AccountId,
+            transaction: TransactionId,
+        ) {
             if self
                 .confirmations
                 .insert((transaction, confirmer), ())
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -66,7 +66,7 @@
 
 use ink_lang as ink;
 
-#[ink::contract(version = "0.1.0", env = MyEnv)]
+#[ink::contract(version = "0.1.0", env = ink_core::env::DefaultEnvTypes)]
 mod multisig_plain {
     use ink_core::{
         env,
@@ -75,8 +75,6 @@ mod multisig_plain {
     use ink_prelude::vec::Vec;
     use scale::Output;
 
-    /// When using custom types in your runtime. Here is the place to declare them.
-    type MyEnv = env::DefaultEnvTypes;
     /// Tune this to your liking but be that many owners will not perform well.
     const MAX_OWNERS: u32 = 50;
 
@@ -252,7 +250,7 @@ mod multisig_plain {
         fn execute_transaction(&mut self, trans_id: TransactionId) -> Result<(), ()> {
             self.ensure_confirmed(trans_id);
             let t = self.take_transaction(trans_id).expect(WRONG_TRANSACTION_ID);
-            env::call::CallParams::<MyEnv, ()>::invoke(t.callee, t.selector.into())
+            env::call::CallParams::<EnvTypes, ()>::invoke(t.callee, t.selector.into())
                 .gas_limit(t.gas_limit)
                 .transferred_value(t.transferred_value)
                 .push_arg(&CallInput(&t.input))
@@ -363,7 +361,7 @@ mod multisig_plain {
     mod tests {
         use super::*;
         use ink_core::env::test;
-        type Accounts = test::DefaultAccounts<MyEnv>;
+        type Accounts = test::DefaultAccounts<EnvTypes>;
 
         fn default_accounts() -> Accounts {
             test::default_accounts()
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -194,26 +194,17 @@ mod multisig_plain {
         transaction: TransactionId,
     }
 
-    /// Emitted when a transaction was executed without evaluating its output.
+    /// Emitted when a transaction was executed.
     #[ink(event)]
-    struct Invokation {
+    struct Execution {
         /// The transaction that was executed.
         #[ink(topic)]
         transaction: TransactionId,
-        /// Indicates whether the transaction executed successfully.
+        /// Indicates whether the transaction executed successfully. If so the `Ok` value holds
+        /// the output in bytes. The Option is `None` when the transaction was executed through
+        /// `invoke_transaction` rather than `evaluate_transaction`.
         #[ink(topic)]
-        result: Result<(), ()>,
-    }
-
-    /// Emitted when a transaction was executed and its output is evaluated.
-    #[ink(event)]
-    struct Evaluation {
-        /// The transaction that was executed.
-        #[ink(topic)]
-        transaction: TransactionId,
-        /// The output of the transaction as bytes.
-        #[ink(topic)]
-        result: Result<Vec<u8>, ()>,
+        result: Result<Option<Vec<u8>>, ()>,
     }
 
     /// Emitted when an owner is added to the wallet.
@@ -442,9 +433,9 @@ mod multisig_plain {
             )
             .fire()
             .map_err(|_| ());
-            self.env().emit_event(Invokation {
+            self.env().emit_event(Execution {
                 transaction: trans_id,
-                result,
+                result: result.map(|_| None),
             });
             result
         }
@@ -464,9 +455,9 @@ mod multisig_plain {
             )
             .fire()
             .map_err(|_| ());
-            self.env().emit_event(Evaluation {
+            self.env().emit_event(Execution {
                 transaction: trans_id,
-                result: result.clone(),
+                result: result.clone().map(|val| Some(val)),
             });
             result
         }
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -99,10 +99,10 @@ mod multisig_plain {
         #[ink(message)]
         fn add_owner(&mut self, new_owner: AccountId) {
             self.ensure_from_wallet();
-            self.ensure_no_owner(&owner);
+            self.ensure_no_owner(&new_owner);
             ensure_requirement(self.owners.len() + 1, *self.requirement);
-            self.is_owner.insert(owner, ());
-            self.owners.push(owner);
+            self.is_owner.insert(new_owner, ());
+            self.owners.push(new_owner);
         }
 
         #[ink(message)]
@@ -121,24 +121,25 @@ mod multisig_plain {
         #[ink(message)]
         fn replace_owner(&mut self, old_owner: AccountId, new_owner: AccountId) {
             self.ensure_from_wallet();
-            self.ensure_owner(&owner);
+            self.ensure_owner(&old_owner);
             self.ensure_no_owner(&new_owner);
-            self.owners.replace(self.owner_index(&owner), || new_owner);
-            self.is_owner.remove(&owner);
+            self.owners
+                .replace(self.owner_index(&old_owner), || new_owner);
+            self.is_owner.remove(&old_owner);
             self.is_owner.insert(new_owner, ());
-            self.clean_owner_confirmations(&owner);
+            self.clean_owner_confirmations(&old_owner);
         }
 
         #[ink(message)]
         fn change_requirement(&mut self, new_requirement: u32) {
             self.ensure_from_wallet();
-            ensure_requirement(self.owners.len(), requirement);
-            self.requirement.set(requirement);
+            ensure_requirement(self.owners.len(), new_requirement);
+            self.requirement.set(new_requirement);
         }
 
         #[ink(message)]
         fn submit_transaction(&mut self, transaction: Transaction) {
-            self.ensure_from_owner();
+            self.ensure_caller_is_owner();
             let id = self.transactions.put(transaction);
             self.confirmation_count.insert(id, 0);
             self.internal_confirm(id);
@@ -152,14 +153,14 @@ mod multisig_plain {
 
         #[ink(message)]
         fn confirm_transaction(&mut self, id: TransactionId) {
-            self.ensure_from_owner();
+            self.ensure_caller_is_owner();
             self.ensure_transaction_exists(id);
             self.internal_confirm(id);
         }
 
         #[ink(message)]
         fn revoke_confirmation(&mut self, id: TransactionId) {
-            self.ensure_from_owner();
+            self.ensure_caller_is_owner();
             if self
                 .confirmations
                 .remove(&(id, self.env().caller()))
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -141,7 +141,7 @@ mod multisig_plain {
             self.requirement.set(requirement);
         }
 
-        /// Add a new owner to the contract. Fails is the owner already exists.
+        /// Add a new owner to the contract. Panics is the owner already exists.
         /// Only callable by the wallet itself.
         #[ink(message)]
         fn add_owner(&mut self, new_owner: AccountId) {
@@ -155,7 +155,7 @@ mod multisig_plain {
         /// Remove an owner from the contract.
         /// Only callable by the wallet itself. If by doing this the amount of owners
         /// would be smaller than the requirement it is adjusted to be exactly the
-        /// number of owners.
+        /// number of owners. Panics if `owner` is no owner of the wallet.
         #[ink(message)]
         fn remove_owner(&mut self, owner: AccountId) {
             self.ensure_from_wallet();
@@ -169,7 +169,8 @@ mod multisig_plain {
             self.clean_owner_confirmations(&owner);
         }
 
-        /// Replace an owner from the contract with a new one.
+        /// Replace an owner from the contract with a new one. Panics if `old_owner`
+        /// is no owner or if `new_owner` already is one.
         /// Only callable by the wallet itself.
         #[ink(message)]
         fn replace_owner(&mut self, old_owner: AccountId, new_owner: AccountId) {
@@ -205,6 +206,7 @@ mod multisig_plain {
 
         /// Remove a transaction from the contract.
         /// Only callable by the wallet itself.
+        /// Panics if `trans_id` is no valid transaction id.
         #[ink(message)]
         fn cancel_transaction(&mut self, trans_id: TransactionId) {
             self.ensure_from_wallet();
@@ -213,6 +215,7 @@ mod multisig_plain {
 
         /// Confirm a transaction for the sender that was submitted by any owner.
         /// This can be called by any owner.
+        /// Panics if `trans_id` is no valid transaction id.
         #[ink(message)]
         fn confirm_transaction(&mut self, trans_id: TransactionId) {
             self.ensure_caller_is_owner();
@@ -222,6 +225,7 @@ mod multisig_plain {
 
         /// Revoke the senders confirmation.
         /// This can be called by any owner.
+        /// Panics if `trans_id` is no valid transaction id.
         #[ink(message)]
         fn revoke_confirmation(&mut self, trans_id: TransactionId) {
             self.ensure_caller_is_owner();
@@ -251,6 +255,8 @@ mod multisig_plain {
                 .map_err(|_| ())
         }
 
+        /// Set the `transaction` as confirmed by `confirmer`. Idempotent operation
+        /// regarding an already confirmed `transaction` by `confirmer`.
         fn confirm_by_caller(
             &mut self,
             confirmer: AccountId,
@@ -266,6 +272,8 @@ mod multisig_plain {
             }
         }
 
+        /// Get the index of `owner` in `self.owners`. Panics if `owner` is not found
+        /// in `self.owners`.
         fn owner_index(&self, owner: &AccountId) -> u32 {
             self.owners.iter().position(|x| *x == *owner).expect(
                 "This is only called after it was already verified that the id is
@@ -273,6 +281,8 @@ mod multisig_plain {
             ) as u32
         }
 
+        /// Remove the transaction identified by `trans_id` from `self.transactions` and
+        /// removes all confirmation state associated with it.
         fn take_transaction(&mut self, trans_id: TransactionId) -> Option<Transaction> {
             let transaction = self.transactions.take(trans_id);
             if transaction.is_some() {
@@ -281,6 +291,8 @@ mod multisig_plain {
             transaction
         }
 
+        /// Remove all confirmation state associated with `owner` and adjust the
+        /// `self.confirmation_count` variable.
         fn clean_owner_confirmations(&mut self, owner: &AccountId) {
             for (trans_id, _) in self.transactions.iter() {
                 if self.confirmations.remove(&(trans_id, *owner)).is_some() {
@@ -290,6 +302,7 @@ mod multisig_plain {
             }
         }
 
+        /// This removes all confirmation state associated with `transaction`.
         fn clean_transaction_confirmations(&mut self, transaction: TransactionId) {
             for owner in self.owners.iter() {
                 self.confirmations.remove(&(transaction, *owner));
@@ -297,6 +310,8 @@ mod multisig_plain {
             self.confirmation_count.remove(&transaction);
         }
 
+        /// Panic if transaction `trans_id` is not confirmed by at least
+        /// `self.requirement` owners.
         fn ensure_confirmed(&self, trans_id: TransactionId) {
             assert!(
                 self.confirmation_count
@@ -306,17 +321,21 @@ mod multisig_plain {
             );
         }
 
+        /// Panic of the transaction `trans_id` does not exit.
         fn ensure_transaction_exists(&self, trans_id: TransactionId) {
             self.transactions.get(trans_id).expect(WRONG_TRANSACTION_ID);
         }
 
+        /// Panic if the sender is no owner of the wallet.
         fn ensure_caller_is_owner(&self) {
-            assert!(self.is_owner.contains_key(&self.env().caller()));
+            self.ensure_owner(&self.env().caller());
         }
 
+        /// Panic if the sender is not this wallet.
         fn ensure_from_wallet(&self) {
             assert!(self.env().caller() == self.env().account_id());
         }
+
 
         fn ensure_owner(&self, owner: &AccountId) {
             assert!(self.is_owner.contains_key(owner));
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -467,11 +467,17 @@ mod multisig_plain {
                 .expect("Test environment is expected to be initialized.")
         }
 
+        fn build_contract() -> MultisigPlain {
+            let accounts = default_accounts();
+            let owners = ink_prelude::vec![accounts.alice, accounts.bob, accounts.eve];
+            MultisigPlain::new(owners, 2)
+        }
+
         #[test]
         fn construction_works() {
             let accounts = default_accounts();
             let owners = ink_prelude::vec![accounts.alice, accounts.bob, accounts.eve];
-            let contract = MultisigPlain::new(owners.clone(), 2);
+            let contract = build_contract();
 
             assert_eq!(contract.owners.len(), 3);
             assert_eq!(*contract.requirement.get(), 2);
@@ -485,6 +491,36 @@ mod multisig_plain {
             assert_eq!(contract.confirmations.len(), 0);
             assert_eq!(contract.confirmation_count.len(), 0);
             assert_eq!(contract.transactions.len(), 0);
+        }
+
+        #[test]
+        #[should_panic]
+        fn empty_owner_construction_fails() {
+            MultisigPlain::new(vec![], 0);
+        }
+
+        #[test]
+        #[should_panic]
+        fn zero_requirement_construction_fails() {
+            let accounts = default_accounts();
+            MultisigPlain::new(vec![accounts.alice, accounts.bob], 0);
+        }
+
+        #[test]
+        #[should_panic]
+        fn too_large_requirement_construction_fails() {
+            let accounts = default_accounts();
+            MultisigPlain::new(vec![accounts.alice, accounts.bob], 3);
+        }
+
+        #[test]
+        fn add_owner_works() {
+            let accounts = default_accounts();
+            let mut contract = build_contract();
+            let owners = contract.owners.len();
+            contract.add_owner(accounts.frank);
+            assert_eq!(contract.owners.len(), owners + 1);
+            assert!(contract.is_owner.get(&accounts.frank).is_some());
         }
     }
 }
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -108,7 +108,7 @@ mod multisig_plain {
         callee: AccountId,
         /// The selector bytes that identifies the function of the callee that should be called.
         selector: [u8; 4],
-        /// The raw parameters that are passed to the called function.
+        /// The SCALE encoded parameters that are passed to the called function.
         input: Vec<u8>,
         /// The amount of chain balance that is transferred to the callee.
         transferred_value: Balance,
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -399,6 +399,7 @@ mod multisig_plain {
         /// This can be called by any owner.
         ///
         /// # Panics
+        ///
         /// If `trans_id` is no valid transaction id.
         #[ink(message)]
         fn confirm_transaction(&mut self, trans_id: TransactionId) -> ConfirmationStatus {
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -80,7 +80,7 @@ mod multisig_plain {
     use ink_prelude::vec::Vec;
     use scale::Output;
 
-    /// Tune this to your liking but be wary that many owners will not perform well.
+    /// Tune this to your liking but be wary that allowing too many owners will not perform well.
     const MAX_OWNERS: u32 = 50;
 
     type TransactionId = u32;
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -7,7 +7,9 @@
 //! Each instantiation of this contract has a set of `owners` and a `requirement` of
 //! how many of them need to agree on a `Transaction` for it to be able to be executed.
 //! Every owner can submit a transaction and when enough of the other owners confirm
-//! it will be able to be executed.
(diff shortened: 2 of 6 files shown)


athei:
Can you please comment on if you think that `execute_transaction` is correct? I assumed that this allows calling another contract with the contract itself ("the wallet") as the origin. I am not sure if this is correct.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+            env::call::CallParams::<MyEnv, ()>::invoke(t.callee, t.selector.into())
+                .gas_limit(t.gas_limit)
+                .transferred_value(t.transferred_value)
+                .push_arg(&CallInput(&t.input))
+                .fire()
+                .map(|_| ())
+                .map_err(|_| ())
+        }
+
+        fn internal_confirm(&mut self, id: TransactionId) {
```
Robbepop: needs better name, e.g. `confirm_transaction_by_caller`

Robbepop: but also I'd like to keep internal methods more descriptive, e.g. adding this `confirmer: AccountId` to the params and invoking it with `self.env().caller()` on call-site instead of using `self.env().caller()` in the implementation. but this is details.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        fn ensure_transaction_exists(&self, id: TransactionId) {
+            self.transactions.get(id).unwrap();
+        }
+
+        fn ensure_from_owner(&self) {
+            assert!(self.is_owner.contains_key(&self.env().caller()));
+        }
+
+        fn ensure_from_wallet(&self) {
```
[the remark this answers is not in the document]
Robbepop: also what about call stacks where in-between the caller is not the same `account_id`?
E.g. `multi-sig` calls contract `A` calls contract `multi-sig`, then this is not identifiable.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+                .insert((id, self.env().caller()), ())
+                .is_none()
+            {
+                self.confirmation_count
+                    .mutate_with(&id, |count| *count += 1);
+            }
+        }
+
+        fn owner_index(&self, owner: &AccountId) -> u32 {
+            self.owners.iter().position(|x| *x == *owner).unwrap() as u32
```
Robbepop: use `expect` instead of `unwrap`

Robbepop: might it makes sense for `self.owners` to be of type `storage::HashMap` or `storage::BTreeMap` instead? Why is it a `storage::Vec`?

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+                self.clean_transaction_confirmations(id);
+            }
+            transaction
+        }
+
+        fn clean_owner_confirmations(&mut self, owner: &AccountId) {
+            for (id, _) in self.transactions.iter() {
+                if self.confirmations.remove(&(id, *owner)).is_some() {
+                    self.confirmation_count
+                        .mutate_with(&id, |count| *count -= 1);
```
Robbepop: Couldn't we simply do this after clean-up of owners and set it directly to `0`? Would be more gas efficient.

Robbepop: I see this being a performance hazard in case of many owners and transactions due to double-loop. We have to make sure that this is as efficient as possible and won't be a trap for gas costs.

athei:
> All in all looks good but needs adjustments and definitely documentation and plenty of tests.

Yes there will be definitely documentation and thorough testing (as soon as the problem with the tests is resolved).

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use ink_lang as ink;
+
+#[ink::contract(version = "0.1.0", env = MyEnv)]
```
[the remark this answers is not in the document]
athei: It did not work outside of the module. I guess the user of this type will be generated inside of he module. Maybe with an absolute path it would work when the type is declared outside.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+mod multisig_plain {
+    use ink_core::{
+        env,
+        storage,
+    };
+    use ink_prelude::vec::Vec;
+    use scale::Output;
+
+    type TransactionId = u32;
+    type MyEnv = env::DefaultEnvTypes;
```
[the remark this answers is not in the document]
athei: Yes doing it that way feels awkward. The only place where it is set should be as a parameter to the proc macro.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+        callee: AccountId,
+        selector: [u8; 4],
+        input: Vec<u8>,
+        transferred_value: Balance,
+        gas_limit: u64,
+    }
+
+    #[ink(storage)]
+    struct MultisigPlain {
+        confirmations: storage::HashMap<(TransactionId, AccountId), ()>,
```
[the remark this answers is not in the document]
athei: I think this is reasonable. Multi dimensional maps are not that exotic.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+        confirmation_count: storage::HashMap<TransactionId, u32>,
```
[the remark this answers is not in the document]
athei: I guess this is only for the time being? Because as an example the `HashMap` is the more favorable structure in this case, right?

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+        required: storage::Value<u32>,
+    }
+
+    impl MultisigPlain {
+        #[ink(constructor)]
+        fn new(&mut self, owners: Vec<AccountId>, required: u32) {
+            for owner in owners.iter() {
+                self.is_owner.insert(*owner, ());
+                self.owners.push(*owner);
+            }
```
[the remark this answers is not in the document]
athei: Seems like a good idea. Loops are ugly.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
Robbepop: This was more about the `()` as value type.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+            self.confirmation_count.remove(&transaction);
+        }
+
+        fn ensure_confirmed(&self, id: TransactionId) {
+            assert!(self.confirmation_count.get(&id).unwrap() >= self.required.get());
+        }
+
+        fn ensure_transaction_exists(&self, id: TransactionId) {
+            self.transactions.get(id).unwrap();
```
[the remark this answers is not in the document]
athei: Are you sure about that? Basically what we are assuming here is: "Okay the user gave us an invalid id. We do not care: Just rollback". So there is no real proof behind that that this cannot happen. It is totally plausible that this happens. It's not like in runtime where we actually `expect` it to not happen.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: You are right. I did not read properly. This also reasonable :D

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
Robbepop: Well that's what you would write into `expect` to make clear that panicking upon this situation is actually the way this API works. Otherwise contract devs looking at this code might think that this is a logic error etc.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
Robbepop: Generally yes but until we don't know how to implement an efficient non-attackable `storage::HashMap` with our current `storage` data structure design we should advocate users to use `storage::BTreeMap` for all cases where normally `storage::HashMap` would better be used.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: That seems reasonable. I will remove all unwraps.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: https://github.com/paritytech/ink/issues/330

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: https://github.com/paritytech/ink/issues/332

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: https://github.com/paritytech/ink/issues/333

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+        fn ensure_owner(&self, owner: &AccountId) {
+            assert!(self.is_owner.contains_key(owner));
+        }
+
+        fn ensure_no_owner(&self, owner: &AccountId) {
+            assert!(!self.is_owner.contains_key(owner));
+        }
+    }
+
+    fn ensure_requirement(owners: u32, required: u32) {
```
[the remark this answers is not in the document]
athei: Done.

examples/multisig_plain/lib.rs
athei: Are you sure you are commenting on the correct line? Is see no double loop here.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: I am not sure how to progress with this item. Should I crate an issue for this? I do not find that behavior surprising. (I am trying to get all items to resolved).

examples/multisig_plain/lib.rs
athei: Done. Resolve if satisfied.

riusricardo:
Just a heads-up. You need to implement a few more traits for `Transaction` to build the metadata. 
`cargo contract generate-metadata` fails to build.

athei:
Yeah I noticed this already. Thanks for pointing that out. I will add that to the TODO of this PR.

athei:
Yeah I noticed this already. Thanks for pointing that out. I will add that to the TODO of this PR.

examples/multisig_plain/lib.rs
athei: Because I have to iterate over them in `clean_transaction_confirmations`. As matter of fact there is also a owner HashMap called `is_owner`. This is used to avoid iteration on the more common operations.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: I do not see the problem. Then the second `multi-sig` sees `A` as the sender which is exactly what I want to happen. Also I do not see how the chosen name is not good, I want to ensure that specific functions call only be called by the contract.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        fn clean_transaction_confirmations(&mut self, transaction: TransactionId) {
+            for owner in self.owners.iter() {
+                self.confirmations.remove(&(transaction, *owner));
+            }
+            self.confirmation_count.remove(&transaction);
+        }
+
+        fn ensure_confirmed(&self, id: TransactionId) {
+            assert!(self.confirmation_count.get(&id).unwrap() >= self.required.get());
```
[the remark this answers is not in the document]
athei: Done. Reopen if necessary.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: Done. Reopen if necessary.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: https://github.com/paritytech/ink/issues/334

athei:
Does anyone understand why the `examples-fmt` CI job fails? I cannot find a helpful message in the output.

Robbepop:
> 
> 
> Does anyone understand why the `examples-fmt` CI job fails? I cannot find a helpful message in the output.

```
  //! ```0 < requirement && requirement <= owners && owners <= MAX_OWNERS```
-//!
+//! 
```
Is the output. So it is unhappy about some double empty lines or some whitespace that follows after a new doc-line. Also I would definitely put the contents of the ````` into another line, e.g.

athei:
You are right. It just strikes me as odd that it does not tell you what is wrong. Only where something went wrong.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
Robbepop: outdated comments from what I can tell. resolved

examples/multisig_plain/lib.rs
Robbepop: still requires change to `expect` with proper message

athei: It is. This snippet is just outdated.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: I switched to `BTreeMap`

athei:
From my point of view all TODOs and comments are dealt with. Requested Review.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+                self.owners.push(*owner);
+            }
+            self.ensure_requirement_is_valid(self.owners.len(), requirement);
+            assert!(self.is_owner.len() == self.owners.len());
+            self.requirement.set(requirement);
+        }
+
+        /// Add a new owner to the contract.
+        /// Panics if the owner already exists.
+        /// Only callable by the wallet itself.
```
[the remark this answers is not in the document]
Robbepop: The above might go too much into detail but it is a general doc-style preferred by the ink! codebase and Rust in general.
First line is short description, then empty line, followed by an optional longer description or note.
Also we make use of the general Rust docs-design principles to put `# Errors`, `# Panics`, `# Examples`, `# Note`, `# Safety` etc. sections in docs if needed.
Please apply to the other docs as well.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+        #[ink(message)]
+        fn add_owner(&mut self, new_owner: AccountId) {
+            self.ensure_from_wallet();
+            self.ensure_no_owner(&new_owner);
+            self.ensure_requirement_is_valid(self.owners.len() + 1, *self.requirement);
```
Robbepop: Btw.: I really like the defensive programming style used in this PR. :) Really good example of how contracts should be written.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+            self.is_owner.insert(new_owner, ());
+            self.owners.push(new_owner);
+            self.env().emit_event(OwnerAddition { owner: new_owner });
+        }
+
+        /// Remove an owner from the contract.
+        /// Only callable by the wallet itself. If by doing this the amount of owners
+        /// would be smaller than the requirement it is adjusted to be exactly the
+        /// number of owners. Panics if `owner` is no owner of the wallet.
```
Robbepop: Is this the semantics of the original multi-sig because I kinda dislike this implied effect but maybe I just don't know why this is useful to have. I'd generally just trap (panic) in such cases. If it is the original semantics we might just keep it as it is though.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+        #[ink(message)]
+        fn remove_owner(&mut self, owner: AccountId) {
+            self.ensure_from_wallet();
+            self.ensure_owner(&owner);
+            let len = self.owners.len() - 1;
+            let requirement = u32::min(len, *self.requirement.get());
```
Robbepop:
```suggestion
            let requirement = u32::min(len, *self.requirement);
```

Robbepop: Maybe I am wrong but it should generally work without the `.get()`

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+            let result = env::call::CallParams::<EnvTypes, ()>::invoke(
+                t.callee,
+                t.selector.into(),
+            )
+            .gas_limit(t.gas_limit)
+            .transferred_value(t.transferred_value)
+            .push_arg(&CallInput(&t.input))
+            .fire()
+            .map(|_| ())
+            .map_err(|_| ());
```
Robbepop: Have a hard time to believe that this is what `rustfmt` actually formatted this code to. 0.o

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+                    transaction: trans_id,
+                });
+            }
+        }
+
+        /// Confirm a transaction for the sender that was submitted by any owner.
+        /// This can be called by any owner.
+        /// Panics if `trans_id` is no valid transaction id.
+        #[ink(message)]
+        fn confirm_transaction(&mut self, trans_id: TransactionId) {
```
Robbepop: Maybe this could profit from a result that tells the caller if the transaction is ready for execution (has enough approvals). What do you think? This way you can more easily emulate the old-style of executing such a transaction right away. Or we could provide another method that does this job of the old semantics.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+                self.env().emit_event(Revokation {
+                    transaction: trans_id,
+                    from: caller,
+                });
+            }
+        }
+
+        /// Execute a confirmed execution.
+        /// Its return type indicates whether the called transaction was succesful.
+        /// This can be called by anyone.
```
Robbepop: Does the original multi-sig return the result of the called function? Because this implementation doesn't but I think we could return some bytes back to the caller. This, however, should be another call since that's more work done. What do you think? Or should we keep it simple?

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+                });
+            }
+        }
+
+        /// Get the index of `owner` in `self.owners`.
+        /// Panics if `owner` is not found in `self.owners`.
+        fn owner_index(&self, owner: &AccountId) -> u32 {
+            self.owners.iter().position(|x| *x == *owner).expect(
+                "This is only called after it was already verified that the id is
+                actually an owner.",
```
Robbepop:
```suggestion
                 actually an owner.",
```

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+        fn ensure_confirmed(&self, trans_id: TransactionId) {
+            assert!(
+                self.confirmation_count
+                    .get(&trans_id)
+                    .expect(WRONG_TRANSACTION_ID)
+                    >= self.requirement.get()
+            );
+        }
+
+        /// Panic of the transaction `trans_id` does not exit.
```
Robbepop:
```suggestion
        /// Panic if the transaction `trans_id` does not exit.
```

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+        K: Eq + Ord + scale::Codec + Copy,
+        V: scale::Codec + Copy,
+        F: FnOnce(&mut V),
+    {
+        let mut count = *map
+            .get(key)
+            .expect("User is responsible for only supplying existing keys.");
+        f(&mut count);
+        map.insert(*key, count);
+    }
```
Robbepop: Should we maybe have this in our `storage::BTreeMap` implementation? Is there a Rust `BTreeMap` equivalent

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+            let owners = contract.owners.len();
+            contract.add_owner(accounts.frank);
+            assert_eq!(contract.owners.len(), owners + 1);
+            assert!(contract.is_owner.get(&accounts.frank).is_some());
+            assert_eq!(test::recorded_events().count(), 1);
+        }
+
+        #[test]
+        #[should_panic]
+        fn add_owner_existing_fails() {
```
Robbepop:
```suggestion
        fn add_existing_owner_fails() {
```

Robbepop: Please no [Yoda](https://en.wikipedia.org/wiki/Yoda_conditions) identifiers. ;)

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+        fn cancel_transaction_works() {
+            let mut contract = submit_transaction();
+            set_from_wallet();
+            contract.cancel_transaction(0);
+            assert_eq!(contract.transactions.len(), 0);
+            assert_eq!(test::recorded_events().count(), 3);
+        }
+
+        #[test]
+        fn cancel_transaction_nonexisting() {
```
Robbepop: So it doesn't panic when I cancel a non-existing transaction?

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+            let mut contract = submit_transaction();
+            let accounts = default_accounts();
+            set_sender(accounts.django);
+            contract.revoke_confirmation(0);
+        }
+
+        #[test]
+        fn execute_transaction_works() {
+            // Execution of calls is currently unsupported in off-chain test.
+            // Calling execute_transaction panics in any case.
```
Robbepop: So sad. :(

examples/multisig_plain/.cargo/config
```diff
@@ -0,0 +1,4 @@
+[target.wasm32-unknown-unknown]
+rustflags = [
+	"-C", "link-args=-z stack-size=65536 --import-memory"
+]
```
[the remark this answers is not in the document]
athei: Aye. It was created by `cargo contract`. Is that already fixed there?

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+//! `submit_transaction`, `cancel_transaction`, `confirm_transaction`,
+//! `revoke_confirmation` and `execute_transaction` are the bread and butter messages
+//! of this contract. Use them to dispatch arbitrary messages to other contracts
+//! with the wallet as a sender.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use ink_lang as ink;
+
+#[ink::contract(version = "0.1.0", env = ink_core::env::DefaultEnvTypes)]
```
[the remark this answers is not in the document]
athei: I know. I just wanted to leave that because there is no other documentation on how to specify an `EnvType`.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+
+    /// A Transaction is what every `owner` can submit for confirmation by other owners.
+    /// If enough owners agree it will be executed by the contract.
+    #[derive(scale::Encode, scale::Decode, storage::Flush)]
+    #[cfg_attr(feature = "ink-generate-abi", derive(type_metadata::Metadata))]
+    #[cfg_attr(feature = "std", derive(Debug, PartialEq, Eq))]
+    pub struct Transaction {
+        /// The AccountId of the contract that is called in this transaction.
+        callee: AccountId,
+        /// The raw selector which is the function name of the `callee`that is called.
```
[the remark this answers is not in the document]
athei: I think for an learner primitive types are certainly easier to grasp because there is no need to look them up. This was my motivation to choose them here.

examples/multisig_plain/lib.rs
athei: Yes this is the gnosis semantic. But we can change it. We changed other things already.

examples/multisig_plain/lib.rs
athei: You are right.

examples/multisig_plain/lib.rs
athei: This makes sense. Now that the confirmation check is cheap this might be viable to return something useful here.

examples/multisig_plain/lib.rs
athei: Ethereum contracts cannot return anything AFAIK. You completely rely on events there. Therefore you would detect the failure by listening for the callees events.

I think you are right. We should make use of return values as its possible in substrate.

If you put this in a separate call you must store the result in contract storage. I do not know if this would be so good.

examples/multisig_plain/lib.rs
athei: It did and the CI agrees that it's OK.

examples/multisig_plain/lib.rs
athei: I think we should. I was surprised that there was no `mutate_with` like with he `HashMap`.

examples/multisig_plain/lib.rs
athei: Comply I must.

examples/multisig_plain/lib.rs
Robbepop: But on the other handside it has an entry API.

examples/multisig_plain/lib.rs
athei: Yep. It just does nothing. The code is just a little simpler that way and uses less gas in the common case (at least I think it does).

examples/multisig_plain/lib.rs
athei: Maybe we can change it from panic to "nothing at all". Then I can at least test if it tried to execute or failed for some other reason.

Robbepop: We could try but on the other hand I don't think that this feature is even on the horizon.

athei: I guess we leave it that way then and re-iterate on the example later.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+    type TransactionId = u32;
+    const WRONG_TRANSACTION_ID: &str =
+        "The user specified an invalid transaction id. Abort.";
+
+    /// A wrapper that allows us to pass untyped parameters as blob to a `CallBuilder`
+    struct CallInput<'a>(&'a [u8]);
+
+    impl<'a> scale::Encode for CallInput<'a> {
+        fn encode_to<T: Output>(&self, dest: &mut T) {
+            dest.write(self.0);
```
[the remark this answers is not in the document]
athei: Are you sure? My understanding is that `Output::write` only writes out the bytes and that `Output::push` would do the SCALE encoding of the slice.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+        transferred_value: Balance,
+        /// Gas limit for the transation.
+        gas_limit: u64,
+    }
+
+    #[ink(storage)]
+    struct MultisigPlain {
+        /// Every entry in this map represents the confirmation of an owner for a
+        /// transaction. This is effecively a set rather than a map.
+        confirmations: storage::BTreeMap<(TransactionId, AccountId), ()>,
```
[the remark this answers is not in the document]
athei: I decided against that when implementing this contract for the following reason: You need stable owner indices for that. If you do enough owner removals additions and replacements you finally run out of indices and your contract is broken because you ran out of indices.

Maybe we can live with that because 64 ids is plenty but I wanted to opt for the safe approach in this example.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
Robbepop: Ah I think you are right and I confused it with `push`. Sorry!

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+        /// The amount of confirmations for every transaction. This is a redundant
+        /// information this kept in order to prevent iterating through the
+        /// confirmation set to check if a transaction is confirmed.
+        confirmation_count: storage::BTreeMap<TransactionId, u32>,
+        /// Just the list of transactions. It is a stash as stable ids are necessary
+        /// for referencing them in confirmation calls.
+        transactions: storage::Stash<Transaction>,
+        /// The list is a vector because iterating over it is necessary when cleaning
+        /// up the confirmation set.
```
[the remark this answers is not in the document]
athei: I switched to `BTreeMap` which I cannot iterate either. Should I wait then for the iterator? Or iterate after merge?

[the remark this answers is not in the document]
Robbepop: If even `BTreeMap` is missing iteration then just keep it as it is for now.

examples/multisig_plain/lib.rs
athei: I think we should leave it the way it is for the following reason: If we introduce the panic behavior you need to lower the requirement before removing an owner. This introduces a race where after lowering the requirement a transaction can get confirmed by too few people (including the owner you want to remove).

examples/multisig_plain/lib.rs
athei: I removed the function and switched to the entry API.

examples/multisig_plain/lib.rs
athei: I split `execute_transaction` into `eval_transaction` and `invoke_transaction`. The former one retrieves the output of the call as bytes.

examples/multisig_plain/lib.rs
```diff
@@ -0,0 +1,837 @@
...
+        /// The new requirement value.
+        new_requirement: u32,
+    }
+
+    impl MultisigPlain {
+        /// The only constructor of the contract.
+        /// A list of owners must be supplied and a number of how many of them must
+        /// confirm a transaction. Duplicate owners are silently dropped.
+        #[ink(constructor)]
+        fn new(&mut self, owners: Vec<AccountId>, requirement: u32) {
```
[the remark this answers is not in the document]
athei: Done.

examples/multisig_plain/lib.rs
athei: I added a `ConfirmationStatus`type that is returned from `submit_transaction` and `confirm_transaction`. It is also included in the confirmation event.

examples/multisig_plain/lib.rs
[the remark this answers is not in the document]
athei: I have adjusted the documentation of the public items accordingly (and added an example).

examples/multisig_plain/lib.rs
```diff
@@ -186,6 +193,17 @@ mod multisig_plain {
...
+    #[ink(event)]
+    struct Evaluation {
+        /// The transaction that was executed.
+        #[ink(topic)]
+        transaction: TransactionId,
+        /// The output of the transaction as bytes.
+        #[ink(topic)]
+        result: Result<Vec<u8>, ()>,
+    }
+
```
Robbepop: I think it would be more user friendly to have a single event for this that unifies both events.
E.g. convert `result` type to `Result<Option<Vec<u8>>, ()>` or something like that to signal if there was a return type or not.

examples/multisig_plain/lib.rs
```diff
@@ -255,8 +260,36 @@ mod multisig_plain {
...
+        ///     input: call.params().to_owned(),
+        ///     transferred_value: 0,
+        ///     gas_limit: 5000
+        /// };
+        ///
+        /// // submit the transaction for confirmation
+        /// let (id, _) = submit_transaction(transaction);
+        /// // wait until all required owners have confirmed and then execute the transaction
+        /// invoke_transaction(id);
+        /// ```
```
Robbepop: nice

examples/multisig_plain/lib.rs
```diff
@@ -339,8 +395,11 @@ mod multisig_plain {
         }
 
         /// Confirm a transaction for the sender that was submitted by any owner.
+        ///
         /// This can be called by any owner.
-        /// Panics if `trans_id` is no valid transaction id.
+        ///
+        /// # Panics
+        /// If `trans_id` is no valid transaction id.
```
Robbepop:
```suggestion
        /// # Panics
        ///
        /// If `trans_id` is no valid transaction id.
```

examples/multisig_plain/lib.rs
athei: Done.

examples/multisig_plain/lib.rs
athei: I reworked it to use `CallParams` to call into the contract. So it's less pseudo code and will actually compile once rustdocs will try to compile the example.

athei:
I rebased the branch onto the current master. I guess the force push removed your approval. Wondering if a merge would have done that, too. Should I avoid force pushing to PRs?

Robbepop:
> I rebased the branch onto the current master. I guess that removed your approval. Wondering if a merge would have done that, too. Should I avoid force pushing to PRs?

Since we squash merge in the end a merge wouldn't be too bad. ;)

Robbepop:
> > I rebased the branch onto the current master. I guess that removed your approval. Wondering if a merge would have done that, too. Should I avoid force pushing to PRs?
> 
> Since we squash merge in the end a merge wouldn't be too bad. ;)

I would have loved a CI output for this new example contract before merge but unfortunately CI is too broken. :( @TriplEight we really need a fix for this.

ascjones:
> I guess that removed your approval

Yeah this repo is configured to "dimiss" previous reviews when any more pushes are made to the PR branch, so requires a fresh approval

TriplEight:
These CI errors are not project-specific, restart of the job helps with them. I'm working on them, but right now have no clue what is the cause.

TriplEight:
Now, these examples failures are something you will want to fix.

athei:
> Now, these examples failures are something you will want to fix.

I am not sure what you are saying. Is there anything that I need to do right now?

TriplEight:
> > These CI errors are not project-specific, restart of the job helps with them. I'm working on > them, but right now have no clue what is the cause.
> 
> > Now, these examples failures are something you will want to fix.
> 
> I am not sure what you are saying. Is there anything that I need to do right now?

Nope, this kind of error is fixed in https://github.com/paritytech/ink/pull/343

athei:
Nice. I will rebase this branch onto this changes as soon as they hit master.

Robbepop:
Clippy warnings in this PR. We can merge once fixed.

athei:
Rebased on current master and fixed the clippy warning. CI still fails:
```
 error: failed to load source for a dependency on `derive_more`
Caused by:
  Unable to update registry `https://github.com/rust-lang/crates.io-index`
Caused by:
  failed to fetch `https://github.com/rust-lang/crates.io-index`
Caused by:
  missing delta bases; class=Indexer (15)
```

TriplEight:
@athei this is a different bug, I'm not sure what's the cause, but I'm on it. This is something random that happens on several projects CIs.

PR #343, opened 2020-03-09
CI fixes

TriplEight:
- running 
```
cargo doc --no-deps --all-features -p type-metadata -p ink_abi -p ink_abi_derive -p ink_core -p ink_core_derive -p ink_primitives -p ink_prelude -p ink_lang -p ink_lang_macro
```
to cover all packages.
- also set `GIT_DEPTH` to 100 to avoid some potential failures I've had with substrate, it will not affect CI speed more than on a second.
- had to `cargo upgrade` to satisfy `trybuild v1.0.23` which now expects everything to be up-to-date. (and [the pipeline](https://gitlab.parity.io/parity/ink/pipelines/82198) succeded)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -227,7 +227,7 @@ publish-docs:
     - git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
     - git fetch origin gh-pages
     # Generating Docs
-    - time cargo doc --no-deps --all-features 
+    - time cargo +nightly doc --no-deps --all-features 
         -p type-metadata -p ink_abi -p ink_abi_derive -p ink_core -p ink_core_derive 
         -p ink_primitives -p ink_prelude -p ink_lang -p ink_lang_macro
     # saving README and docs
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,7 +13,7 @@ stages:
 
 variables:
   GIT_STRATEGY:                    fetch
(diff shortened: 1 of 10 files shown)


.gitlab-ci.yml
```diff
@@ -212,10 +212,10 @@ publish-docs:
...
   variables:
     GIT_DEPTH:                     0
-  only:
-    - master
-    - schedules
-    - tags
+  # only:
+  #   - master
+  #   - schedules
+  #   - tags
```
athei: Please delete code that should be removed instead of commenting it.

.gitlab-ci.yml
```diff
@@ -13,7 +13,7 @@ stages:
 
 variables:
   GIT_STRATEGY:                    fetch
-  GIT_DEPTH:                       3
+  GIT_DEPTH:                       100
```
athei: Why is this increased to such a huge number? Also it seems unrelated to the PR.

TriplEight: also checking one theory, will place WIP here to solve some issues

.gitlab-ci.yml
TriplEight: it's commented for testing purposes, shouldn't be removed

TriplEight:
I too do not like it that much, but it's not that bad since it's not strict bound to exact versions (as with `"=1.0.23"`. Besides, after fixing that bug in `trybuild` we will be able to return to the more convenient versioning.

PR #345, opened 2020-03-11
Revert precise version specs and use new trybuild

ascjones:
Latest version of `trybuild` includes this fix https://github.com/dtolnay/trybuild/pull/54, so should no longer require the precise versions introduced in #343.


core/derive/tests/compile_tests.rs
```diff
@@ -1,22 +0,0 @@
...
-// limitations under the License.
-
-#[test]
-fn compile_tests() {
-    let t = trybuild::TestCases::new();
-    t.pass("tests/ui/pass/flush.rs");
-    t.pass("tests/ui/pass/allocate_using.rs");
-    t.compile_fail("tests/ui/fail/flush_empty_enum.rs");
-    t.compile_fail("tests/ui/fail/allocate_using_enum.rs");
-}
```
Robbepop: Not sure why you would delete all of our tests.

ascjones: Troubleshooting, they'll be back

.gitlab-ci.yml
```diff
@@ -13,7 +13,7 @@ stages:
 
 variables:
   GIT_STRATEGY:                    fetch
-  GIT_DEPTH:                       100
+#  GIT_DEPTH:                       100
```
Robbepop: Best would be if we could make this `1`.

ascjones: I'm just testing whether having it set it is causing the issue we are seeing

PR #349, opened 2020-03-16
Reworkaround clippy bug

TriplEight:
removed exact Rust nightly versions since workaround is not needed anymore.


TriplEight:
CI will ne rgeener after https://github.com/paritytech/scripts/pull/164

PR #351, opened 2020-03-17
Add support for ext_transfer

Robbepop:
Adds support for `ext_transfer` call introduced in https://github.com/paritytech/substrate/pull/5169.

### DIFF
--- a/core/src/env/api.rs
+++ b/core/src/env/api.rs
@@ -395,6 +395,33 @@ pub fn restore_contract<T>(
     })
 }
 
+/// Transfers value from the contract to the destination account ID.
+///
+/// # Note
+///
+/// This is more efficient and simpler than the alternative to make a no-op
+/// contract call or invoke a runtime function that has performs the
+/// transaction.
+///
+/// # Errors
+///
+/// If the contract doesn't have sufficient funds.
+pub fn transfer<T>(
+    destination: T::AccountId,
+    value: T::Balance,
+) -> Result<()>
+where
+    T: EnvTypes,
+{
+    <EnvInstance as OnInstance>::on_instance(|instance| {
+        TypedEnv::transfer::<T>(
+            instance,
+            destination,
+            value,
+        )
+    })
+}
+
 /// Returns the input to the executed contract.
 ///
 /// # Note
--- a/core/src/env/api.rs
+++ b/core/src/env/api.rs
@@ -406,19 +406,12 @@ pub fn restore_contract<T>(
 /// # Errors
 ///
 /// If the contract doesn't have sufficient funds.
-pub fn transfer<T>(
-    destination: T::AccountId,
-    value: T::Balance,
-) -> Result<()>
+pub fn transfer<T>(destination: T::AccountId, value: T::Balance) -> Result<()>
 where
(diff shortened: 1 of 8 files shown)


core/src/env/api.rs
```diff
@@ -395,6 +395,26 @@ pub fn restore_contract<T>(
     })
 }
 
+/// Transfers value from the contract to the destination account ID.
+///
+/// # Note
+///
+/// This is more efficient and simpler than the alternative to make a no-op
+/// contract call or invoke a runtime function that has performs the
```
ascjones:
```suggestion
/// contract call or invoke a runtime function that performs the
```

core/src/env/engine/on_chain/impls.rs
```diff
@@ -330,6 +330,21 @@ impl TypedEnv for EnvInstance {
...
 
+    fn transfer<T>(&mut self, destination: T::AccountId, value: T::Balance) -> Result<()>
+    where
+        T: EnvTypes,
+    {
+        // Reset the contract-side buffer to append onto clean slate.
+        self.reset_buffer();
+        // Append the encoded `account_id`, `code_hash` and `rent_allowance`
+        // and `filtered_keys` in order and remember their encoded regions
+        // within the buffer.
```
ascjones: See CI for compiler errors in this method, also this comment needs to be updated. Looks like a copy paste error.

jonahb27:
Is there any documentation and/or example of using this function? Also, how would you transfer to a contract as opposed to an account?

jonahb27:
Big picture, I am looking for the Ink! parallel to these Solidity functions:
payable: keyword to allow contract to receive crypto
msg.value : to get the amount sent to the contract
msg.sender.send(amount): to send crypto to the calling account
beneficiary.transfer(amount): to transfer crypto to an account

athei:
Copied from @Robbepop answered in out riot channel:

> payable: keyword to allow contract to receive crypto

this is going to be introduced in ink! 3.0. That said you can send value to any contract or plain account for that matter. Payable is merely an ABI construct that tells you that you shouldn't send any value and the contract will revert the transaction when you send it via a call. Transfer is mainly for depositing money into plain accounts.

> msg.value : to get the amount sent to the contract

this is called transferred_value in ink!
`self.env().transferred_balance()`
https://paritytech.github.io/ink/ink_lang/struct.EnvAccess.html#method.transferred_balance

> msg.sender.send(amount): to send crypto to the calling account

this can be done using the transfer host function
calling contract can be queried using caller host function
`self.env().transfer(self.env().caller(), 1000.into())`
https://paritytech.github.io/ink/ink_lang/struct.EnvAccess.html#method.transfer

`beneficiary.transfer(amount): to transfer crypto to an account`
same as above but with beneficiary being some concrete account ID

jonahb27:
UNBELIEVABLE! THANK YOU FOR THE TREMENDOUS HELP!

jonahb27:
Is 1000 an arbitrary amount? Or does it mean something?

athei:
The first one: Just some random amount.

jonahb27:
Awesome... thanks!

PR #359, opened 2020-03-20
making CI faster by ~1 min; 

TriplEight:
- `needs:` makes the job not to wait for the previous stage to be over, just a mentioned job
- `cargo test` does not need to run `--release`, it's faster without it, and also it doesn't run `release` profile, it runs `bench`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -114,7 +114,7 @@ test:
     - check-std
   script:
     - for crate in ${ALL_CRATES}; do
-        cargo test --verbose --all-features --manifest-path ${crate}/Cargo.toml;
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -114,7 +114,7 @@ test:
     - check-std
   script:
     - for crate in ${ALL_CRATES}; do
-        cargo test --verbose --all-features --release --manifest-path ${crate}/Cargo.toml;
+        cargo test --verbose --all-features --manifest-path ${crate}/Cargo.toml;
```
Robbepop: This should not be faster in net result since test runs normally rum much slower in non `--release` than compilation takes.

TriplEight: I did some measurements and testing with or without `--release` with caching, do not make much difference for ink.

Issue #358, opened 2020-03-20
Implement contracts pallet provided crypto hash functions


Robbepop:
Implement new crypto hash functions natively supported in contracts pallet.

Substrate issue: https://github.com/paritytech/substrate/issues/5258
Implementing PR: https://github.com/paritytech/substrate/pull/5326

This shall implement:

- `keccak_256`
- `blake2_256`
- `blake2_128`
- `twox_256`
- `twox_128`
- `twox_64`
- `sha2_256`

All of them return `[u8; 8]`, `[u8; 16]` or `[u8; 32]`.

Robbepop:
Implemented in https://github.com/paritytech/ink/pull/360 and merged. Closed.

PR #360, opened 2020-03-21
Implement crypto hash functions supported by contracts pallet

Robbepop:
Implements https://github.com/paritytech/ink/issues/358.

We intentionally do not implement those hash functions on the `EnvAccess` and thus do not allow the `self.hash_fn` syntax in contracts. The intended use case for invoking those hash functions is either directly through the environment or through hash abstractions that we are going to add in the future.

## Screenshots

2020-03-21-150154_1613x596_scrot
2020-03-21-150242_1442x304_scrot

## Future Tasks

We might want to add new example contracts that demonstrate the usage of those hash functions.

### DIFF
--- a/core/src/env/backend.rs
+++ b/core/src/env/backend.rs
@@ -77,6 +77,27 @@ pub trait Env {
 
     /// Prints the given contents to the console log.
     fn println(&mut self, content: &str);
+
+    /// Returns the SHA2 256-bit hash of the given input.
+    fn hash_sha2_256(input: &[u8]) -> [u8; 32];
+
+    /// Returns the KECCAK 256-bit hash of the given input.
+    fn hash_keccak_256(input: &[u8]) -> [u8; 32];
+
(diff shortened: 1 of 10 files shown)


core/src/env/backend.rs
```diff
@@ -77,6 +77,27 @@ pub trait Env {
 
     /// Prints the given contents to the console log.
     fn println(&mut self, content: &str);
+
+    /// Returns the SHA2 256-bit hash of the given input.
+    fn hash_sha2_256(input: &[u8]) -> [u8; 32];
```
athei: Is there a specific reason that you do not use a mutable reference to array as an out parameter? Does the purer interface justify the additional copy of 32 bytes?

Robbepop: You are right that we should provide a lower-level API for the functions implemented here. We can then implement some pure functions of top of those. Thanks.

Issue #362, opened 2020-03-21
Provide high-level hashing API


Robbepop:
With https://github.com/paritytech/ink/pull/360 we added a fairly low-level hashing API to ink!.
While contract writers can already make use of this API and its efficient cryptographic hash functions ink! should provide a more user friendly alternative out of the box.
This includes pure versions of the same functions as well as implementations for Rust's [`Hasher`](https://doc.rust-lang.org/std/hash/trait.Hasher.html) trait.

Robbepop:
Implemented in https://github.com/paritytech/ink/pull/363. Closed.

PR #363, opened 2020-03-21
Initial implementation of high-level hashing infrastructure

Robbepop:
Implements https://github.com/paritytech/ink/issues/362.

**Note:** Might be a bit overtuned. Please review API design.

## Overview

The API supports the following cryptographic hashes:

- `keccak_256`
- `blake2_256`
- `blake2_128`
- `twox_256`
- `twox_128`
- `twox_64`
- `sha2_256`

There are 4 different root hash functions available in `HashBuilder`:

- `hash_raw_using`
    - Takes a byte slice as input and stores the resulting hash into a given output buffer.
- `hash_raw`
    - Takes a byte slice as input and directly returns the resulting hash.
- `hash_encoded_using`
    - Takes an encodable entity as input and stores the resulting hash into a given output buffer.
- `hash_encoded`
    - Takes an encodable entity as input and returns the resulting hash.

## Accumulator

With the concept of `Accumulator` types I tried to make it possible for users to have more control over allocations performed during the hashing of hashable generic entities which is basically done by accumulating their hash state into a growing buffer. If a user wants to avoid heap allocations at all they can potentially provide a custom defined `Accumulator` types and use it via `*_using` APIs in order to do so.

By default only Rust's `Vec<u8>` and `&mut Vec<u8>` implement the `Accumulator` trait.
By default the crate already implements `Accumulator` for the following types:

- Rust's `Vec<u8>`
- The `crate`'s `Wrap` type which allows wrapping mutable byte slices
- All `&mut T` where `T: Accumulator` which is handy for `&mut Vec<u8>`

Since ink! already has quite a huge static bytes buffer (for encoding/decoding purposes) we could in theory use it to avoid allocations. However, this would induce shared state across APIs. Practically this shouldn't be too bad since ink! already is pretty constraint in being single-threaded.

## Hash Support

If users want to use those cryptographic hashes via Rust's `Hash` trait they can since they have built-in support for Rust's `Hash` trait via `CryptoHasher` and `CryptoBuildHasher`.

## Incremental Hash Builders

The current implementation does not support built-in incremental hash builders since contracts pallet does not provide ink! with the low-level primitives for incremental hash building (yet).
However, this design is suited to allow easy integration for built-in supported incremental hash builders once the low-level API for it exists.

The rough outline for adding this support is as follows:

Incremental support by:

1. Adding the following methods to the `Hasher` trait:
```rust
    fn initialize_incremental();
    fn write_incremental(input: &[u8]);
    fn finalize_incremental(output: &mut Self::Output);
```
2. Add `Incremental` marker type: `pub struct Incremental;`
3. Add `impl scale::Output for `HashBuilder<H, Incremental> where H: Hasher` as follows:
```rust
impl<H> scale::Output for HashBuilder<H, Incremental>
where
    H: Hasher,
{
        <H as Hasher>::write_incremental(bytes)
    }
}
```
4. Add `impl Finalize for HashBuilder<H, Incremental> where H: Hasher`.
5. Add calls to `H::initialize_incremental` upon `HashBuilder<H, Incremental>` construction.

## Screenshots of Docs

2020-03-24-142315_1470x1199_scrot
2020-03-24-142302_1453x709_scrot
2020-03-24-142248_1289x829_scrot


core/src/hash/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+    /// This is the simplest way to call this API but does not avoid heap
+    /// allocations.
+    ///
+    /// ```
+    /// # use ink_core::hash::sha2_256_into;
+    /// # const EXPECTED: [u8; 32] = [
+    /// #     123,  38, 102, 217, 176, 174, 109,  30, 100, 71, 85, 214, 203,
+    /// #     212,  16,  26,  88,   5, 194, 156, 138, 243, 34, 74,  67, 115,
+    /// #     178, 200, 65, 118, 14, 253
+    /// # ];
```
cmichi: Why the `#` at the beginning of above lines?

core/src/hash/hasher.rs
```diff
@@ -0,0 +1,360 @@
...
+/// Generic cryptographic hasher.
+pub struct CryptoHasher<T, A>
+where
+    T: markers::Sealed,
+{
+    /// The accumulator buffer.
+    ///
+    /// A bytes buffer that is used to accumulate a hashing state.
+    accumulator: A,
+    /// Tricks compiler into thinking that `Self` uses `T`.
```
NikVolf: seems over-documenting, it is not even a public member

core/src/hash/hasher.rs
```diff
@@ -0,0 +1,360 @@
...
+
+impl Accumulator for Vec<u8> {
+    fn reset(&mut self) {
+        <Vec<_>>::clear(self)
+    }
+
+    fn write(&mut self, bytes: &[u8]) {
+        // This could theoretically be speed-up by using `unsafe` `set_len`
+        // and `[u8]` `copy_from_slice` methods.
+        <Vec<_>>::extend_from_slice(self, bytes)
```
NikVolf:
```suggestion
        self.extend_from_slice(bytes)
```

core/src/hash/hasher.rs
```diff
@@ -0,0 +1,360 @@
...
+    fn reset(&mut self);
+    /// Writes the given bytes into the buffer.
+    fn write(&mut self, bytes: &[u8]);
+    /// Returns a shared reference to the slice of the current state of the buffer.
+    fn as_slice(&self) -> &[u8];
+}
+
+impl Accumulator for Vec<u8> {
+    fn reset(&mut self) {
+        <Vec<_>>::clear(self)
```
NikVolf:
```suggestion
        self.clear()
```

core/src/hash/hasher.rs
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    fn write(&mut self, bytes: &[u8]) {
+        // This could theoretically be speed-up by using `unsafe` `set_len`
+        // and `[u8]` `copy_from_slice` methods.
+        <Vec<_>>::extend_from_slice(self, bytes)
+    }
+
+    fn as_slice(&self) -> &[u8] {
+        <Vec<_>>::as_slice(self)
```
NikVolf:
```suggestion
        self.as_slice()
```

core/src/hash/hasher.rs
```diff
@@ -0,0 +1,360 @@
...
+
+impl<'a> Accumulator for Wrap<'a> {
+    fn reset(&mut self) {
+        self.len = 0;
+    }
+
+    fn write(&mut self, bytes: &[u8]) {
+        let len = self.len;
+        let bytes_len = bytes.len();
+        <[u8]>::copy_from_slice(&mut self.buffer[len..(len + bytes_len)], bytes);
```
NikVolf:
```suggestion
        self.buffer[len..(len + bytes_len)].copy_from_slice(bytes);
```

core/src/hash/hasher.rs
```diff
@@ -0,0 +1,360 @@
...
+    }
+}
+
+/// Hash functions that can put their results into an output buffer
+/// described by the `Output` type.
+///
+/// # Note
+///
+/// This is an internal trait used to deduplicate some implementations.
+pub trait FinishInto<Output> {
```
NikVolf: Can `FinishInto` be implemented with multiple `Output` types for some definite type? I guess no. So it should be associated type for `FinishInto`, not a generic parameter?

core/src/hash/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+                $( #[$doc] )*
+                pub fn $name(input: &[u8]) -> [u8; $output_len] {
+                    let mut output = [0x00_u8; $output_len];
+                    self::[<$name _into>](input, &mut output);
+                    output
+                }
+            }
+        )*
+    };
+}
```
NikVolf:
```suggestion
}

```

core/src/hash/hasher.rs
Robbepop: I can and it is: E.g. for `[u8; 32]`, `[u8; 16]` and `[u8; 8]` for the TwoX hasher.

core/src/hash/mod.rs
Robbepop: We adhere to `rustfmt` in this crate.

core/src/hash/hasher.rs
Robbepop: I can understand your argument here and I am undecided about this. So far I always documented minor things like this also to teach new Rust people along reading source code. Less questions, more progress for new Rustaceans here and there.

core/src/hash/hasher.rs
Robbepop: I deliberately made this more explicit coming from a heavy macros background but in this situation you are right and we should apply your suggestions for the direct `Vec<u8>` implementations.

core/src/hash/hasher.rs
```diff
@@ -0,0 +1,360 @@
...
+impl<A> FinishInto<[u8; 16]> for TwoxHasher<A>
+where
+    A: Accumulator,
+{
+    fn finish_into(&self, output: &mut [u8; 16]) {
+        super::twox_128_raw_into(self.accumulator.as_slice(), output)
+    }
+}
+
+impl<A> FinishInto<[u8; 8]> for TwoxHasher<A>
```
NikVolf: Oh, I see here now that there are multiple `FinishInto<T>` implementations for one type.

core/src/hash/hasher.rs
```diff
@@ -0,0 +1,360 @@
...
+    /// Returns the `u64` result of the accumulated hash.
+    fn finish(&self) -> u64;
+}
+
+impl<A> FinishU64 for Sha2x256Hasher<A>
+where
+    A: Accumulator,
+{
+    fn finish(&self) -> u64 {
+        let [h0, h1, h2, h3, h4, h5, h6, h7, ..] =
```
NikVolf: this looks like it needs a helper function

with hashes, it is better to never to shuffle bytes individually, since errors are sometimes hard to catch - hash bytes are uniform (look random)

Robbepop: How to better extract the first 8 bytes out of a byte slice?
Note that
```
let first_8_bytes: [u8; 8] = <Self as Finish<[u8; 32]>>::finish(self)[0..8];
```
Doesn't work since Rust cannot infer `[u8; 8]` type here, just a `[u8]` so you'd need to unsafely cast the type or use some crate that does the job for you but then again you don't want unnecessary dependencies.

NikVolf:
This need test vectors from public sources also, otherwise looks good

Also, if you look at substrate, concrete hasher is designed have concrete output, so consider doing the same. But I am not sure about pros and cons of the change.

core/src/hash/hasher.rs
NikVolf: Yeah, it's a known Rust problem. However, you can just add this:

u64_from_slice_first_8_bytes(bytes: &[u8]) -> u64 {
   let mut first8bytes = [0u8; 8];
   firstBytes[..].copy_from_slice(&bytes[..8];
   u64::from_le_bytes(firstBytes)
}

and use it 4 times here 🤷‍♀️

Robbepop: Ah that's what you meant. xD Yeah of course.

Robbepop: done

core/src/hash/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+    (
+        $(
+            $( #[$doc:meta] )*
+            fn $name:ident($output_len:literal);
+        )*
+    ) => {
+        $(
+            paste::item! {
+                $( #[$doc] )*
+                pub fn $name(input: &[u8]) -> [u8; $output_len] {
```
athei: IMO this is redundant. u8 slices implement `Hash` which means there is no reason to use this function the non-raw function. It would be nice to cut down the large number of functions.

Robbepop: Yes I thought so as well but I was not sure exactly _HOW_ `u8` slices are hashes. E.g. are they hashed using their `len` + `hash(u8)` for each `u8` they contain just like a generic `[T]` slice or are `[u8]` actually specialized in NOT doing so. My fear is that there is no guarantee given by Rust how `[u8]` is really hashed and therefore we cannot get rid of the `_raw` APIs.

Robbepop:
> My only grumble is with the naming. I feel that the word `Into` is loaded with special meaning for rust developers. And that meaning is usually "take something and turn it **into** something else" which does not fit the notion of "encode **into** some buffer". I would have expected the `_into` functions to return something which hampered my understanding of this code.
> 
> I would suggest renaming all occurrences of `Into` to something else. Maybe `Buffer` or `MutBuffer` or `WithBuffer`.
> 
> Apart from that I very much like the code.

I actually took over the naming from Substrate. :D 
Normally I am using the word `using` for situations like these:
`hash_using(buffer)` etc.

core/src/hash/mod.rs
athei: Wouldn't you just use the `_raw_into` function it that rather low level use case? Does it really justify a new function? I am just genuinely asking. I feel like the number if overwhelming.

athei:
> I actually took over the naming from Substrate. :D
> Normally I am using the word using for situations like these:
> hash_using(buffer) etc.

You prefer sticking with it for consistency reasons?

Robbepop:
> > I actually took over the naming from Substrate. :D
> > Normally I am using the word using for situations like these:
> > hash_using(buffer) etc.
> 
> You prefer sticking with it for consistency reasons?

I don't care too much with consistency with other codebases. We should be more consistent within our own codebase. So I might change it to `using` or similar. Thanks for the pointers.

Robbepop:
> @kirushik should we omit the insecure hash functions? They seem to be security footguns to me.

What insecure hash functions do you mean? I think we should support all of the hash functions supported already by this PR since users might want to combine them invarious ways, e.g.:
```
Twox128(module_prefix) ++ Twox128(storage_prefix) ++ Blake2x128(encode(key))
```
Where `++` means append.

athei:
> @kirushik should we omit the insecure hash functions? They seem to be security footguns to me.

Good point. Do we even need them? Hash maps would be the prime example for non-cryptographic hash functions in a trusted environment which does not apply here.

At least we should make a clear distinction and not just list them right next to cryptographic hash functions.

DemiMarie-parity:
> > @kirushik should we omit the insecure hash functions? They seem to be security footguns to me.
> 
> What insecure hash functions do you mean?

xxHash

> I think we should support all of the hash functions supported already by this PR since users might want to combine them invarious ways, e.g.:
> 
> ```
> Twox128(module_prefix) ++ Twox128(storage_prefix) ++ Blake2x128(encode(key))
> ```
> 
> Where `++` means append.

That’s only safe if `module_prefix` and `storage_prefix` are trusted, which in practice means that they are compile-time constants, in which case the hashing isn’t needed to begin with.

DemiMarie-parity:
> > @kirushik should we omit the insecure hash functions? They seem to be security footguns to me.
> 
> Good point. Do we even need them (xxHash)? Hash maps would be the prime example for non-cryptographic hash functions in a trusted environment which does not apply here.

Agreed.  Smart contracts are not a trusted environment, and hash maps are not a good idea in smart contracts.  We could provide a smart contract from the host environment, but it would not be able to safely support iteration.

> At least we should make a clear distinction and not just list them right next to cryptographic hash functions.

:+1:

ascjones:
> That’s only safe if module_prefix and storage_prefix are trusted, which in practice means that they are compile-time constants, in which case the hashing isn’t needed to begin with.

This is from the substrate storage map implementation, where indeed those are compile-time constants.

https://github.com/paritytech/substrate/blob/master/frame/support/src/storage/generator/map.rs#L50

Robbepop:
> > > @kirushik should we omit the insecure hash functions? They seem to be security footguns to me.
> > 
> > 
> > Good point. Do we even need them (xxHash)? Hash maps would be the prime example for non-cryptographic hash functions in a trusted environment which does not apply here.
> 
> Agreed. Smart contracts are not a trusted environment, and hash maps are not a good idea in smart contracts. We could provide a smart contract from the host environment, but it would not be able to safely support iteration.
> 
> > At least we should make a clear distinction and not just list them right next to cryptographic hash functions.
> 
> +1

We are going to remove all instances of xx hash completely from high-level and low-level APIs for contracts.

DemiMarie-parity:
> > That’s only safe if module_prefix and storage_prefix are trusted, which in practice means that they are compile-time constants, in which case the hashing isn’t needed to begin with.
> 
> This is from the substrate storage map implementation, where indeed those are compile-time constants.
> 
> https://github.com/paritytech/substrate/blob/master/frame/support/src/storage/generator/map.rs#L50

Then why is the hash not computed at compile-time?

Robbepop:
> > > That’s only safe if module_prefix and storage_prefix are trusted, which in practice means that they are compile-time constants, in which case the hashing isn’t needed to begin with.
> > 
> > 
> > This is from the substrate storage map implementation, where indeed those are compile-time constants.
> > https://github.com/paritytech/substrate/blob/master/frame/support/src/storage/generator/map.rs#L50
> 
> Then why is the hash not computed at compile-time?

I think this is going to happen soonish(TM).

DemiMarie-parity:
> > > > That’s only safe if module_prefix and storage_prefix are trusted, which in practice means that they are compile-time constants, in which case the hashing isn’t needed to begin with.
> > > 
> > > 
> > > This is from the substrate storage map implementation, where indeed those are compile-time constants.
> > > https://github.com/paritytech/substrate/blob/master/frame/support/src/storage/generator/map.rs#L50
> > 
> > 
> > Then why is the hash not computed at compile-time?
> 
> I think this is going to happen soonish(TM).

In which case switching to a cryptographic hash function is free.

core/src/hash/hasher.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// This is a byte slice with varying lengths, e.g. `[u8; 32]`, [`u8; 16]`, etc.
+    type Output: Default;
+
+    /// Finalizes the hash using the underlying procedure.
+    fn finalize_immediate(input: &[u8], output: &mut Self::Output);
+}
+
+macro_rules! impl_hasher_for {
+    (
+        $( #[$doc:meta] )*
```
DemiMarie-parity: This throws away the documentation comments, which is bad.

Robbepop: good catch!

DemiMarie-parity: Thank you!

PR #378, opened 2020-04-16
[lang] remove EnvAccess inherent emit_event to avoid override

### LINKED ISSUE #377: Events seem to not be serialized with respect to their discriminant.
There was a bug report stating that the ERC20 `Transfer` event is serialized incorrectly at contract instantiation. It apparently is missing its discriminant before encoding its concrete parameters.

ink! normally packs all distinctly defined event types into a single overarching event enum thus providing discriminants upon serialization automatically.

Suspicion: Event serialization happens without packing the event type into the event enum.

```
0x
00 //None
01 //Some
d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d //Alice
0080c6a47e8d03000000000000000000 //1000 unit
```

ascjones:
Fixes #377 

Renamed `EnvAccess::emit_events` inherent impl to `emit_events_inner` (please suggest a better name).

Previously it wasn't calling the `EmitEvent::emit_event` implementation because it was defaulting to the inherent definition of the same name.

### DIFF
--- a/lang/macro/src/codegen/events.rs
+++ b/lang/macro/src/codegen/events.rs
@@ -119,7 +119,7 @@ impl EventHelpers<'_> {
                     where
                         E: Into<Self::Event>,
                     {
-                        ink_lang::EnvAccess::<EnvTypes>::emit_event(self, event.into())
+                        ink_lang::EnvAccess::<EnvTypes>::emit_event_inner(self, event.into())
                     }
                 }
             };
--- a/lang/macro/src/codegen/events.rs
+++ b/lang/macro/src/codegen/events.rs
@@ -119,7 +119,7 @@ impl EventHelpers<'_> {
                     where
                         E: Into<Self::Event>,
                     {
-                        ink_lang::EnvAccess::<EnvTypes>::emit_event_generic(self, event.into())
+                        ink_lang::EnvAccess::<EnvTypes>::emit_event_generic::<Self::Event>(self, event.into())
                     }
                 }
             };
--- a/lang/macro/src/codegen/events.rs
+++ b/lang/macro/src/codegen/events.rs
@@ -135,7 +135,7 @@ impl EventHelpers<'_> {
             .collect::<Vec<_>>();
 
         quote! {
-            #[derive(scale::Encode)]
+            #[derive(scale::Encode, scale::Decode)]
             pub enum Event {
                 #( #event_idents(#event_idents), )*
             }
@@ -182,7 +182,7 @@ impl EventHelpers<'_> {
 ///
 /// - making all fields `pub`
 /// - strip `#[ink(..)]` attributes
-/// - add `#[derive(scale::Encode)]`
+/// - add `#[derive(scale::Encode, scale::Decode)]`
 ///
 /// # Note
 ///
@@ -223,7 +223,7 @@ impl EventStructs<'_> {
             quote_spanned!(span =>
                 #conflic_depedency_cfg
                 #(#attrs)*
-                #[derive(scale::Encode)]
+                #[derive(scale::Encode, scale::Decode)]
                 pub struct #ident
                     #fields
             )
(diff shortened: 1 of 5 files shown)


lang/macro/src/codegen/events.rs
```diff
@@ -135,7 +135,7 @@ impl EventHelpers<'_> {
             .collect::<Vec<_>>();
 
         quote! {
-            #[derive(scale::Encode)]
+            #[derive(scale::Encode, scale::Decode)]
```
Robbepop, line 138: Why do we need `Decode` here?

lang/macro/src/codegen/storage.rs
```diff
@@ -54,13 +54,6 @@ impl GenerateCode for Storage<'_> {
         let message_impls = self.generate_message_impls();
         let storage_struct = self.generate_storage_struct();
 
-        let use_emit_event = if !self.contract.events.is_empty() {
-            // Required to allow for `self.env().emit_event(..)` in messages and constructors.
-            quote! { use __ink_private::EmitEvent as _; }
-        } else {
-            quote! {}
-        };
-
```
Robbepop, line 63: How can we do without this?

lang/macro/src/codegen/events.rs
```diff
@@ -119,7 +119,7 @@ impl EventHelpers<'_> {
                     where
                         E: Into<Self::Event>,
                     {
-                        ink_lang::EnvAccess::<EnvTypes>::emit_event(self, event.into())
+                        ink_lang::EnvAccess::<EnvTypes>::emit_event_inner(self, event.into())
```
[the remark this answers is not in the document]
Robbepop: I think it could already help if we instead do:
```rust
ink_lang::EnvAccess::<EnvTypes>::emit_event::<Self::Event>(self, event.into())
```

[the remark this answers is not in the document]
ascjones: The problem is that calling `self.env().emit_event()` from the contract was calling the wrong `emit_event`. It was calling the inherent one rather than the `impl EmitEvent`

lang/macro/src/codegen/events.rs
ascjones: For the test to decode the Event. I could add it under `test-env` but it should get pruned anyway by the build if not used.

lang/macro/src/codegen/storage.rs
ascjones: Define an empty `Event` enum when no events are defined, rather than no enum at all?

lang/src/env_access.rs
```diff
@@ -190,7 +190,7 @@ where
     /// # Note
     ///
     /// For more details visit: [`ink_core::env::emit_event`]
-    pub fn emit_event<Event>(self, event: Event)
+    pub fn emit_event_inner<Event>(self, event: Event)
```
Robbepop, line 193: Can we also rename this to `emit_generic_event`?

lang/macro/src/codegen/events.rs
```diff
@@ -119,7 +119,7 @@ impl EventHelpers<'_> {
                     where
                         E: Into<Self::Event>,
                     {
-                        ink_lang::EnvAccess::<EnvTypes>::emit_event_inner(self, event.into())
+                        ink_lang::EnvAccess::<EnvTypes>::emit_event_generic(self, event.into())
```
Robbepop, line 122:
```suggestion
                        ink_lang::EnvAccess::<EnvTypes>::emit_generic_event(self, event.into())
```

lang/src/env_access.rs
```diff
@@ -190,7 +190,7 @@ where
     /// # Note
     ///
     /// For more details visit: [`ink_core::env::emit_event`]
-    pub fn emit_event<Event>(self, event: Event)
+    pub fn emit_event_generic<Event>(self, event: Event)
```
Robbepop, line 193:
```suggestion
    pub fn emit_generic_event<Event>(self, event: Event)
```

lang/macro/src/codegen/storage.rs
ascjones: Heh sorry I realise I misunderstood your question originally. The answer is I don't know - it just worked. The trait must be in scope somehow. Anyway I have restored this.

PR #381, opened 2020-04-23
Add scale::Output to Wrap (no_std) + minor improvements

Robbepop:
- Add `scale::Output` impl to `Wrap` in `no_std`
- Minor improvements

### DIFF
--- a/core/src/hash/accumulator.rs
+++ b/core/src/hash/accumulator.rs
@@ -79,6 +79,23 @@ pub struct Wrap<'a> {
     len: usize,
 }
 
+impl Wrap<'_> {
+    /// Returns the capacity of the underlying buffer.
+    fn capacity(&self) -> usize {
+        self.buffer.len()
+    }
+
+    /// Returns the length of the underlying buffer.
+    fn len(&self) -> usize {
+        self.len
+    }
+
+    /// Returns `true` if the underlying buffer is empty.
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
+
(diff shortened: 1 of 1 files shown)


core/src/hash/accumulator.rs
```diff
@@ -103,6 +120,7 @@ impl<'a> Accumulator for Wrap<'a> {
     }
 
     fn write(&mut self, bytes: &[u8]) {
+        assert!(self.len() + bytes.len() <= self.capacity());
```
athei, line 123: Isn't that redundant? Your slicing operation will panic anyways when this happens.

core/src/hash/accumulator.rs
```diff
@@ -113,3 +131,16 @@ impl<'a> Accumulator for Wrap<'a> {
...
 }
+
+#[cfg(not(feature = "std"))]
+impl<'a> scale::Output for Wrap<'a> {
+    fn write(&mut self, bytes: &[u8]) {
+        <Self as Accumulator>::write(self, bytes)
+    }
+
+    fn push_byte(&mut self, byte: u8) {
+        assert!(self.len() < self.capacity());
```
athei, line 142: As above. I think this is just redundant noise (IMHO).

core/src/hash/accumulator.rs
Robbepop: true

core/src/hash/accumulator.rs
Robbepop: true

PR #382, opened 2020-04-24
Fix single threaded tests in off-chain environment

Robbepop:
Currently some tests that rely on interaction with the off-chain test environment share some state when run in a single thread. This PR resolves these conflicts by conditionally resetting the off-chain test environment for every test incarnation if the off-chain test environment has already been initialized at that time.

Tests are normally run in parallel each in their own thread, however, some tools such as `miri` run all tests single threaded.

Tests that have been run using:
`cargo test -- --test-threads=1`

### DIFF
--- a/core/src/env/engine/off_chain/db/accounts.rs
+++ b/core/src/env/engine/off_chain/db/accounts.rs
@@ -78,6 +78,11 @@ impl AccountsDb {
         }
     }
 
+    /// Resets the account DB to uninitialized state.
+    pub fn reset(&mut self) {
+        self.accounts.clear()
+    }
+
     /// Returns the account at the given account ID or creates it.
     pub fn get_or_create_account<T>(&mut self, at: &T::AccountId) -> &mut Account
     where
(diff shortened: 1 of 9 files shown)


Robbepop:
> From your description it is not obvious to me _why_ they share state in a single threaded environment. I guess because you rely on `thread_local!` to separate the state.
> 
> Maybe you can clarify that before merging.
> 
> When you expect every test to have a fresh environment why do not code that straightforward by creating a new instance in every test? I think even when running the tests per default it is not guaranteed that no thread is recycled. At least it is a assumption on something that is too much an implementation detail for my taste.

The problem is that the environment in ink! is a static entity. This is okay because ink! is normally run single threaded (as Wasm contract), however, in the off-chain environment is a bit more complicated.
On the one hand side we want to mimic semantics of the on-chain environment (being static, thus we opted into being `thread_local`) and on the other hand side it might be run in parallel.
Tests in ink! normally use the `run_test` API which now simply performs a reset of the underlying `thread_local` off-chain environment if it has been initialized already. So this is rather straight forward. Since each test is either run single threaded and in sequence of paralell and in their own thread this always works.

athei:
> This shouldn't be a problem since then the `thread_local` off-chain environment would simply reset.

I know this is why I gave my approval. I was just wondering why it was done in this more complicated way:  Static instance that is rested instead of just spawning a new one on the stack for every test.

If you say there are good reasons for that it is fine by me.

PR #388, opened 2020-05-04
file updates for license recognition

TovarishFin:
## what does this do?
* moves `LICENSE_TEMPLATE` to `FILE_TEMPLATE`
* adds a new line to `LICENSE`

## why?
In preparation for the new contributor license agreement bot, repos need to have licenses which are recognized by GitHub. 

Having multiple files starting with "*license*" (`LICENSE_TEMPLATE` & `LICENSE`) seems to confuse license recognition. Renaming `LICENSE_TEMPLATE` to `FILE_TEMPLATE` should resolve this problem.

Adding a new line to `LICENSE`, nudges the license detector the check the `LICENSE` file again for a license type. It doesn't seem to recheck this on its own...


TovarishFin:
> can you please rename FILE_TEMPLATE to TEMPLATE_LICENSE instead? describes better what it actually is.

Could we call it `FILE_HEADER` instead? I would rather avoid license in the name altogether in order to be sure that there is no confusion by the license detector.

`file_header.txt` (instead of `license_header.txt`) has been the replacement for other repos with a similar problem:
https://github.com/paritytech/polkadot/blob/master/file_header.txt

The name should appropriately name what it is in this case as well. Right?

Robbepop:
> 
> 
> > can you please rename FILE_TEMPLATE to TEMPLATE_LICENSE instead? describes better what it actually is.
> 
> Could we call it `FILE_HEADER` instead? I would rather avoid license in the name altogether in order to be sure that there is no confusion by the license detector.
> 
> `file_header.txt` (instead of `license_header.txt`) has been the replacement for other repos with a similar problem:
> https://github.com/paritytech/polkadot/blob/master/file_header.txt
> 
> The name should appropriately name what it is in this case as well. Right?

I am okay with `FILE_HEADER`, too.

PR #392, opened 2020-05-12
Make CI test ink! without optimizations

Robbepop:
Closes https://github.com/paritytech/devops/issues/545.


Robbepop:
> LGTM
> Fun fact: without `--release` you can use the same cache to run `cargo build ...` and use those artifacts for, say, e2e tests.

I am wiling to accept code suggestions of yours that implement this.

PR #399, opened 2020-05-19
Follow-up: Implement storage2 trait derive proc. macros

Robbepop:
Follow up for #311 .
Adds derive proc. macros for the `SpreadLayout` and `PackedLayout` traits of the new `ink_core::storage2` module.

Must not be merged before #311 is merged.

This PR also improves the structure of the `ink_core/derive` crate.
Moves `AllocateUsing` and `Flush` derives into their own files.
Move to `synstructure` tests for the new trait derive proc. macros instead of the current test framework.

### DIFF
--- a/core/derive/src/spread_layout.rs
+++ b/core/derive/src/spread_layout.rs
@@ -55,8 +55,7 @@ fn footprint(s: &synstructure::Structure) -> TokenStream2 {
                 })
         })
         .collect::<Vec<_>>();
-    let footprint = max_n(&variant_footprints[..]);
-    footprint
+    max_n(&variant_footprints[..])
 }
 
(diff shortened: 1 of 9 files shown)


core/derive/src/spread_layout.rs
```diff
@@ -0,0 +1,211 @@
...
+/// Generates the tokens to compute the maximum of the numbers given via
+/// their token streams at compilation time.
+///
+/// # Note
+///
+/// Since Rust currently does not allow conditionals in const contexts
+/// we use the array indexing trick to compute the maximum element:
+///
+/// ```no_compile
+/// max(a, b) = [a, b][(a < b) as usize]
```
ascjones, line 27: TIL

PR #413, opened 2020-05-22
Eliminate some unsafe in storage2 module

Robbepop:
This PR introduces a new thin wrapper around `UnsafeCell` which allows us to safely drop some `unsafe` blocks here and there for the lazy storage abstractions. Besides that the new `CacheCell` abstractions has very similar characteristics compared to the `UnsafeCell` but its API is simply optimized for our use case.

### DIFF
--- a/core/src/storage2/lazy/cache_cell.rs
+++ b/core/src/storage2/lazy/cache_cell.rs
@@ -1,0 +1,103 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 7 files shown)


core/src/storage2/lazy/cache_cell.rs
```diff
@@ -0,0 +1,103 @@
...
+        //         requires exclusive reference access to the wrapping `self`
+        //         disallowing aliasing exclusive references.
+        unsafe { &mut *self.inner.get() }
+    }
+
+    /// Returns a mutable pointer to the inner value.
+    ///
+    /// # Safety
+    ///
+    /// This is unsafe since it allows to mutably borrow the inner value through
```
DemiMarie-parity, line 95: This method is actually safe.  It is using the returned raw pointer that is unsafe.

PR #417, opened 2020-05-23
Add PushOnDrop storage abstraction

Robbepop:
Experimental implementation for the `PushOnDrop` storage abstraction.

It was planned that this will help with automatically pushing dynamic storage allocator state to the contract storage upon end of execution as well as automatically pushing the static smart contract state (`#[ink(storage)]`) to the contract storage upon end of execution.
While the latter probably works the former doesn't since the dynamic storage allocator is a static and static items never call their `Drop` implementation: https://doc.rust-lang.org/reference/items/static-items.html

> Static items do not call drop at the end of the program.

### DIFF
--- a/core/src/storage2/mod.rs
+++ b/core/src/storage2/mod.rs
@@ -19,6 +19,7 @@ pub mod collections;
 pub mod lazy;
 mod memory;
 mod pack;
+mod push_on_drop;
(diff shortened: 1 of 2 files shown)


Robbepop:
Closed since not really needed anymore.

PR #418, opened 2020-05-23
Add method to finalize the dynamic storage allocator

Robbepop:
So far it has not been possible to write back the accumulated state of the global dynamic storage allocator instance. With this PR we not only eliminate some code duplication but also provide this additional API in order to provide users with a way to command finalization of the global allocator for storage.

### DIFF
--- a/core/src/storage2/alloc/init.rs
+++ b/core/src/storage2/alloc/init.rs
@@ -182,11 +182,19 @@ cfg_if! {
 
         /// Forwards to the `initialize` of the global dynamic storage allocator instance.
         pub fn initialize(phase: ContractPhase) {
+            // SAFETY: Accessing the global allocator in Wasm mode is single
+            //         threaded and will not return back a reference to its
+            //         internal state. Also the `initialize` method won't
+            //         re-enter the dynamic storage in any possible way.
             unsafe { &mut GLOBAL_INSTANCE }.initialize(phase);
         }
 
         /// Forwards to the `finalize` of the global dynamic storage allocator instance.
         pub fn finalize() {
+            // SAFETY: Accessing the global allocator in Wasm mode is single
+            //         threaded and will not return back a reference to its
+            //         internal state. Also the `finalize` method won't
+            //         re-enter the dynamic storage in any possible way.
             unsafe { &mut GLOBAL_INSTANCE }.finalize();
         }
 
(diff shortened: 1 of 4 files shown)


core/src/storage2/alloc/init.rs
```diff
@@ -66,47 +75,127 @@ impl From<ContractPhase> for DynamicAllocatorState {
...
-        pub fn initialize_for(phase: ContractPhase) {
-            let instance = unsafe { &mut GLOBAL_INSTANCE };
-            // We do not allow reinitialization for Wasm targets for performance reasons.
-            if let DynamicAllocatorState::Initialized(_) = instance {
-                panic!("cannot reinitialize dynamic storage allocator instance in Wasm");
-            }
-            *instance = phase.into();
+        /// Forwards to the `initialize` of the global dynamic storage allocator instance.
+        pub fn initialize(phase: ContractPhase) {
+            unsafe { &mut GLOBAL_INSTANCE }.initialize(phase);
```
athei, line 185: Prove or remove ;) Would be nice to have some reasoning why this is safe. The single threaded aspect might be obvious but I am not so sure about reentrancy on the top of my head.

core/src/storage2/alloc/mod.rs
```diff
@@ -75,27 +75,51 @@ mod init;
...
+///   in contract instantiations and calls.
+/// - This has to be issued before the first interaction with the global allocator.
+/// - The actual instantiation will happen only upon the first interaction with
+///   the global allocator, e.g. using its `alloc` or `free` calls. Until then
+///   it remains uninitialized.
+///
+/// If this function is not called before the first global allocator interaction
+/// then the default initialization scheme is for contract instantiation.
+/// However, this behavior might change and must not be relied upon.
+pub fn initialize(phase: ContractPhase) {
```
athei, line 112: Who is responsible for calling this? Code generated by the `#[contract]` macro? Would be nice to have that explained in the comment.

core/src/storage2/alloc/mod.rs
```diff
@@ -75,27 +75,51 @@ mod init;
...
+}
+
+/// Finalizes the global dynamic storage allocator instance.
+///
+/// This pushes all the accumulated state from this contract execution back to
+/// the contract storage to be used in the next contract execution for the same
+/// contract instance.
+///
+/// The global dynamic storage allocator must not be used after this!
+pub fn finalize() {
```
athei, line 123: See above.

core/src/storage2/alloc/init.rs
Robbepop: Good catch! Will add a comment there.

core/src/storage2/alloc/mod.rs
Robbepop: Mostly auto generate code will call `initialize` and `finalize` users that use ink! (the eDSL) will not have to ever call this.

core/src/storage2/alloc/mod.rs
Robbepop: see above :P

PR #432, opened 2020-05-27
Companion #422: New storage layout trait (ink! metadata)

Robbepop:
Adds a new `StorageLayout` trait to `ink_core::storage2` that is going to be implemented by many storage entities and storage data structures in order for the ink! metadata to tell how the static contract storage layout looks like.
This is useful for 3rd party tools to access the contract storage in order to make queries and extract state information.

This also adds a new `layout2` module to the `ink_abi` crate that fixes many of the problems with the previous data structures provided for the same underlying metadata problem.

Note that the new `StorageLayout` does not take a `&self` reference compared to the old trait that did. We simply do not need an instance to tell about the static fields. Instead we now employ a `KeyPtr` parameter that guides the associated storage regions as we did with the other traits such as `SpreadLayout` and `PackedLayout`.


core/src/storage2/traits/layout/impls.rs
```diff
@@ -0,0 +1,242 @@
...
+    V: Metadata + 'static,
+{
+    fn layout(key_ptr: &mut KeyPtr) -> Layout {
+        Layout::Cell(CellLayout::new::<Self>(LayoutKey::from(
+            key_ptr.advance_by(1),
+        )))
+    }
+}
+
+// impl<T> StorageLayout for BinaryHeap<T>
```
Robbepop, line 208: These are commented out because they ideally should exist but are currently somehow not supported by the `type_metadata` crate which they should.

Robbepop:
@ascjones already started reviewing https://github.com/paritytech/ink/pull/422 so I will simply close this since https://github.com/paritytech/ink/pull/422 superseeds this PR.

PR #440, opened 2020-06-08
Implement storage2::Stash::remove_occupied unsafe function

### LINKED ISSUE #411: Add storage2::Stash::remove method
The `storage2::Stash` data structure is currently missing a method to remove elements without forcing a lookup into the storage. For example removing an element of a `storage::Stash` instance via its `take` method will potentially cause a contract storage read for the same element because `take` returns the old element if any.

This overhead cannot be avoided even if a caller is not interested into the old returned value. For the purpose of effectively avoiding the storage read we need to add a new API to the `storage2::Stash`:
```rust
fn remove(&mut self, index: Index) -> Option<()> { ... }
```
This API acts very similar to the `take` API and even still returns an `Option`. However, it guarantees to make no contract storage reads to the indexed element and will only write to its internal low-level lazy cache that the element at the given index is going to be removed at the end of the contract execution.
Calling `storage2::Stash::remove` with an index out of bounds for the `storage2::Stash` should return `None` and does not remove the element otherwise it returns `Some`.

## ToDo List

- [ ] Implement `storage2::Stash::remove`.
- [ ] Write unit tests to proof correctness of the new `storage2::Stash::remove`.
- [ ] **Bonus:** Write performance tests to show the difference between `storage2::Stash::take` and `storage2::Stash::remove`.

cmichi:
Closes #411.

The benchmark:
```
     Running /home/michi/projects/ink/target/release/deps/bench-fb9c90643691b4d3
Gnuplot not found, using plotters backend
RemoveMustOutperformTake/Remove/0                                                                             
                        time:   [496.21 ns 522.22 ns 551.32 ns]
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild
RemoveMustOutperformTake/Take/0                                                                             
                        time:   [611.62 ns 657.69 ns 705.12 ns]
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
```

Removing the elements via `take()` took about 100 ns longer than via `remove()`.


Robbepop:
The benchmarks look good but you could provide another benchmark that compares the worst case for take where it has to load the returned value from the contract storage.

core/benches/bench.rs
```diff
@@ -0,0 +1,72 @@
...
+
+fn take_from_filled_worst_case(test_values: &[u8; 6]) {
+    let mut stash = test_values.iter().copied().collect::<StorageStash<_>>();
+
+    for (index, _value) in test_values.iter().enumerate() {
+        let val = stash.take(index as u32);
+        let v = val.expect("must exist");
+    }
+    assert_eq!(stash.len(), 0);
+}
```
Robbepop, line 50: I think there was a misunderstanding here. Let me explain.
In order to enforce that the `storage2::Stash` actually loads the `taken` value from the contract storage we have to instantiate an instance of such a `test_values` storage Stash and then `push_spread` it onto the contract storage using some known `Key`.
When performing the benchmarks we then have to lazily instantiate such a storage Stash using the `pull_spread` method using the same key. It will just load lazily and won't pull anything from the storage, yet. So using `take` from this instance will actually initiate loading from the emulated contract storage. This way we can compare `remove` and `take` operations with the worst-case access for the `take` operation which should yield a significantly higher performance difference since `remove` guarantees to never load from contract storage.

core/src/storage2/collections/stash/mod.rs
```diff
@@ -466,6 +466,26 @@ where
...
+    ///
+    /// Calling this method with an index out of bounds for the returns `None` and
+    /// does not `remove` the element, otherwise it returns `Some(())`.
+    pub fn remove(&mut self, at: Index) -> Option<()> {
+        if at >= self.len_entries() {
+            // Early return since `at` index is out of bounds.
+            return None
+        }
+        self.entries.put(at, None);
+        self.header.len -= 1;
```
Robbepop, line 485: Hmmm this is actually problematic since we don't really know if decreasing this counter is correct. It might be incorrect if we removed an already removed (vacant) entry for example. So this is actually pretty bad and I haven't thought about this before. It might even be that we cannot implement this functionality for `storage::Stash` because of this. Also if we remove an occupied entry we have to relink the vacant entries since they form something like a doubly-linked list of vacant entries in order to quickly find a position for insertion upon `put`. So yeah, this is a bit more involved than I initially thought it'd be.

core/src/storage2/collections/stash/mod.rs
```diff
@@ -466,6 +466,26 @@ where
...
+    ///
+    /// This method acts similar to the take API and even still returns an Option.
+    /// However, it guarantees to make no contract storage reads to the indexed
+    /// element and will only write to its internal low-level lazy cache that the
+    /// element at the given index is going to be removed at the end of the contract
+    /// execution.
+    ///
+    /// Calling this method with an index out of bounds for the returns `None` and
+    /// does not `remove` the element, otherwise it returns `Some(())`.
+    pub fn remove(&mut self, at: Index) -> Option<()> {
```
Robbepop, line 479: After some thinking I decided that we must make this method `unsafe` as well as rename it to `remove_occupied`. The method must trust the caller that it is about to remove an occupied entry. This way we can make it as efficient as it guarantees to be. However, the current implementation forgot to relink the vacant entries that form the doubly linked list. This can be done similarly as it is done in the `take` method. We might want to pull out the relinking code into another method to avoid code duplication.

DemiMarie-parity: That is fine so long as we provide safe wrappers.

Robbepop: > That is fine so long as we provide safe wrappers.

Why would we additionally need safe wrappers? We have safe alternative accessors but also provide this `unsafe` marked accessor in case the caller can be certain about some required invariants. Also this might be useful for other ink! provided data structures to provide more efficient access to themselves. For example ink!'s `storage2::HashMap` could potentially profit here.

core/benches/bench.rs
cmichi: Thanks, I added it. Things is because of the issue with the linked list from the comment below the difference currently can't be measured.

core/src/storage2/collections/stash/tests.rs
```diff
@@ -106,6 +106,35 @@ fn take_out_of_bounds_works() {
...
+
+#[test]
+fn remove_out_of_bounds_works() {
+    let mut stash = [b'A', b'B', b'C']
+        .iter()
+        .copied()
+        .collect::<StorageStash<_>>();
+    assert_eq!(stash.remove(3), None);
+}
+
```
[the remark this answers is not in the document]
cmichi: Good idea, I added this test.

cmichi:
Current benches are below. Not a huge diff, but some.
```
Compare: `remove_occupied_all` and `take_all` (populated cache)/remove_occupied_all                        
                        time:   [484.80 ns 534.43 ns 593.66 ns]
                        change: [-25.859% -17.368% -7.6058%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  5 (5.00%) high mild
  10 (10.00%) high severe
Compare: `remove_occupied_all` and `take_all` (populated cache)/take_all                                                                             
                        time:   [530.10 ns 577.03 ns 632.72 ns]
                        change: [-6.6070% +5.9117% +20.450%] (p = 0.39 > 0.05)
Found 11 outliers among 100 measurements (11.00%)
  11 (11.00%) high mild

Compare: `remove_occupied_all` and `take_all` (empty cache)/remove_occupied_all                                                                             
                        time:   [8.0454 us 8.9226 us 9.9526 us]
                        change: [+3.7486% +17.577% +32.294%] (p = 0.01 < 0.05)
                        Performance has regressed.
Found 13 outliers among 100 measurements (13.00%)
  2 (2.00%) high mild
  11 (11.00%) high severe
Compare: `remove_occupied_all` and `take_all` (empty cache)/take_all                                                                             
                        time:   [10.624 us 11.765 us 13.002 us]
                        change: [-0.8077% +11.064% +24.657%] (p = 0.08 > 0.05)
Found 14 outliers among 100 measurements (14.00%)
  6 (6.00%) high mild
  8 (8.00%) high severe
```

core/src/storage2/collections/stash/mod.rs
```diff
@@ -344,49 +344,9 @@ where
...
-        // - There are no vacant entries before.
-        if at >= self.len_entries() {
-            // Early return since `at` index is out of bounds.
-            return None
-        }
-        // Precompute prev and next vacant entries as we might need them later.
-        // Due to borrow checker constraints we cannot have this at a later stage.
-        let (prev, next) = if let Some(index) = self.last_vacant_index() {
+    /// Returns the previous and next vacant entry.
+    fn fetch_prev_and_next_vacant_entry(&mut self, at: Index) -> (Index, Index) {
```
Robbepop, line 348: Doesn't need to be `&mut self`, `&self` should be enough.

core/src/storage2/collections/stash/mod.rs
```diff
@@ -344,49 +344,9 @@ where
...
-        // - There are vacant entries already.
-        // - There are no vacant entries before.
-        if at >= self.len_entries() {
-            // Early return since `at` index is out of bounds.
-            return None
-        }
-        // Precompute prev and next vacant entries as we might need them later.
-        // Due to borrow checker constraints we cannot have this at a later stage.
-        let (prev, next) = if let Some(index) = self.last_vacant_index() {
+    /// Returns the previous and next vacant entry.
```
Robbepop, line 347: Needs more documentation around the `at` parameter. Why is it returned as `(at, at)` if `self.last_vacant_index` is `None` etc.

core/src/storage2/collections/stash/mod.rs
```diff
@@ -463,6 +477,41 @@ where
...
+    /// element at the given index is going to be removed at the end of the contract
+    /// execution.
+    ///
+    /// Calling this method with an index out of bounds for the returns `None` and
+    /// does not `remove` the element, otherwise it returns `Some(())`.
+    ///
+    /// # Safety
+    ///
+    /// It must be ensured that `at` is an occupied index, otherwise this function
+    /// will panic.
```
Robbepop, line 494:
```suggestion
    /// The caller must ensure that `at` refers to an occupied index. Behavior is unspecified if `at` refers to a vacant index and could seriously damage the contract storage integrity.
```

Robbepop: maybe the above comment suggestion should be spread over multiple lines.

core/src/storage2/collections/stash/mod.rs
```diff
@@ -463,6 +477,41 @@ where
...
+    pub unsafe fn remove_occupied(&mut self, at: Index) -> Option<()> {
+        // This function is written similar to `take()`, with the exception
```
Robbepop, line 496:
```suggestion
        // This function is written similar to [`Stash::take`], with the exception
```

core/src/storage2/collections/stash/mod.rs
```diff
@@ -463,6 +477,41 @@ where
...
+        // that we take it for granted that `at` is occupied and hence
```
Robbepop, line 497:
```suggestion
        // that the caller has to ensure that `at` refers to an occupied entry whereby the procedure can avoid loading the occupied entry which might be handy if the stored `T` is especially costly to load from contract storage.
```

core/src/storage2/collections/stash/mod.rs
```diff
@@ -463,6 +477,41 @@ where
...
+        // can remove some logic under this assumption.
```
Robbepop, line 498:
```suggestion
```

core/src/storage2/collections/stash/mod.rs
cmichi: Nice catch, thx!

Issue #442, opened 2020-06-09
Do not eagerly use dynamic storage allocator


Robbepop:
## Current State

In the current development and research branch that can be found [here](https://github.com/paritytech/ink/tree/robin-research) the dynamic storage allocator is in place and can be used by ink! smart contracts.

This is done by automatically initializing and finalizing the dynamic storage allocator in the ink! macros. So the user has to do nothing in order to simply use it, e.g. by creating and using some [`storage::Box`](https://github.com/paritytech/ink/blob/robin-research/core/src/storage2/alloc/boxed/mod.rs#L49) instances.

The code that automatically initializes and finalizes the dynamic storage allocator can be found in several places of [this file](), for example

- [initialization](https://github.com/paritytech/ink/blob/robin-research/lang/src/dispatcher.rs#L75): via
    - `alloc::initialize(ContractPhase::Call);` or
    - `alloc::initialize(ContractPhase::Deploy);`
- [finalization](https://github.com/paritytech/ink/blob/robin-research/lang/src/dispatcher.rs#L79): via `alloc::finalize();`

The problem with the current approach is that making these calls to initialize or finalize the dynamic storage allocator let's the Rust compiler compile all of the dynamic storage allocator into the final resulting Wasm binary which is increased by approx. 5kB just by this. For example if we comment these calls out of the `ink_lang` crate smart contracts that never make use of the dynamic storage allocator drop in Wasm file size significantly from 7kB to 2kB.

We urgently need a solution to produce better code that makes it possible for the Rust compiler to better detect if the dynamic storage allocator is unused by the smart contract and being able to completely throw it out of the final Wasm binary.

## Solutions

There are currently mainly two competing potential solutions.

### Manual Intervention

We could simply implement another contract annotation that can be optionally provided at the header of an ink! smart contract similar to the already existing mandatory `version` attribute in
```rust
#[ink::contract(version = "0.1.0")]
mod foo { ... }
```
With this new optional annotation a user could manually enable or disable the dynamic storage allocator. For example, if we name the annotation `use-dynamic-storage` the contract writer could then specify the ink! header in the following way in order to completely disable automatic initialization and finalization of the dynamic storage allocator:
```rust
#[ink::contract(
    version = "0.1.0",
    dynamic-storage = false,
)]
mod foo { ... }
```
[...]

PR #449, opened 2020-06-13
Avoid heap allocations for contract dispatch

Robbepop:
This PR implements the contract message and constructor dispatch without the need for heap allocations.

Closes: https://github.com/paritytech/ink/issues/442


ascjones:
Conflicts with master need resolving

Robbepop:
> Conflicts with master need resolving

Will resolve today! :+1:

lang/macro/src/codegen/dispatch.rs
```diff
@@ -70,6 +77,238 @@ impl GenerateCode for Dispatch<'_> {
...
+        }
+    }
+
+    fn generate_message_dispatch_enum(&self) -> TokenStream2 {
+        let storage_ident = &self.contract.storage.ident;
+        let message_variants = self
+            .contract
+            .functions
+            .iter()
+            .filter(|function| function.is_message())
```
ascjones, line 143: This code block filtering message functions is repeated three times here, could be done once.

lang/macro/src/codegen/dispatch.rs
```diff
@@ -70,6 +77,238 @@ impl GenerateCode for Dispatch<'_> {
...
+            .filter(|function| function.is_constructor())
+            .map(|message| self.generate_dispatch_variant_arm(message, "Constructor"));
+        let decode_message = self
+            .contract
+            .functions
+            .iter()
+            .filter(|function| function.is_constructor())
+            .map(|message| self.generate_dispatch_variant_decode(message, "Constructor"));
+        let execute_variants = self
+                .contract
```
ascjones, line 248: indentation

lang/macro/src/codegen/dispatch.rs
```diff
@@ -70,6 +77,238 @@ impl GenerateCode for Dispatch<'_> {
...
+        }
+    }
+
+    fn generate_constructor_dispatch_enum(&self) -> TokenStream2 {
+        let storage_ident = &self.contract.storage.ident;
+        let message_variants = self
+            .contract
+            .functions
+            .iter()
+            .filter(|function| function.is_constructor())
```
ascjones, line 239: Same as above, could remove repetition here of this block

lang/src/dispatcher.rs
```diff
@@ -15,229 +15,111 @@
...
-            ManuallyDrop::new(pull_spread_root::<<M as FnState>::State>(&root_key));
-        let result = <M as MessageRef>::CALLABLE(&state, args);
-        alloc::finalize();
-        if TypeId::of::<<M as FnOutput>::Output>() != TypeId::of::<()>() {
-            ink_core::env::output::<<M as FnOutput>::Output>(&result)
-        }
-        Ok(())
+    alloc::initialize(ContractPhase::Call);
+    let root_key = Key::from([0x00; 32]);
+    let state = ManuallyDrop::new(pull_spread_root::<<M as FnState>::State>(&root_key));
```
ascjones, line 75: Can you explain the use of `ManuallyDrop` here?

lang/macro/src/codegen/dispatch.rs
Robbepop: strange that this is formatted with `cargo fmt` :thinking:

lang/macro/src/codegen/dispatch.rs
Robbepop: I will see what I can do about it.

lang/src/dispatcher.rs
Robbepop: When `Drop::drop` is called on a storage entity the storage entity tries to clean-up its state on the storage. By packing it right into a `ManuallyDrop` we prevent it from doing this. Note that we generally want this behavior of immediate clean-up upon `Drop::drop` call, however in this case we are looking at a special case that we explicitely want to avoid this for our static storage fields.

Issue #441, opened 2020-06-08
Design Proposal: Trait Support


Robbepop:
# Design Proposal: ink! Trait Support

It has been a long time since trait support for ink! has been proposed the first time.
Now we finally got ink! into the shape where we can think about a concrete design for this feature.

Trait support contains both:
- Being able to define traits and implement traits for the contract's `#[ink(storage)]` annotated struct.
- Being able to call trait implementations cross-contract.

For this ink! does not only have to identify trait implementations but also generate dispatch and cross-calling code for them. In order for this to work we have to provide ink! with `#[ink(..)]` hints as always.

## Example: Flipper

The syntactic proposal looks like the following applied to the well known Flipper smart contract:

```rust
use ink_lang as ink;

#[ink::contract(version = "0.1.0")]
mod flipper {
    #[ink(storage)]
    struct Flipper {
        value: bool,
    }

    #[ink(export)]
    impl Default for Flipper {
        #[ink(constructor)]
        fn default() -> Self {
            Self::new(Default::default())
        }
    }

    impl Flipper {
        #[ink(constructor)]
        fn new(init_value: bool) -> Self {
            Self { value: init_value }
        }

        #[ink(message)]
        fn flip(&mut self) {
            self.value = !self.value;
        }

        #[ink(message)]
        fn get(&self) -> bool {
            self.value
        }
    }
}
```

In the above example we implement the Rust prelude's `Default` trait for the Flipper's storage struct. We tell ink! that this is a trait implementation for which it shall generate cross-calling and dispatch code via the `#[ink(export)]` annotation.
We also provide a `#[ink(constructor)]` on the `default` trait method implementation in order to make ink! check the function as applicable for ink! constructors.

## Mandatory Annoations

For the first iteration we want to make this additional `#[ink(constructor)]` annotation mandatory and might think about relaxing these strict annotation rules for later revisions of ink! making them optional since ink! could very well infer them since in the case of trait implementations either all or none of the provided methods must be exported to the contracts API - this is the main difference to normal `impl` blocks where the ink! annotations for constructors and messages are mandatory and required.

## Disambiguation

There is a problem with equally named traits having equally named methods where ink! would fail to generate proper disambiguated code in certain use cases. To properly disambiguate we present different strategies out of which one will be implemented.

Technically speaking: In order to disambiguate we need to define different unique identifiers for all exported messages and constructors from all trait implementations and inherent implementations.

### Strategy 1: Salt

[...]

PR #451, opened 2020-06-14
Implement trait support - Part 1/3 - New ink! IR

Robbepop:
Implements & closes https://github.com/paritytech/ink/issues/441.

## tl;dr

The PR packs all of the new definitions into a new ink! crate, called `ink_lang_ir`.
This new crate provides abstractions to parse, analyze and inspect ink! smart contract code on an IR level.
Other utilities, such as the planned new ink! codegen can then depend on `ink_lang_ir` and use its API to do their jobs.

A rough outline of the API surface is shown in the screenshot below:

2020-06-28-202734_1069x1163_scrot

## ToDo List

- [x] Write unit tests for conflicting ink! attributes for:
    - [x] `Event`
    - [x] `Storage`
    - [x] `Message`
    - [x] `Constructor`
- [x] Implement interface for composed selectors.
    - [x] Avoid accessing user provided selector. (This is just confusing anyways.)
    - [x] Integrate composed selector access with iteration over impl block messages and constructors.
    - [ ] Write tests for composed selector access through this iteration interface.
- [x] Rename `salt` to `namespace`.
- [x] Implement check to ensure no selectors overlap
    - [x] Implement check
    - [x] Implement tests
    - [x] Add additional test for selectors of different trait impls with same name methods.
- [x] Make all in-doc code examples compile.
- [x] Implement selector computation that treats trait implementation blocks different from inherent implementation blocks.
    - [x] Implement unit tests.
- [x] Implement event topics.
    - [x] Write failure unit tests for event topics.
    - [x] Write unit tests for event fields iterator.
    - [x] Register which fields are topic fields.
    - [x] Provide iterator that yields `(&Field, bool)` where if the `bool` is `true` the `&Field` is a topic field.
- [x] Implement `TryFrom<syn::Module> for ir2::Module`.
    - [x] Implement success unit tests.
    - [x] Implement failure unit tests.
        - [x] Missing storage struct
        - [x] Too many storage structs
        - [x] Missing constructor
        - [x] Missing message
        - [x] Invalid out-of-line module
        - [x] Conflicting ink! attributes (any)
    - [x] Implement check for missing and multiple `#[ink(storage)]` structs.
    - [x] Implement check for missing `#[ink(constructor)]`.
    - [x] Implement check for missing `#[ink(message)]`.
- [x] Implement `Constructor` and `Message` checks from within `ir2::ItemImpl`.
    - [x] Check that `Message` of inherent `ItemImpl` has public visibility.
    - [x] Check that `Message` of trait `ItemImpl` has inherited visibility. (Maybe not needed.)

In order to sanely implement traits we have to change the ink!'s IR to no longer be message & constructor focused but implementation block centric. That way we can also generate code that is more similar to the original user input. Also this allows to have more fine grained control over the output of ink! as a user.

By rewriting the ink! IR we are trying to find a design and implementation that suits other issues in the issue tracker, e.g. being able to apply multiple ink! markers to entities, e.g.:
```rust
#[ink(message)]
#[ink(payable)]
fn my_message(&self) { ... }

// or shorthand
#[ink(message, payable)]
fn my_message(&self) { ... }
```

## Main Challenge & Problem

The main challenge behind ink!'s trait support is to somehow enforce on a syntactical level (we are still just a macro) that either none (normal trait implementation) or all methods lead to some properly specified ink! trait implementation.

E.g. when looking at a trait with default members or with super traits it is easy that we could implement a trait like this:
```rust
pub trait MyTrait {
    // This needs to be implemented!
    fn required_method(&self);

    // This can be omitted in an implementation.
    fn forward(&self) {
        self.required_method(&self);
    }
}

impl MyTrait for MyContract {
    #[ink(message)]
    fn required_method(&self) { ... }
}
```
While the above example is safe and ink! would be able to provide proper accessors to it so that users cannot misuse it or do cross-calls that end in nowhere the following example is more involved for ink! (or even impossible to safeguard).

```rust
// Might be defined in another crate so ink! doesn't necessarily have access to its contents.
pub trait SuperTrait {
    fn super_foo(&self);
}

// Might be defined in another crate so ink! doesn't necessarily have access to its contents.
pub trait BaseTrait: SuperTrait {
    fn foo(&sefl);
    fn bar(&self) {
        self.super_foo()
    }
}

impl BaseTrait for MyContract {
    #[ink(message)]
    fn foo(&self) { ... }
}
```
In this example, since ink! only ever has access to the syntactical portions of all items the ink! macro cannot know the internal structure of the `SuperTrait` or even the `BaseTrait` and only sees the required `BaseTrait::foo` method for which it creates the correct code. However, for `BaseTrait::bar` it will generate incorrect code that calls `SuperTrait::super_foo` without it being a proper cross-contract call on the chain. This is a huge problem and I currently do not see any solution to this. Since we cannot spot this we cannot even generate a warning in its place. The only potential solution is to require traits to be defined in the same ink! module where they have been implemented which kind of destroys the gains of having them in the first place.

We still think that the above mentioned problem can be fixed if we don't implement non-ink! marked traits for the cross-calling wrappers of a contract. We believe that this way we can again guarantee that all trait implementations will result in a proper cross-contract call.


lang/ir/src/ast/config.rs
```diff
@@ -0,0 +1,298 @@
...
+impl ToTokens for MetaNameValue {
+    fn to_tokens(&self, tokens: &mut TokenStream2) {
+        self.name.to_tokens(tokens);
+        self.eq_token.to_tokens(tokens);
+        self.value.to_tokens(tokens);
+    }
+}
+
+impl MetaNameValue {
+    /// Like Path::parse_mod_style but accepts keywords in the path.
```
cmichi, line 99:
```suggestion
    /// Like `Path::parse_mod_style` but accepts keywords in the path.
```

lang/ir/src/ast/config.rs
```diff
@@ -0,0 +1,298 @@
...
+        ParseStream,
+    },
+    punctuated::Punctuated,
+    Token,
+};
+
+/// The attribute arguments for the configuration of an ink! smart contract.
+///
+/// These are the segments `version = "0.1.0"` and `env_types = DefaultEnvTypes`
+/// in `#[ink::contract(version = "0.1.0", env_types = DefaultEnvTypes)]`.
```
cmichi, line 33: `version` attribute is obsolete.

lang/ir/src/ir/item_impl/mod.rs
```diff
@@ -0,0 +1,355 @@
...
+    impl_token: syn::token::Impl,
+    generics: syn::Generics,
+    trait_: Option<(Option<syn::token::Bang>, syn::Path, syn::token::For)>,
+    self_ty: Box<syn::Type>,
+    brace_token: syn::token::Brace,
+    items: Vec<ImplItem>,
+    /// A namespace to disambiguate trait implementation blocks with equal
+    /// names. Generally can be used to change computation of message and
+    /// constructor selectors of the implementation block.
+    namespace: Option<ir::Namespace>,
```
cmichi, line 81: Awesome, thanks!

lang/ir/src/ir/item_impl/tests.rs
```diff
@@ -0,0 +1,256 @@
...
+        "ink! message in trait impl blocks must have inherited visibility",
+    );
+    assert_try_from_item_impl_fails(
+        syn::parse_quote! {
+            impl MyTrait for MyStorage {
+                #[ink(constructor)]
+                pub fn my_public_constructor() -> Self {}
+            }
+        },
+        "ink! constructor in trait impl blocks must have inherited visibility",
```
cmichi, line 206: Could you explain what the issue with the code above is? In principle constructors should be public, but in this case it should be private since the property is somehow inherited?

lang/ir/src/ast/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+
+//! Types and abstractions for ink! definitions that require custom syntax.
+//!
+//! # Note
+//!
+//! In general we try not to require any sort of custom non-standard Rust
+//! syntax.
+//!
+//! At the time of this writing we currently only use this for the argument
+//! parsing of ink! config header `#[ink(version = "0.1.0", etc...)]` in order
```
cmichi, line 23: `version` is obsolete.

lang/ir/src/ir/contract.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    /// Returns the identifier of the ink! inline module definition.
+    ///
+    /// # Note
+    ///
+    /// The ink! inline module definition is the module that comprises the
+    /// whole ink! smart contract, e.g.:
+    ///
+    /// ```no_compile
+    /// #[ink::contract(version = "0.1.0")]
```
cmichi, line 84: `version` is obsolete.

lang/ir/src/ir/contract.rs
```diff
@@ -0,0 +1,116 @@
...
+        &self.item
+    }
+
+    /// Returns the configuration of the ink! smart contract.
+    ///
+    /// # Note
+    ///
+    /// The configuration is given via the `#[ink::contract(config))]` attribute
+    /// macro annotation itself within the `(config)` part. The only mandatory
+    /// field there is `version`. Other fields include but are not limited to:
```
cmichi, line 101: Hmm, is `version` really obsolete? This comment states it's mandatory?

lang/ir/src/ir/attrs.rs
```diff
@@ -0,0 +1,979 @@
...
+    /// `#[ink(impl)]`
+    ///
+    /// This attribute supports a niche case that is rarely needed.
+    ///
+    /// Can be applied on ink! implementation blocks in order to make ink! aware
+    /// of them. This is useful if such an implementation block doesn't contain
+    /// any other ink! attributes, so it would be flagged by ink! as a Rust item.
+    /// Adding `#[ink(impl)]` on such implementation blocks makes them treated
+    /// as ink! implementation blocks thus allowing to access the environment
+    /// etc. Note that ink! messages and constructors still need to be explicitely
```
ascjones, line 310:
```suggestion
    /// etc. Note that ink! messages and constructors still need to be explicitly
```

lang/ir/src/ir/attrs.rs
```diff
@@ -0,0 +1,979 @@
...
+                Attribute::Other(other_attr) => Either::Right(other_attr),
+            }
+        });
+    Attribute::ensure_no_duplicate_attrs(&ink_attrs)?;
+    Ok((ink_attrs, others))
+}
+
+/// Sanitizes the given attributes.
+///
+/// This paritions the attributes into ink! and non-ink! attributes.
```
ascjones, line 417:
```suggestion
/// This partitions the attributes into ink! and non-ink! attributes.
```

lang/ir/src/ir/attrs.rs
```diff
@@ -0,0 +1,979 @@
...
+                    "encountered duplicate ink! attribute arguments"
+                )
+                .into_combine(format_err!(
+                    seen.span(),
+                    "first equal ink! attribute argument here"
+                )))
+            }
+            seen.insert(arg);
+        }
+        Ok(())
```
ascjones, line 555: Duplicate code with `ensure_no_duplicate_args` (and maybe `ensure_no_duplicate_attrs`)

lang/ir/src/ir/config.rs
```diff
@@ -0,0 +1,275 @@
...
+};
+use core::convert::TryFrom;
+use syn::spanned::Spanned;
+
+/// The ink! configuration.
+#[derive(Debug, Default, PartialEq, Eq)]
+pub struct Config {
+    /// If `true` enables the dynamic storage allocator
+    /// facilities and code generation of the ink! smart
+    /// contract. Does incure some overhead. The default is
```
ascjones, line 27:
```suggestion
    /// contract. Does incur some overhead. The default is
```

lang/ir/src/ir/item_impl/tests.rs
Robbepop: This is because for trait implementation blocks you cannot specify a visibility. This rule is coming from Rust.
E.g.
```rust
impl MyTrait for MyStorage {
    fn my_trait_method() {} // <-- implicitely `pub`, so cannot manually set it to `pub`
}
```

lang/ir/src/ir/item_impl/callable.rs
```diff
@@ -0,0 +1,557 @@
...
+/// - Avoid directly setting the selector of an ink! message or constuctor.
+///   Only do this if nothing else helps and you need a very specific selector,
+///   e.g. in case of backwards compatibility.
+/// - Do not use the namespace unless required to disambiguate.
+pub fn compose_selector<C>(item_impl: &ir::ItemImpl, callable: &C) -> ir::Selector
+where
+    C: Callable,
+{
+    if let Some(selector) = callable.user_provided_selector() {
+        return *selector
```
ascjones, line 266: Could warn if providing both `selector` and `namespace`, that `namespace` will be ignored.

lang/ir/src/ir/contract.rs
Robbepop: The comment is as deprecated as `version`. ;) Thanks

lang/ir/src/ir/item_impl/callable.rs
Robbepop: Sadly as of today we cannot emit compile time warnings. :(
https://internals.rust-lang.org/t/pre-rfc-add-compile-warning-macro/9370

lang/ir/src/ir/attrs.rs
Robbepop: Thanks! Eliminated the duplicate. However, while similar, `ensure_no_duplicate_attrs` actually does some slightly different thing.

lang/ir/src/ast/config.rs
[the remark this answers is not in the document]
Robbepop: done

lang/ir/src/ir/item/event.rs
```diff
@@ -0,0 +1,386 @@
...
+            | syn::Visibility::Restricted(_)
+            | syn::Visibility::Crate(_) => {
+                return Err(format_err_spanned!(
+                    &item_struct.vis,
+                    "non `pub` ink! event structs are not supported",
+                ))
+            }
+            _ => (),
+        }
+        'outer: for field in item_struct.fields.iter() {
```
ascjones, line 106: TIL, I thought it was a lifetime. Though as I understand this is for nested loops, where is the `'inner`?

lang/ir/src/ir/item/storage.rs
```diff
@@ -0,0 +1,233 @@
...
+// limitations under the License.
+
+use crate::ir;
+use core::convert::TryFrom;
+use proc_macro2::Ident;
+use syn::spanned::Spanned as _;
+
+/// An ink! storage struct definition.
+///
+/// Noticed by ink! throught the `#[ink(storage)]` annotation.
```
ascjones, line 22:
```suggestion
/// Noticed by ink! through the `#[ink(storage)]` annotation.
```

lang/ir/src/ir/item/storage.rs
```diff
@@ -0,0 +1,233 @@
...
+///
+/// # Example
+///
+/// ```
+/// # use core::convert::TryFrom;
+/// # let event = <ink_lang_ir::Storage as TryFrom<syn::ItemStruct>>::try_from(syn::parse_quote! {
+/// #[ink(storage)]
+/// pub struct MyStorage {
+///     my_value: bool,
+//      counter: u32,
```
ascjones, line 37:
```suggestion
///      counter: u32,
```

lang/ir/src/ir/item/storage.rs
```diff
@@ -0,0 +1,233 @@
...
+/// # Note
+///
+/// An ink! smart contract must have exactly one storage definition.
+/// The storage definition must be found in the root of the ink! module.
+///
+/// # Example
+///
+/// ```
+/// # use core::convert::TryFrom;
+/// # let event = <ink_lang_ir::Storage as TryFrom<syn::ItemStruct>>::try_from(syn::parse_quote! {
```
ascjones, line 33:
```suggestion
/// # let storage = <ink_lang_ir::Storage as TryFrom<syn::ItemStruct>>::try_from(syn::parse_quote! {
```

lang/ir/src/ir/item_mod.rs
```diff
@@ -0,0 +1,616 @@
...
+                        pub fn my_constructor() -> Self {}
+
+                        #[ink(message, selector = "0xDEADBEEF")]
+                        pub fn my_message(&self) {}
+                    }
+                }
+            })
+            .is_ok()
+        )
+    }
```
ascjones, line 615: Would be good to have a test with overlapping selectors from different traits with the same method, not just using `selector`

lang/ir/src/ir/item_impl/mod.rs
```diff
@@ -0,0 +1,355 @@
...
+                return Ok(true)
+            }
+        }
+        // Check if any of the implementation block's methods either resembles
+        // an ink! constructor or an ink! message:
+        'outer: for item in &item_impl.items {
+            match item {
+                syn::ImplItem::Method(method_item) => {
+                    if !ir::contains_ink_attributes(&method_item.attrs) {
+                        continue 'outer
```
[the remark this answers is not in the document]
Robbepop: yes and also for robustness in case someone changes the code in the future with additional loops or whatever. I just find the explicitly labeled continue or break statements to be much more readable in general.

lang/ir/src/ir/attrs.rs
```diff
@@ -0,0 +1,979 @@
...
+            | Item::Struct(syn::ItemStruct { attrs, .. })
+            | Item::Trait(syn::ItemTrait { attrs, .. })
+            | Item::TraitAlias(syn::ItemTraitAlias { attrs, .. })
+            | Item::Type(syn::ItemType { attrs, .. })
+            | Item::Union(syn::ItemUnion { attrs, .. })
+            | Item::Use(syn::ItemUse { attrs, .. }) => attrs,
+            _ => &[],
+        }
+    }
+}
```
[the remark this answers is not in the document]
Robbepop: https://github.com/dtolnay/syn/issues/866

lang/ir/src/ir/contract.rs
```diff
@@ -0,0 +1,116 @@
...
+        let module = syn::parse2::<syn::ItemMod>(ink_module)?;
+        let ink_config = ir::Config::try_from(config)?;
+        let ink_module = ir::ItemMod::try_from(module)?;
+        Ok(Self {
+            item: ink_module,
+            config: ink_config,
+        })
+    }
+
+    /// Returns the identifier of the ink! inline module definition.
```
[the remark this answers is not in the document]
Robbepop: oh yeah it seems that the doc comment it outdated. good find!

lang/ir/src/ir/item/event.rs
Robbepop: Yeah I use it always when having to use `continue` or `break` (goto) statements because to me the code is more readable and more robust in case a person changes the code and might introduce new loops etc.

Robbepop: Maybe it is bad style so it is more of a personal thing. If there are good reasons to avoid those annotations in general comments are very welcome!

lang/ir/src/ir/item/storage.rs
Robbepop: 2020-07-02-123712_1242x251_scrot
your suggestion looks bugged to me :S

lang/ir/src/ir/item/storage.rs
Robbepop: let us simply remove the let binding altogether.

lang/ir/src/ir/item/storage.rs
ascjones: Weird - must be the dark theme :grin: 

image

lang/ir/src/ir/item_impl/callable.rs
```diff
@@ -0,0 +1,557 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.callable
+    }
+}
+
+/// An ink! callable.
+///
+/// This is either an ink! message or an ink! constructor.
+/// Used to share common behavior between different callable types.
+pub trait Callable {
```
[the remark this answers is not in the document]
Robbepop: I chose this design to avoid further nesting of items. Technically you are definitely right that we could merge those types. However, I think the current design through the trait definition is fine. Maybe we can explore this other design in some follow-up PR?

lang/ir/src/ir/item_impl/constructor.rs
```diff
@@ -0,0 +1,571 @@
...
+            method_item.span(),
+            method_item.attrs.clone(),
+            &ir::AttributeArgKind::Constructor,
+            |kind| {
+                match kind {
+                    ir::AttributeArgKind::Constructor
+                    | ir::AttributeArgKind::Payable
+                    | ir::AttributeArgKind::Namespace(_)
+                    | ir::AttributeArgKind::Selector(_) => false,
+                    _ => false,
```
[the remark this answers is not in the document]
Robbepop: seems like we are missing tests here :upside_down_face:

lang/ir/src/ir/item_impl/mod.rs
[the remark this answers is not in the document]
ascjones: Yeah it is useful actually to let your eye skip to the loop, just never seen it used before

lang/ir/src/ir/item/event.rs
ascjones: All good I just had never seen it before - as in the other comment I agree it is good for readability and robustness

lang/ir/src/ir/item_mod.rs
Robbepop: i agree

lang/ir/src/ir/item_impl/callable.rs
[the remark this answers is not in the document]
ascjones: Good enough for me - was just curious about the reasoning because I always go back and forth on this and usually *slightly* favour not using traits.

lang/ir/src/ir/item_mod.rs
Robbepop: created an action item in the above ToDo list

PR #452, opened 2020-06-15
Fix compilation of multisig example doctest

athei:
This fixes the compilation of the doctest and changes the attribute from `ignore` to `no_run` in order to have it compiled on test. However, **running** the doctest is not feasible as the off-chain environment does support it.

### DIFF
--- a/examples/multisig_plain/lib.rs
+++ b/examples/multisig_plain/lib.rs
@@ -301,7 +301,7 @@ mod multisig_plain {
         ///
         /// // first create the transaction that adds `alice` through `add_owner`
         /// let alice: AccountId = [1u8; 32].into();
-        /// let mut call = CallData::new(Selector::from_str("add_owner"));
+        /// let mut call = CallData::new(Selector::new([166, 229, 27, 154])); // add_owner
         /// call.push_arg(&alice);
         /// let transaction = Transaction {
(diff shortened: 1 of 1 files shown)


examples/multisig_plain/lib.rs
```diff
@@ -307,13 +308,19 @@ mod multisig_plain {
         /// };
         ///
         /// // submit the transaction for confirmation
-        /// let mut submit = CallParams::eval(wallet_id, Selector::from_str("submit_transaction"));
-        /// let (id, _)  = submit.push_arg(&transaction)
+        /// let mut submit = CallParams::<Env, _, _>::eval(
+        ///     wallet_id,
+        ///     Selector::from_str("submit_transaction")
```
Robbepop, line 317: Can you replace this call with `Selector::new([u8; 4])`? We are about to remove `Selector::from_str` since computing the selector is actually an ink! implementation detail exposed by the metadata. So what you could do instead is to compute it manually on your machine and simply insert the resulting bytes. The hash should be BLAKE-2 and the input to it are simply the string as bytes. Especially for the upcoming trait support we need to extend or change this system which is why we do not want to expose some stable interface for it (`Selector::from_str`).

Robbepop:
Thanks for the fix!

PR #456, opened 2020-06-18
[examples] remove config made redundant by cargo-contract

ascjones:
Remove stuff which is now generated by `cargo-contract`:
- abi/metadata packages (https://github.com/paritytech/cargo-contract/pull/58)
- `profile.release` sections (https://github.com/paritytech/cargo-contract/pull/55)

**Note:** `cargo-contract` will not be released to crates.io until `ink 3.0`, so this PR changes the CI to install from a github tag.

## todo

- [x] fix `dns`
- [x] fix `delegator`
- [ ] update `check-examples` to use `cargo-contract`


examples/dns/Cargo.toml
```diff
@@ -19,8 +19,6 @@ path = "lib.rs"
 crate-type = [
     # Used for normal contract Wasm blobs.
     "cdylib",
```
Robbepop, line 21: We can simpliy this to `crate-type = "cdylib"` now.

Robbepop: Or does this actually work if we still want to execute the contract crate via `cargo test`?

Robbepop:
> Nice! I really like this change.
> I'm already hating you @ascjones because it is a breaking change facepalm
> However it is okay to keep it for ink! 3.0.

It is really important that we get rid of all the cruft from ink! before the next big milestones.

examples/dns/Cargo.toml
ascjones: `cargo test` still works

ascjones: > We can simpliy this to `crate-type = "cdylib"` now.

Arpparently it needs to be an array `invalid type: string "cdylib", expected a sequence for key `lib.crate-type``

examples/delegator/lib.rs
```diff
@@ -14,7 +14,6 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-pub use self::delegator::Delegator;
```
Robbepop, line 17: I think it is fine that this PR deletes this line of code but in general we might need some strategy to expose public items from a smart contract. E.g. for `cargo doc`, otherwise it will display an empty smart contract API if there are no public items.

PR #460, opened 2020-06-22
Extern function to generate metadata

ascjones:
Currently the `main` function of the `abi_gen` package needs to know the name of the storage struct in order to invoke the metadata generation. This is fragile because it depends on the struct being publicly accessble at a known location. 

This PR exposes `__ink_generate_metadata()` with `#[no_mangle]` so it can be called via an `extern`.
This makes it independent of where the user defines their `#[ink::contract]` module.

Also it will enable the `cargo-contract` feature to autogenerate the `abi_gen` package: https://github.com/paritytech/cargo-contract/pull/58


examples/incrementer/.ink/abi_gen/main.rs
```diff
@@ -1,6 +1,18 @@
...
+
+extern "Rust" {
+    fn __ink_generate_metadata() -> ink_abi::InkProject;
+}
+
+extern crate contract;
+
+extern "Rust" {
+    fn __ink_generate_metadata() -> ink_abi::InkProject;
+}
```
Robbepop, line 11: duplicate code

PR #463, opened 2020-06-29
[core] update arithmetic traits, add saturating impls

ascjones:
Changes `SimpleArithmetic` to `AtLeast32BitsUnsigned` to match equivalent substrate traits. Also introduces `Saturating` and default impls. See: https://github.com/paritytech/substrate/blob/4be954a8463e6f33e0ec854c4b8bcd940af260fc/primitives/arithmetic/src/traits.rs#L23

Will be useful for #461 emulating gas price calculation in the offchain environment.


ascjones:
Indeed I should know about the "monomorphization" of env types - I did the original work on that. You're right that they are only useful for the generic offchain environment.

The original motivation was to allow gas price calculation emulation offchain - and I figured that the saturating impls were useful in general. But since they are not used in the contract itself after all, could just do something like: https://github.com/paritytech/ink/pull/461/commits/6a883531d6835be05031e742e00ef138af21853b

If that is acceptable I can either close this PR or just remove the saturating impls, at least it might be useful to align the names of the traits with substrate (i.e. `SimpleArithmetic` no longer exists).

PR #466, opened 2020-07-01
[core] Add storage2::Lazy::set(…)

### LINKED ISSUE #426: Add storage2::Lazy::set method
Currently we have no way to efficiently update the value of a `storage2::Lazy` entity.
Instead we have to mutably dereference the instance in the hopes that this won't trigger a storage read.
Adding the `Lazy::set` API is a guarantee that no storage read is ever going to occur due to update the internal value.
The `set` API must be an inherent method since `Lazy` implements `Deref` and `DerefMut` so that no methods of the inner `T` get shadowed.

## ToDo List

- [ ] Add `storage2::Lazy::set(this: &mut Self, new_value: T);` method
- [ ] Write unit tests for the new `set` API
- [ ] **Bonus:** Write benchmarks proving that the new `set` API is really more efficient than using the current `DerefMut` in some situations.

cmichi:
Closes #426.

Current benches (tl;dr is that `set` is faster than `deref_mut`):
```
Benchmarking Compare: `set` and `deref_mut` (populated cache)/set/0: Warming up for 3.0000 sclippy
Compare: `set` and `deref_mut` (populated cache)/set/0                                                                             
                        time:   [218.24 ns 227.33 ns 237.28 ns]                
                        change: [-8.2839% -2.5945% +3.5291%] (p = 0.38 > 0.05) 
Found 9 outliers among 100 measurements (9.00%)                                
  8 (8.00%) high mild                                                          
  1 (1.00%) high severe   
Compare: `set` and `deref_mut` (populated cache)/deref_mut/0                                                                             
                        time:   [273.25 ns 297.83 ns 325.22 ns]
                        change: [-1.6218% +7.9414% +18.099%] (p = 0.11 > 0.05)
Found 6 outliers among 100 measurements (6.00%)                                
  6 (6.00%) high mild                                                          
                                                                               
Compare: `set` and `deref_mut` (empty cache)/set/0                                                                             
                        time:   [205.83 ns 233.05 ns 268.66 ns]                
                        change: [-4.3310% +10.258% +29.037%] (p = 0.20 > 0.05)                                                                                 
Found 21 outliers among 100 measurements (21.00%)                                                                                                              
  1 (1.00%) high mild                                                                                                                                          
  20 (20.00%) high severe                                                                                                                                      
Compare: `set` and `deref_mut` (empty cache)/deref_mut/0                                                                            
                        time:   [355.29 ns 370.78 ns 390.67 ns]                                                                                                
                        change: [-31.978% -25.741% -19.115%] (p = 0.00 < 0.05) 
                        Performance has improved.            
Found 14 outliers among 100 measurements (14.00%)
  5 (5.00%) high mild                                                          
  9 (9.00%) high severe
```


core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -307,6 +308,33 @@ where
...
+    ///
+    /// If accessing the inner value fails.
+    pub fn set(&mut self, new_value: T) {
+        // SAFETY: This is critical because we mutably access the entry.
+        let cache = unsafe { &mut *self.cache.get_ptr().as_ptr() };
+        if cache.is_none() {
+            // Write value to storage
+            let key = self
+                .key
+                .expect("key must exist when attempting to set a value");
```
cmichi, line 328: I'm not entirely sure if this assumption always holds true. Probably requires further discussion.

core/benches/bench_cell.rs
```diff
@@ -0,0 +1,106 @@
...
+/// for only one op the criterion benchmark might just show 0 ps and thus
+/// make it hard to estimate the difference between `deref_mut` and `set`.
+mod populated_cache {
+    use super::*;
+
+    pub fn set() {
+        let mut cell = <LazyCell<i32>>::new(Some(1));
+        for i in 0..100 {
+            black_box(cell.set(i));
+        }
```
Robbepop, line 48: I am not sure the loops here hurt the actual measurement. Are they needed? If yes, why?

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -307,6 +308,33 @@ where
     pub fn get_mut(&mut self) -> Option<&mut T> {
         self.load_entry_mut().value_mut().into()
     }
+
+    /// Sets the value in this cell to `value`, without executing any reads.
+    ///
+    /// # Note
```
Robbepop, line 314: Should note that this method should be preferred over get_mut in case the returned value is of no interest to the caller.

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -307,6 +308,33 @@ where
...
+    ///
+    /// No loads will be executed.
```
Robbepop, line 316: "no reads from contract storage"

Robbepop, line 316: "no reads from contract storage"

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -307,6 +308,33 @@ where
...
+    /// If accessing the inner value fails.
+    pub fn set(&mut self, new_value: T) {
+        // SAFETY: This is critical because we mutably access the entry.
+        let cache = unsafe { &mut *self.cache.get_ptr().as_ptr() };
+        if cache.is_none() {
+            // Write value to storage
+            let key = self
+                .key
+                .expect("key must exist when attempting to set a value");
+            push_spread_root_opt::<T>(Some(&new_value), &key);
```
Robbepop, line 329: The `set` operation just as any other non-`PushSpread` operation should never directly write to the contract storage. This is subject to the setup and tear-down process of a contract. So what you want upon `set` if the cache is already populated is to simply overwrite its already existing value. If the cache is empty you simply set the cache to the value. This also means that the `key` does not need to exist for `set` to work since it only needs to write the value into its cache and is done.

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -416,4 +444,24 @@ mod tests {
...
+        run_test::<env::DefaultEnvTypes, _>(|_| {
+            let mut cell = <LazyCell<u8>>::lazy(Key::from([0x42; 32]));
+            let value = cell.get();
+            assert_eq!(value, None);
+
+            cell.set(13);
+            assert_eq!(cell.get(), Some(&13));
+            Ok(())
+        })
+    }
```
Robbepop, line 466: Maybe some other test would be nice that test this functionality over the following:
1. Initialize a LazyCell<i32> with None
2. Push it to K
3. Pull another instance V from K, check that it is None
4. Set V using `set` to an actual value
5. Push V it to K
6. Load V2 from K
7. Check that V2 is the set value

core/benches/bench_cell.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    pub fn deref_mut() {
+        let mut cell = <LazyCell<i32>>::new(Some(1));
+        for i in 0..100 {
+            let v = black_box(cell.get_mut().unwrap());
+            black_box(*v = i);
+        }
```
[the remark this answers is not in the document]
cmichi: Yeah, good point, I changed it to bench against `Lazy` instead.

core/benches/bench_cell.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+fn bench_set_populated_cache(c: &mut Criterion) {
+    let mut group = c.benchmark_group("Compare: `set` and `deref_mut` (populated cache)");
+    group.bench_function(BenchmarkId::new("set", 0), |b| {
+        b.iter(|| populated_cache::set())
+    });
+    group.bench_function(BenchmarkId::new("deref_mut", 0), |b| {
```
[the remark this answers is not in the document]
cmichi: IIUC it is sufficient to just change the benchmark of `deref_mut` to use `Lazy` instead of `LazyCell` (without changing anything about benching `set` with `LazyCell`)?

core/benches/bench_cell.rs
cmichi: The comment above the `mod` describes the reason:
```
/// For the populated cache benchmarks we use a loop of 100 ops, since
/// for only one op the criterion benchmark might just show 0 ps and thus
/// make it hard to estimate the difference between `deref_mut` and `set`.
```
I agree with you though, but if I remove the loop criterion always panics because of too few meaningful samples. I investigated this in more detail for a while yesterday and don't yet fully understand the issue. I'll look more into it next week, but from what I understand so far it could also be a bug/issue with criterion (or of course the benchmark is wrong).

core/benches/bench_cell.rs
```diff
@@ -0,0 +1,106 @@
...
+        b.iter(|| populated_cache::deref_mut())
+    });
+    group.finish();
+}
+
+/// Pushes a value to contract storage and creates a `LazyCell` pointing to it.
+fn push_storage_cell(value: i32) -> LazyCell<i32> {
+    let root_key = Key::from([0x00; 32]);
+    SpreadLayout::push_spread(&value, &mut KeyPtr::from(root_key));
+    <LazyCell<i32>>::lazy(root_key)
```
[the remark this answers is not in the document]
cmichi: Can you elaborate what you mean by a silent failure? Since both benchmark methods use `SpreadLayout::push_spread` the comparison should not be affected?

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -325,12 +324,13 @@ where
...
-            // Write value to storage
-            let key = self
-                .key
-                .expect("key must exist when attempting to set a value");
-            push_spread_root_opt::<T>(Some(&new_value), &key);
+            // Cache is empty, so we simply set the cache to the value.
+            // The key does not need to exist for this to work, we only need to
+            // write the value into the cache and are done. Writing to contract
+            // storage happens during setup/teardown of a contract.
+            *cache = Some(Entry::new(Some(new_value), EntryState::Preserved));
```
Robbepop, line 331:
```suggestion
            *cache = Some(Entry::new(Some(new_value), EntryState::Mutated));
```

Robbepop: The `EntryState` must be `Mutated`, otherwise the `Entry`'s value is not going to be written back on contract tear-down. `Preserved` means that the cache holds a value that mirror what can already be found in the contract storage.

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -307,6 +307,37 @@ where
...
+        } else {
+            //  Cache is already populated we simply overwrite its already existing value.
+            cache
+                .as_mut()
+                .expect("unpopulated cache entry")
+                .put(Some(new_value));
+        }
+        debug_assert!(cache.is_some());
```
Robbepop, line 339: That's a good idea! We should have more `debug_assert` statements in the whole codebase for such purposes.

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -307,6 +307,37 @@ where
...
+    /// This method should be preferred over dereferencing or `get_mut`
+    /// in case the returned value is of no interest to the caller.
+    ///
+    /// # Panics
+    ///
+    /// If accessing the inner value fails.
+    pub fn set(&mut self, new_value: T) {
+        // SAFETY: This is critical because we mutably access the entry.
+        let cache = unsafe { &mut *self.cache.get_ptr().as_ptr() };
+        if cache.is_none() {
```
Robbepop, line 326: I wonder if pattern matching `match cache` or an `if let Some(cache)` would be more practical at this point for the `Some` case.

core/benches/bench_cell.rs
```diff
@@ -0,0 +1,104 @@
...
+    black_box,
+    criterion_group,
+    criterion_main,
+    BenchmarkId,
+    Criterion,
+};
+use ink_core::{
+    env,
+    storage2::{
+        lazy::Lazy,
```
Robbepop, line 25: If we now benchmark `Lazy` instead of `LazyCell` we should rename the benchmark file.

core/benches/bench_cell.rs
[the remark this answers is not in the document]
Robbepop: We do not guarantee that pushing of type `T` to location `k` and then pulling another type `V` from the same `k` results in sane behavior. But this is what you currently do here where `T == i32` and `V == LazyCell<i32>`. It only succeeds because apparently both types currently do the same encoding but this is never guaranteed.

cmichi:
Updated benches below.

Strange thing is that runs of populated cache `deref_mut` are reproducible always faster than `set`. Can you make any sense of that?

```
Compare: `set` and `deref_mut` (populated cache)/set/0                                                                             
                        time:   [2.8211 ns 2.9129 ns 3.0109 ns]
                        change: [-9.2827% -5.5126% -1.2291%] (p = 0.01 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
Compare: `set` and `deref_mut` (populated cache)/deref_mut/0                                                                             
                        time:   [2.3792 ns 2.4548 ns 2.5361 ns]
                        change: [-9.3920% -5.1172% -0.7223%] (p = 0.03 < 0.05)
                        Change within noise threshold.
Found 9 outliers among 100 measurements (9.00%)
  9 (9.00%) high mild

Compare: `set` and `deref_mut` (empty cache)/set/0                                                                            
                        time:   [330.08 ns 367.04 ns 408.14 ns]
                        change: [-15.375% -5.8701% +5.1110%] (p = 0.27 > 0.05)
Found 18 outliers among 100 measurements (18.00%)
  4 (4.00%) high mild
  14 (14.00%) high severe
Compare: `set` and `deref_mut` (empty cache)/deref_mut/0                                                                             
                        time:   [524.14 ns 563.18 ns 605.31 ns]
                        change: [-18.508% -10.482% -2.0013%] (p = 0.02 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  6 (6.00%) high mild
  2 (2.00%) high severe
```

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -307,6 +307,34 @@ where
...
+    ///
+    /// No reads from contract storage will be executed.
+    ///
+    /// This method should be preferred over dereferencing or `get_mut`
+    /// in case the returned value is of no interest to the caller.
+    ///
+    /// # Panics
+    ///
+    /// If accessing the inner value fails.
+    pub fn set(&mut self, new_value: T) {
```
Robbepop, line 323: mark as `#[inline]`

core/src/storage2/lazy/lazy_cell.rs
```diff
@@ -307,6 +307,34 @@ where
...
+        // SAFETY: This is critical because we mutably access the entry.
+        let cache = unsafe { &mut *self.cache.get_ptr().as_ptr() };
+        if let Some(cache) = cache.as_mut() {
+            //  Cache is already populated we simply overwrite its already existing value.
+            cache.put(Some(new_value));
```
Robbepop, line 328: One difference that could have a performance impact is that we are using `Entry::put` here but `Entry::value_mut` + assign in `get_mut`. Maybe the latter is more efficient?

cmichi: Good catch, but I tried it out and it actually slows `set` down :/.

core/benches/bench_lazy.rs
```diff
@@ -0,0 +1,100 @@
...
+    use super::*;
+
+    pub fn set() {
+        let mut lazy = push_storage_lazy(1);
+        black_box(Lazy::set(&mut lazy, 13));
+    }
+
+    pub fn deref_mut() {
+        let mut lazy = push_storage_lazy(1);
+        black_box(*lazy = 13);
```
Robbepop, line 83:
```suggestion
        black_box(*lazy) = 13;
```

Robbepop: Please try this and see if it changes something in the benchmarks.

Robbepop: We want to keep the scope of our `black_box` as big as needed but as small as possible.

core/benches/bench_lazy.rs
```diff
@@ -0,0 +1,100 @@
...
+criterion_main!(populated_cache, empty_cache);
+
+mod populated_cache {
+    use super::*;
+    use core::ops::DerefMut;
+
+    pub fn set() {
+        let mut lazy = <Lazy<i32>>::new(1);
+        let lazy_mut = black_box(&mut lazy);
+        black_box(Lazy::set(lazy_mut, 17));
```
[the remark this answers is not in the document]
cmichi: This additional `black_box` was the change which fixed the benchmarks 0 ps :-).

core/benches/bench_lazy.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    pub fn deref_mut() {
+        let mut lazy = <Lazy<i32>>::new(1);
+        let i32_mut = black_box(lazy.deref_mut());
+        black_box(*i32_mut = 17);
```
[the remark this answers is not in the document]
cmichi: This additional `black_box` was the change which fixed the benchmarks 0 ps :-).

core/benches/bench_lazy.rs
cmichi:
```
error[E0070]: invalid left-hand side of assignment
  --> core/benches/bench_lazy.rs:93:26
   |
93 |         black_box(*lazy) = 13;
   |         ---------------- ^
   |         |
   |         cannot assign to this expression
```
I had tried to get variations of this to work as well. So seems like this is the smallest scope to which we can get.

Issue #299, opened 2019-12-12
Additional contract metadata


ascjones:
As discussed in https://github.com/paritytech/ink/pull/296#issuecomment-564531512.

Add extra data to at the `InkProject` struct level. e.g.

- `version` as suggested by @jacogr 
- `language` e.g. `ink!`
- Hashes over the code to properly infer if the supplied code matches the metadata
- the author of the contract to make it possible to contact maintainers
- the used Rust (or other compiler) version to have "more deterministic" way to rebuild the contract
- etc.

**Update:** see https://github.com/paritytech/ink/issues/299#issuecomment-651132343 below with outcome of meeting with @Robbepop @seanyoung

## todo:

- [x] `scale-info`
  - [x] remove string table (https://github.com/paritytech/scale-info/pull/17)
  - [x] release `v0.3.0`
- [x] `ink!`
  - [x] ~~Define types and add to `InkProject` struct~~, 
  - [x] ~~Accept instances of contract metadata types via `_ink_generate_metadata`~~
  - [x] Merge https://github.com/paritytech/ink/pull/467
- [x] `cargo-contract`
  - [x] Extract ink contract metadata as json via call to extern `_ink_generate_metadata`
  - [x] Combine with additional metadata extracted from cargo metadata
  - [x] Merge https://github.com/paritytech/cargo-contract/pull/62

Robbepop:
For `version` we might just use the one we already have in the ink! macro header and its value should be a string with the following formatting:
```
version: "0.1.0"
```

Also we might be interested in the underlying compiler, so we might want to include the Rust version into `version` somehow.
For `language` we might want to use `ink!` for an ink! project. Its value is simply a string:
```
language: "ink!"
```

Not sure how we are going to compute the hashes but we should standardize a specific hash for that first.
The `authors` field should be an array of `string` where `string` should have the following formatting (as in Cargo):
```
username <user@email.domain>
```
So we end up having something like:
```
authors:  [ foo <foo@bar.baz>, qed <qed.boom@meow.woof> ]
```

Also we maybe even want the date of creation of the metadata:
```
date: "2019-12-31"
```

Maybe even a description of the whole contract might be handy:
```
description: "This is my description of what this contract does and does not."
```

seanyoung:
How about including the compiler version in the string which identifies the compiler.

```
compiler: "solang 0.1.0 (llvm 8.0.1)"
```

which could also be:

```
compiler: "ink! 2.0 (rust 1.33.0)"
```

[...]

PR #467, opened 2020-07-02
[metadata] update to scale-info 0.3, removes string interning

ascjones:
Rel #299.

- Updates to latest `scale-info` 0.3 which removes the string interning, making all strings inline.
- ~~Adds struct name to the layout.~~ Removed because this is an `ink!` specific concept and not currently required by 3rd party tools

*Note:* Previously this PR contained additional metadata data structures, which have now been moved over to `cargo-contract`. See https://github.com/paritytech/cargo-contract/pull/62


metadata/Cargo.toml
```diff
@@ -21,9 +21,11 @@ ink_primitives = { version = "2.1.0", path = "../primitives/", default-features
 serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
 serde_json = "1.0"
 derive_more = { version = "0.99", default-features = false, features = ["from"] }
+proc-macro2 = "1.0"
 scale-info = { version = "0.3", default-features = false, features = ["derive"] }
 semver = { version = "0.10.0", features = ["serde"] }
 url = { version = "2.1.1", features = ["serde"] }
+quote = "1.0"
```
Robbepop: `proc-macro2` and `quote` (as well as `syn`) dependencies are normally reserved for proc macro crates and since they are not `no_std` compatible should be avoided for ink! core crates. Maybe there is still a good reason for them here but from what I see right now this should not be included.

ascjones: The idea is for `cargo-contract` to create an instance of `InkProjectExtension` [here](https://github.com/paritytech/cargo-contract/pull/62/files#diff-e4f600b805550e47676a5a537a3bf4b7R87) (which contains the additional metadata), and then use that instance to generate the code to place in `main.rs` which will construct an equivalent instance and pass it to `__ink_generate_metadata`.

Ways forward:

1. Continue with this approach but make sure `proc_macro2` and `quote` are for `std` feature only
2. This approach but just use our own trait in place of `ToTokens` and strings to generate code
3. Make all fields public and implement `ToTokens` on `cargo-contract` side
4. Scrap this approach entirely and load the contract as a DLL using https://docs.rs/libloading/0.6.2/li__ink_generate_metadata(extension)bloading/ or similar, and calling `__ink_generate_metadata(extension)`

ascjones: This has now all been removed. The additional metadata data structures have all been moved to `cargo-contract` https://github.com/paritytech/cargo-contract/pull/62

.gitlab-ci.yml
```diff
@@ -255,7 +255,7 @@ publish-docs:
     - git fetch origin gh-pages
     # Generating Docs
     - time cargo doc --no-deps --all-features
-        -p scale-info -p ink_metadata -p ink_metadata_derive -p ink_core -p ink_core_derive
+        -p scale-info -p ink_metadata -p ink_core -p ink_core_derive
```
ascjones, line 258: This should fix the master branch CI

metadata/Cargo.toml
```diff
@@ -19,12 +19,14 @@ ink_prelude = { version = "2.1.0", path = "../prelude/", default-features = fals
 ink_primitives = { version = "2.1.0", path = "../primitives/", default-features = false }
 
 serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
+serde_json = "1.0"
 derive_more = { version = "0.99", default-features = false, features = ["from"] }
-scale-info = { version = "0.2", default-features = false, features = ["derive"] }
+scale-info = { version = "0.3", default-features = false, features = ["derive"] }
+semver = { version = "0.10.0", features = ["serde"] }
+url = { version = "2.1.1", features = ["serde"] }
```
Robbepop, line 26: What are these (and `serde_json`) new dependencies good for?

metadata/src/lib.rs
```diff
@@ -49,25 +50,26 @@ use scale_info::{
...
-    /// Creates a new ink! project.
     pub fn new<L, S>(layout: L, spec: S) -> Self
     where
         L: Into<layout2::Layout>,
         S: Into<ContractSpec>,
     {
         let mut registry = Registry::new();
-        Self {
+
+        InkProject {
```
Robbepop, line 72: Why not `Self`? It is generally better not to repeat yourself if possible.

metadata/Cargo.toml
ascjones: Good for nothing, a hangover from when the extended types were defined here. Good catch!

Robbepop: `serde_json` should probably be a `dev-dependency` again.

ascjones: Yeah I've moved it back, my commits have not made it through to this PR yet...

Issue #469, opened 2020-07-04
Generalize EnvTypes to just Env, add MAX_EVENT_TOPICS constant


Robbepop:
In order to properly implement a compile-time check for ink! smart contracts to define not more event topics for any given user defined event than is allowed by the target smart contracts chain we need to include another constant value in our `EnvTypes` trait.
For that purpose we should also consider renaming `EnvTypes` to just `Env` or something similar since it will no longer just contain some type aliases.
With the so-called `MAX_EVENT_TOPICS` constant set in place the ink! codegen will be able to generate code in a way that checks that there are no user defined events with more topics than are allowed given this constant value.

## ToDo List

- [x] Add `MAX_EVENT_TOPICS` constant to `EnvTypes`
- [x] Rename `EnvTypes` to something that suits better, e.g. `Env` (not sure if `Env` is actually a good name here)
- [x] Implement ink! codegen to check if no user defined events has more topics than specified by `MAX_EVENT_TOPICS` constant.

Robbepop:
Implemented. Closed.

Issue #105, opened 2019-05-30
Make ink! 2.0 event topics work


Robbepop:
## ink! event topics

The current ink! 2.0 syntax has support for event topics in the following way:

```rust
#[ink(event)]
struct Transferred {
    #[ink(topic)]
    from: Option<AccountId>,
    #[ink(topic)]
    to: Option<AccountId>,
    #[ink(topic)]
    amount: Balance,
}
```

## Current state

However, the current implementation currently does not completely implement the topic semantics for the marked event parameters.
What is currently been done in ink!'s codegen is that the `ink_lang2::Topics` trait is being implemented for the given events, however the implementation currently always returns an empty array instead of the properly calculated event topics.

## Steps towards full implementation

To make `ink!` work with event topics we need to implement the following steps:

- [x] Handle input errors, like receiving `#[ink(topic)]` twice.
- [ ] Codegen shall implement the `ink_lang2::Topics` trait correctly for all given events. Those calculations must all be done at compile time as possible.

## Background

Topics in ink! or Substrate work pretty much the same as in Ethereum. For more information about details in how event topics are generated, refer to the event topics documentation of Solidity.

Robbepop:
This has been implemented in #510 . Closed.

PR #470, opened 2020-07-04
Implement trait support - Part 2/3 - New ink! codegen

Robbepop:
## Trait Support - New ink! Codegen

This PR implements the new ink! codegen to provide support for ink! trait implementations.

## Follow-up PRs

- [ ] Make use of the new `ink_lang_ir` and `ink_lang_codegen` crates from `ink_lang_macro`. (Part 3)
- [ ] Convert existing example smart contracts to ink! 3.0 syntax.
    - [ ] Flipper
    - [ ] Incrementer
    - [ ] ERC-20
    - [ ] ERC-721
    - [ ] Delegator
    - [ ] Multi-Signature
    - [ ] Domain Name Service (DNS)
- [ ] Write and convert all the acceptance tests for ink! 3.0.
- [ ] Make ink! event topics work, finally: https://github.com/paritytech/ink/issues/105
    - [ ] Find a way to generate event topic implementations in a way that they do not require heap allocations but instead make use of ink!'s static 16kB buffer. This might imply providing yet another high-level `ink_core` API specific to this purpose.
    - Use BLAKE2 cryptographic hash.
    - Avoid BLAKE2 hashing in case the SCALE encoded values fit into `Env::Hash`.
    - Encode `topics[0]` as the event's hashed signature if the ink! event is not anonymous.
    - Since Seal only supports up to `MAX_EVENT_TOPICS` but cannot distinguish between existing special-cased `topics[0]` or not we need to make sure that non-anonymous events can actually only have up to `MAX_EVENT_TOPICS - 1` topics.

## ToDo List

- [x] Think about `#[ink(anonymous)]` ink! event definitions.
    - An anonymous ink! event definition would not encode `topics[0]` as the hash of its signature.
    - Discussed with some Solidity experts: They concluded that anonymous events are not very commonly used. So instead of implementing this feature we will wait until people come up with serious feature requests for it.
- [x] Improve codegen if all ink! messages are not payable.
    - Instead of performing the check after the message dispatch we can do it once before.
- [x] Think about better default value for dynamic storage allocator config.
- [x] Rename `storage_alloc` ink! config to `dynamic_storage_allocator`.
- [x] Add check for every event to guard against maximum number of topics.
- [x] Adjust `EnvTypes` trait: https://github.com/paritytech/ink/issues/469
- [x] ink! constructors must not be `payable`. They implicitly are always `payable` due to endowment.
- [x] Implement codegen for non-payable ink! messages.
- [x] Implement codegen for enabled or disabled dynamic storage allocator.
- [x] Improve codegen of `Env` type alias. Should be no longer needed ideally and instead rely on `ink_lang::ContractEnv` trait implementation for the storage struct.
- [x] Fix potential `BaseEvent` bug: https://github.com/paritytech/ink/issues/483
- [x] `#[ink::trait_definition]` proc. macro that asserts that the given trait definition respects certain rules.
    - [x] implemented analysis
    - [x] implemented codegen
- [x] Experiment if `proc-macro-error` crate is actually useful for us for error reporting in the `ink_lang_ir` crate.
    - **Result:** It probably would not be of too much help for ink!. However, we could need the ability to output multiple errors at once.
- [x] Refactor core facilities for cross-contract calling to make them trait-impl proof:
    - [x] Refactor `CallBuilder` et. al.
    - [x] Refactor `InstantiateBuilder` et. al. (now called `CreateBuilder`)
- [x] Implement the entire rest of ink! codegen:
    - [x] `EnvTypes` trait
    - [x] ink! storage (`#[ink(storage)]`) struct
    - [x] ink! event definitions
        - [x] event struct generation + trait impls
        - [x] `EmitEvent` trait implementation
    - [x] ink! implementation blocks
        - [x] ink! messages
        - [x] ink! constructors
        - [x] `#[ink(impl)]` defined impl blocks with no ink! specific items
        - [x] non-ink! specific impl items
    - [x] metadata generation procedures
    - [x] cross-calling definitions
        - [x] For ink! messages
        - [x] For ink! trait messages
            - [x] for short-hand syntax
            - [x] for long-hand syntax
        - [x] For ink! constructors
        - [x] For ink! trait contructors
            - [x] for short-hand syntax
            - [x] for long-hand syntax
    - [x] dispatch routines
        - [x] ink! messages dispatch enum
        - [x] ink! constructors dispatch enum
        - [x] dispatch trait impls
        - [x] dispatch entry points
        - [x] `DispatchUsingMode` impl
        - [x] implement support for trait implementations in dispatch trait impls
        - [x] Make `Constr` and `Msg` implementations more concrete around trait impls
    - [x] All other non-ink! specific (Rust) items

## Description

The entire codegen library is going to be based on the new ink! IR that has been added in [this previous PR](https://github.com/paritytech/ink/pull/451).

In the end the `lang_lang_macro` crate that is the actual proc. macro will only have dependencies to `ink_lang_ir` and `ink_lang_codegen` in order to completely do its job of generating the ink! smart contract code.

## API

The target API is for `ink_lang_codegen` is kept minimal:

2020-07-04-163638_1462x209_scrot

### Default Trait Impls

Read more about this potential footgun here:
https://gist.github.com/Robbepop/de8a16d6b476a7d0b3bf88f343ac8384

## New ink_core::env::call module

The `ink_core::env::call` module got significantly cleaned up. Usage should be more clear now.
Also its builders are now checking more properly the construction process and should have even less overhead than before.

2020-07-11-132834_1773x1096_scrot

## Generated Metadata

Here is the metadata for the Flipper example contract generated with the new codegen:

```json
{
  "metadata_version": "0.1.0",
  "source": {
    "hash": "0x5f9c6583fab4323caebd7db0e57266c6973e3299406dc5ac3324f71017c4edbe",
    "language": "ink! 2.1.0",
    "compiler": "rustc 1.47.0-nightly"
  },
  "contract": {
    "name": "flipper",
    "version": "2.1.0",
    "authors": [
      "Parity Technologies <admin@parity.io>"
    ]
  },
  "spec": {
    "constructors": [
      {
        "args": [
          {
            "name": "init_value",
            "type": {
              "displayName": [
                "bool"
              ],
              "id": 1
            }
          }
        ],
        "docs": [],
        "name": [
          "new"
        ],
        "selector": "0xd183512b"
      },
      {
        "args": [],
        "docs": [],
        "name": [
          "default"
        ],
        "selector": "0x6a3712e2"
      }
    ],
    "docs": [],
    "events": [],
    "messages": [
      {
        "args": [],
        "docs": [],
        "isPayable": false,
        "mutates": true,
        "name": [
          "flip"
        ],
        "returnType": null,
        "selector": "0xc096a5f3"
      },
      {
        "args": [],
        "docs": [],
        "isPayable": false,
        "mutates": false,
        "name": [
          "get"
        ],
        "returnType": {
          "displayName": [
            "bool"
          ],
          "id": 1
        },
        "selector": "0x1e5ca456"
      }
    ]
  },
  "storage": {
    "struct": {
      "fields": [
        {
          "layout": {
            "cell": {
              "key": "0x0000000000000000000000000000000000000000000000000000000000000000",
              "ty": 1
            }
          },
          "name": "value"
        }
      ]
    }
  },
  "types": [
    {
      "def": {
        "primitive": "bool"
      }
    }
  ]
}
```


core/derive/src/tests/packed_layout.rs
```diff
@@ -21,8 +21,7 @@ fn unit_struct_works() {
             struct UnitStruct;
         }
         expands to {
-            #[allow(non_upper_case_globals)]
-            const _DERIVE_ink_core_storage2_traits_PackedLayout_FOR_UnitStruct: () = {
+            const _: () = {
```
cmichi, line 24: Why do you get rid of all these names with informational value?

Robbepop: because these names do not really bear informational value and it is a feature that they can no longer potentially clash with other names.

core/src/env/call/execution_input.rs
```diff
@@ -15,6 +15,7 @@
 use crate::env::call::Selector;
 
 /// The input data for a smart contract execution.
+#[derive(Debug)]
```
cmichi, line 18: Intentional Debug derive here and below?

lang/codegen/src/generator/contract.rs
```diff
@@ -0,0 +1,75 @@
...
+            #( #attrs )*
+            #vis mod #ident {
+                #env
+                #storage
+                #events
+                #dispatch
+                #item_impls
+                #cross_calling
+                #metadata
+                // #event_structs
```
cmichi, line 70: I guess this is for the upcoming event topics?

lang/codegen/src/generator/cross_calling.rs
```diff
@@ -0,0 +1,816 @@
...
+    fn generate_code(&self) -> TokenStream2 {
+        let storage = self.generate_storage();
+        let standard_impls = self.generate_standard_impls();
+        let call_forwarder = self.generate_call_forwarders();
+        let impl_blocks = self.generate_impl_blocks();
+        quote! {
+            #storage
+            #standard_impls
+            #call_forwarder
+            // #short_hand_impls
```
cmichi, line 62: Forgotten or intentional?

metadata/src/specs.rs
```diff
@@ -372,9 +425,9 @@ impl<M, R> MessageSpecBuilder<Missing<state::Selector>, M, R> {
     }
 }
 
-impl<S, R> MessageSpecBuilder<S, Missing<state::Mutates>, R> {
+impl<S, P, R> MessageSpecBuilder<S, Missing<state::Mutates>, P, R> {
     /// Sets if the message is mutable, thus taking `&mut self` or not thus taking `&self`.
```
cmichi, line 429: Is this comment outdated? The function takes `self`.

core/src/env/call/create_builder.rs
```diff
@@ -0,0 +1,338 @@
...
+where
+    T: EnvTypes,
+{
+    /// Creates the contract instance from the account ID of the already instantiated contract.
+    fn from_account_id(account_id: <T as EnvTypes>::AccountId) -> Self;
+}
+
+/// Builds up contract instantiations.
+#[derive(Debug)]
+pub struct CreateParams<E, Args, R>
```
ascjones, line 50: Should it not still be `Instantiate` rather than `Create` so it matches the name from the contracts pallet? Also all comments say `instantiate`

core/src/env/call/call_builder.rs
```diff
@@ -0,0 +1,447 @@
...
+pub trait IndicateReturnType: Default + self::seal::Sealed {}
+impl IndicateReturnType for () {}
+impl<T> IndicateReturnType for ReturnType<T> {}
+
+impl<E, Callee, GasLimit, TransferredValue, Args>
+    CallBuilder<E, Callee, GasLimit, TransferredValue, Args, Unset<ReturnType<()>>>
+where
+    E: EnvTypes,
+{
+    /// Sets the value transferred upon the execution of the call.
```
ascjones, line 318: Incorrect comment: should be about the return type

core/src/env/call/call_builder.rs
```diff
@@ -0,0 +1,447 @@
...
+{
+    /// Finalizes the call builder to call a function.
+    pub fn params(self) -> CallParams<E, Args, RetType> {
+        CallParams {
+            callee: self.callee.value(),
+            gas_limit: self.gas_limit.unwrap_or_else(|| 0),
+            transferred_value: self
+                .transferred_value
+                .unwrap_or_else(|| E::Balance::from(0)),
+            return_type: Default::default(),
```
ascjones, line 400: Set the return type?

ascjones (commented):
WIP

lang/codegen/src/generator/events.rs
```diff
@@ -0,0 +1,235 @@
...
+                const __ink_MAX_EVENT_TOPICS: usize = <
+                    <#storage_ident as ::ink_lang::ContractEnv>::Env as ::ink_core::env::EnvTypes
+                >::MAX_EVENT_TOPICS;
+
+                fn __ink_ensure_max_event_topics<T>(_: T)
+                where
+                    T: __ink_RenameBool,
+                    <T as __ink_RenameBool>::Type: ::ink_lang::False,
+                {}
+                let _ = __ink_ensure_max_event_topics::<[(); (#len_topics <= __ink_MAX_EVENT_TOPICS) as usize]>;
```
ascjones, line 176: :eyes: so what will the error look like if I exceed the max event topics?

lang/ir/src/ir/item/event.rs
```diff
@@ -92,16 +92,17 @@ impl TryFrom<syn::ItemStruct> for Event {
...
-            syn::Visibility::Inherited
-            | syn::Visibility::Restricted(_)
-            | syn::Visibility::Crate(_) => {
-                return Err(format_err_spanned!(
-                    &item_struct.vis,
-                    "non `pub` ink! event structs are not supported",
-                ))
-            }
-            _ => (),
+        let bad_visibility = match &item_struct.vis {
```
ascjones, line 95: Duplicate code fragment with `./storage.rs`. Not sure whether worth extracting, just flagging it.

lang/ir/src/ir/item/storage.rs
```diff
@@ -91,16 +91,17 @@ impl TryFrom<syn::ItemStruct> for Storage {
...
-            syn::Visibility::Inherited
-            | syn::Visibility::Restricted(_)
-            | syn::Visibility::Crate(_) => {
-                return Err(format_err_spanned!(
-                    &item_struct.vis,
-                    "non `pub` ink! storage structs are not supported",
-                ))
-            }
-            _ => (),
+        let bad_visibility = match &item_struct.vis {
```
ascjones, line 94: Duplicate of code in `event.rs`

lang/ir/src/ir/item_impl/constructor.rs
```diff
@@ -586,6 +553,12 @@ mod tests {
                 #[ink(event)]
                 fn my_constructor() -> Self {}
             },
+            // constructor + payable
+            syn::parse_quote! {
+                #[ink(constructor)]
+                #[ink(payable)]
```
ascjones, line 559: Should we allow `payable` on constructors now, since it is payable by default it could be misleading

metadata/src/specs.rs
```diff
@@ -289,16 +306,23 @@ impl ConstructorSpecBuilder<state::Selector> {
...
+    /// In case of trait provided messages and constructors the prefix
+    /// by convention in ink! is the name of the trait.
+    name: Vec<&'static str>,
     /// The selector hash of the message.
     #[serde(serialize_with = "serialize_as_byte_str")]
     selector: [u8; 4],
     /// If the message is allowed to mutate the contract state.
     mutates: bool,
+    /// If the message is payable by the caller.
+    #[serde(rename = "isPayable")]
```
ascjones, line 320: `[serde(rename_all = "camelCase")]` should handle this

metadata/src/specs.rs
```diff
@@ -289,16 +306,23 @@ impl ConstructorSpecBuilder<state::Selector> {
...
+    is_payable: bool,
     /// The parameters of the message.
     args: Vec<MessageParamSpec<F>>,
     /// The return type of the message.
+    #[serde(rename = "returnType")]
```
ascjones, line 325: Ditto

lang/codegen/src/generator/events.rs
Robbepop:
```rust
error[E0277]: the trait bound `erc20::_::EventTopicsWithinBounds: False` is not satisfied
  --> /home/robrob/projects/ink/examples/erc20/lib.rs:35:5
   |
35 | /     pub struct Transfer {
36 | |         #[ink(topic)]
37 | |         from: Option<AccountId>,
38 | |         #[ink(topic)]
...  |
45 | |         b: i32,
46 | |     }
   | |     ^
   | |     |
   | |_____the trait `False` is not implemented for `erc20::_::EventTopicsWithinBounds`
   |       required by this bound in `erc20::_::__ink_ensure_max_event_topics`
```

Robbepop: Problem is we cannot query the `MAX_EVENT_TOPICS` value at contract compile time. So we need to force the Rust compiler into doing the check for us.

metadata/src/specs.rs
Robbepop: yeah thanks and actually I wanted to rename this to just `payable`.

lang/ir/src/ir/item_impl/constructor.rs
Robbepop: Payable constructors are actually not allowed and specifying this is a violation.

lang/ir/src/ir/item/event.rs
Robbepop: yeah not sure either. let me see if deduplication is easy.

core/src/env/call/create_builder.rs
```diff
@@ -0,0 +1,338 @@
...
+    E: EnvTypes,
+{
+    /// Sets the value transferred upon the execution of the call.
+    #[inline]
+    pub fn endowment(
+        self,
+        endowment: E::Balance,
+    ) -> CreateBuilder<E, CodeHash, GasLimit, Set<E::Balance>, Args, R> {
+        CreateBuilder {
+            env_types: Default::default(),
```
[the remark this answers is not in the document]
Robbepop: it is just a `PhantomData` used to simply mark the `EnvTypes` type in order to derive `E::Balance` etc.

core/src/env/call/create_builder.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+}
+
+#[cfg(
+    // We do not currently support cross-contract instantiation in the off-chain
+    // environment so we do not have to provide these getters in case of
+    // off-chain environment compilation.
+    all(not(feature = "std"), target_arch = "wasm32")
+)]
+impl<E, Args, R> CreateParams<E, Args, R>
```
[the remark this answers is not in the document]
Robbepop: good catch

core/src/env/call/call_builder.rs
Robbepop: again the return type field is only used to compile time signature derivation. Not really a thing anymore during runtime.

core/src/env/call/create_builder.rs
Robbepop: The main reason I chose `Create` over `Instantiate` was that it is shorter and simpler to type. Blame me for my inability to use a keyboard. :D

metadata/src/specs.rs
Robbepop: Yeah this is a builder pattern so we can construct the metadata of the ink! entity in question. So `self` is not what we are referring to with it.

lang/codegen/src/generator/contract.rs
Robbepop: Those are living in `#events` and this is just me forgetting to delete a comment.

core/src/env/call/execution_input.rs
Robbepop: Yes actually `Debug` derives are normally not bad since they allow for debugging and typically do not slow down compilation too much. For our release builds they are compiled out of the resulting Wasm file.

lang/codegen/src/generator/dispatch.rs
```diff
@@ -0,0 +1,697 @@
...
+            }
+            ir::Receiver::Ref => {
+                (None, quote! { MessageRef }, quote! { execute_message })
+            }
+        };
+        let selector_id = cws.composed_selector().unique_id();
+        let namespace = Self::dispatch_trait_impl_namespace(ir::CallableKind::Message);
+        // If all ink! messages deny payment we can move the payment check to before
+        // the message dispatch which is more efficient.
+        let accepts_payments = cws.is_payable() || self.all_messages_deny_payment();
```
ascjones, line 535:
```suggestion
        let accepts_payments = cws.is_payable();
```

ascjones (commented):
53/62

lang/codegen/src/generator/dispatch.rs
Robbepop: This change would be wrong. If all messages deny payment we can move the check for each of them individually to before dispatch which is more optimal in this case.

ascjones: My confusion stems from the naming of `accepts_payments`, it appears illogical. Now I understand what you mean, that if `self.all_messages_deny_payment() == true`, then the check happens here:

(https://github.com/paritytech/ink/blob/7f1bb35f1ed2652e5eb07e0cd5e6728a564909ea/lang/codegen/src/generator/dispatch.rs#L110) 

However, even with the comment this was not immediately obvious. Might be better to be named `check_accepts_payments` or similar.

Robbepop: Yeah even I was confused during development. Maybe it really is the bad naming here.

lang/codegen/src/generator/events.rs
ascjones: It's probably good enough for a first iteration

lang/codegen/src/generator/dispatch.rs
Robbepop: Maybe renaming to `deny_payments` is better here. However, normally I try to formulate everything "positively" with regards to checks etc. to avoid statements like "check against non unpayable non-message". :D

lang/ir/src/ir/item/storage.rs
Robbepop: resolved. thanks for pointing out! duplicate code often happens during prototyping of new features etc. so this is super helpful!

lang/ir/src/ir/item/event.rs
Robbepop: resolved. thanks for pointing out! duplicate code often happens during prototyping of new features etc. so this is super helpful!

lang/codegen/src/generator/events.rs
Robbepop: I think changing it to
```
error[E0277]: the trait bound `erc20::_::EventTopicsWithinBounds: True` is not satisfied
```
makes more sense. what do you think?

ascjones (approved):
LGTM

lang/macro/src/trait_def.rs
```diff
@@ -0,0 +1,29 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use ink_lang_codegen::generate_code;
+use proc_macro2::TokenStream as TokenStream2;
+use syn::Result;
+
+pub fn analyse(attr: TokenStream2, input: TokenStream2) -> TokenStream2 {
```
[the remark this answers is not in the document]
Robbepop: You are telling this to me who doesn't even know the difference for all the words. :upside_down_face: 
Thanks for telling! Will fix.

lang/codegen/src/generator/dispatch.rs
ascjones: I think that's just equivalent to `accept_payments`. The problem is that `true` either means "don't perform the check because this method is payable" or "don't perform the check because it is done somewhere else". I get that the comment is supposed to explain that, but I didn't get it at first.

lang/macro/src/trait_def.rs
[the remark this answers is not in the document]
Robbepop: fixed

lang/codegen/src/generator/dispatch.rs
Robbepop: Yeah I can totally understand your confusion. I am confused as well. :S Not sure what name would be better though.

lang/codegen/src/generator/events.rs
ascjones: Actually yes that does make a bit more sense! :smile:

cmichi (commented):
LGTM!

PR #472, opened 2020-07-09
[core] Optimize dynamic storage allocator

### LINKED ISSUE #407: Optimize dynamic storage allocator
In the dynamic storage allocator provided with the new `storage2` module there is an obvious place where we could improve performance.
Currently when searching for a free bit within the bit vector representing our existing live allocations we actually perform a search on the `CountFree` storage vector elements.

They contain 32 `u8` instances and a single `u32` instance.
The 32 `u8` instances each represent the number of set (`1`) bits in the associated 256-bit chunk whereas the `u32` is simply put the 9th bit for each of the 32 `u8` instances.
So every bit in the `u32` tells if a 256-bit block is already completely filled with existing live allocations if it is set to `1`.

To search for a new free spot we, however, iterate over the 32 `u8`. What we could do instead is to abuse the implied behavior of the 9th bit stored in the single `u32` in order to iterate much much faster to find a free bit. There we simply have to perform a search for the first unset bit within the `u32` and can even skip the whole block if the `u32` is equal to `u32::MAX`. That's a single comparison compared to 32 comparisons.

In order to be sure that this is an actual improvement and doesn't screw with the correctness of the underlying algorithm this issue is also about adding performance tests and maybe some more unit tests to uncover situations that could be problematic for this optimized use case.

## ToDo List

- Apply the optimization described above.
- Implement potentially missing unit tests to actually test if the optimized version is still correct.
- Implement benchmark tests and compare the new optimized version with the old as "proof" that the optimized version is actually more efficient.

cmichi:
Closes #407.

## Benches for 100k `BitStash::put` ops (updated)

### New version (no more loop):

```
Bench: populated cache/fill_bitstash                                                                             
                        time:   [16.253 ms 16.922 ms 17.659 ms]                                                                                                
                        change: [-10.013% -3.6306% +3.2150%] (p = 0.29 > 0.05)                                                                                 
Found 13 outliers among 100 measurements (13.00%)                              
  5 (5.00%) high mild                                                                                                                                          
  8 (8.00%) high severe                                                                                                                                        
Bench: populated cache/one_put                                                                             
                        time:   [465.78 ns 500.94 ns 538.52 ns]                
                        change: [-97.830% -97.481% -97.089%] (p = 0.00 < 0.05)                                                                                 
                        Performance has improved.                                                                                                              
Found 3 outliers among 100 measurements (3.00%)                                                                                                                
  3 (3.00%) high mild                                                                                                                                          
                                                                                                                                                               
Bench: empty cache/fill_bitstash                                                                                                                               
                        time:   [17.082 ms 17.946 ms 18.910 ms]                                                                                                
                        change: [-1.5820% +5.6911% +13.130%] (p = 0.11 > 0.05)                                                                                 
Found 12 outliers among 100 measurements (12.00%)
  7 (7.00%) high mild
  5 (5.00%) high severe
Bench: empty cache/one_put                                                                             
                        time:   [7.7743 us 8.3617 us 9.0137 us]
                        change: [-5.6624% +3.8722% +14.787%] (p = 0.47 > 0.05)
Found 18 outliers among 100 measurements (18.00%)
  5 (5.00%) high mild
  13 (13.00%) high severe
```

### Running the old version (loop) subsequently after the new version:
```
Bench: populated cache/fill_bitstash                                                                            
                        time:   [28.347 ms 29.501 ms 30.765 ms]
                        change: [+64.246% +74.338% +85.703%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 13 outliers among 100 measurements (13.00%)
  2 (2.00%) high mild
  11 (11.00%) high severe
Bench: populated cache/one_put                                                                            
                        time:   [757.24 ns 810.25 ns 867.63 ns]
                        change: [+46.578% +61.744% +77.221%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

Bench: empty cache/fill_bitstash                                                                            
                        time:   [27.392 ms 28.532 ms 29.816 ms]
                        change: [+48.523% +58.986% +69.641%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) high mild
  8 (8.00%) high severe
Bench: empty cache/one_put                                                                            
                        time:   [8.2094 us 8.9335 us 9.7604 us]
                        change: [-4.1931% +6.8383% +21.234%] (p = 0.27 > 0.05)
Found 19 outliers among 100 measurements (19.00%)
  1 (1.00%) high mild
  18 (18.00%) high severe
```

### DIFF
--- a/core/benches/bench_bitstash.rs
+++ b/core/benches/bench_bitstash.rs
@@ -89,9 +89,9 @@ fn bench_populated_cache(c: &mut Criterion) {
         b.iter(|| populated_cache::fill_bitstash())
     });
     group.bench_function("one_put", |b| {
-        b.iter_batched(
+        b.iter_batched_ref(
             || create_large_stash(),
-            |mut stash| one_put(&mut stash),
+            |stash| black_box(one_put(stash)),
             BatchSize::SmallInput,
         )
     });
--- a/core/benches/bench_bitstash.rs
+++ b/core/benches/bench_bitstash.rs
@@ -84,7 +84,7 @@ mod populated_cache {
 }
 
 fn bench_populated_cache(c: &mut Criterion) {
-    let mut group = c.benchmark_group("Bench: `fill_bitstash` (populated cache)");
+    let mut group = c.benchmark_group("Bench: populated cache");
     group.bench_function("fill_bitstash", |b| {
         b.iter(|| populated_cache::fill_bitstash())
     });
@@ -116,7 +116,7 @@ mod empty_cache {
 /// benchmark iteration.
 fn bench_empty_cache(c: &mut Criterion) {
     let _ = env::test::run_test::<env::DefaultEnvTypes, _>(|_| {
-        let mut group = c.benchmark_group("Bench: `fill_bitstash` (empty cache)");
+        let mut group = c.benchmark_group("Bench: empty cache");
         group
             .bench_function("fill_bitstash", |b| b.iter(|| empty_cache::fill_bitstash()));
         group.bench_function("one_put", |b| {
--- a/core/benches/bench_bitstash.rs
+++ b/core/benches/bench_bitstash.rs
@@ -16,6 +16,7 @@ use criterion::{
     black_box,
     criterion_group,
     criterion_main,
+    BatchSize,
     Criterion,
 };
 
@@ -50,14 +51,29 @@ fn pull_stash() -> BitStash {
     <BitStash as SpreadLayout>::pull_spread(&mut KeyPtr::from(root_key))
 }
 
+/// Returns a stash on which `100_000` `put` operations have been executed.
+fn create_large_stash() -> BitStash {
+    let mut stash = BitStash::default();
+    for _ in 0..100_000 {
+        stash.put();
+    }
+    stash
+}
+
 mod populated_cache {
     use super::*;
 
+    /// Executes `put` operations on a new `BitStash` exactly `BENCH_ALLOCATIONS` times.
     pub fn fill_bitstash() {
         let mut stash = BitStash::default();
         for _ in 0..BENCH_ALLOCATIONS {
             black_box(stash.put());
         }
+    }
+
+    /// Executes only a single `put` operation on the stash.
+    pub fn one_put(stash: &mut BitStash) {
+        black_box(stash.put());
     }
 }
 
@@ -65,6 +81,14 @@ fn bench_populated_cache(c: &mut Criterion) {
     let mut group = c.benchmark_group("Bench: `fill_bitstash` (populated cache)");
     group.bench_function("fill_bitstash", |b| {
         b.iter(|| populated_cache::fill_bitstash())
+    });
+
+    group.bench_function("one_put", |b| {
+        b.iter_batched(
+            || create_large_stash(),
+            |mut stash| populated_cache::one_put(&mut stash),
+            BatchSize::SmallInput,
+        )
     });
     group.finish();
 }
@@ -82,6 +106,11 @@ mod empty_cache {
             black_box(stash.put());
         }
     }
+
+    /// Executes only a single `put` operation on the stash.
+    pub fn one_put(stash: &mut BitStash) {
+        black_box(stash.put());
+    }
 }
 
 /// In this case we lazily instantiate a `BitStash` by first creating and storing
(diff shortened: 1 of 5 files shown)


core/src/storage2/collections/bitstash/counts.rs
```diff
@@ -63,24 +63,57 @@ impl FullMask {
...
+
+        Some(i as u8)
+
+        // Old version
+        // for (i, count) in self.counts.iter_mut().enumerate() {
+        // if !self.full.is_full(i as u8) {
+        // return Some(i as u8)
+        // }
+        // }
+        // None
```
cmichi, line 98: I'll remove this commented code before we merge.

core/src/storage2/collections/bitstash/counts.rs
```diff
@@ -63,24 +63,57 @@ impl FullMask {
...
+        }
+
+        let i = (!self.full.0).leading_zeros();
+        if i == 32 {
+            unreachable!(
+                "if full we must already have exited right at the start of this function"
+            );
+        }
+
+        Some(i as u8)
```
Robbepop, line 90: Have you experimented around with removing `self.full.is_completely_full()` and instead make `unreachable!` just return `None`? Would look a bit clearer and in the common case we drop one integer comparison.

core/benches/bench_bitstash.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    /// In this case we lazily load the stash from storage using `pull_spread`.
+    /// This will just load lazily and won't pull anything from the storage.
+    /// `take` will then result in loading from storage.
+    pub fn fill_bitstash() {
+        push_stash();
+        let mut stash = pull_stash();
+        for _ in 0..BENCH_ALLOCATIONS {
+            black_box(stash.put());
+        }
```
Robbepop, line 83: It would actually be more interesting to see how one single `stash.put()` performs in case of a lazily loaded `BitStash` with 100k elements in it. So this one load would probably need approx. 12 storage reads which should sum up quite heavily in the benchmarks. So: For every iteration push upfront (outside the bench-loop) a filled instance to the storage and pull it lazily (empty cache). And then benchmark how a single additional `put` performs.

Robbepop: We actually should deploy the same benchmark for a populated cache as well.

core/src/storage2/collections/bitstash/counts.rs
```diff
@@ -70,17 +70,30 @@ impl CountFree {
     ///
     /// Returns `None` if all counts are `0xFF`.
     pub fn position_first_zero(&mut self) -> Option<u8> {
```
Robbepop, line 72: does no longer need to be `&mut self`

core/src/storage2/collections/bitstash/counts.rs
```diff
@@ -70,17 +70,30 @@ impl CountFree {
...
+    /// - If the increment would cause an overflow.
+    pub fn inc(&mut self, index: usize) {
+        assert!(index < 32, "index is out of bounds");
+        if self.counts[index] == !0 {
+            self.full.set_full(index as u8);
+        } else {
+            let new_value = self.counts[index]
+                .checked_add(1)
+                .expect("set bits increment overflowed");
+            self.counts[index] = new_value;
```
Robbepop, line 95: This is doubly checking the addition since we already have a check for `== !0` in the if. Not needed.

core/benches/bench_bitstash.rs
```diff
@@ -82,6 +106,11 @@ mod empty_cache {
             black_box(stash.put());
         }
     }
+
+    /// Executes only a single `put` operation on the stash.
+    pub fn one_put(stash: &mut BitStash) {
+        black_box(stash.put());
+    }
```
Robbepop, line 113: code dupe

core/benches/bench_bitstash.rs
```diff
@@ -92,6 +121,14 @@ fn bench_empty_cache(c: &mut Criterion) {
         let mut group = c.benchmark_group("Bench: `fill_bitstash` (empty cache)");
         group
             .bench_function("fill_bitstash", |b| b.iter(|| empty_cache::fill_bitstash()));
+
+        group.bench_function("one_put", |b| {
+            b.iter_batched(
+                || create_large_stash(),
+                |mut stash| empty_cache::one_put(&mut stash),
```
Robbepop, line 128: Not testing empty cache.

Robbepop: You need to pull a new instance from the storage for every iteration to make it have empty cache. Can't re-use instances otherwise we populate cache.

PR #475, opened 2020-07-10
[core] Add simple fuzz test for storage2::BitStash

cmichi:
Can be run via `cd core/ && cargo test --features ink-fuzz-tests -- bitstash::fuzz_tests`.

### DIFF
--- a/core/src/storage2/collections/bitstash/fuzz_tests.rs
+++ b/core/src/storage2/collections/bitstash/fuzz_tests.rs
@@ -1,0 +1,75 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


core/src/storage2/collections/bitstash/fuzz_tests.rs
```diff
@@ -0,0 +1,75 @@
...
+use super::BitStash;
+use crate::env;
+
+/// Conducts repeated insert and remove operations into the stash by iterating
+/// over `xs`. Typically the `xs` and `inserts_each` arguments are provided
+/// by our fuzzing engine in an iterative manner.
+///
+/// For each odd `x` in `xs` a number of put operations are executed.
+/// For each even `x` it is asserted that the previously inserted elements
+/// are in the stash and they are taken out subsequently.
```
Robbepop, line 24: Btw.: Just curious. Is this kind of pattern common in fuzz tests?

cmichi: Nope, I haven't seen it elsewhere. I started this pattern while implementing `BTreeMap` and found a couple of corner cases. The idea is to maximize randomness for insert/remove operations. Yeah so it has worked well so far, but of course I'm open to other ideas!

Robbepop: Would be very nice if you could investigate this further in how other crates that are doing fuzz tests are using their fuzzers to randomize their inputs.

PR #477, opened 2020-07-16
[core] Add Entry API for storage2::HashMap

### LINKED ISSUE #427: Implement Entry API for storage2::HashMap
The current re-implementation of the `storage::HashMap` for the `storage2` module does not yet have a so-called `Entry` API.
It is important to note that only the former `storage::BTreeMap` was actually having one since the former `storage::HashMap` wasn't even a proper collection.
Since `storage2::HashMap` now is a proper collection type we can also add the `Entry` API to it inspired by the Rust standard library `Entry` API that can be found here: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry

The pros of the `Entry` API are that it allows for cleaner accesses to the underlying collection and also provides more efficient accesses in terms of calculating the hash or querying the tree or interacting with the contract storage. We really want this API back!

## ToDo List

- [ ] Implement the `Entry` API for the `storage2::HashMap`.
- [ ] Write unit tests for the new `Entry` API.
- [ ] **Bonus:** Write benchmarks showing the performance difference between `Entry` API and equivalent accesses without `Entry` API.

cmichi:
Closes #427.

Benches:
```
$ cargo bench -- entry_api 2>&1 | egrep "^Compare|time" | grep -v Warning
Compare: `insert_and_inc` and `insert_and_inc_entry_api` (populated cache)/insert_and_inc
                        time:   [263.95 us 275.97 us 289.26 us]
Compare: `insert_and_inc` and `insert_and_inc_entry_api` (populated cache)/insert_and_inc_entry_api
                        time:   [271.81 us 283.34 us 295.88 us]

Compare: `remove` and `remove_entry_api` (populated cache)/remove
                        time:   [171.53 us 177.15 us 183.52 us]
Compare: `remove` and `remove_entry_api` (populated cache)/remove_entry_api
                        time:   [179.98 us 190.69 us 202.47 us]

Compare: `insert_and_inc` and `insert_and_inc_entry_api` (empty cache)/insert_and_inc
                        time:   [827.08 us 872.82 us 927.68 us]
Compare: `insert_and_inc` and `insert_and_inc_entry_api` (empty cache)/insert_and_inc_entry_api
                        time:   [825.21 us 861.89 us 901.10 us]

Compare: `remove` and `remove_entry_api` (empty cache)/remove
                        time:   [567.94 us 600.89 us 633.95 us]
Compare: `remove` and `remove_entry_api` (empty cache)/remove_entry_api
                        time:   [519.43 us 540.76 us 564.59 us]
```

I used [codecov on `storage2/HashMap/mod.rs`](https://codecov.io/gh/paritytech/ink/pull/477/src/core/src/storage2/collections/hashmap/mod.rs?before=core/src/storage2/collections/hashmap/mod.rs) to guide the unit tests which I wrote.

### DIFF
--- a/core/benches/bench_hashmap.rs
+++ b/core/benches/bench_hashmap.rs
@@ -1,0 +1,160 @@
+// Copyright 2019-2020 Parity Technologies (UK) Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use criterion::{
+    black_box,
+    criterion_group,
+    criterion_main,
+    Criterion,
+};
+
+use ink_core::{
+    env,
+    storage2::{
+        collections::HashMap as StorageHashMap,
+        traits::{
+            KeyPtr,
+            SpreadLayout,
+        },
+    },
+};
+use ink_primitives::Key;
+
+criterion_group!(populated_cache, bench_insert_populated_cache,);
+criterion_group!(empty_cache, bench_insert,);
+criterion_main!(populated_cache, empty_cache,);
+
+/// Returns some test values for use in benchmarks.
+fn test_values() -> Vec<(i32, i32)> {
+    let mut v = Vec::new();
+    for index in 0..500 {
+        v.push((index, index));
+    }
+    v
+}
+
+/// Creates a storage stash from the given slice.
+fn hashmap_from_slice(slice: &[(i32, i32)]) -> StorageHashMap<i32, i32> {
+    slice.iter().copied().collect::<StorageHashMap<i32, i32>>()
+}
+
+/// Returns always the same `KeyPtr`.
+fn key_ptr() -> KeyPtr {
+    let root_key = Key::from([0x42; 32]);
+    KeyPtr::from(root_key)
+}
+
+/// Creates a storage stash and pushes it to the contract storage.
+fn push_storage_hashmap() {
+    let test_values = test_values();
+    let stash = hashmap_from_slice(&test_values[..]);
+    SpreadLayout::push_spread(&stash, &mut key_ptr());
+}
+
+/// Pulls a lazily loading storage stash instance from the contract storage.
+fn pull_storage_hashmap() -> StorageHashMap<i32, i32> {
+    <StorageHashMap<i32, i32> as SpreadLayout>::pull_spread(&mut key_ptr())
+}
+
+mod populated_cache {
+    use super::*;
+
+    pub fn insert_if_nonexistent() {
+        let test_values = test_values();
+        let mut hmap = hashmap_from_slice(&test_values[..]);
+        for key in 0..1000 {
+            black_box({
+                if !hmap.contains_key(&key) {
+                    hmap.insert(key, key);
+                }
+                *hmap.get_mut(&key).unwrap() += 1;
+            });
+        }
+    }
+
+    pub fn insert_if_nonexistent_entry_api() {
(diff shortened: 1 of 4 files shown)


core/benches/bench_hashmap.rs
```diff
@@ -0,0 +1,302 @@
...
+/// the `StorageHashMap` used in these benchmarks.
+const ENTRIES: i32 = 500;
+
+/// Returns some test values for use in benchmarks.
+fn test_values() -> Vec<(i32, i32)> {
+    let mut v = Vec::new();
+    for index in 0..ENTRIES {
+        v.push((index, index));
+    }
+    v
```
Robbepop, line 56:
```suggestion
    (0..ENTRIES).iter().map(|index| (index, index)).collect::<Vec<_>>()
```

core/benches/bench_hashmap.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// Iteratively checks if an entry is in the `StorageHashMap`. If not, it
+    /// is inserted. In either case it is incremented afterwards.
+    pub fn insert_and_inc(hmap: &mut StorageHashMap<i32, i32>) {
+        for key in 0..ENTRIES * 2 {
+            black_box({
+                if !hmap.contains_key(&key) {
+                    hmap.insert(key, key);
+                }
+                *hmap.get_mut(&key).unwrap() += 1;
+            });
```
Robbepop, line 99: As with `unsafe` I try to make the scope of `black_box` always as small as possible but as big as needed. However, I am not sure about best practices around usage of `black_box` to be honest.

core/benches/bench_hashmap.rs
```diff
@@ -0,0 +1,302 @@
...
+    pub fn remove_entry_api(hmap: &mut StorageHashMap<i32, i32>) {
+        for key in 0..ENTRIES {
+            black_box({
+                if let Entry::Occupied(o) = hmap.entry(key) {
+                    o.remove();
+                }
+            });
+        }
+    }
+}
```
Robbepop, line 241: Correct me if I am wrong but this looks like code dupe from populated cache. We could just use the same methods for both benchmarks but with different setups, right?

core/benches/bench_hashmap.rs
```diff
@@ -0,0 +1,293 @@
...
+
+fn bench_insert_and_inc(hmap: &mut StorageHashMap<i32, i32>) {
+    for key in 0..ENTRIES * 2 {
+        black_box({
+            if !hmap.contains_key(&key) {
+                hmap.insert(key, key);
+            }
+            *hmap.get_mut(&key).unwrap() += 1;
+        });
```
cmichi, line 93: So I'm also still learning about `black_box`'es, but the [docs](https://docs.rs/criterion/0.2.3/criterion/fn.black_box.html) say
>[…] used to prevent the compiler from optimizing away computations in a benchmark.

In the case above we want to ensure that each operation is definitely executed. Hence it seems reasonable to me to have this scope here.

core/benches/bench_hashmap.rs
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    /// Iteratively checks if an entry is in the `StorageHashMap`. If yes, it
+    /// is taken out.
+    ///
+    /// Uses the Entry API.
+    pub fn remove_entry_api(hmap: &mut StorageHashMap<i32, i32>) {
+        bench_removal_entry_api(hmap);
+    }
+}
```
Robbepop, line 148: `populated_cache` and `empty_cache` completely duplicated code

core/src/storage2/collections/hashmap/mod.rs
```diff
@@ -337,4 +384,204 @@ where
...
+    }
+
+    /// Inserts `value` into `entry`.
+    fn insert(value: V, entry: VacantEntry<'a, K, V, H>) -> &'a mut V {
+        let old_value = entry.base.insert(entry.key.clone(), value);
+        debug_assert!(old_value.is_none());
+        match entry.base.entry(entry.key) {
+            Entry::Vacant(_) => unreachable!("entry was just inserted; qed"),
+            Entry::Occupied(entry) => entry.into_mut(),
+        }
```
Robbepop, line 487: Couldn't this be replaced by `entry.base.get_mut(entry_key).expect("encountered invalid vacant entry")` ?

Issue #474, opened 2020-07-10
Sync with most recent Substrate changes


Robbepop:
The `contracts-pallet` (Seal) has changed significantly in [this PR](https://github.com/paritytech/substrate/issues/6355).
We need to adjust ink! to incorporate adjustments for all the changes necessary to make it work with the next upcoming released version of Substrate.

This issue is soon to be extended with a detailed ToDo list to track sub-tasks for this process.

## ToDo List

- `ext_gas_price` -> `ext_weight_to_fee(ctx, gas: u64, out_ptr: u32, out_len_ptr: u32)`
- Remove `ext_scratch_read`, `ext_scratch_write` and `ext_scratch_len`.
- Add `ext_input` and use it instead of `ext_scratch_read` + `ext_scratch_len`.
- Replace scratch buffer with `(pointer, len)` tuple:
    - [ ] `ext_get_storage(ctx, key_ptr: u32, out_ptr: u32, out_len_ptr: u32) -> ReturnCode`
        - `ReturnCode::KeyNotFound`
    - [ ] `ext_transfer(account_ptr: u32, account_len: u32, value_ptr: u32, value_len: u32)`
    - [ ] `ext_call`: Add `output_ptr` and `output_len_ptr` as well as change return type to `ReturnCode`.
        ```rust
        ext_call(
            callee_ptr: u32,
            callee_len: u32,
            gas: u64,
            value_ptr: u32,
            value_len: u32,
            input_data_ptr: u32,
            input_data_len: u32,
            output_ptr: u32,
            output_len_ptr: u32
        ) -> ReturnCode
        ```
        - `ReturnCode::CalleeReverted`
        - `ReturnCode::CalleeTrapped`
    - [ ] `ext_instantiate`: Add `output_ptr` and `output_len_ptr` as well as change return type to `ReturnCode`.
        ```rust
        ext_instantiate(code_hash_ptr: u32,
            code_hash_len: u32,
            gas: u64, value_ptr: u32,
            value_len: u32,
            input_data_ptr: u32,
            input_data_len: u32,
            address_ptr: u32,
            address_len_ptr: u32,
            output_ptr: u32,
            output_len_ptr: u32,
        ) -> ReturnCode
        ```
        - `ReturnCode::CalleeReverted`
        - `ReturnCode::CalleeTrapped`
    - [ ] `ext_input(ctx, buf_ptr: u32, buf_len_ptr: u32)`
[...]

PR #478, opened 2020-07-23
Sync with Substrate RC6

Robbepop:
Closes https://github.com/paritytech/ink/issues/474


core/src/env/api.rs
```diff
@@ -238,25 +238,6 @@ pub fn clear_contract_storage(key: &Key) {
...
-/// - If the called runtime function does not exist.
-pub fn invoke_runtime<T>(params: &T::Call) -> Result<()>
-where
-    T: EnvTypes,
-{
-    <EnvInstance as OnInstance>::on_instance(|instance| {
-        TypedEnv::invoke_runtime::<T>(instance, params)
-    })
-}
-
```
ascjones, line 259: So we decided on the "delete" vs feature-flag (which is what I prefer too). Could do this as a discrete PR then bringing them back could be as simple as a `git revert`

Robbepop: We won't bring them back since we now instead have
```rust
pub fn ext_call_chain_extension(
    func_id: u32,
    input_ptr: u32,
    input_len: u32,
    output_ptr: u32,
    output_len_ptr: u32,
) -> u32;
```

ascjones: Perfect :+1:

core/src/env/engine/on_chain/buffer.rs
```diff
@@ -29,63 +26,123 @@ impl StaticBuffer {
...
+        self.buffer
+    }
+}
+
+impl<'a> scale::Output for EncodeScope<'a> {
     fn write(&mut self, bytes: &[u8]) {
-        if self.len + bytes.len() > Self::CAPACITY {
-            panic!("static buffer overflowed")
-        }
+        assert!(self.len() + bytes.len() <= self.capacity());
```
cmichi, line 77: How about retaining the message?
```suggestion
        assert!(self.len() + bytes.len() <= self.capacity(), "static buffer overflowed");
```

core/src/env/engine/on_chain/buffer.rs
```diff
@@ -29,63 +26,123 @@ impl StaticBuffer {
...
         let len_bytes = bytes.len();
         self.buffer[start..(start + len_bytes)].copy_from_slice(bytes);
         self.len += len_bytes;
     }
 
     fn push_byte(&mut self, byte: u8) {
-        if self.len == Self::CAPACITY {
-            panic!("static buffer overflowed")
-        }
+        assert_ne!(self.len(), self.capacity());
```
cmichi, line 85: Same here.
```suggestion
        assert_ne!(self.len(), self.capacity(), "static buffer overflowed");
```

core/src/env/engine/on_chain/buffer.rs
Robbepop: I will retain them, however, they are only of use for us programmers since this is happening within the on-chain environment that has no panic output.

core/src/env/engine/off_chain/chain_extension.rs
```diff
@@ -0,0 +1,112 @@
...
+            Box::new(move |encoded_input: Vec<u8>| {
+                let decoded_input = scale::Decode::decode(&mut &encoded_input[..])?;
+                let decoded_output = extension.call(&decoded_input)?;
+                Ok(scale::Encode::encode(&decoded_output))
+            }),
+        );
+    }
+
+    /// Invokes the chain extension with the given parameters.
+    pub fn invoke<I, O>(&mut self, func_id: FuncId, input: &I) -> Result<O>
```
athei, line 95: AFAIK ink! uses "call" as verb in other places when a result is returned and not invoke.

core/src/env/engine/off_chain/chain_extension.rs
```diff
@@ -0,0 +1,112 @@
...
+    where
+        I: scale::Codec + 'static,
+        O: scale::Codec + 'static,
+    {
+        use std::collections::hash_map::Entry;
+        match self.registered.entry(func_id) {
+            Entry::Occupied(mut occupied) => {
+                let encoded_input = scale::Encode::encode(input);
+                let encoded_output = occupied.get_mut()(encoded_input)?;
+                scale::Decode::decode(&mut &encoded_output[..]).map_err(Into::into)
```
athei, line 105: Why to encode it here just to have it decoded again in the closure in order to pass it to the Trait implementer? Why not just pass it through here and remove the decoding pass in the closure?

Robbepop: The problem is that the off-chain environment is a static object that cannot be generic so it needs to marshall all values as it would be done in the real Seal. I see that this is bad but unfortunately I haven't come up with a better idea.

core/src/env/engine/off_chain/chain_extension.rs
Robbepop: This might be a good idea. The problem I have with `call` is that it doesn't tell the user anything about any evaluation of return type which to my understanding is better expressed by invoke/eval pairs.

core/src/env/engine/on_chain/ext.rs
```diff
@@ -79,27 +154,37 @@ mod sys {
...
-        pub fn ext_scratch_size() -> u32;
-        pub fn ext_scratch_read(dst_ptr: u32, offset: u32, len: u32);
-        pub fn ext_scratch_write(src_ptr: u32, len: u32);
+        pub fn ext_call_chain_extension(
+            func_id: u32,
+            input_ptr: u32,
+            input_len: u32,
+            output_ptr: u32,
+            output_len_ptr: u32,
+        ) -> u32;
```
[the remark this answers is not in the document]
Robbepop: Having this in the set of `ext_` calls won't invalidate compiled contracts as long as they do not use this functionality. We could simply `cfg` it until it has been properly implemented on the Substrate side. Or we could just leave it as it is so that ink! magically already supports the whole thing as soon as it is implemented on Substrate.

core/src/env/engine/off_chain/chain_extension.rs
athei: So it should be called eval then?

core/src/env/engine/on_chain/buffer.rs
```diff
@@ -29,63 +26,130 @@ impl StaticBuffer {
...
+impl<'a> From<&'a mut [u8]> for ScopedBuffer<'a> {
+    fn from(buffer: &'a mut [u8]) -> Self {
+        Self { buffer }
     }
 }
 
-impl<I: core::slice::SliceIndex<[u8]>> core::ops::IndexMut<I> for StaticBuffer {
-    fn index_mut(&mut self, index: I) -> &mut Self::Output {
-        core::ops::IndexMut::index_mut(&mut self.buffer[..self.len], index)
+impl<'a> ScopedBuffer<'a> {
```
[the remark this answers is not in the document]
Robbepop: That's the thing! We do use `wee_alloc` but we do not want to have it compiled in (1-2kb of Wasm space) if we do not need it. So avoiding `wee_alloc` where possible provided us with a drop of around 1-2 kB of Wasm file size for some contracts that don't rely (by definition) on heap memory allocation. This is also far simpler than a full fledged allocator and thus has less overhead. Also it works slightly different since we allocate (shift) in chunks but deallocate all at once, so it is more comparable to an arena in that regard.

core/src/env/engine/off_chain/chain_extension.rs
Robbepop: That's actually a fair point! :)

athei:
Upcoming API changes to `ext_transfer, ext_instantiate, ext_call`. But I guess this PR only targets RC5, right?

core/src/env/engine/on_chain/buffer.rs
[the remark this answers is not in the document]
athei: Yes its a bump allocation arena. Just keep in mind that when this implementation grows in the future it could make sense to switch to bumpalo.

Just out of curiosity: Is there a feature I need to opt-in in order to make use of allocations?

core/src/env/engine/on_chain/impls.rs
```diff
@@ -135,47 +93,40 @@ impl Env for EnvInstance {
...
 
-    fn get_contract_storage<R>(&mut self, key: &Key) -> Option<Result<R>>
+    fn get_contract_storage<R>(&mut self, key: &Key) -> Result<Option<R>>
     where
         R: scale::Decode,
     {
-        if ext::get_storage(key.as_bytes()).is_err() {
-            return None
+        let output = &mut self.scoped_buffer().take_rest();
+        if ext::get_storage(key.as_bytes(), output).is_err() {
```
ascjones, line 105: This would be stricter and clearer if we matched on the `KeyNotFound` error

core/src/env/engine/on_chain/ext.rs
```diff
@@ -97,7 +97,7 @@ type Result = core::result::Result<(), Error>;
 
 mod sys {
     extern "C" {
```
athei, line 99:
```suggestion
    #[link(wasm_import_module = "seal0")]
    extern "C" {
```

Robbepop: thanks, already implemented!

Robbepop:
> Upcoming API changes to `ext_transfer, ext_instantiate, ext_call`: [paritytech/substrate#6773](https://github.com/paritytech/substrate/pull/6773)
> 
> But I guess this PR only targets RC5, right?

This PR mainly targets to fully sync with Substrate's latest changes. So we will include those as well so that we are in sync again (finally).

core/src/env/engine/on_chain/buffer.rs
```diff
@@ -29,63 +26,130 @@ impl StaticBuffer {
...
+
+impl<'a> scale::Output for EncodeScope<'a> {
     fn write(&mut self, bytes: &[u8]) {
-        if self.len + bytes.len() > Self::CAPACITY {
-            panic!("static buffer overflowed")
-        }
+        assert!(
+            self.len() + bytes.len() <= self.capacity(),
+            "encode scope buffer overflowed",
+        );
```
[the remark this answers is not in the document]
Robbepop: yes it is kinda redundant but improves error messages a bit. however, not really needed tbh.

core/src/env/engine/on_chain/buffer.rs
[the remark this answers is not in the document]
Robbepop: > Just out of curiosity: Is there a feature I need to opt-in in order to make use of allocations?

Not yet. Also there are two kinds of allocations: Heap allocations and dynamic storage allocations both of which require their dedicated allocators. For heap allocations we use `wee_alloc` so far but we want to experiment with an even simpler bump-allocator since contract calls are always short lived and probably do not need functioning dealloc anyways. For dynamic storage allocations we have the dynamic storage allocator provided by ink!. It consumes some fair amount of Wasm file size when used and it is currently not possible to disable it, however, since it uses this fair amount of size we really want to disable it by default and only use it when really needed. This, however is more difficult than originally anticipated.

TIL: `bumpalo` But I don't think `bumpalo` suits our needs since our bump allocator above is especially needed for different amounts of raw bytes whereas from what I currently see `bumpalo` is just a normal arena for one specific type `T`.

core/src/env/engine/on_chain/buffer.rs
```diff
@@ -29,63 +26,130 @@ impl StaticBuffer {
...
 
     fn push_byte(&mut self, byte: u8) {
-        if self.len == Self::CAPACITY {
-            panic!("static buffer overflowed")
-        }
+        assert_ne!(
+            self.len(),
+            self.capacity(),
+            "encode scope buffer overflowed",
+        );
```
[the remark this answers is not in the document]
Robbepop: I am going to convert those asserts into `debug_assert` for a compromise.

core/src/env/engine/on_chain/buffer.rs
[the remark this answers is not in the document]
Robbepop: turned the assertions into debug assertions

Robbepop:
> > Upcoming API changes to `ext_transfer, ext_instantiate, ext_call`: [paritytech/substrate#6773](https://github.com/paritytech/substrate/pull/6773)
> > But I guess this PR only targets RC5, right?
> 
> This PR mainly targets to fully sync with Substrate's latest changes. So we will include those as well so that we are in sync again (finally).

Changes have been implemented and I have been able to build the Flipper contract using `cargo-contract` again:

2020-08-25-105809_627x163_scrot

core/src/env/engine/on_chain/ext.rs
```diff
@@ -16,58 +16,133 @@
...
-        ) -> u32;
-
-        pub fn ext_transfer(
+            transferred_value_ptr: u32,
+            transferred_value_len: u32,
+            input_ptr: u32,
+            input_len: u32,
+            output_ptr: u32,
+            output_len_ptr: u32,
+        ) -> ReturnCode;
```
athei, line 124: This is unsound. Seal is allowed to **extend** this enum. Older contracts on newer chains then may produce an invalid discriminant insides the enum. You have to use use `u32` everywhere and convert that to the enum using the `Unknown` variant as fallback with your macro.

core/src/env/engine/on_chain/ext.rs
```diff
@@ -16,58 +16,133 @@
...
+
+        impl From<ReturnCode> for Result {
+            #[inline]
+            fn from(return_code: ReturnCode) -> Self {
+                match return_code {
+                    ReturnCode::Success => Ok(()),
+                    $(
+                        ReturnCode::$name => Err(Error::$name),
+                    )*
+                    ReturnCode::UnknownError => Err(Error::UnknownError),
```
athei, line 61: How would that work? Seal does not produce this variant. What you need is a generated:
```
impl From<u32> for Result
```

where `_ => Error::UnknownError`.

athei (changes requested):
My only complaint is the `ReturnCode` casting which seems unsound.

alloc/src/handlers.rs
```diff
@@ -17,9 +17,7 @@ pub fn panic(_info: &core::panic::PanicInfo) -> ! {
     core::intrinsics::abort()
 }
 
-// `extern` fn uses type `core::alloc::Layout`, which is not FFI-safe
-#[allow(improper_ctypes)]
 #[alloc_error_handler]
-pub extern "C" fn oom(_: core::alloc::Layout) -> ! {
```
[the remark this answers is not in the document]
Robbepop: you are right that the `panic_handler` actually shouldn't be part of `ink_alloc`. Not sure where to put it else though. Maybe best place is in `ink_core` then.

core/src/env/backend.rs
```diff
@@ -86,7 +104,7 @@ pub trait Env {
     /// The setting of this property must be the last interaction between
     /// the executed contract and its environment.
     /// The environment access asserts this guarantee.
```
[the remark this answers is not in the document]
Robbepop: yeah, outdated docs. good find!

alloc/src/handlers.rs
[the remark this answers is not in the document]
athei: Seems about right.

core/src/env/engine/on_chain/ext.rs
Robbepop: implemented through `RawReturnCode` which is basically a `u32`.

athei (approved):

cmichi (approved):

PR #480, opened 2020-07-24
Implement Entry API for storage2::LazyHashMap

### LINKED ISSUE #479: Implement Entry API for storage2::LazyHashMap
While implementing #427 we found that the performance of the new `storage2::LazyHashMap` Entry API lacks behind our expectations. We presume that this is because the `HashMap` has to use techniques to work around the fact that `storage2::LazyHashMap` doesn't have an Entry API. Implementing the `LazyHashMap` Entry API hence should improve the performance of the `HashMap` Entry API benchmarks significantly.

## ToDo List

- [ ] Implement the Entry API for `storage2::LazyHashMap`.
- [ ] Write unit tests for the new Entry API.
- [ ] Write benchmarks showing the performance difference between Entry API and equivalent accesses without Entry API.
- [ ] Make `storage2::HashMap` use the new `storage2::LazyHashMap`.
- [ ] Run the benchmarks for the `storage2::HashMap` Entry API with and without the new `storage2::LazyHashMap` Entry API.

cmichi:
Closes #479.

# Current benchmarks

## Before `HashMap` was using the new `LazyHashMap` Entry API

```
$ cargo bench -- StorageHashMap 2>&1 | egrep " Compare|time" | egrep -v "Benchmarking|Warning"
StorageHashMap<i32, i32> Compare: `insert_and_inc` and `insert_and_inc_entry_api` (populated cache)/...
                        time:   [216.33 us 224.39 us 233.19 us]
StorageHashMap<i32, i32> Compare: `insert_and_inc` and `insert_and_inc_entry_api` (populated cache)/... #2
                        time:   [200.38 us 205.88 us 212.27 us]
StorageHashMap<i32, i32> Compare: `remove` and `remove_entry_api` (populated cache)/remove
                        time:   [140.09 us 147.61 us 156.13 us]
StorageHashMap<i32, i32> Compare: `remove` and `remove_entry_api` (populated cache)/remove_entry_api
                        time:   [134.15 us 137.28 us 140.45 us]
StorageHashMap<i32, i32> Compare: `insert_and_inc` and `insert_and_inc_entry_api` (empty cache)/inse...
                        time:   [609.94 us 622.39 us 635.53 us]
StorageHashMap<i32, i32> Compare: `insert_and_inc` and `insert_and_inc_entry_api` (empty cache)/inse... #2
                        time:   [704.19 us 729.47 us 755.61 us]
StorageHashMap<i32, i32> Compare: `remove` and `remove_entry_api` (empty cache)/remove
                        time:   [721.95 us 742.12 us 764.39 us]
StorageHashMap<i32, i32> Compare: `remove` and `remove_entry_api` (empty cache)/remove_entry_api
                        time:   [705.04 us 726.37 us 750.54 us]
```

## Now that `HashMap` is using the new `LazyHashMap` Entry API
```
$ cargo bench -- StorageHashMap 2>&1 | egrep " Compare|time" | egrep -v "Benchmarking|Warning"
StorageHashMap<i32, i32> Compare: `insert_and_inc` and `insert_and_inc_entry_api` (populated cache)/...
                        time:   [205.43 us 209.77 us 215.22 us]
StorageHashMap<i32, i32> Compare: `insert_and_inc` and `insert_and_inc_entry_api` (populated cache)/... #2
                        time:   [117.55 us 118.76 us 120.20 us]
StorageHashMap<i32, i32> Compare: `remove` and `remove_entry_api` (populated cache)/remove
                        time:   [121.13 us 123.00 us 125.32 us]
StorageHashMap<i32, i32> Compare: `remove` and `remove_entry_api` (populated cache)/remove_entry_api
                        time:   [81.564 us 82.229 us 83.036 us]
StorageHashMap<i32, i32> Compare: `insert_and_inc` and `insert_and_inc_entry_api` (empty cache)/inse...
                        time:   [594.74 us 602.38 us 611.13 us]
StorageHashMap<i32, i32> Compare: `insert_and_inc` and `insert_and_inc_entry_api` (empty cache)/inse... #2
                        time:   [496.79 us 505.89 us 519.49 us]
StorageHashMap<i32, i32> Compare: `remove` and `remove_entry_api` (empty cache)/remove
                        time:   [611.75 us 629.06 us 651.67 us]
StorageHashMap<i32, i32> Compare: `remove` and `remove_entry_api` (empty cache)/remove_entry_api
                        time:   [566.02 us 587.06 us 613.11 us]
```


core/src/storage2/lazy/mod.rs
```diff
@@ -44,7 +44,12 @@ pub use self::{
         LazyArrayLength,
     },
     lazy_cell::LazyCell,
-    lazy_hmap::LazyHashMap,
+    lazy_hmap::{
+        Entry,
+        LazyHashMap,
+        OccupiedEntry,
+        VacantEntry,
+    },
```
Robbepop, line 52: You should not expose the `Entry`, `OccupiedEntry` and `VacantEntry` to the `lazy` module and instead should make the `lazy_hmap` module simply public. This way we would not have needed the `Entry` -> `InternalEntry` rename. The std lib does the same for those utility types. So you have an `Entry` per map collection there as well.
However, I agree that to counteract confusion we should rename `Entry` to something like `StorageEntry` instead of `InternalEntry` because that describes better what it is.

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -395,7 +394,7 @@ where
     Key: From<<H as Hasher>::Output>,
 {
     /// Gets the given key's corresponding entry in the map for in-place manipulation.
-    pub fn entry(&mut self, key: K) -> Entry<K, V, H> {
+    pub fn entry(&mut self, key: K) -> Entry<K, V> {
```
Robbepop, line 401: The `unsafe` scope is bigger than it needs to be and also you are simply loading the entry from the cache while forgetting to eventually load it from storage first.

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -272,6 +353,12 @@ where
         self.key.as_ref()
     }
 
+    /// Returns the length of the underlying entries.
+    #[cfg(test)]
+    pub fn len(&self) -> usize {
```
Robbepop, line 358: You cannot and should not call this `len` since len normally refers to the number of items in a collection so this is highly confusing. What this actually is is the number of currently cached or persisted items.

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -296,8 +383,58 @@ where
...
+                    }
+                }
+            }
+            BTreeMapEntry::Vacant(vacant) => {
+                let value = self
+                    .key_at(&key)
+                    .map(|key| pull_packed_root_opt::<V>(&key))
+                    .unwrap_or(None);
+                vacant.insert(Box::new(StorageEntry::new(value, EntryState::Preserved)));
+                self.entry(key)
```
Robbepop, line 435: If you call this method recursively you won again nothing in the benchmarks since you query the underlying caching b-tree map multiple times now which is exactly what the entry API tries to avoid. So you have to find another method in order to return your entry properly without querying the underlying b-tree map multiple times just as you did in the occupied state.

core/src/storage2/collections/hashmap/mod.rs
```diff
@@ -165,6 +165,12 @@ where
         self.keys.len()
     }
 
+    /// Returns the number of key-value pairs stored in the hash map.
```
Robbepop, line 168:
```suggestion
    /// Returns the number of key-value pairs stored in the cache.
    ///
    /// # Note
    ///
    /// This is a test-only API and not intended for use outside of testing.
```

core/src/storage2/collections/hashmap/mod.rs
```diff
@@ -165,6 +165,12 @@ where
...
+    #[cfg(test)]
+    pub fn len_cached_entries(&self) -> u32 {
```
Robbepop:
```suggestion
    pub(crate) fn len_cached_entries(&self) -> u32 {
```

Robbepop: just makes the intent a bit clearer even though not necessary since it is `#[cfg(test)]`

core/src/storage2/hashmap_entry_api_tests.rs
```diff
@@ -307,6 +307,122 @@ macro_rules! gen_tests_for_backend {
...
+                let hmap1 = prefilled_hmap();
+                assert_eq!(hmap1.get(&b'A'), Some(&13));
+                push_hmap(&hmap1);
+
+                // when
+                let mut hmap2 = pull_hmap();
+
+                // then
+                match hmap2.entry(b'A') {
+                    Entry::Occupied(o) => assert_eq!(o.get(), &mut 13),
```
Robbepop, line 324: `get` returns a `&mut` ?

core/src/storage2/hashmap_entry_api_tests.rs
```diff
@@ -307,6 +307,122 @@ macro_rules! gen_tests_for_backend {
...
+                match hmap2.entry(b'A') {
+                    Entry::Occupied(mut o) => {
+                        assert_eq!(o.get_mut(), &mut 13);
+                    }
+                    Entry::Vacant(_) => panic!(),
+                }
+
+                Ok(())
+            })
+        }
```
Robbepop, line 353: Very similar to the above test so you could as well just have both asserts for `get` and `get_mut` in the same test imo.

core/src/storage2/hashmap_entry_api_tests.rs
```diff
@@ -307,6 +307,122 @@ macro_rules! gen_tests_for_backend {
...
+                assert_eq!(hmap1.get(&b'A'), Some(&13));
+                push_hmap(&hmap1);
+
+                // when
+                let mut hmap2 = pull_hmap();
+
+                // then
+                match hmap2.entry(b'A') {
+                    Entry::Occupied(mut o) => {
+                        assert_eq!(o.insert(999), 13);
```
Robbepop, line 369: `o.insert` returns by value and not `&mut T` such as in the standard library?

core/src/storage2/hashmap_entry_api_tests.rs
```diff
@@ -307,6 +307,122 @@ macro_rules! gen_tests_for_backend {
...
+        #[test]
+        fn value_not_in_cache_but_in_storage_remove_entry() -> env::Result<()> {
+            env::test::run_test::<env::DefaultEnvTypes, _>(|_| {
+                // given
+                let hmap1 = prefilled_hmap();
+                assert_eq!(hmap1.get(&b'A'), Some(&13));
+                push_hmap(&hmap1);
+
+                // when
+                let mut hmap2 = pull_hmap();
```
Robbepop, line 388: maybe you could provide a utility function since this is duplicated for at least 4 times now.

core/src/storage2/hashmap_entry_api_tests.rs
Robbepop: ah I think I am wrong since this just returned the old value by value. nvm then.

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -426,13 +455,32 @@ where
...
                 let value = self
                     .key_at(&key)
                     .map(|key| pull_packed_root_opt::<V>(&key))
                     .unwrap_or(None);
-                vacant.insert(Box::new(StorageEntry::new(value, EntryState::Preserved)));
-                self.entry(key)
+                match value.is_some() {
+                    // The entry was not in the cache, but in the storage. This results in
+                    // a problem: We only have `Vacant` here, but need to return `Occupied`,
+                    // to reflect this.
```
Robbepop, line 466: This comment block should be shifted into the `true => { ... }` block.

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -814,40 +862,77 @@ where
     }
 
     /// Take the ownership of the key and value from the map.
-    pub fn remove_entry(mut self) -> (K, V) {
-        let old = core::mem::replace(self.entry.get_mut().value_mut(), None)
-            .expect("entry behind `OccupiedEntry` must always exist");
+    pub fn remove_entry(self) -> (K, V) {
+        let old = match self.entry {
+            EntryOrMutableValue::Entry(mut entry) => {
+                core::mem::replace(entry.get_mut().value_mut(), None)
```
Robbepop, line 868: For option there is also the `take` method instead of `replace`.

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -814,40 +862,77 @@ where
...
+                    .expect("entry behind `OccupiedEntry` must always exist")
+            }
+            EntryOrMutableValue::MutableValue(v_mut) => {
+                core::mem::replace(v_mut.value_mut(), None)
```
Robbepop, line 872: better to use `take` method here

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -92,16 +93,39 @@ pub struct LazyHashMap<K, V, H> {
...
+    Entry(E),
+    /// A reference to the mutable value behind a cache entry.
+    MutableValue(V),
+}
+
+/// An occupied `EntryMap` entry that holds a value.
+type OccupiedCache<'a, K, V> = BTreeMapOccupiedEntry<'a, K, Box<StorageEntry<V>>>;
+
 /// A vacant entry with previous and next vacant indices.
 pub struct OccupiedEntry<'a, K, V>
```
Robbepop: Comment and type doesn't stick together.

Robbepop: Please look if you overlooked any copy'n'paste doc comments.

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -92,16 +93,39 @@ pub struct LazyHashMap<K, V, H> {
...
+/// `Occupied` here, since the object is now in the cache. We could do this
+/// by querying the cache another time -- but this would be an additional
+/// search. So what we do instead is to save a reference to the inserted
+/// cache value in the `Occupied`. As a consequence all Entry API operations
+/// (`get`, `remove`, ...) need to distinguish both cases.
+enum EntryOrMutableValue<E, V> {
+    /// An occupied `EntryMap` entry that holds a value.
+    Entry(E),
+    /// A reference to the mutable value behind a cache entry.
+    MutableValue(V),
```
Robbepop, line 113: The name should make it more clear that this represents the case where the element was not in the cache but in the storage.

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -814,40 +862,77 @@ where
...
+                let new_value =
+                    Box::new(StorageEntry::new(Some(new_value), EntryState::Mutated));
+                entry
+                    .insert(new_value)
+                    .into_value()
+                    .expect("entry behind `OccupiedEntry` must always exist")
+            }
+            EntryOrMutableValue::MutableValue(v_mut) => {
+                core::mem::replace(v_mut.value_mut(), Some(new_value))
+                    .expect("entry behind `MutableValue` must always exist")
```
Robbepop, line 933: I am not sure right now about this but this routine should make sure that the value is not only replaced but also that the cache is notified that it is no longer preserving the value if this is not already the case so that the cache properly flushes the value to the storage upon the end of execution.

Robbepop (commented):
looks good overall, some nits

core/src/storage2/lazy/lazy_hmap.rs
cmichi: It's flushed, I added a test for this now here: https://github.com/paritytech/ink/pull/480/commits/f5d011df1b948902610d5d77012c03f565b67ae9.

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -86,6 +93,72 @@ pub struct LazyHashMap<K, V, H> {
...
+/// search. So what we do instead is to save a reference to the inserted
+/// cache value in the `Occupied`. As a consequence all Entry API operations
+/// (`get`, `remove`, ...) need to distinguish both cases.
+enum EntryOrMutableValue<E, V> {
+    /// An occupied `EntryMap` entry that holds a value.
+    /// This represents the case where the key was in the cache.
+    Entry_ElementWasInCache(E),
+    /// A reference to the mutable value behind a cache entry.
+    /// This represents the case where the key was not in the cache, but in storage.
+    MutableValue_ElementWasNotInCache(V),
```
cmichi, line 115: No really happy with both those names, feel free to suggest better ones if you have a better idea.

core/src/storage2/lazy/lazy_hmap.rs
```diff
@@ -86,6 +93,72 @@ pub struct LazyHashMap<K, V, H> {
...
+/// search. So what we do instead is to save a reference to the inserted
+/// cache value in the `Occupied`. As a consequence all Entry API operations
+/// (`get`, `remove`, ...) need to distinguish both cases.
+enum EntryOrMutableValue<E, V> {
+    /// An occupied `EntryMap` entry that holds a value.
+    /// This represents the case where the key was in the cache.
+    EntryElementWasInCache(E),
+    /// A reference to the mutable value behind a cache entry.
+    /// This represents the case where the key was not in the cache, but in storage.
+    MutableValueElementWasNotInCache(V),
```
cmichi, line 115: Not really happy with both those names, feel free to suggest better ones if you have a better idea.

Robbepop: at least they speak for themselves :D

Robbepop (approved):
LGTM

PR #481, opened 2020-07-24
[core] storage2::BinaryHeap

### LINKED ISSUE #410: Port storage::BinaryHeap to storage2
The new `storage2` module is missing the important `BinaryHeap` that operates efficiently on contract storage.
It has been implemented by @cmichi in the past for the old `storage` module and needs to be ported with some adjustments over to `storage2`. Of course, if needed also a rewrite is possible here and might be useful since all `storage2` data structure have been complete rewrites so far to accomodate for the new requirements around the new traits and new low-level lazy data structures.

Also we might want to add some new APIs to the `storage::BinaryHeap`.
So, what is going to be needed for this ...

## Internal Structure

Similar to the `storage2::Vec` the `storage2::BinaryHeap` should be based on the low-level `LazyIndexMap` and should require a `len` field of type `Lazy<u32>` to store the number of elements that are currently stored in the storage binray heap.

## ToDo List

Subsets of this ToDo list can suffice to craft a PR so we can implement the `storage2::BinaryHeap` in many PRs incrementally.

- [ ] Implement `BinaryHeap` for the `storage2` module with the following methods:
    - [x] `new`: Creates a new `BinaryHeap` for contract inistantiation purposes (see other data structures for `storage2` as examples)
    - [x] `len`: Returns the current length of the binary heap.
    - [x] `is_empty`: Returns `true` if the binray heap is empty.
    - [x] `peek`: Returns a shared reference to the greatest element of the binary heap if any.
    - [x] `peek_mut`: Returns an exclusive reference to the greatest element of the binary heap if any.
    - [x] `push`: Pushes the given element to the binary heap.
    - [x] `pop`: Pops the greatest element from the binary heap.
    - [x] `iter`: Iterates over the values of the binary heap.
        Strangely the old `storage::BinaryHeap` has two iterator methods `iter` and `values` but I think we only need `iter`.
    - [x] **Bonus:** `clear`: Clears all elements from the binary heap.
- [ ] Implement all important standard Rust traits for the `storage2::BinaryHeap`:
    - [x] `Default`: For simply contract instantiation.
    - [x] `Debug`: Should be derivable.
    - [x] `PartialEq` + `Eq` if `T: PartialEq` and `T: Eq`: Mostly important for testing but anyways nice to have in general.
    - [x] `Drop`: Must be implemented in order to command the underlying `LazyIndexMap` to clear its cells. For an example how to implement this properly take a look at the `storage2::Vec`.
    - [x] `Extend` and `FromIterator`: Useful traits to initialize a `storage2::BinaryHeap` from an iterator.
    - [ ] `IntoIterator` for `&BinaryHeap` and `&mut BinaryHeap`: Generally useful for testing but also nice to have in general.
- [x] Implement `SpreadLayout` trait for the `storage2::BinaryHeap`.
- [x] Implement unit tests for the `storage2::BinaryHeap`.
    - For an inspiration look at the unit tests provided for the other `storage2` data structure.
- [x] **Bonus:** Add performance tests for the `storage2::BinaryHeap` using the `criterion` crate.
- [ ] **Bonus:** Add optimizations to merge child elements to use the same storage entry. This is useful since most often both child elements are accessed together.

ascjones:
Closes #410.

This PR is a minimal `BinaryHeap` implementation, using a `StorageVec` to store the underlying elements. The basic algorithms to reorganize the tree were "copied" from the Rust `BinaryTree` (with the `Hole` abstraction removed):

- [`sift_up`](https://github.com/rust-lang/rust/blob/39e29ce4d08674734e2f2759607b1486db7d0fde/src/liballoc/collections/binary_heap.rs#L512) on `push()`: 
- [`sift_down`](https://github.com/rust-lang/rust/blob/39e29ce4d08674734e2f2759607b1486db7d0fde/src/liballoc/collections/binary_heap.rs#L530) on `pop()`

I did begin to work on some optimizations, including porting the idea of storing elements in pairs from @cmichi's original implementation.

I propose this implementation for the first iteration: it is very simple and provides the worst case storage read/write complexity of:

|    |Reads  |Writes | 
|----|---------|---------|
| `push` | `O(log n)` | `O(log n)`|
| `pop` | `O(2log(n))` | `O(log n)`|

The obvious area for improvement there is the `O(2log(n))` reads for `push` which is the main improvement from storing child nodes in pairs in the original implementation - with the trade-off of increased *code* complexity.

Optimizations can be submitted and reviewed independently as follow up PRs, together with updated benchmarks and complexity measurements to compare with the baselines set in this PR. 

### Baseline (16cef007)
|                 | 8         | 16        | 32        | 64        |
|-----------------|-----------|-----------|-----------|-----------|
| **Empty**           |           |           |           |           |
| Push (Largest)  | 903.90 ns | 1.2531 us | 1.4664 us | 1.8076 us |
| Push (Smallest) | 499.25 ns | 521.94 ns | 546.15 ns | 546.15 ns |
| Pop             | 1.4516 us | 1.9274 us | 2.4798 us | 3.0575 us |
| **Populated**       |           |           |           |           |
| Push (Largest)  | 249.81 ns | 408.62 ns | 582.56 ns | 603.01 ns |
| Push (Smallest) | 131.50 ns | 222.89 ns | 250.39 ns | 257.56 ns |
| Pop             | 211.49 ns | 373.89 ns | 584.64 ns | 663.50 ns |

## todo

- [x] Rebase commits onto original master (this branch appears to be based on an old PR branch)
- [x] Run and record baseline performance
- [x] Quickcheck property tests
- [x] More unit tests


core/src/storage2/collections/binary_heap/reverse.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use crate::storage2::traits::{
+    KeyPtr,
+    PackedLayout,
+    SpreadLayout,
+};
+use ink_prelude::vec::Vec;
+use ink_primitives::Key;
+
+/// Wrapper for [`core::cmp::Reverse`] for using a [`BinaryHeap`](`super::BinaryHeap`) as a min-heap
```
cmichi, line 31: Nitpick, decide for yourself, but for consistency reasons I would appreciate having one line comments end with `.` (there are other instances in this file as well).

core/src/storage2/collections/binary_heap/tests.rs
```diff
@@ -0,0 +1,378 @@
...
+            expected_reads,
+            expected_writes,
+        )?;
+    }
+    Ok(())
+}
+
+#[cfg(feature = "ink-fuzz-tests")]
+#[quickcheck]
+fn pop_always_returns_largest_element(xs: Vec<i32>) {
```
cmichi, line 365: Awesome that you added a fuzz test!

core/src/storage2/collections/binary_heap/tests.rs
```diff
@@ -0,0 +1,378 @@
...
+fn drop_works() {
+    env::test::run_test::<env::DefaultEnvTypes, _>(|_| {
+        let root_key = Key::from([0x42; 32]);
+        {
+            let heap = heap_from_slice(&[23, 25, 65]);
+            SpreadLayout::push_spread(&heap, &mut KeyPtr::from(root_key));
+
+            let _ = <BinaryHeap<u8> as SpreadLayout>::pull_spread(&mut KeyPtr::from(
+                root_key,
+            ));
```
cmichi, line 233: This test could also panic here, resulting in a false positive. I would add an assertion after it to prevent this from happening.

core/src/storage2/collections/binary_heap/tests.rs
```diff
@@ -0,0 +1,378 @@
...
+        // write back to storage so we can see how many writes required
+        SpreadLayout::push_spread(&lazy_heap, &mut KeyPtr::from(root_key));
+
+        let (reads, writes) = env::test::get_contract_storage_rw::<env::DefaultEnvTypes>(
+            &contract_account,
+        )?;
+        let net_reads = reads - base_reads;
+        let net_writes = writes - base_writes;
+
+        // println!("{}: reads {}, writes {}", heap_size, net_reads, net_writes);
```
cmichi, line 293: Accidentally left in?

Robbepop (approved):
We will merge this now so @cmichi can rebase his optimization PR on top of the `master` branch.

PR #487, opened 2020-09-17
Migrate examples to new APIs

### LINKED ISSUE #485: Migrate examples to ink! 3.0
With ink! 3.0 we have a couple neat new storage APIs. The examples should be adapted to use those.

## ToDo

- [ ] Migrate examples to use the new `storage2::StorageHashMap` Entry API, introduced in #480 
- [ ] Migrate examples to use the new `storage2::Lazy::set(…)`, introduced in #466 
- [ ] Migrate examples to use the ops which were introduced for `storage2::Vec` in #453 
- [ ] If possible, migrate examples to use `storage2::Stash::remove_occupied(…)` introduced in #440

cmichi:
Closes #485 

Left these two ToDos out:
```
Migrate examples to use the ops which were introduced for storage2::Vec in #453
```
There are no examples which could use `Vec::set()` or `Vec::clear()` at the moment.

```
If possible, migrate examples to use storage2::Stash::remove_occupied(…) introduced in #440
```
The method is `unsafe` and we try to use only `unsafe` in examples when it's strictly necessary.

In terms of the new HashMap Entry API there are more places where we could use them ‒ at the cost of readability though.

### DIFF
--- a/examples/erc721/lib.rs
+++ b/examples/erc721/lib.rs
@@ -428,17 +428,8 @@ mod erc721 {
 
     /// Increase token counter from the `of` AccountId.
     fn increase_counter_of(entry: Entry<AccountId, u32>) -> Result<(), Error> {
-        match entry {
-            Entry::Occupied(mut occupied) => {
-                let count = occupied.get_mut();
-                *count += 1;
-                Ok(())
-            }
-            Entry::Vacant(vacant) => {
-                vacant.insert(1);
-                Ok(())
-            }
-        }
+        entry.and_modify(|v| *v += 1).or_insert(1);
+        Ok(())
     }
 
     /// Unit tests
@@ -756,8 +747,6 @@ mod erc721 {
         #[test]
         fn burn_fails_token_not_found() {
             run_test(|| {
-                let accounts = env::test::default_accounts::<env::DefaultEnvTypes>()
-                    .expect("Cannot get accounts");
                 // Create a new contract instance.
                 let mut erc721 = Erc721::new();
                 // Try burning a non existent token
--- a/examples/erc721/lib.rs
+++ b/examples/erc721/lib.rs
@@ -418,10 +418,10 @@ mod erc721 {
     }
 
     fn decrease_counter_of(
-        foo: &mut StorageHashMap<AccountId, u32>,
+        hmap: &mut StorageHashMap<AccountId, u32>,
         of: &AccountId,
     ) -> Result<(), Error> {
-        let count = (*foo).get_mut(of).ok_or(Error::CannotFetchValue)?;
+        let count = (*hmap).get_mut(of).ok_or(Error::CannotFetchValue)?;
         *count -= 1;
         Ok(())
     }
--- a/examples/erc721/lib.rs
+++ b/examples/erc721/lib.rs
@@ -69,7 +69,10 @@ use ink_lang as ink;
 #[ink::contract(version = "0.1.0")]
 mod erc721 {
     #[cfg(not(feature = "ink-as-dependency"))]
-    use ink_core::storage2::collections::HashMap as StorageHashMap;
+    use ink_core::storage2::collections::{
+        hashmap::Entry,
+        HashMap as StorageHashMap,
+    };
     use scale::{
         Decode,
         Encode,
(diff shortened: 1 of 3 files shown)


cmichi:
This is an example of a code snippet which would get more complicated using the new Entry API: https://github.com/paritytech/ink/blob/d4df4e7ebadc449fc6d63cebb115d546d3b57ca5/examples/erc721/lib.rs#L266-L274.

examples/multisig_plain/lib.rs
```diff
@@ -457,14 +457,9 @@ mod multisig_plain {
...
-                    if let Some(value) = self.confirmation_count.get_mut(&trans_id) {
-                        *value -= 1;
-                    } else {
-                        unreachable!()
-                    }
-                } else {
-                    self.confirmation_count.insert(trans_id, 0);
+                let v_mut = self.confirmation_count.entry(trans_id).or_insert(1);
+                if *v_mut > 0 {
+                    *v_mut -= 1;
```
Robbepop, line 462: Wouldn't this be the perfect opportunity to make use of https://doc.rust-lang.org/std/collections/btree_map/enum.Entry.html#method.and_modify instead of the `if`?

Robbepop:
```rust
self.confirmation_count
    .entry(trans_id)
    .and_modify(|v| if v > 0 { *v -= 1; })
    .or_insert(1);
```

Also this would allow us to remove the let binding from what I can see.

examples/multisig_plain/lib.rs
```diff
@@ -532,13 +527,7 @@ mod multisig_plain {
...
             transaction: TransactionId,
         ) -> ConfirmationStatus {
-            if !self.confirmation_count.contains_key(&transaction) {
-                self.confirmation_count.insert(transaction, 0);
-            }
-            let count = self
-                .confirmation_count
-                .get_mut(&transaction)
-                .expect("just inserted if missing");
+            let count = self.confirmation_count.entry(transaction).or_insert(0);
```
Robbepop, line 530: This change does not preserve semantics. Before the change we panicked upon not receiving a confirmation count and thus could tell if there was a bug and now we silently add a `0` to the confirmation count hash table which adds "gargabe" data.

examples/multisig_plain/lib.rs
```diff
@@ -598,13 +587,7 @@ mod multisig_plain {
...
             {
                 if self.confirmations.take(&(trans_id, *owner)).is_some() {
-                    if !self.confirmation_count.contains_key(&trans_id) {
-                        self.confirmation_count.insert(trans_id, 0);
-                    }
-                    *self
-                        .confirmation_count
-                        .get_mut(&trans_id)
-                        .expect("just inserted if missing") += 1;
+                    *self.confirmation_count.entry(trans_id).or_insert(0) += 1;
```
Robbepop, line 590: Same as above: Does not preserve semantics and is less "secure" by adding garbage to our storage instead of sanely panicking when things are messed up.

examples/erc721/lib.rs
```diff
@@ -420,6 +417,30 @@ mod erc721 {
...
+            Entry::Occupied(mut occupied) => {
+                let count = occupied.get_mut();
+                *count += 1;
+                Ok(())
+            }
+            Entry::Vacant(vacant) => {
+                vacant.insert(1);
+                Ok(())
+            }
+        }
```
Robbepop, line 441: Could be simplified by `and_modify` and `or_insert` calls.

Robbepop (changes requested):
All in all looks good but contains some non semantic preserving changes that should either be rewritten or removed.

examples/multisig_plain/lib.rs
cmichi: I think you missed the three lines before `let count = …` which I also replaced:
```
if !self.confirmation_count.contains_key(&transaction) {
    self.confirmation_count.insert(transaction, 0);
}
```
So my change is equivalent and this "garbage 0" was also added prior to it.

examples/multisig_plain/lib.rs
cmichi: See comment above, same three lines were replaced here.

examples/multisig_plain/lib.rs
Robbepop: yep, thanks! sorry

Robbepop (approved):
LGTM

PR #488, opened 2020-09-22
Deserialize metadata

ascjones:
Implements deserialization for the contract metadata, and expose fields via public getters. Required for https://github.com/paritytech/cargo-contract/pull/79.

## Public fields

**Edit** Previously this PR used public fields, it has now been converted to using public getters to enforce read-only.


Robbepop (commented):
All in all looks good, but one nit pick about hex-code serialization format for empty byte sequences remains.

Robbepop (approved):
LGTM

PR #493, opened 2020-09-28
Finalize ink! 3.0 integration

Robbepop:
## Finalization: ink! 3.0

This PR finalizes the implementation and integration of the new ink! 3.0.

## ToDo List

- [ ] Rework `ink_core/env` provided hashes
    - [ ] For every crypto hash provide two APIs:
        1. `hash_raw` to hash some raw input bytes
        2. `hash_encoded` to hash some SCALE encodable entity of type `T`
    - [ ] For every crypto hash provide a type that provides simple access to the above 2 APIs.
    - [ ] Remove the `ink_core/hash` module and move all linkers to the above definitions.
- [ ] Hide definitions in `ink_lang` using `#[doc(hidden)]` that are not user facing definitions but only required for the underlying code generation.
- [ ] **Experiment:** Try including `scale` and `scale-info` into `ink_lang` so that contract writers can simply use that one instead of having to depend on those crates directly themselves.
    - [ ] Adjust `cargo-contract new` template for the changes.
- [ ] Organizational
    - [ ] Move all ink! crates into `crates` directory.
    - [ ] Split `ink_core` into `ink_env` and `ink_storage` crate. Maybe rename `ink_env` to `ink_runtime`.
    - [ ] Rename `ink_prelude` to `ink_memory`, mirroring the planned `ink_storage` crate.
    - [x] Rename `ink_core/storage2` -> `storage`
    - [x] Rename `ink_metadata/layout2` -> `layout`
- [ ] Make ink! event topics work, finally: https://github.com/paritytech/ink/issues/105
    - [ ] Find a way to generate event topic implementations in a way that they do not require heap allocations but instead make use of ink!'s static 16kB buffer. This might imply providing yet another high-level `ink_core` API specific to this purpose.
    - Use BLAKE2 cryptographic hash.
    - Avoid BLAKE2 hashing in case the SCALE encoded values fit into `Env::Hash`.
    - Encode `topics[0]` as the event's hashed signature if the ink! event is not anonymous.
    - Since Seal only supports up to `MAX_EVENT_TOPICS` but cannot distinguish between existing special-cased `topics[0]` or not we need to make sure that non-anonymous events can actually only have up to `MAX_EVENT_TOPICS - 1` topics.
- [x] Write and convert all the acceptance tests for ink! 3.0.
- [x] Implement lint against `__ink_` prefixed identifiers generically to be usable by all ink! language macros.
- [x] Port `#[ink::test]` proc. macro to newer `ink_lang_ir` and `ink_lang_codegen` crates.
- [x] Update all ink! crates to actually use version `3.0.0`.
- [x] Make use of the new `ink_lang_ir` and `ink_lang_codegen` crates from `ink_lang_macro`. (Part 3)
- [x] Convert existing example smart contracts to ink! 3.0 syntax.
    - [x] Flipper
    - [x] Flipper (trait based)
    - [x] Incrementer
    - [x] ERC-20
    - [x] ERC-721
    - [x] Delegator
    - [x] Multi-Signature
    - [x] Domain Name Service (DNS)


examples/delegator/accumulator/lib.rs
```diff
@@ -17,30 +17,30 @@
 pub use self::accumulator::Accumulator;
 use ink_lang as ink;
 
-#[ink::contract(version = "0.1.0")]
-mod accumulator {
-    /// Holds a simple i32 value that can be incremented and decremented.
+#[ink::contract]
+pub mod accumulator {
+    // / Holds a simple i32 value that can be incremented and decremented.
```
cmichi, line 22:
```suggestion
    /// Holds a simple i32 value that can be incremented and decremented.
```

cmichi: I see there are way more occurrences of this `// /`. Is this intentional?

Robbepop: It is intentional since the doc comments screw up `cargo expand` output magically.

examples/dns/lib.rs
```diff
@@ -27,7 +27,7 @@ mod dns {
 
     /// Emitted whenever a new name is being registered.
     #[ink(event)]
-    struct Register {
+    pub struct Register {
```
cmichi, line 30: Do structs now have to be public when annotated with ` #[ink(event)]`?

examples/erc20/lib.rs
```diff
@@ -25,14 +25,14 @@ mod erc20 {
     };
 
     #[ink(storage)]
-    struct Erc20 {
+    pub struct Erc20 {
```
cmichi, line 28: Why public now?

Cargo.lock.old
```diff
@@ -0,0 +1,1355 @@
+# This file is automatically @generated by Cargo.
```
cmichi, line 1: Why is this file in here?

examples/delegator/accumulator/lib.rs
cmichi: Mhm, do I understand you right that the only purpose of this hack is to work around a bug in `cargo expand`?

Cargo.lock.old
Robbepop: oups :upside_down_face:

examples/dns/lib.rs
Robbepop: That's an interesting question actually: why would you think `#[ink(event)]` definitions should not required to be public?

examples/delegator/accumulator/lib.rs
Robbepop: It is rather a bug in the Rust compiler from what I know since it is having some hiccups with internal attributes. (Everything that is a `#![...]`).

lang/ir/src/ir/item_impl/callable.rs
```diff
@@ -362,12 +363,17 @@ pub(super) fn ensure_callable_invariants(
     method_item: &syn::ImplItemMethod,
     kind: CallableKind,
 ) -> Result<(), syn::Error> {
-    if !matches!(method_item.vis, syn::Visibility::Public(_) | syn::Visibility::Inherited)
-    {
-        return Err(format_err_spanned!(
-            method_item.vis,
+    let bad_visibility = match &method_item.vis {
```
ascjones, line 366: Can use `utils::ensure_pub_visibility`

metadata/src/specs.rs
```diff
@@ -321,6 +321,7 @@ pub struct MessageSpec<F: Form = MetaForm> {
     /// The parameters of the message.
     args: Vec<MessageParamSpec<F>>,
     /// The return type of the message.
+    #[serde(rename = "returnType")]
```
ascjones, line 324: Redundant because of (`#[serde(rename_all = "camelCase")]`)

ascjones (approved):
LGTM :taco:

examples/delegator/accumulator/lib.rs
cmichi: Ok, I see. Yeah I think it's rather ugly and goes against our "as close to Rust as possible" mantra. Hmm, not sure how to handle this. We could maybe track it in a follow-up issue?

cmichi (approved):
👏👏👏

lang/ir/src/ir/item_impl/callable.rs
Robbepop: Unfortunately we cannot make use of that API since the callables also allow inherited visibility.

examples/delegator/accumulator/lib.rs
Robbepop: We actually cannot fix this. This needs to be fixed in the Rust compiler.

PR #499, opened 2020-09-30
Implement new crypto hash API

Robbepop:
This is going to replace the current `ink_core::env` provided cryptographic hashing procedures.
Also this is going to deprecate the `ink_core::hash` module entirely.

The main reason for the integration of `hash_encoded` into `ink_core::env` is to allow encoded hashing without using the heap allocator for our on-chain environment which is important for proper event topics support.

# Implementation Hurdles

All in all the implementation of the new hash API and the replacement of the old hash module was not very difficult. However, there was one pitfall with `ink_core::storage::alloc::DynamicAllocation::key` implementation. As soon as we based it onto the new `ink_core::env::hash_encoded` or `hash_bytes` API some tests would fail indicating some double `RefCell` mutable borrows. I don't know exactly why this happened but I assume it is due to some `Drop` code calling things in an order that is uncommon in normal user code.
I fixed this by basing the implementation on `CryptoHash::hash` internals which isn't too bad either.

# Screenshots

## CryptoHash trait definition

2020-09-30-140604_1117x855_scrot

## The new hash_encoded API

2020-09-30-161052_1266x631_scrot

## The new hash_bytes API

2020-09-30-140418_1822x345_scrot


ascjones (approved):
I think this API is much nicer.

Robbepop:
> I think this API is much nicer.

I too think so. The new API builds on top of the `ScopedBuffer` API which did not exist while I designed the old hashing API.

PR #503, opened 2020-10-02
Debug codecov

TriplEight:
# investigating why coverage misses the entire files

putting reports to the artifacts output for the further analysis.

maybe it's due to some syntax cov tool can't parse

PR where it has initially started  https://github.com/paritytech/ink/pull/487

the massive loss appears here https://codecov.io/gh/paritytech/ink/pull/487/tree/lang/macro/src/codegen , all files to 0%

For the most up-to-date coverage report all these files are not covered even though we definitely extensively test them:
https://codecov.io/gh/paritytech/ink/tree/4770498c95e07e50cf7f64fb7ad33d05ee43d537/crates/lang/codegen/src

And these are actually unit tests (so should be 100%) and showing just single digit coverage. (Actually they should be excluded from coverage testing)
https://codecov.io/gh/paritytech/ink/tree/4770498c95e07e50cf7f64fb7ad33d05ee43d537/crates/lang/macro/tests/ui/pass

All of these also are covered with tests:
https://codecov.io/gh/paritytech/ink/tree/4770498c95e07e50cf7f64fb7ad33d05ee43d537/crates/lang/macro/src

reference to how to report to codecov: https://docs.codecov.io/docs/unexpected-coverage-changes

reference on how to read reports:  http://ltp.sourceforge.net/coverage/lcov/geninfo.1.php


TriplEight:
Closes paritytech/devops#697

Robbepop:
Is this on ice or can it be closed? I do not see that it improves the situation with ink! test coverage. What are the problems? How can we tackle those?

TriplEight:
This PR is needed, just not as burning as some others. And frankly, it's not clear what to do here.

TriplEight:
Can close it since I can't rise the priority for it right now, but will mark it as unfinished like this: https://github.com/paritytech/ci_cd/issues/13

PR #504, opened 2020-10-02
Add proc. macro documentation

Robbepop:
Adds missing documentation for all of the proc. macro provided by the `ink_lang_macro` crate.

## Screenshots

Some screenshots of the new rendered documentation below:

2020-10-02-190324_1303x1108_scrot

2020-10-02-190248_970x1274_scrot

2020-10-02-190213_1289x1259_scrot

2020-10-02-190150_1961x373_scrot

### DIFF
--- a/crates/env/src/types.rs
+++ b/crates/env/src/types.rs
@@ -216,6 +216,16 @@ pub trait Clear {
     fn clear() -> Self;
 }
 
+impl Clear for [u8; 32] {
+    fn is_clear(&self) -> bool {
+        self.as_ref().iter().all(|&byte| byte == 0x00)
+    }
+
+    fn clear() -> Self {
+        [0x00; 32]
+    }
+}
+
 impl Clear for Hash {
     fn is_clear(&self) -> bool {
         self.as_ref().iter().all(|&byte| byte == 0x00)
--- a/crates/env/src/types.rs
+++ b/crates/env/src/types.rs
@@ -228,10 +228,10 @@ impl Clear for [u8; 32] {
 
 impl Clear for Hash {
     fn is_clear(&self) -> bool {
-        self.as_ref().iter().all(|&byte| byte == 0x00)
+        <[u8; 32] as Clear>::is_clear(&self.0)
     }
 
     fn clear() -> Self {
-        Self([0x00; 32])
+        Self(<[u8; 32] as Clear>::clear())
     }
 }
--- a/crates/lang/macro/src/lib.rs
+++ b/crates/lang/macro/src/lib.rs
@@ -284,6 +284,9 @@ use proc_macro::TokenStream;
 ///     Authors of ink! smart contracts can make an ink! message payable by adding the `payable`
 ///     flag to it. An example below:
 ///
+///     Note that ink! constructors are always implicitely payable and thus cannot be flagged
+///     as such.
+///
 ///     ```
 ///     # use ink_lang as ink;
 ///     # #[ink::contract]
--- a/crates/lang/macro/src/lib.rs
+++ b/crates/lang/macro/src/lib.rs
@@ -23,6 +23,76 @@ pub fn contract(attr: TokenStream, item: TokenStream) -> TokenStream {
     contract::generate(attr.into(), item.into()).into()
 }
 
+/// Marks trait definitions to ink! as special ink! trait definitions.
+///
+/// There are some restrictions that apply to ink! trait definitions that
+/// this macro checks. Also ink! trait definitions are required to have specialized
+/// structure so that the main [`#[ink::contract]`](`macro@crate::contract`) macro can
+/// properly generate code for its implementations.
+///
+/// # Example: Definition
+///
+/// ```
+/// use ink_lang as ink;
+/// # type Balance = <ink_env::DefaultEnvTypes as ink_env::EnvTypes>::Balance;
+///
+/// #[ink::trait_definition]
+/// pub trait Erc20 {
+///     /// Constructors a new ERC-20 compliant smart contract using the initial supply.
+///     #[ink(constructor)]
+///     fn new(initial_supply: Balance) -> Self;
+///
+///     /// Returns the total supply of the ERC-20 smart contract.
+///     #[ink(message)]
+///     fn total_supply(&self) -> Balance;
+///
+///     // etc.
+/// }
+/// ```
+///
+/// # Example: Implementation
+///
+/// Given the above trait definition you can implement it as shown below:
+///
+/// ```
+/// # use ink_lang as ink;
+/// #
+/// #[ink::contract]
+/// mod base_erc20 {
+/// #    // We somehow cannot put the trait in the doc-test crate root due to bugs.
+/// #    #[ink_lang::trait_definition]
+/// #    pub trait Erc20 {
+/// #        /// Constructors a new ERC-20 compliant smart contract using the initial supply.
+/// #        #[ink(constructor)]
+/// #        fn new(initial_supply: Balance) -> Self;
+/// #
+/// #        /// Returns the total supply of the ERC-20 smart contract.
+/// #        #[ink(message)]
+/// #        fn total_supply(&self) -> Balance;
+/// #    }
+/// #
+///     #[ink(storage)]
+///     pub struct BaseErc20 {
+///         total_supply: Balance,
+///         // etc ..
+///     }
+///
+///     impl Erc20 for BaseErc20 {
+///         #[ink(constructor)]
+///         fn new(initial_supply: Balance) -> Self {
+///             Self { total_supply: initial_supply }
+///         }
+///
+///         /// Returns the total supply of the ERC-20 smart contract.
+///         #[ink(message)]
+///         fn total_supply(&self) -> Balance {
+///             self.total_supply
+///         }
+///
+///         // etc ..
+///     }
+/// }
+/// ```
 #[proc_macro_attribute]
 pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
     trait_def::analyze(attr.into(), item.into()).into()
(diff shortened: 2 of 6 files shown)


crates/lang/macro/src/lib.rs
```diff
@@ -18,20 +18,600 @@ mod trait_def;
...
+///     ```
+///
+///     **Default value:** `false`
+///
+/// - `compile_as_dependency: bool`
+///
+///     Tells the ink! code generator to always or never
+///     compile the smart contract as if it was used as a dependency of another ink!
+///     smart contract.
+///     Normally this flag is only really useful for ink! developers of people that
```
cmichi, line 95:
```suggestion
///     Normally this flag is only really useful for ink! developers that
```

crates/lang/macro/src/lib.rs
```diff
@@ -18,20 +18,600 @@ mod trait_def;
...
+///         # #[ink(message)]
+///         # pub fn message(&self) {}
+///     }
+///     # }
+///     ```
+///
+/// - There must be at least one `#[ink(message)]` defined method.
+///
+///     Methods flagged with `#[ink(message)]` are special in that they are dispatchable
+///     upon contract invokation. The set of ink! messages defined for an ink! smart contract
```
cmichi, line 238:
```suggestion
///     upon contract invocation. The set of ink! messages defined for an ink! smart contract
```

crates/lang/macro/src/lib.rs
```diff
@@ -18,20 +18,600 @@ mod trait_def;
...
+///         pub fn get(&self) -> bool {
+///             self.value
+///         }
+///     }
+///     # }
+///     ```
+///
+///     **Payable Messages:**
+///
+///     An ink! message by default will reject calls that additional fund the smart contract.
```
cmichi, line 283: I don't get this sentence, do you mean "that additionally fund the smart contract"?

crates/lang/macro/src/lib.rs
```diff
@@ -18,20 +18,600 @@ mod trait_def;
...
+///
+///         # /// Flips the current value.
+///         # #[ink(message)]
+///         # #[ink(payable)] // You can either specify payable out-of-line.
+///         # pub fn flip(&mut self) {
+///         #     self.value = !self.value;
+///         # }
+///         #
+///         /// Returns the current value.
+///         #[ink(message, selector = "0xFEEDBEEF")] // ... or specify payable inline.
```
cmichi, line 348:
```suggestion
///         #[ink(message, selector = "0xFEEDBEEF")] // ... or specify selector inline.
```

crates/lang/macro/src/lib.rs
```diff
@@ -18,20 +18,600 @@ mod trait_def;
...
+///     impl Flipper {
+///         #[ink(constructor)]
+///         #[ink(selector = "0xDEADBEEF")]
+///         pub fn new(initial_value: bool) -> Self {
+///             Flipper { value: false }
+///         }
+///
+///         # /// Flips the current value.
+///         # #[ink(message)]
+///         # #[ink(payable)] // You can either specify payable out-of-line.
```
cmichi, line 342:
```suggestion
///         # #[ink(selector = "0xFEEDBEEF")] // You can either specify selector out-of-line.
```

crates/lang/macro/src/lib.rs
```diff
@@ -18,20 +18,600 @@ mod trait_def;
...
-
+/// Defines a unit test that makes use of ink!'s off-chain testing capabilities.
+///
+/// If your unit test does not require the existence of an off-chain environment
+/// it is fine to not use this macro since it bears some overhead with the test.
+///
+/// Note that this macro is not required to run unit tests that require ink!'s
+/// off-chain testing capabilities but merely improves code readability.
+///
+/// ## How do you find out if your test requires the off-chain environment?
```
cmichi, line 579: I'm not sure if we overcomplicate everything by making the distinction between `[ink::test]` and `#[test]`.

It comes down to a trade-off: performance vs. ease-of-use. I tend towards ease-of-use since the overhead should not be that big if people just use `[ink::test]` everywhere (haven't measured it though). We could get rid of this entire distinction above then.

What do you think?

crates/lang/macro/src/lib.rs
Robbepop: yes, good catch!

crates/lang/macro/src/lib.rs
Robbepop: I totally get your point. I think it is just the purist in me trying to make people really understand the difference. If we don't tell them in the first place I fear that confusion might be bigger.

crates/lang/macro/src/lib.rs
```diff
@@ -18,20 +18,600 @@ mod trait_def;
...
+///
+/// The `#[ink::contract]` macro fully analyses its input smart contract
+/// against invalid arguments and structure.
+///
+/// Some example rules include but are not limited to:
+///
+/// - There must be exactly one `#[ink(storage)]` struct.
+///
+///     This struct defined the layout of the storage that the ink! smart contract operates on.
+///     The user is able to use a variety of built-in facitilies, combine them in various way
```
ascjones, line 181:
```suggestion
///     The user is able to use a variety of built-in facilities, combine them in various ways
```

Maybe facilities could be replaced by `storage data structures`, but that does appear later in the sentence

crates/lang/macro/src/lib.rs
```diff
@@ -18,20 +18,600 @@ mod trait_def;
...
+///     or even provide their own implementations of storage data structures.
+///
+///     For more information the user shall visit the `ink_storage` crate documentation.
```
ascjones, line 184:
```suggestion
///     For more information visit the `ink_storage` crate documentation.
```
Would be great to have a doc link here too

crates/lang/macro/src/lib.rs
```diff
@@ -18,20 +18,600 @@ mod trait_def;
...
+///
+/// # Example: Definition
+///
+/// ```
+/// use ink_lang as ink;
+/// # type Balance = <ink_env::DefaultEnvTypes as ink_env::EnvTypes>::Balance;
+///
+/// #[ink::trait_definition]
+/// pub trait Erc20 {
+///     /// Constructors a new ERC-20 compliant smart contract using the initial supply.
```
ascjones, line 511:
```suggestion
///     /// Constructs a new ERC-20 compliant smart contract using the initial supply.
```

ascjones (approved):
Sieht gut aus

cmichi (approved):
Looks good in general, some nitpicks regarding clarity and language.

crates/lang/macro/src/lib.rs
cmichi: I don't yet fully get it, could you elaborate on what it means that a call additionally funds the smart contract?

Robbepop: When you submit a transaction to call a smart contract you can not only set gas limit but also how much actual value you want to transfer to the call.

crates/env/src/types.rs
```diff
@@ -216,6 +216,16 @@ pub trait Clear {
     fn clear() -> Self;
 }
 
+impl Clear for [u8; 32] {
+    fn is_clear(&self) -> bool {
+        self.as_ref().iter().all(|&byte| byte == 0x00)
```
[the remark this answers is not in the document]
Robbepop: yes it should!

crates/lang/macro/src/lib.rs
```diff
@@ -18,20 +18,600 @@ mod trait_def;
...
+///
+/// - `compile_as_dependency: bool`
+///
+///     Tells the ink! code generator to always or never
+///     compile the smart contract as if it was used as a dependency of another ink!
+///     smart contract.
+///     Normally this flag is only really useful for ink! developers of people that
+///     want to inspect code generation of ink! smart contracts.
+///     The author is not aware of any particular practical use case for users that
+///     makes use of this flag.
```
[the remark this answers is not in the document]
Robbepop: can you make a suggestions? :sweat_smile: Or tell me what is unclear to you?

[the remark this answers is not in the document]
ascjones: Isn't this flag used for when cross-calling contracts, the callee contract is compiled as a dependency? Although is it intended that traits should be used in this case?

Anyway it isn't clear from this comment what the flag is actually for, only that there is no "practical use case". As a user I would not be further enlightened.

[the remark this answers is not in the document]
Robbepop: Well maybe there are use cases that we don't know yet but normally you compile ink! smart contracts as dependency through their crate features `ink-as-dependency` and not via this flag because you do not necessarily always want to compile them as dependency. Or maybe you do? I want to leave this question open for further research. After all using ink! smart contracts yields many discoveries.

crates/lang/macro/src/lib.rs
```diff
@@ -89,13 +94,18 @@ use proc_macro::TokenStream;
...
+///
 ///     Normally this flag is only really useful for ink! developers who
 ///     want to inspect code generation of ink! smart contracts.
 ///     The author is not aware of any particular practical use case for users that
-///     makes use of this flag.
+///     makes use of this flag but contract writers are encouraged to disprove this.
+///
+///     Note that it is recommended to make use of the built-in crate feature
+///     `ink-as-dependency` to flag smart contract dependencies listed in a contract's
+///     `Cargo.toml` as actual dependencies to ink! so that ink!.
```
ascjones, line 108: so that ink!....what?

crates/lang/macro/src/lib.rs
```diff
@@ -284,6 +298,9 @@ use proc_macro::TokenStream;
 ///     Authors of ink! smart contracts can make an ink! message payable by adding the `payable`
 ///     flag to it. An example below:
 ///
+///     Note that ink! constructors are always implicitely payable and thus cannot be flagged
```
ascjones, line 301:
```suggestion
///     Note that ink! constructors are always implicitly payable and thus cannot be flagged
```

crates/lang/macro/src/lib.rs
Robbepop: ouf thanks

crates/lang/macro/src/lib.rs
Robbepop: i will probably never learn this word

PR #509, opened 2020-10-05
[ci] Iterate through trybuild versions

cmichi:


cmichi:
Found no difference in coverage.

PR #510, opened 2020-10-07
Implement event topic code generation

Robbepop:
**NOTE:** From what I can tell CI is failing due to a recent change in error message displays of the Rust compiler.
Minified code example for better understanding of this PR's techniques:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=be091af94b8407d122d4a092ea4cbb09

The most important thing to review in this PR is the implementation of the on-chain backend:
https://github.com/paritytech/ink/blob/robin-implement-topics/crates/env/src/engine/on_chain/impls.rs#L130

The challenge was to find a design that does not make use of any of the following:
- heap memory allocations
- dynamic dispatch (e.g. of `Encode` trait)

Also ideally the implementation should be as low-overhead as possible.
I am not sure if there is yet a better and more efficient way to implement this but at least we got topics in now.
The current implementation is more complex than I wish it was, happy for comments to reduce complexity before we merge if possible without losing upon any of the properties mentioned above.

The implementation makes heavy use of the scoped buffer provided by the on-chain environment.
For this purpose this PR adds some additional APIs to it: `append_encoded` and `take_appended`
These new APIs allow us to incrementally build up the encoding of the event topics.

Note that we do not encode event topics via `Vec<Hash>` but we do the same encoding manually guided by the event itself but we secure at the same time that the resulting encodings are the same as if it was directly encoded as `Vec<Hash>`: `Compact<u32> + N * Hash`

## ToDo List

- [x] Add test case for ensuring correct encoded topics.
- [x] Include event signature (`"ContractName::EventName"`) - optionally hashed - as first argument to the event topics.
- [x] Think about adding anonymous event definitions in order to prevent the above overhead.

Due to the upfront complexity of our definitions in the `ink_env` crate we are able to minimize the generated code for our event topic implementations:

ERC-20: `BaseEvent` impl

```rust
const _: () = {
    pub enum __ink_UndefinedAmountOfTopics {}
    impl ::ink_env::topics::EventTopicsAmount for __ink_UndefinedAmountOfTopics {
        const AMOUNT: usize = 0;
    }
    #[cfg(not(feature = "ink-as-dependency"))]
    impl ::ink_env::Topics for __ink_EventBase {
        type RemainingTopics = __ink_UndefinedAmountOfTopics;
        fn topics<E, B>(
            &self,
            builder: ::ink_env::topics::TopicsBuilder<::ink_env::topics::state::Uninit, E, B>,
        ) -> <B as ::ink_env::topics::TopicsBuilderBackend<E>>::Output
        where
            E: ::ink_env::EnvTypes,
            B: ::ink_env::topics::TopicsBuilderBackend<E>,
        {
            match self {
                Self::Transfer(event) => {
                    <Transfer as ::ink_env::Topics>::topics::<E, B>(event, builder)
                }
                Self::Approval(event) => {
                    <Approval as ::ink_env::Topics>::topics::<E, B>(event, builder)
                }
            }
        }
    }
};
```

And ERC-20 `Transfer` event impl

```rust
#[cfg(not(feature = "ink-as-dependency"))]
const _: () = {
    impl ::ink_env::Topics for Transfer {
        type RemainingTopics = [::ink_env::topics::state::HasRemainingTopics; 3usize];
        fn topics<E, B>(
            &self,
            builder: ::ink_env::topics::TopicsBuilder<::ink_env::topics::state::Uninit, E, B>,
        ) -> <B as ::ink_env::topics::TopicsBuilderBackend<E>>::Output
        where
            E: ::ink_env::EnvTypes,
            B: ::ink_env::topics::TopicsBuilderBackend<E>,
        {
            builder
                .build::<Self>()
                .push_topic::<Option<AccountId>>(&self.from)
                .push_topic::<Option<AccountId>>(&self.to)
                .push_topic::<Balance>(&self.value)
                .finish()
        }
    }
};
```


crates/env/src/topics.rs
```diff
@@ -0,0 +1,204 @@
...
+            state: Default::default(),
+        }
+    }
+}
+
+#[doc(hidden)]
+pub mod state {
+    /// The topic builder is uninitialized and needs to be provided with the
+    /// expected number of topics that need to be constructed.
+    pub enum Uninit {}
```
athei, line 69: Just a matter of taste. I find those empty braces ugly.

```suggestion
    pub struct Uninit;
```

crates/env/src/topics.rs
```diff
@@ -0,0 +1,204 @@
...
+where
+    E: EnvTypes,
+    B: TopicsBuilderBackend<E>,
+{
+    /// Initializes the topics builder and informs it about how many topics it must expect to serialize.
+    ///
+    /// The number of expected topics is given implicitely by the `E` type parameter.
+    pub fn build<Event: Topics>(
+        mut self,
+    ) -> TopicsBuilder<<Event as Topics>::RemainingTopics, E, B> {
```
athei, line 86: I don't get how this works. `<Event as Topics>::RemainingTopics` does not implement `SomeRemainingTopics`.

crates/env/src/topics.rs
Robbepop: There is a semantics difference. You can have instances of `struct Uninit;` but you cannot have live instances of `enum Uninit {}`. So if you want to use those types on a purely type level it is cleaner to have them as empty enums.

crates/env/src/topics.rs
Robbepop: It does not need to. That is only going to be important for the `push_topics` method.
With this abstraction we simply can make sure that we do not specify more event topics than there actually are. However, we do all this on a purely type level.

athei: But how is a `TopicsBuilder` to which the impl block with `push_topics` applies created? Thought it should be created through `build`.

crates/env/src/topics.rs
```diff
@@ -0,0 +1,204 @@
...
+// limitations under the License.
+
+//! Docs
+
+use crate::EnvTypes;
+
+/// The concrete implementation that is guided by the topics builder.
+///
+/// To be implemented by the on-chain and off-chain environments respectively.
+#[doc(hidden)]
```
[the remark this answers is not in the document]
Robbepop: True but I think marking each and every component that shall be hidden within such a small module increases readability since it restates the intent of the hidden entities locally. For example the `Topics` trait is not hidden and maybe some others will follow. If they were not hidden some developer in the future might think - "oh damn, we forgot to re-export those to the crate level, let's do that!". And that would be bad.

crates/env/src/topics.rs
Robbepop: Not sure I get your question but the on/off chain environment first initialize the `TopicsBuilder` in an uninitialized state. Then, through `Topics` trait it gets first initialized with the amount of expected topics and with every call to `push_topic` a type-based counter is decreased so that exactly `N` topics are going to be serialized before the implementer of `Topics` can finally call `finish` to yield the result back to the environment for doing the rest of the serialization efforts.

Robbepop: Note that this type state approach was taken as an extra safety check between environment implementation and ink! codegen and to also eliminate some assertions that would have been necessary otherwise.

ascjones (approved):
LGTM. 

I would like to see at least one simple sanity test that the topics are produced, even in one of the example contracts e.g. https://github.com/paritytech/ink/blob/cfe52e98667a85f3ebb94ee7e23a05823f7d3872/examples/erc20/lib.rs#L166-L184

crates/env/src/topics.rs
athei: I see. Makes sense.

crates/env/src/topics.rs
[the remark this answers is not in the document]
athei: Makes sense.

crates/env/src/topics.rs
athei: Thanks for the explainer. I got it now.

crates/env/src/engine/on_chain/buffer.rs
```diff
@@ -106,19 +109,31 @@ impl<'a> scale::Output for EncodeScope<'a> {
...
+            offset: 0,
+            buffer: &mut self.buffer[self.offset..],
+        }
+    }
+
     /// Returns the first `len` bytes of the buffer as mutable slice.
     pub fn take(&mut self, len: usize) -> &'a mut [u8] {
-        assert!(len <= self.buffer.len());
+        debug_assert_eq!(self.offset, 0);
+        debug_assert!(len <= self.buffer.len());
```
cmichi, line 136: Nice to see all those `debug_assert`s!

crates/env/src/topics.rs
```diff
@@ -0,0 +1,204 @@
...
+where
+    E: EnvTypes,
+    B: TopicsBuilderBackend<E>,
+{
+    /// Finalizes the topics builder.
+    ///
+    /// No more event topics can be serialized afterwards but the environment will be
+    /// able to extract the information collected by the topics builder in order to
+    /// emit the serialized event.
+    // pub fn finish(self) -> TopicsBuilder<state::Finish, B>
```
cmichi, line 131: Leftover comment.
```suggestion
```

crates/lang/codegen/src/generator/events.rs
```diff
@@ -203,16 +217,50 @@ impl<'a> Events<'a> {
     fn generate_topics_impls(&'a self) -> impl Iterator<Item = TokenStream2> + 'a {
         let no_cross_calling_cfg =
             self.generate_code_using::<generator::CrossCallingConflictCfg>();
+        // let contract_ident = self.contract.module().storage().ident();
```
cmichi, line 220: Leftover comment.
```suggestion
```

Robbepop:
> LGTM.
> 
> I would like to see at least one simple sanity test that the topics are produced, even in one of the example contracts e.g.
> 
> https://github.com/paritytech/ink/blob/cfe52e98667a85f3ebb94ee7e23a05823f7d3872/examples/erc20/lib.rs#L166-L184

good point! i added some elaborate tests to ERC-20 example contract.

crates/env/src/topics.rs
Robbepop: cannot apply your suggestion somehow :S

cmichi (approved):
Thanks for the ERC-20 tests!

ascjones (approved):

PR #512, opened 2020-10-08
Pre pop cargo home cache

TriplEight:
`CARGO_HOME` cache was moved to the same "project/cache_type/branch_name/job_name" level as
`CARGO_TARGET_DIR` because of frequent weird data-race issues. This just means that the same cache that 
would have been used for the entire pipeline will be duplicated for each job.


scripts/pre_cache.sh
```diff
@@ -5,16 +5,35 @@ set -u
...
+#     -type d -name "$CI_JOB_NAME"  -exec stat --printf="%Y\t%n\n" {} \; |sort -n -r |head -1 |cut -f2);
+#   if [[ -d $FRESH_TARGET_CACHE ]]; then
+#     echo "____Using" "$FRESH_TARGET_CACHE" "to prepopulate the cache____";
+#     time cp -r "$FRESH_TARGET_CACHE" "$CARGO_TARGET_DIR";
+#   else
+#     echo "_____No such targets dir, proceeding from scratch_____";
+#   fi
+# else
+#   echo "____No need to prepopulate CARGO_TARGET_DIR cache____";
+# fi
```
Robbepop, line 20: Happy to merge when this is removed.

Robbepop (approved):
LGTM

PR #515, opened 2020-10-09
Prepare for ink! 3.0-rc1 release

Robbepop:
- [x] Change version of all ink! 3.0 crates to `3.0.0-rc1`.
- [x] Add release notes for ink! 3.0 release.
- [x] Making sure that all ink! crates can potentially be deployed to crates.io.
    - A `--dry-run` looked promising already.

Rendered release notes: https://github.com/paritytech/ink/blob/robin-prepare-for-ink3.0-rc1/RELEASES.md


RELEASES.md
```diff
@@ -1,3 +1,436 @@
+# Version 3.0 (2020-10-09)
+
+Be prepared for the ink! 3.0 release notes because the whole version was basically a rewrite of
+all the major components that make up ink!. With our experience gained by the last released versions
+of ink! we were able to detect weak spots of the design and provided ink! with more tools,
+more features and more efficiency as ever. Read more below ...
```
cmichi, line 6:
```suggestion
more features and more efficiency as ever. Read more below…
```

Robbepop: is that how you actually have to do it according to typists?

RELEASES.md
```diff
@@ -1,3 +1,440 @@
...
+
+## Just. Be. Rust. 3.0
+
+In the 3.0 update we further explored the space for ink! to just feel like it was plain Rust.
+With this regard we changed the syntax slightly in order to better map from ink! to the generated
```
ascjones, line 11:
```suggestion
With this in mind we changed the syntax slightly in order to better map from ink! to the generated
```

RELEASES.md
```diff
@@ -1,3 +1,440 @@
...
+The storage module has been reworked entirely.
+Also it no longer lives in the `ink_core` crate but instead is defined as its own `ink_storage` crate.
+
+In a sense it acts as the standard storage library for ink! smart contracts in that it provides all the
+necessary tools and data structures to organize and operate the contract's storage intuitively and efficiently.
+
+### Lazy
+
+The most fundamental change in how you should think about data structures provided by the new `ink_storage`
+crate is that they are inherently lazy.
```
ascjones, line 59: I can relate

ascjones: Would be good to define what we mean by `lazy` near the top here in case it is not immediately obvious. 

(I see that you do a bit further down :point_down:)

RELEASES.md
```diff
@@ -1,3 +1,440 @@
...
+```
+Now `a` and `b` are only loaded when the contract really needs their values.
+Note that `offset` remained `i32` since it is always needed and could spare the minor overhead of the `ink_storage::Lazy` wrapper.
+
+### HashMap
+
+In the follow we explore the differences between the high-level `ink_storage::collections::HashMap`
+and the low-level `ink_storage::lazy::LazyHashMap`. Both provide very similar functionality in that they map some generic key to some storage entity.
+
+However, their APIs look very differently. Whereas the `HashMap` provides a rich and high-level API that is comparable to that of Rust's very own `HashMap`, the `LazyHashMap` provides only a fraction of the API and also operates on `Option<T>` values types instead of `T` directly. It is more similar Solidity mappings than to Rust's `HashMap`.
```
ascjones, line 119:
```suggestion
However, their APIs look very different. Whereas the `HashMap` provides a rich and high-level API that is comparable to that of Rust's very own `HashMap`, the `LazyHashMap` provides only a fraction of the API and also operates on `Option<T>` values types instead of `T` directly. It is more similar Solidity mappings than to Rust's `HashMap`.
```

RELEASES.md
```diff
@@ -1,3 +1,440 @@
...
+| `HashMap<K, V>` | high-level | yes | yes | `V` (key type `K`) | Rust map-like container |
+| `LazyHashMap<K, V>` | low-level | yes | yes | `Option<V>` (key type `K`) | similar to Solidity mapping |
+
+There are many more! For more information about the specifics please take a look into [the `ink_storage` crate documentation](https://paritytech.github.io/ink/ink_storage/).
+
+### Spread & Packed Modes
+
+Storing or loading complex data structures to and from contract storage can be done in many different ways. You could store all information into a single storage cell or you could try to store all information into as many different cells as possible. Both strategies have pros and cons under different conditions.
+
+For example it might be a very good idea to store all the information under the same cell if all the information is very compact. For example when we are dealing with a byte vector that is expected to never be larger than approx a tousand elements it would probably be more efficient if we store all those tousand bytes in the same cell and especially if we often access many of those (or all) in our contract messages.
```
ascjones, line 139:
```suggestion
For example it might be a very good idea to store all the information under the same cell if all the information is very compact. For example when we are dealing with a byte vector that is expected to never be larger than approx a thousand elements it would probably be more efficient if we store all those thousand bytes in the same cell and especially if we often access many of those (or all) in our contract messages.
```

RELEASES.md
cmichi: Yup: https://en.wikipedia.org/wiki/Ellipsis

cmichi: I mean it's really just a nitpick, but for me in the same class as using "->" vs "➝", so yeah just decide for yourself ;-).

RELEASES.md
```diff
@@ -1,3 +1,440 @@
...
+
+On the other handside spreading information to as many cells as possible might be much more efficient if we are dealing with big data structures or a lot of information that is not compact or when messages that operate on the data always only need a small fraction of the whole data.
```
ascjones, line 141:
```suggestion
On the other hand spreading information across as many cells as possible might be much more efficient if we are dealing with big data structures, a lot of information that is not compact, or when messages that operate on the data always only need a small fraction of the whole data.
```
Sentence is a bit long, have added some commas

RELEASES.md
Robbepop: Advises a space between "below` and the three dots though: Read more below …

ascjones (commented):
WIP

RELEASES.md
Robbepop: It is good to have a reviewer that knows where to place commas in english sentences. :D

RELEASES.md
Robbepop: Hmmm, I think it might be enough to state that we will explain what this means below.

Robbepop: done

RELEASES.md
```diff
@@ -1,3 +1,440 @@
...
+```rust
+use ink_storage::Vec as StorageVec;
+
+#[ink(storage)]
+pub struct Matrix {
+    values: StorageVec<StorageVec<i32>>,
+}
+```
+
+However, this will fail compilation with an error indicating that `StorageVec<T>` requires for its `T` to be packed (`T: PackedLayout`) which `StorageVec<T>` itself does not since it stores all of its elements into different cells always. The same applies to many other storage data sturctures provided by `ink_storage` and is a trade-off the ink! team decided for the case of efficiency of the overall system.
```
ascjones, line 246:
```suggestion
However, this will fail compilation with an error indicating that `StorageVec<T>` requires for its `T` to be packed (`T: PackedLayout`) which `StorageVec<T>` itself does not since it always stores all of its elements into different cells. The same applies to many other storage data sturctures provided by `ink_storage` and is a trade-off the ink! team decided for the case of efficiency of the overall system.
```

RELEASES.md
```diff
@@ -1,3 +1,440 @@
...
+Instead what a user can do in order to get their vector-of-vector to be working is making use of ink!'s dynamic storage allocator capabilities.
```
ascjones, line 247:
```suggestion
Instead what a user can do in order to get their vector-of-vector to be working is to make use of ink!'s dynamic storage allocator capabilities.
```

RELEASES.md
Robbepop: done, thanks

RELEASES.md
```diff
@@ -1,3 +1,440 @@
...
+| `#[ink(event)]` | On `struct` definitions. | Defines an ink! event. A contract can define multiple such ink! events. |
+| `#[ink(anonymous)]` **new** | Applicable to ink! events. | Tells the ink! codegen to treat the ink! event as anonmyous which omits the event signature as topic upon emitting. Very similar to anonymous events in Solidity. |
+| `#[ink(topic)]` | Applicate on ink! event field. | Tells the ink! codegen to provide a topic hash for the given field. Every ink! event can only have a limited number of such topic field. Similar semantics as to indexed event arguments in Solidity. |
+| `#[ink(message)]` | Applicable to methods. | Flags a method for the ink! storage struct as message making it available to the API for calling the contract. |
+| `#[ink(constructor)]` | Applicable to method. | Flags a method for the ink! storage struct as constructor making it available to the API for instantiating the contract. |
+| `#[ink(payable)]` **new** | Applicable to ink! messages. | Allow transactions while calling this ink! message. ink! constructors are implicitely payable. |
```
ascjones, line 325: Can be more specific than "transaction", something about receiving value as part of the call

Robbepop: done

ascjones (approved):
LGTM

PR #523, opened 2020-10-16
[examples] Add ERC-20 Trait-ified example

cmichi:
…and migrate the existing ERC-20 example to use proper Rust `Result` types instead of returning `bool`.

### DIFF
--- a/examples/erc20/lib.rs
+++ b/examples/erc20/lib.rs
@@ -152,7 +152,7 @@ mod erc20 {
         }
 
         fn balance_of_or_zero(&self, owner: &AccountId) -> Balance {
-            *self.balances.get(owner).unwrap_or(&0)
+            self.balances.get(owner).copied().unwrap_or(0)
         }
 
         fn allowance_of_or_zero(
@@ -160,7 +160,10 @@ mod erc20 {
             owner: &AccountId,
             spender: &AccountId,
         ) -> Balance {
-            *self.allowances.get(&(*owner, *spender)).unwrap_or(&0)
+            self.allowances
+                .get(&(*owner, *spender))
+                .copied()
+                .unwrap_or(0)
         }
     }
 
--- a/examples/erc20/lib.rs
+++ b/examples/erc20/lib.rs
@@ -51,6 +51,17 @@ mod erc20 {
         value: Balance,
     }
 
+    /// The ERC-20 error types.
+    #[derive(Debug, PartialEq, Eq, scale::Encode)]
+    #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+    pub enum Error {
+        /// Returned if not enough balance to fulfill a request is available.
+        InsufficientBalance,
+    }
+
+    /// The ERC-20 result type.
+    pub type Result<T> = core::result::Result<T, Error>;
+
     impl Erc20 {
         #[ink(constructor)]
         pub fn new(initial_supply: Balance) -> Self {
@@ -86,13 +97,13 @@ mod erc20 {
         }
 
         #[ink(message)]
-        pub fn transfer(&mut self, to: AccountId, value: Balance) -> bool {
+        pub fn transfer(&mut self, to: AccountId, value: Balance) -> Result<()> {
             let from = self.env().caller();
             self.transfer_from_to(from, to, value)
         }
 
         #[ink(message)]
-        pub fn approve(&mut self, spender: AccountId, value: Balance) -> bool {
+        pub fn approve(&mut self, spender: AccountId, value: Balance) -> Result<()> {
             let owner = self.env().caller();
             self.allowances.insert((owner, spender), value);
             self.env().emit_event(Approval {
@@ -100,7 +111,7 @@ mod erc20 {
                 spender,
                 value,
             });
-            true
+            Ok(())
         }
 
         #[ink(message)]
@@ -109,11 +120,11 @@ mod erc20 {
             from: AccountId,
             to: AccountId,
             value: Balance,
-        ) -> bool {
+        ) -> Result<()> {
             let caller = self.env().caller();
             let allowance = self.allowance_of_or_zero(&from, &caller);
             if allowance < value {
-                return false
+                return Err(Error::InsufficientBalance)
             }
             self.allowances.insert((from, caller), allowance - value);
             self.transfer_from_to(from, to, value)
@@ -124,10 +135,10 @@ mod erc20 {
             from: AccountId,
             to: AccountId,
             value: Balance,
-        ) -> bool {
+        ) -> Result<()> {
             let from_balance = self.balance_of_or_zero(&from);
             if from_balance < value {
-                return false
+                return Err(Error::InsufficientBalance)
             }
             self.balances.insert(from, from_balance - value);
             let to_balance = self.balance_of_or_zero(&to);
@@ -137,7 +148,7 @@ mod erc20 {
                 to: Some(to),
                 value,
             });
-            true
+            Ok(())
         }
 
         fn balance_of_or_zero(&self, owner: &AccountId) -> Balance {
@@ -289,7 +300,7 @@ mod erc20 {
 
             assert_eq!(erc20.balance_of(accounts.bob), 0);
             // Alice transfers 10 tokens to Bob.
-            assert_eq!(erc20.transfer(accounts.bob, 10), true);
+            assert_eq!(erc20.transfer(accounts.bob, 10), Ok(()));
             // Bob owns 10 tokens.
             assert_eq!(erc20.balance_of(accounts.bob), 10);
 
@@ -340,7 +351,10 @@ mod erc20 {
             );
 
             // Bob fails to transfers 10 tokens to Eve.
-            assert_eq!(erc20.transfer(accounts.eve, 10), false);
+            assert_eq!(
+                erc20.transfer(accounts.eve, 10),
+                Err(Error::InsufficientBalance)
+            );
             // Alice owns all the tokens.
             assert_eq!(erc20.balance_of(accounts.alice), 100);
             assert_eq!(erc20.balance_of(accounts.bob), 0);
@@ -367,9 +381,12 @@ mod erc20 {
                     .expect("Cannot get accounts");
 
             // Bob fails to transfer tokens owned by Alice.
-            assert_eq!(erc20.transfer_from(accounts.alice, accounts.eve, 10), false);
+            assert_eq!(
+                erc20.transfer_from(accounts.alice, accounts.eve, 10),
+                Err(Error::InsufficientBalance)
+            );
             // Alice approves Bob for token transfers on her behalf.
-            assert_eq!(erc20.approve(accounts.bob, 10), true);
+            assert_eq!(erc20.approve(accounts.bob, 10), Ok(()));
 
             // The approve event takes place.
             assert_eq!(ink_env::test::recorded_events().count(), 2);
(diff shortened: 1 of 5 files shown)


examples/trait-erc20/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "erc20"
```
cmichi, line 2: Should we change the name of trait-ified examples to `name = "trait-erc20"`? The `trait-flipper` crate is currently also still named `flipper`.

Robbepop: I think that is fine.

examples/trait-erc20/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use ink_env::{
+    DefaultEnvironment,
+    Environment,
+};
+use ink_lang as ink;
+
+pub use erc20::Erc20;
```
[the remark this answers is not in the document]
Robbepop: not really needed

examples/trait-erc20/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+    fn approve(&mut self, spender: AccountId, value: Balance) -> Result<()>;
+
+    #[ink(message)]
+    fn transfer_from(
+        &mut self,
+        from: AccountId,
+        to: AccountId,
+        value: Balance,
+    ) -> Result<()>;
+}
```
Robbepop, line 67: We could also move this entire section into the contract module so that we do not have to specify `Balance` and `AccountId` which is ugly. Also we wouldn't need to import of the trait. What do you think? Which is better as learning material?

examples/trait-erc20/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+            self.env().emit_event(Transfer {
+                from: Some(from),
+                to: Some(to),
+                value,
+            });
+            Ok(())
+        }
+
+        fn balance_of_or_zero(&self, owner: &AccountId) -> Balance {
+            *self.balances.get(owner).unwrap_or(&0)
```
Robbepop, line 208:
```suggestion
            self.balances.get(owner).copied().unwrap_or(0)
```

Robbepop: I think this is more readable.

examples/trait-erc20/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+        }
+
+        fn allowance_of_or_zero(
+            &self,
+            owner: &AccountId,
+            spender: &AccountId,
+        ) -> Balance {
+            *self.allowances.get(&(*owner, *spender)).unwrap_or(&0)
```
Robbepop, line 216:
```suggestion
            self.allowances.get(&(*owner, *spender)).copied().unwrap_or(0)
```

Robbepop: also more readable

examples/trait-erc20/lib.rs
cmichi: I think we should omit anything that is not strictly necessary to demonstrate basic Trait functionality. As to make it easily understandable and reduce causes for confusion. Yeah, so I'll adapt it.

examples/erc20/lib.rs
```diff
@@ -70,64 +90,94 @@ mod erc20 {
...
             let from = self.env().caller();
             self.transfer_from_to(from, to, value)
         }
 
+        /// Allows `spender` to withdraw from the caller's account multiple times, up to
+        /// the `value` amount.
+        ///
+        /// If this function is called again it overwrites the current allowance with `value`.
+        ///
+        /// An `Approval` event is emitted.
```
Robbepop, line 128: Missing `# Errors` section that describes what errors are possible.

examples/erc20/lib.rs
```diff
@@ -70,64 +90,94 @@ mod erc20 {
...
         pub fn allowance(&self, owner: AccountId, spender: AccountId) -> Balance {
             self.allowance_of_or_zero(&owner, &spender)
         }
 
+        /// Transfers `value` amount of tokens from the caller's account to account `to`.
+        ///
+        /// On success a `Transfer` event is emitted.
+        ///
+        /// Returns `Err(Error:InsufficientBalance)` if there are not enough tokens on
+        /// the caller's account balance.
```
Robbepop, line 116: Missing `# Errors` section that describes what errors are possible.

examples/erc20/lib.rs
```diff
@@ -70,64 +90,94 @@ mod erc20 {
...
 
+        /// Transfers `value` tokens on the behalf of `from` to the account `to`.
+        ///
+        /// This can be used to allow a contract to transfer tokens on ones behalf and/or
+        /// to charge fees in sub-currencies, for example.
+        ///
+        /// On success a `Transfer` event is emitted.
+        ///
+        /// Returns `Err(Error:InsufficientBalance)` if there are not enough tokens on
+        /// the caller's account balance.
```
Robbepop, line 149:
```suggestion
        /// # Errors
        /// 
        /// Returns `InsufficientBalance` error if there are not enough tokens on
        /// the caller's account balance.
```

examples/erc20/lib.rs
Robbepop: look in one of my suggestions for an example

examples/erc20/lib.rs
Robbepop: look in one of my suggestions for an example

examples/erc20/lib.rs
```diff
@@ -70,64 +90,94 @@ mod erc20 {
...
+        /// Returns the total token supply.
         #[ink(message)]
         pub fn total_supply(&self) -> Balance {
             *self.total_supply
         }
 
+        /// Returns the account balance for the specified `owner`.
         #[ink(message)]
         pub fn balance_of(&self, owner: AccountId) -> Balance {
             self.balance_of_or_zero(&owner)
```
Robbepop, line 102: I think we should simply make the implementation of `balance_of_or_zero` inplace. We did not have this inplace in the early days because it was not possible to call ink! messages internally. Nowadays I see no reason why we shouldn't have yet another function at this point.

examples/erc20/lib.rs
```diff
@@ -70,64 +90,94 @@ mod erc20 {
...
         }
 
+        /// Returns the amount which `spender` is still allowed to withdraw from `owner`.
         #[ink(message)]
         pub fn allowance(&self, owner: AccountId, spender: AccountId) -> Balance {
             self.allowance_of_or_zero(&owner, &spender)
```
Robbepop, line 108: I think we should simply make the implementation of `allowance_of_or_zero` inplace. We did not have this inplace in the early days because it was not possible to call ink! messages internally. Nowadays I see no reason why we shouldn't have yet another function at this point.

examples/erc20/lib.rs
```diff
@@ -70,64 +90,94 @@ mod erc20 {
...
             self.allowances.insert((from, caller), allowance - value);
             self.transfer_from_to(from, to, value)
         }
 
+        /// Transfers `value` amount of tokens from the caller's account to account `to`.
+        ///
+        /// On success a `Transfer` event is emitted.
+        ///
+        /// Returns `Err(Error:InsufficientBalance)` if there are not enough tokens on
+        /// the caller's account balance.
```
Robbepop, line 171:
```suggestion
        /// # Errors
        /// 
        /// Returns `InsufficientBalance` error if there are not enough tokens on
        /// the caller's account balance.
```

examples/erc20/lib.rs
```diff
@@ -70,64 +90,94 @@ mod erc20 {
...
             from: AccountId,
             to: AccountId,
             value: Balance,
-        ) -> bool {
+        ) -> Result<()> {
             let caller = self.env().caller();
             let allowance = self.allowance_of_or_zero(&from, &caller);
             if allowance < value {
-                return false
+                return Err(Error::InsufficientBalance)
```
Robbepop, line 160: I think we might want another error here. Something like `InsufficientAllowance` is more applicable here. Please also update the docs respectively.

examples/erc20/lib.rs
Robbepop: We actually want to return `InsufficientAllowance` instead of `InsufficientBalance` here.

examples/erc20/lib.rs
```diff
@@ -70,64 +90,94 @@ mod erc20 {
...
             }
             self.allowances.insert((from, caller), allowance - value);
             self.transfer_from_to(from, to, value)
```
Robbepop, line 163: Also I am not sure about the order of operations here but what if `self.transfer_from_to` fails? Then we already subtracted from `self.allowances` might we have the need to swap these operations?

examples/erc721/lib.rs
```diff
@@ -118,7 +118,7 @@ mod erc721 {
         id: TokenId,
     }
 
-    /// Event emited when a token approve occurs.
+    /// Event emmited when a token approve occurs.
```
Robbepop, line 121: Actually both are wrong as it is `emitted`: https://www.dict.cc/?s=emitted

:D

Robbepop (changes requested):
Have not yet looked at the traitified ERC-20 but we should first fix the standard ERC-20 and then apply only the necessary changes to the traitified one so that we keep the changes as minimum as possible.

examples/erc20/lib.rs
cmichi: Good catch, I updated the code and added a test which would have catched this.

examples/trait-erc20/lib.rs
```diff
@@ -533,5 +533,49 @@ mod erc20 {
...
+            assert_eq!(
+                ink_env::test::push_execution_context::<ink_env::DefaultEnvironment>(
+                    accounts.bob,
+                    callee,
+                    1000000,
+                    1000000,
+                    data
+                ),
+                ()
+            );
```
Robbepop, line 563: What exactly is this `assert_eq` trying to achieve?

cmichi: That's a good question, it's used in surrounding code and ERC-721 as well. I can't make sense of it, since `push_execution_context` doesn't have a return type. So asserting for the unit type will always be true. I've created a commit to remove the assert for all these instances.

Robbepop: Hard to reconstruct where it originated from due to squashed merge history. I begin to believe that squashing may not always be what we should actually be doing sometimes ...

examples/trait-erc20/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+            to: AccountId,
+            value: Balance,
+        ) -> Result<()> {
+            let caller = self.env().caller();
+            let allowance = self.allowance(from, caller);
+            if allowance < value {
+                return Err(Error::InsufficientAllowance)
+            }
+            self.transfer_from_to(from, to, value)?;
+            self.allowances.insert((from, caller), allowance - value);
```
Robbepop, line 223: Have you swapped the order because the test verified my assumption that it is broken otherwise?

Robbepop (approved):
LGTM

One final question before we can merge.

examples/trait-erc20/lib.rs
cmichi: Yes, the test fails before swapping the order and succeeds after the fix has been applied. So yeah you were right, it was a bug in the existing example contract.

Robbepop: okay thank you for the clarification!

PR #528, opened 2020-10-21
Fix lazy bug

Robbepop:
The problem was that if a `Lazy` instance has not been used in an execution (which it its whole point of existence) then it corrupted the `KeyPtr` for all following computations.

Unfortunately this bug went unnoticed in our extensive test suite.

- [x] Fixed the code causing the issue.
- [x] Cleaned up the surrounding code.
- [x] Added regression unit test.

### DIFF
--- a/crates/storage/src/lazy/entry.rs
+++ b/crates/storage/src/lazy/entry.rs
@@ -204,7 +204,6 @@ where
     pub fn push_packed_root(&self, root_key: &Key) {
         let old_state = self.replace_state(EntryState::Preserved);
         if old_state.is_mutated() {
-            self.replace_state(EntryState::Preserved);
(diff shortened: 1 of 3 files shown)


ascjones:
Is it possible to make the tests fail for this case in the offchain env?

ascjones (approved):
Confirmed fixes erc20. Still curious about whether the offchain env should catch such bugs

Robbepop:
> Is it possible to make the tests fail for this case in the offchain env?

I just added a regression test that fails without this fix.
Also cleaned-up implementation of `LazyCell` and the underlying `StorageEntry`.

cmichi (approved):
Thanks for adding the test!

PR #529, opened 2020-10-21
[storage] Add some tests for storage primitives

### LINKED ISSUE #403: Add more unit tests for storage2 primitive implementations
One of the few unit tests that have been left out of the initial implementation of the `storage2` module have been unit tests that check if the provided implementation of all Rust primitives, arrays, tuples and prelude types is actually correct.

These implementations are critical enough that they should not be taken for granted.
This issue is about writing unit tests for their `SpreadLayout` and `PackedLayout` implementations.

A show case for unit tests of all Rust primitives such as `i32` and also ink! primitives such as `Key` does already exist and can be taken as inspiration.

## ToDo List

- [ ] Get familiar with writing unit tests using the already existing unit tests for ink! and Rust primitives for `SpreadLayout` and `PackedLayout`.
    - Find them here: https://github.com/paritytech/ink/blob/master/core/src/storage2/traits/impls/prims.rs#L236
- [ ] Write unit tests for Rust arrays.
- [ ] Write unit tests for Rust tuples.
- [ ] Write unit tests for Rust's `Option` and `Result` types.
- [ ] Write unit tests for the rest prelude types, e.g. `Box`, `BTreeMap`, etc.

cmichi:
Closes #403.

### DIFF
--- a/crates/storage/src/traits/impls/fuzz_tests.rs
+++ b/crates/storage/src/traits/impls/fuzz_tests.rs
@@ -1,0 +1,81 @@
+// Copyright 2018-2020 Parity Technologies (UK) Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Fuzz tests for some storage primitives.
+
+#[cfg(all(test, feature = "std", feature = "ink-fuzz-tests"))]
+use quickcheck::TestResult;
+
+#[cfg(all(test, feature = "std", feature = "ink-fuzz-tests"))]
+use std::convert::AsMut;
+
+///  Receives a slice, returns an array.
+fn clone_into_array<A, T>(slice: &[T]) -> A
+where
+    A: Default + AsMut<[T]>,
+    T: Clone,
+{
+    let mut a = A::default();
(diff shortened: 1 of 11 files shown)


crates/storage/src/traits/impls/fuzz_tests.rs
```diff
@@ -0,0 +1,81 @@
...
+///  Receives a slice, returns an array.
+fn clone_into_array<A, T>(slice: &[T]) -> A
+where
+    A: Default + AsMut<[T]>,
+    T: Clone,
+{
+    let mut a = A::default();
+    <A as AsMut<[T]>>::as_mut(&mut a).clone_from_slice(slice);
+    a
+}
```
Robbepop, line 32: Pretty cool!

crates/storage/src/traits/impls/fuzz_tests.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// Tests if a fuzzed `[i32; 32]` array results in the same object when
+/// pushed/pulled from storage (for `spread` and `packed`).
+#[quickcheck]
+fn fuzz_pull_push_pull_array(x: Vec<i32>) -> TestResult {
+    if x.len() < 32 {
+        return TestResult::discard()
+    }
```
Robbepop, line 40: Why discard in this case?

crates/storage/src/traits/impls/fuzz_tests.rs
Robbepop: However, why not use `Vec<u8>`? ^^

crates/storage/src/traits/impls/fuzz_tests.rs
Robbepop: If we used `Vec<_>` we were fine.

Robbepop (commented):
All in all looks good to me. Some questions though.

crates/storage/src/traits/impls/fuzz_tests.rs
cmichi: > Why discard in this case?

I added a clarifying comment.

Robbepop (approved):
LGTM

Issue #519, opened 2020-10-10
Fixes for ink! 3.0-rc2


Robbepop:
- [x] Update `CONTRIBUTING.md` file. (By @cmichi )
- [x] Fix `Cargo.toml` `documentation` links to link to our deployed documentation or crates.io documentation.
- [x] Fix `README` `coverage` report statistics that should always point to current `master` instead of latest `commit`.
- [x] Include `ink_storage::BinaryHeap` (by @ascjones) + its optimizations (by @cmichi).
- [x] Re-export `ink_storage::lazy::lazy_hmap::LazyHashMap` at `ink_storage::lazy` for easier access.
- [x] Add note about splitting `ink_core` into `ink_env` and `ink_storage` to `RELEASE_NOTES.md`.
- [x] Fix `RELEASE_NOTES.md` `#[ink(constructors)]` and `#[ink(storage)]` instead of `#[ink(constructor)]` in first example.
- [x] `ink_storage::Box` needs to implement `TypeInfo` or storage data structures such as `StorageVec` need to lose this trait bound. So that we can put a `storage::Box` as `T` in a `storage::Vec<T>`.
    - It was natural to simply implement `TypeInfo` for `ink_storage::Box` since it is SCALE encodable anyways. However, some more fundamental problems have popped up during the implementation for which we will need more investigation in the future.
- [x] Investigate bug with `ink_storage::Lazy` usage not mutating the contract's storage in the ERC-20 example.
- [x] Fix an eventual bug that `ink_storage::collection::HashMap` does not properly mutate storage in on-chain environments. We first have to find out what causes this bug and under which circumstances it exists.
    - We think the reports `HashMap` bugs are caused by the above `ink_storage::Lazy` bug that has now been fixed.

PR #530, opened 2020-10-21
Misc fixes for ink! 3.0-rc1

Robbepop:
Closes https://github.com/paritytech/ink/issues/519.


crates/allocator/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2018"
 license = "Apache-2.0"
 readme = "README.md"
 repository = "https://github.com/paritytech/ink"
-documentation = "https://substrate.dev/substrate-contracts-workshop/#/"
+documentation = "https://docs.rs/ink_allocator/"
```
cmichi, line 10: Just noted that https://docs.rs/ink_allocator/ is completely empty. Maybe at least a module documentation comment would be good?

crates/storage/src/alloc/boxed/storage.rs
```diff
@@ -29,21 +29,27 @@ use ink_primitives::Key;
...
+                    scale_info::Path::from_segments(vec!["ink_storage", "alloc", "Box"])
+                        .expect("encountered invalid Rust path"),
+                )
+                // Unfortunately we cannot encode the type parameters of the box since they
+                // have to be `T: scale::Codec`. However, them not requiring to be encodable
+                // is the purpose of the storage `Box<T>`.
+                // Until we found a solution to this problem we cannot uncomment the below
+                // line of code:
+                //
+                // .type_params(vec![scale_info::MetaType::new::<T>()])
```
cmichi, line 48: What's the trade-off of not being able to un-comment the line?

Robbepop: It does not make sense for the `T` in `ink_storage::alloc::Box<T>` to implement `scale_info::TypeInfo` so having this requirement on the bounds would not make sense, too. The `T: scale_info::TypeInfo` bound means that `T` is somehow `scale` encodable. However, `ink_storage::alloc::Box<T>` actually exists exactly for the purpose of allowing non-`scale` encodable instances within others.

cmichi (approved):
LGTMichi :-)

Issue #507, opened 2020-10-05
Optimization: Reduce read overhead when loading from storage


cmichi:
When loading an object from storage we have an overhead of one read operation.

See [here](https://github.com/paritytech/ink/blob/master/crates/storage/src/traits/optspec.rs#L28-L38):
```
    // In case the contract storage is occupied we handle
    // the Option<T> as if it was a T.
    ink_env::get_contract_storage::<()>(root_key)
        .ok()
        .flatten()
        .map(|_| super::pull_spread_root::<T>(root_key))
```
We read the `root_key` two times from storage.

At the moment there is no API to ask Seal if a storage is occupied, so instead we have to issue the first read in order to find out if it's `None` or `Some(…)`. In the latter case we pull recursively from storage.

## Possible Solution

We probably have to adapt the `SpreadLayout` interface, since we have to cache those reads ourselves. An option could be to implement caching into `KeyPtr`.

Robbepop:
The special challenge behind this is to find a solution that optimally works without using heap memory allocation so that simpler contracts can still get rid of the heap allocator. This is very important since this machinery is very deeply rooted in all ink! smart contracts.

cmichi:
I've removed this ticket from Substrate's "Improve throughput" issue, since this issue rather targets the runtime performance by having one storage read less. The "Improve throughput" issue on the other hand aims to reduce the size of contracts. The ideas we've had so far to solve this very issue here involve more code (i.e. caching) though.

PR #532, opened 2020-10-21
[Prototype] Implement Storage Entry API

### LINKED ISSUE #379: Implement storage entry API
Currently we have 3 different APIs to access a contract's storage in `ink_core`:

- `env::get_contract_storage`: To retrieve a decoded `T` from the contract storage cell.
- `env::set_contract_storage`: To encode a `T` into the contract storage cell.
- `env::clear_contract_storage`: To clear all data from the contract storage cell.

This API is fine, however, it doesn't scale well for additional operations.
For example, adding the ability to query if a contract storage is vacant or occupied requires adding yet another API to the already large set of root functions.

Instead we could unify all accesses to a single contract storage cell using one entry API:

```
fn storage_entry(at: &Key) -> StorageEntry { ... }


pub struct StorageEntry<'a> {
    key: &'a Key,
}

pub enum StorageEntryState {
    Vacant,
    Occupied,
}

impl StorageEntry<'_> {
    pub fn is_vacant(&self) -> bool {
        ...
    }

    pub fn is_occupied(&self) -> bool {
        ...
    }

    pub fn get<R>(&self) -> Result<Option<R>>
    where
        R: scale::Decode,
    {
        ...
    }

    pub fn set<V>(&self, value: &V)
    where
        V: scale::Encode,
    {
        ...
    }
}
```

This is also easily extensible in case we later add the subtrie API for contracts.

cmichi:
Closes #379.

### DIFF
--- a/crates/env/src/engine/on_chain/impls.rs
+++ b/crates/env/src/engine/on_chain/impls.rs
@@ -209,6 +209,15 @@ impl EnvInstance {
 }
 
 impl EnvBackend for EnvInstance {
+    fn is_contract_storage(&mut self, key: &Key) -> bool {
+        let output = &mut self.scoped_buffer().take_rest();
+        match ext::get_storage(key.as_bytes(), output) {
+            Ok(_) => true,
+            Err(ExtError::KeyNotFound) => false,
+            Err(_) => panic!("encountered unexpected error"),
+        }
+    }
+
     fn set_contract_storage<V>(&mut self, key: &Key, value: &V)
     where
         V: scale::Encode,
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -218,39 +218,80 @@ where
     })
 }
 
-/// Writes the value to the contract storage under the given key.
-///
-/// # Panics
-///
-/// - If the encode length of value exceeds the configured maximum value length of a storage entry.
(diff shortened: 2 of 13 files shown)


crates/env/src/engine/on_chain/impls.rs
```diff
@@ -209,6 +209,15 @@ impl EnvInstance {
 }
 
 impl EnvBackend for EnvInstance {
+    fn is_contract_storage(&mut self, key: &Key) -> bool {
+        let output = &mut self.scoped_buffer().take_rest();
+        match ext::get_storage(key.as_bytes(), output) {
```
cmichi, line 214: A seal operation `is_storage` would do us good here as well, same as in https://github.com/paritytech/ink/issues/507. Maybe it's time?

Robbepop:
Thanks for working on this POC @cmichi !
However, I unfortunately see some problems with this API as such in regards to the notion of "leaky abstractions".
We might not want to merge this feature due to lack of backend support although its interfaces have some charm to themselves.

Robbepop (commented):
Will put on ice until we decide how to push forward with this experimental API.

cmichi:
Closing this issue, since we determined to not go down this road. See the comments above for the reasoning.

Nevertheless it was a worthy exploration and we couldn't have come to that conclusion without this PR.

PR #540, opened 2020-10-24
Improve docs on transfer() + minimum_balance()

### LINKED ISSUE #538: Unable to transfer contract balance to another account
Hi,

The `env.transfer()` seems not doing its job in my environment. The balance is not changed, neither in the contract nor in the destination account.  No exception/panic is raised.

I use it in a Ink contract, with something like :

```rust
let value = self.env().balance();
let _ = self.env().transfer(self.charity_pot, value);
```

**Note :** In the same contract, using the same `charity_pot` account, the  `env.terminate_contract()` is correctly transferring the funds to the given account. 

Please let me know if you need additionnal informations. Thanks for your help !

**Environment**

Using  `nightly-2020-08-23` and `cargo-contract 0.7.0` on Ubuntu 20.04.

```toml
[dependencies]
ink_primitives = { version = "3.0.0-rc1", default-features = false }
ink_metadata = { version = "3.0.0-rc1", default-features = false, features = ["derive"], optional = true }
ink_env = { version = "3.0.0-rc1", default-features = false }
ink_storage = { version = "3.0.0-rc1", default-features = false }
ink_lang = { version = "3.0.0-rc1", default-features = false }

scale = { package = "parity-scale-codec", version = "1.3", default-features = false, features = ["derive"] }
scale-info = { version = "0.4", default-features = false, features = ["derive"], optional = true }
```

cmichi:
Closes #538 by clearing up the confusion that lead to it.

@athei I think it makes sense to expose a getter for the `subsistence_threshold` from seal. Otherwise it's not possible to take this variable into consideration in the smart contract when calculating how much to transfer (like in the issue).

```
/// Subsistence threshold is the extension of the minimum balance (aka existential deposit) by the
/// tombstone deposit, required for leaving a tombstone.
```
(from [here](https://github.com/paritytech/substrate/blob/master/frame/contracts/src/lib.rs#L778-L779))

We already have the `minimum_balance`, so it seems consistent to expose the subsistence threshold as well. Wdyt?

### DIFF
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -422,9 +422,11 @@ where
 /// contract call or invoke a runtime function that performs the
 /// transaction.
 ///
-/// # Panics
+/// # Errors
 ///
-/// If the contract doesn't have sufficient funds.
+/// - If the contract doesn't have sufficient funds.
+/// - If the transfer would have brought the sender's total balance below the
+///   subsistence threshold.
 pub fn transfer<T>(destination: T::AccountId, value: T::Balance) -> Result<()>
 where
     T: Environment,
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -169,7 +169,7 @@ where
     })
 }
 
-/// Returns the minimum balance for the contracts chain.
+/// Returns the minimum balance that is required for creating an account.
 ///
 /// # Errors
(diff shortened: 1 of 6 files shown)


Robbepop (approved):
LGTM

Thanks a lot for the docs improvements!

PR #541, opened 2020-10-26
Improve docs on non-payable constructor

cmichi:
I think I'm not the only one who will wonder about this :-).

We could also add more explanations for conflicts.


crates/env/src/engine/off_chain/impls.rs
```diff
@@ -349,7 +349,7 @@ impl TypedEnvBackend for EnvInstance {
         T: Environment,
         Args: scale::Encode,
     {
-        unimplemented!("off-chain environment does not support contract invokation")
+        unimplemented!("off-chain environment does not support contract invocation")
```
Robbepop, line 352: dang! i will never get this word right :D

Robbepop (approved):
Wow, this is really cool, thanks!

Robbepop:
Also thanks a lot for the 2 new UI tests, I just wanted to ask for them ... :D

Robbepop (commented):
While I really really like this PR and its changes one nit is that normally you do not create single big errors with all the information necessary but build them up incrementally appending information to already existing errors. The way that this is normally done is via: https://docs.rs/syn/1.0.48/syn/struct.Error.html#method.combine

There are also examples for usage of this in ink! when looking for duplicate attributes or duplicate storages etc.

To add to this, in the `modular_bitfield` crate I just added these definitions to extend the `syn::Error` API:

```rust
pub trait CombineError {
    fn into_combine(self, another: syn::Error) -> Self;
}

impl CombineError for syn::Error {
    fn into_combine(mut self, another: syn::Error) -> Self {
        self.combine(another);
        self
    }
}
```

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -549,16 +575,13 @@ impl InkAttribute {
     pub fn ensure_no_conflicts<'a, P>(
         &'a self,
         mut is_conflicting: P,
-    ) -> Result<(), syn::Error>
+    ) -> Result<(), &'a ir::AttributeArg>
```
Robbepop, line 578: While I can understand the change behind this I am not sure I like an `ir::AttributeArg` as error return type.

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -230,6 +230,12 @@ impl InkAttribute {
         })
     }
 
+    /// Returns `true` if the ink! attribute contains the `constructor` argument.
+    pub fn is_constructor(&self) -> bool {
+        self.args()
+            .any(|arg| matches!(arg.kind(), AttributeArgKind::Constructor))
```
Robbepop, line 236: While this works a message is only a constructor if the first argument is #[ink(constructor)]`. This is to disambiguate messages and constructors that have been invalidly flagged with both.

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -441,19 +447,21 @@ where
 /// - If there are duplicate ink! attributes.
 /// - If the first ink! attribute is not matching the expected.
 /// - If there are conflicting ink! attributes.
-pub fn sanitize_attributes<I, C>(
+pub fn sanitize_attributes<I, C, R>(
     parent_span: Span,
     attrs: I,
     is_valid_first: &ir::AttributeArgKind,
     mut is_conflicting_attr: C,
+    conflict_reason: R,
```
Robbepop, line 455: Not sure why we still need this whole parameter and a new `InkAttributeConflictErr` type at all with our combine approach.

crates/lang/ir/src/ir/item_impl/constructor.rs
```diff
@@ -342,6 +348,17 @@ mod tests {
         );
     }
 
+    fn assert_try_from_fails_multiple_times(
```
Robbepop, line 351: Please add proper docs.

Robbepop (commented):
Please elaborate clearly why we need or want to have this new error type for conflicting attributes and not simply use syn::Error.

cmichi:
Blocked until #553 has been resolved.

Robbepop:
Superseeded by https://github.com/paritytech/ink/pull/628. Closed.

PR #549, opened 2020-10-27
Fix fuzz setup

cmichi:
Some fixes.

Works now, as can be seen in https://github.com/paritytech/ink/issues/551.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -340,7 +340,7 @@ fuzz-tests:
                 -H "Content-Type: application/json; charset=utf-8" \
                 -d $'{
                     "title": "[ci] Failing fuzz tests on master ('"$( date +"%d %b %Y" )"')",
-                    "body": "The CI job ['"${CI_JOB_ID}"']('"${CI_JOB_URL}"') just failed.\n\nThe offending commit is `['"${CI_COMMIT_TITLE}"'](https://github.com/paritytech/ink/commit/'"${CI_COMMIT_SHA}"')`.",
+                    "body": "The CI job ['"${CI_JOB_ID}"']('"${CI_JOB_URL}"') just failed.\n\nThe offending commit is ['"${CI_COMMIT_TITLE}"'](https://github.com/paritytech/ink/commit/'"${CI_COMMIT_SHA}"').",
                     "assignees": [],
                     "labels": [
                         "P-high"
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -327,15 +327,16 @@ fuzz-tests:
...
-                    "title": "[ci] Failing fuzz tests on master ('"$( date +"%d %b %Y" )"'),
-                    "body": "This CI job '"${CI_JOB_ID}"' just failed: '"${CI_JOB_URL}"'.\n\nThe offending commit is `['"${CI_COMMIT_TITLE}"'](https://github.com/paritytech/ink/commit/'"${CI_COMMIT_SHA}"')`.",
+                    "title": "[ci] Failing fuzz tests on master ('"$( date +"%d %b %Y" )"')",
+                    "body": "The CI job ['"${CI_JOB_ID}"']('"${CI_JOB_URL}"') just failed.\n\nThe offending commit is ['"${CI_COMMIT_TITLE}"'](https://github.com/paritytech/ink/commit/'"${CI_COMMIT_SHA}"').",
                     "assignees": [],
                     "labels": [
                         "P-high"
                     ]
                 }'
+        - exit 1
```
TriplEight, line 342: so now we always fail it?

cmichi: Nope, in line 328 there is an early return if all tests succeeded:
```
- if [ $all_tests_passed -eq 0 ]; then exit 0; fi
```
The issue whiched is fixed with this `- exit 1`is that this `fuzz` CI stage always exited with `0`, since the last statement was `curl` and not the `cargo test …`. So the CI was green, despite of a fuzz test having failed.

TriplEight: maybe then `exit "$all_tests_passed"`?

cmichi: That's a great suggestion! I just changed it.

TriplEight (approved):
Let's see how it works on master then!

TriplEight:
Maybe it would make sense to put all this bash magic into https://github.com/paritytech/ink/tree/master/scripts though.

cmichi:
> Maybe it would make sense to put all this bash magic into https://github.com/paritytech/ink/tree/master/scripts though.

In this case I would rather opt for keeping it in the `.gitlab-ci.yml`, since we recommend contributors to run those scripts as a mean to check if their local changes are fine, so I think it would make things more complicated. Fine with you?

Issue #356, opened 2020-03-18
Create example contract that uses ext_terminate


Robbepop:
We now have the ability for smart contracts to terminate themselves via `ext_terminate`.
We should provide an example smart contract for users to demonstrate this API.

cmichi:
Has already been closed with https://github.com/paritytech/ink/pull/554 some months ago.

PR #554, opened 2020-10-29
Add example which uses ext_transfer + ext_terminate

### LINKED ISSUE #355: Create example contract that uses ext_transfer
We have a new native `ext_transfer` with which a contract can easily transfer value to another account. We should have an example contract to demonstrate this new API.

cmichi:
Closes #355 and #356.


crates/env/src/engine/off_chain/impls.rs
```diff
@@ -232,6 +232,20 @@ impl EnvInstance {
...
+    where
+        T: Environment,
+    {
+        // Send the remaining balance to the beneficiary
+        let all: T::Balance = self.balance::<T>().expect("could not decode balance");
+        self.transfer_impl::<T>(beneficiary, all)?;
+
+        let contract_id = self.account_id::<T>().expect("could not decode account id");
+        self.accounts.remove_account::<T>(contract_id);
+
```
cmichi, line 246: What the on-chain impl would do here is set a tombstone with a code hash and remove storage. AFAIU both is not easily achievable with our current off-chain env, hence I left it out here for the moment.

crates/env/src/backend.rs
```diff
@@ -302,7 +302,7 @@ pub trait TypedEnvBackend: EnvBackend {
     /// # Note
     ///
     /// For more details visit: [`ink_env::terminate_contract`]
-    fn terminate_contract<T>(&mut self, beneficiary: T::AccountId) -> !
+    fn terminate_contract<T>(&mut self, beneficiary: T::AccountId) -> Result<()>
```
cmichi, line 305: Not sure about the return type here. I guess the real thing would indeed never return.

Robbepop: I guess you changed this API to make it possible to deliver an off-chain implementation of it. However, as you already noticed our off-chain environment is kind of limited and the inability to do proper contract terminations is one. Instead of adjusting useful interfaces to an non useful off-chain implementation we should instead strife for improving the off-chain environment so that it is able to handle those things if possible.

Robbepop: This means that for example contracts we cannot really off-chain test them properly if they make use of such behavior.

Robbepop: You example contract is fine to cover both APIs. However, I think for `transferred_balance` we could also create a simple alternative example contract similar to what I did in the release notes of v3.0.0-rc1.

Robbepop (changes requested):

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -363,3 +363,58 @@ where
...
+        let encoded_input: &Vec<u8> =
+            value_any.downcast_ref::<Vec<u8>>().expect("must work");
+        let info: ink_env::test::ContractTerminationResult<AccountId, Balance> =
+            scale::Decode::decode(&mut &encoded_input[..]).expect("must work");
+
+        let expected_beneficiary: AccountId = $beneficiary;
+        assert_eq!(info.beneficiary, expected_beneficiary);
+
+        let expected_balance: Balance = $balance;
+        assert_eq!(info.transferred, expected_balance);
```
Robbepop, line 418: very clever and creative solution! might be acceptable. however, needs further macro hygiene improvements since this is going to be exposed to the users.

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -363,3 +363,58 @@ where
...
+#[derive(scale::Encode, scale::Decode)]
+pub struct ContractTerminationResult<AccountId, Balance>
+where
+    AccountId: scale::Codec,
+    Balance: scale::Codec,
+{
+    /// The beneficiary account who received the remaining value in the contract.
+    pub beneficiary: AccountId,
+    /// The value which was transferred to the `beneficiary`.
+    pub transferred: Balance,
```
Robbepop, line 377:
```suggestion
pub struct ContractTerminationResult<E>
where
    E: Environment,
{
    /// The beneficiary account who received the remaining value in the contract.
    pub beneficiary: <E as Environment>::AccountId,
    /// The value which was transferred to the `beneficiary`.
    pub transferred: <E as Environment>::Balance,
```

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -363,3 +363,58 @@ where
...
+            value_any.downcast_ref::<Vec<u8>>().expect("must work");
+        let info: ink_env::test::ContractTerminationResult<AccountId, Balance> =
+            scale::Decode::decode(&mut &encoded_input[..]).expect("must work");
+
+        let expected_beneficiary: AccountId = $beneficiary;
+        assert_eq!(info.beneficiary, expected_beneficiary);
+
+        let expected_balance: Balance = $balance;
+        assert_eq!(info.transferred, expected_balance);
+    };
```
Robbepop, line 419:
```suggestion
    ) => {{
        let __act_value_any = ::std::panic::catch_unwind($should_terminate)
            .expect_err("contract did not terminate");
        let __act_encoded_input: &::std::vec::Vec<::core::primitive::u8> =
            __act_value_any.downcast_ref::<::std::vec::Vec<::core::primitive::u8>>().expect("must work");
        let __act_info: ::ink_env::test::ContractTerminationResult<AccountId, Balance> =
            ::scale::Decode::decode(&mut &__act_encoded_input[..]).expect("must work");

        let __act_expected_beneficiary: AccountId = $beneficiary;
        assert_eq!(info.beneficiary, __act_expected_beneficiary);

        let __act_expected_balance: Balance = $balance;
        ::std::assert_eq!(info.transferred, __act_expected_balance);
    }};
```

Robbepop: Also we are clearly missing out on `AccountId` and `Balance` types. Those ideally need to be something like `<E as ::ink_env::Environment>::Balance` and `<E as ::ink_env::Environment>::Balance` respectively with an `E` that is fed to the macro invocation somehow. Not sure about the best design here.

examples/contract-terminate/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+        #[ink(constructor)]
+        pub fn new() -> Self {
+            Self {}
+        }
+
+        /// Terminates with the caller as beneficiary.
+        #[ink(message)]
+        pub fn terminate_me(&mut self) {
+            self.env().terminate_contract(self.env().caller());
+        }
```
cmichi, line 39: The correct version of this would be:
```suggestion
        pub fn terminate_me(&mut self) -> ! {
            self.env().terminate_contract(self.env().caller())
        }
```

We then run into `error[E0277]: the trait bound `!: scale_info::TypeInfo` is not satisfied` though. Wdyt?

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -375,3 +375,44 @@ where
...
+/// ink_env::test::assert_contract_termination::<ink_env::DefaultEnvironment, _>(
+///     should_terminate,
+///     expected_beneficiary,
+///     expected_value_transferred_to_beneficiary
+/// );
+/// ```
+///
+/// See `examples/contract-terminate` for a complete usage example.
+#[cfg(feature = "std")]
+pub fn assert_contract_termination<T, F>(
```
Robbepop, line 398: So much better as a function compared to a macro!

Robbepop (approved):
LGTM

Thanks for the new example!

PR #561, opened 2020-11-03
[lang] Improve invalid selector error

cmichi:
Previously it was just:
```
$ cargo build                                                                                                                                                  
  --> lib.rs:19:1                                                                                                                                              
   |                                                                                                                                                           
19 | #[ink::contract]                                                                                                                                          
   | ^^^^^^^^^^^^^^^^                                                                                                                                          
   |                                                                                                                                                           
   = help: message: called `Option::unwrap()` on a `None` value                                                                                                
                                                                                                                                                               
error: aborting due to previous error  
```


crates/lang/macro/tests/ui/fail/M-05-message-invalid-selector.stderr
```diff
@@ -0,0 +1,7 @@
+error: custom attribute panicked
+ --> $DIR/M-05-message-invalid-selector.rs:3:1
+  |
+3 | #[ink::contract]
```
ascjones, line 4: Can the span be across the actual message function?

ascjones (approved):
LGTM

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -580,24 +580,42 @@ impl TryFrom<syn::NestedMeta> for AttributeArg {
...
                                     )
                                 })?;
                                 let str = lit_str.value();
-                                let cap = regex.captures(&str).unwrap();
+                                let cap = regex.captures(&str)
+                                .ok_or_else(||
+                                   format_err_spanned!(
+                                        meta,
+                                        "invalid selector - a selector must consist of four bytes in hex (e.g. `selector = \"0xCAFEBABE\"`)"
+                                    )
```
Robbepop, line 592: Would be nice if we could factor this out into its own little function since there is some code dupe.

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -580,24 +580,42 @@ impl TryFrom<syn::NestedMeta> for AttributeArg {
...
-                                        "encountered non-hex digit at position 2",
-                                    ),
-                                    u8::from_str_radix(&cap[4], 16).expect(
-                                        "encountered non-hex digit at position 3",
-                                    ),
+                                    u8::from_str_radix(&cap[1], 16).map_err(|_| {
+                                        format_err_spanned!(
+                                            meta,
+                                            "encountered non-hex digit at position 0",
+                                        )
```
Robbepop, line 599: Would be nice if we could factor this into its own function with some index parameter because of the code dupe.

Robbepop (changes requested):

crates/lang/ir/src/ir/attrs.rs
cmichi: In this case I think this would lead to complexity which is unnecessary, the diff would be sth like, due to both closures requiring different input:

```
+fn invalid_selector_err_regex(meta: syn::Meta) -> impl FnOnce(regex::Error) -> syn::Error {
+    |_| {
+        format_err_spanned!(
+                meta,
+                "invalid selector - a selector must consist of four bytes in hex (e.g. `selector = \"0xCAFEBABE\"`)"
+            )
+    }
+}
+
+fn invalid_selector_err(meta: syn::Meta) -> impl FnOnce() -> syn::Error {
+    || {
+        format_err_spanned!(
+                meta,
+                "invalid selector - a selector must consist of four bytes in hex (e.g. `selector = \"0xCAFEBABE\"`)"
+            )
+    }
+}
+
 impl TryFrom<syn::NestedMeta> for AttributeArg {
     type Error = syn::Error;
 
@@ -577,20 +595,10 @@ impl TryFrom<syn::NestedMeta> for AttributeArg {
                             if let syn::Lit::Str(lit_str) = &name_value.lit {
                                 let regex = Regex::new(
                                     r"0x([\da-fA-F]{2})([\da-fA-F]{2})([\da-fA-F]{2})([\da-fA-F]{2})"
-                                ).map_err(|_| {
-                                    format_err_spanned!(
-                                        meta,
-                                        "invalid selector - a selector must consist of four bytes in hex (e.g. `selector = \"0xCAFEBABE\"`)"
-                                    )
-                                })?;
+                                ).map_err(invalid_selector_err_regex(meta))?;
                                 let str = lit_str.value();
                                 let cap = regex.captures(&str)
-                                .ok_or_else(||
-                                   format_err_spanned!(
-                                        meta,
-                                        "invalid selector - a selector must consist of four bytes in hex (e.g. `selector = \"0xCAFEBABE\"`)"
-                                    )
-                                )?;
```

So I'd rather leave as is here.

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -565,6 +566,13 @@ impl InkAttribute {
     }
 }
 
+fn non_hex_err<'a>(
```
Robbepop: Maybe I am a bit too pedantic but I really would like to see a comment here.

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -565,6 +566,13 @@ impl InkAttribute {
...
+    meta: &'a syn::Meta,
+    pos: usize,
+) -> impl FnOnce(ParseIntError) -> syn::Error + 'a {
+    move |_| format_err_spanned!(meta, "encountered non-hex digit at position {}", pos)
```
Robbepop, line 573: why `move |_|` ? this is counter intuitive imo. Please just return a `syn::Error` here as usual. the point of making it possible to call it without the `|x| foo` in the invocation is not worse than this.

crates/lang/ir/src/ir/attrs.rs
Robbepop: I think it would be a real improvement if your diff would look more intuitively like this instead:

```rust
fn invalid_selector_err_regex(meta: syn::Meta) -> syn::Error {
    format_err_spanned!(
        meta,
        "invalid selector - a selector must consist of four bytes in hex (e.g. `selector = \"0xCAFEBABE\"`)"
    )
}
```
And use like:
```rust
.map_err(|_| invalid_selector_err_regex(meta))
```
Done. Less complex. More readable.

Also I do not understand why you need `invalid_selector_err_regex` AND `invalid_selector_err` since both are equivalent under my proposal.

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -566,11 +565,15 @@ impl InkAttribute {
...
-fn err_non_hex<'a>(
-    meta: &'a syn::Meta,
-    pos: usize,
-) -> impl FnOnce(ParseIntError) -> syn::Error + 'a {
-    move |_| format_err_spanned!(meta, "encountered non-hex digit at position {}", pos)
+fn err_non_hex(meta: &'_ syn::Meta, pos: usize) -> syn::Error {
+    format_err_spanned!(meta, "encountered non-hex digit at position {}", pos)
+}
+
+fn invalid_selector_err_regex(meta: &'_ syn::Meta) -> syn::Error {
```
Robbepop, line 574: Is the `&'_` lifetime needed here?

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -566,11 +565,15 @@ impl InkAttribute {
     }
 }
 
-fn err_non_hex<'a>(
-    meta: &'a syn::Meta,
-    pos: usize,
-) -> impl FnOnce(ParseIntError) -> syn::Error + 'a {
-    move |_| format_err_spanned!(meta, "encountered non-hex digit at position {}", pos)
+fn err_non_hex(meta: &'_ syn::Meta, pos: usize) -> syn::Error {
```
Robbepop, line 569: is the `&'_` lifetime needed here?

Robbepop (approved):
LGTM

Thanks a lot for the highly anticipated error reporting improvements! :)

PR #567, opened 2020-11-04
Bring discord badge to life

cmichi:
[Preview here](https://github.com/paritytech/ink/tree/cmichi-discord-badge#----paritys-ink-for-writing-smart-contracts)

I talked to the ops people for our Discord channels and they activated a feature which lets us display the number of users online in the badge. IMO it makes sense to do that, since it depicts that the project is full of life.

For Matrix I also talked to the dedicated ops people for that, but for security reasons they won't enable the feature which lets us display the number of users in that badge.


Robbepop (approved):
really cool btw!

PR #569, opened 2020-11-06
Remove iter_mut() for BinaryHeap

cmichi:
Mutations of the entries in the heap can mess up the sort oder. The Rust stdlib doesn't contain `BinaryHeap::iter_mut()` for this reason.

```
#[test]
fn mess_up_heap_order_via_iter_mut() {
    // Largest Element is 300
    let mut heap = heap_from_slice(&[100, 300, 200]);
    let largest = heap.pop().unwrap();
    assert_eq!(largest, 300);

    let mut heap = heap_from_slice(&[100, 300, 200]);
    heap.iter_mut().for_each(|current| {
        // If on iteration we find the element with 300, we change it to 1
        if *current == 300 {
            *current = 1;
        }
    });
    let largest = heap.pop().unwrap();
    // Largest Element will suddenly be 1 here, even though all others are larger
    assert_eq!(largest, 300);
}
```

### DIFF
--- a/crates/storage/src/collections/binary_heap/children_vec.rs
+++ b/crates/storage/src/collections/binary_heap/children_vec.rs
@@ -179,17 +179,6 @@ where
         Iter::new(&self)
     }
 
-    /// Returns an iterator yielding exclusive references to all elements of the heap.
-    ///
-    /// # Note
-    ///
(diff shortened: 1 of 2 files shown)


Robbepop (approved):
Good catch!

PR #570, opened 2020-11-07
Fix Option pointer bug

### LINKED ISSUE #568: [Bug] Contract Trapped Issue due to Ordering of Struct
Currently in https://gist.github.com/shirshak55/504745cc62539ae5f6388491448ae19b if you try to invoke get_entry two times you will get Contract Trapped.

So if you follow println you will see following image giving some false indication like `vector.len()` cannot be called 
https://pasteboard.co/JyPBED9.png

But the error was due to ordering of field which I thought wouldn't matter in Rust and its normal in Rust code.

But if we use this struct it doesn't give Contract Trapped Error.

```rust
#[ink(storage)]
pub struct Charity {
    // User can play only once and added to pool with one submission. HashMap is perfect fit here.
    // users_vec so that we can choose it easily when we draw it
    users: HashMap<User, Balance>,
    users_vec: Vec<User>,
    // The collected money is sent  to predefined address. Which is the one who calls the contract first time
    collector: AccountId,
    // Total amount collected till now so we don't have to iterate over users to know it.
    // Just for optimization
    amount_collected: Balance,
    // Winners List
    winners: Vec<User>,
    countdown: Option<Timestamp>,
}
```

I would like to thank Robin from Discord for figuring out ordering issue. I had already wasted like 24 hour and if he weren't there i bet I wouldn't have found bug for another 24 hour :) Thanks to him.

According to him:

> the `Option<Timestamp>` is suspicious because we do some arcane optimized handling for Option<T> values in ink_storage to avoid some storage reads and writes and apparently things that are complex tend to have bugs.

cmichi:
Closes #568.

The bug was in the the implementation of `SpreadLayout` for `Option`:  
* On `push_spread`, `clear_spread`, `pull_spread` no further forwarding to the inner value `push_spread`, `clear_spread`, `pull_spread` functions happened if the `Option` was `None`.   

* This is problematic because these methods advance the ptr by one ‒ which happened if the `Option` was `Some(…)`. For `None` it didn't happen, thus causing havoc for all objects that subsequently tried to push, pull or clear their content from storage.


Robbepop:
> Closes #568.
> 
> So I'm still working on adding a regression test and making it prettier, but I think we have it.
> 
> The bug was in the the implementation of `SpreadLayout` for `Option`:
> 
>     * On `push_spread`, `clear_spread`, `pull_spread` no further forwarding to the inner value `push_spread`, `clear_spread`, `pull_spread` functions happened if the `Option` was `None`.
> 
>     * This is problematic because these methods advance the ptr by one ‒ which happened if the `Option` was `Some(…)`. For `None` it didn't happen, thus causing havoc for all objects that subsequently tried to push, pull or clear their content from storage.
> 
> 
> (I don't want to commit the contract ofc, will revert that commit before merging).

Thanks for fixing! This is exactly what I thought was wrong. :)

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -545,4 +545,63 @@ mod tests {
...
+        type V2 = u32;
+
+        run_test::<ink_env::DefaultEnvironment, _>(|_| {
+            let root_key = Key::from([0x00; 32]);
+            {
+                // Step 1: Push two valid values one after the other to contract storage.
+                // The first value needs to be an Option, since the bug was messing up
+                // following pointers for an Option.
+                let v1: V1 = None;
+                let v2: V2 = 13;
```
Robbepop, line 561: Why not simply use `Option<u32>` and `u32` instead of `V1` and `V2`?

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -545,4 +545,63 @@ mod tests {
...
+    fn regression_test_for_issue_570() -> ink_env::Result<()> {
+        type V1 = Option<u32>;
+        type V2 = u32;
+
+        run_test::<ink_env::DefaultEnvironment, _>(|_| {
+            let root_key = Key::from([0x00; 32]);
+            {
+                // Step 1: Push two valid values one after the other to contract storage.
+                // The first value needs to be an Option, since the bug was messing up
+                // following pointers for an Option.
```
Robbepop, line 559:
```suggestion
                // following pointers for an `Option::None` value.
```

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -545,4 +545,63 @@ mod tests {
...
+                let v1: V1 = None;
+                let v2: V2 = 13;
+                let mut ptr = KeyPtr::from(root_key);
+
+                let v1_cell = v1;
```
Robbepop, line 564: why do we have this no-op assignment?

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -545,4 +545,63 @@ mod tests {
...
+                SpreadLayout::push_spread(&v1_cell, &mut ptr);
+
+                let v2_cell = v2;
```
Robbepop, line 567: Ẁhy do we have this no-op assignment?

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -545,4 +545,63 @@ mod tests {
             Ok(())
         })
     }
+
+    #[test]
+    fn regression_test_for_issue_570() -> ink_env::Result<()> {
```
Robbepop, line 550: The entire test can be heavily simplified. You only need to push a `(Option, i32)` tuple with values `(None, 1)` to the contract storage and then read it again in the next step and assert that the first value is `None` and the second value is `1`. Done.

Robbepop: Or am I missing something?

Robbepop (commented):
Thanks for the bug fix and clippy suggestion improvement.

crates/storage/src/lazy/lazy_cell.rs
cmichi: So that's a clever idea, I guess you thought the `1` would be written out to the first pointer and on the next `pull_spread` we would try to read a `Some(...)` there and fail. It doesn't work unfortunately because of this: `push_spread` for `Option` writes `is_some()` in the first ptr slot and increments the pointer. Hence on the next `pull_spread` we immediately return the `None` and continue pulling the `i32`.

cmichi: Using a tuple also doesn't simplify the test significantly, because for each tuple element the entire `push_spread` recursion is called. So storage is basically overwritten until I introduce `LazyCell` into the tuple.

So overall I'm quite happy with the simplicity of the test. But yeah lmk what you think.

I mean it could also be simplified to something like
```
assert_eq!(ptr_after_pushing_none, ptr_after_pushing_some);
assert_eq!(ptr_after_pulling_none, ptr_after_pulling_some);
``

Robbepop: Thanks for the elaboration!

Robbepop (approved):
Thanks a lot for this fix! We will probably release an ink! 3.0-rc3 soon with it.

PR #572, opened 2020-11-09
Fix CI on master

cmichi:
CI fails on `master` currently for the fuzz tests stage.

This is because `let` has some strange exit codes if the computation succeeds (`1`).


TriplEight (approved):

Robbepop (approved):
Thanks for the quick fix!

PR #574, opened 2020-11-11
Simplify regression test

cmichi:


crates/primitives/src/key_ptr.rs
```diff
@@ -18,7 +18,7 @@ use crate::Key;
 ///
 /// This wraps a base key and provides an interface to mimic pointer arithmetics.
 /// Mainly used to coordinate keys through static storage structures.
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
```
Robbepop, line 21: If these trait impls are only required for testing we might want to put them under `cfg_attr(test, ..)`

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -549,52 +549,38 @@ mod tests {
...
-                assert_eq!(pulled_v2, 13);
-            }
+            let none: Option<u32> = None;
+            let some: Option<u32> = Some(13);
+
+            // when
+            let mut ptr_push_none = KeyPtr::from(root_key);
+            SpreadLayout::push_spread(&none, &mut ptr_push_none);
+            let mut ptr_pull_none = KeyPtr::from(root_key);
+            let _: Option<u32> = SpreadLayout::pull_spread(&mut ptr_pull_none);
```
Robbepop, line 561: maybe check that this is in fact `None`?

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -549,52 +549,38 @@ mod tests {
...
+            let mut ptr_clear_none = KeyPtr::from(root_key);
+            SpreadLayout::clear_spread(&none, &mut ptr_clear_none);
+
+            let mut ptr_push_some = KeyPtr::from(root_key);
+            SpreadLayout::push_spread(&some, &mut ptr_push_some);
+            let mut ptr_pull_some = KeyPtr::from(root_key);
+            let _: Option<u32> = SpreadLayout::pull_spread(&mut ptr_pull_some);
```
Robbepop, line 568: check that this is `== some`

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -549,52 +549,38 @@ mod tests {
...
+            // was set so a different value if the `Option` was `None` vs. if it was
+            // `Some`.
+            //
+            // if the bug has been fixed the pointer must be the same for `None`
+            // and `Some` after push/pull/clear. otherwise subsequent operations using
+            // the pointer will break as soon as the `Option` is changed to it's
+            // opposite (`None` -> `Some`, `Some` -> `None`).
+            assert_eq!(ptr_push_none, ptr_push_some);
+            assert_eq!(ptr_pull_none, ptr_pull_some);
+            assert_eq!(ptr_clear_none, ptr_clear_some);
```
Robbepop, line 583: Note that this test breaks in case the `Option<T>` implementation is completely screwed atm without additional checks, e.g. never mutates the `KeyPtr`. So I think the former test does actually assert better our intention. This would be fixed if we put the mentioned assertions in and also put some comparison into place with some expected key after the operations.

crates/primitives/src/key_ptr.rs
cmichi: So it's a good idea that I hadn't thought about. But as it turns out when we run tests for crate `storage`, crate `primitives` (which contains the above `KeyPtr`) will not have `cfg_attr(test)` true.

There's discussion about an RFC [here](https://github.com/rust-lang/cargo/issues/2911#issuecomment-625731513).

Robbepop (approved):
LGTM

Thanks for the additional test that makes it more clear what had regressed.

PR #575, opened 2020-11-11
Update to new `cargo-contract` cli

cmichi:


Robbepop:
We should make sure that the `canvas-ui` is also updated first to support the new `cargo contract build` artifacts.

cmichi:
@Robbepop I would still already merge it, because atm the `cargo contract` `master` does not work with the ink! `master` instructions. And the `delegator` build script would fail as well. Imho it's more probable for people to run into this mismatch.

Hmm, I could remove mentions of `.contract` from the PR and we add them back once the UI is updated?

Robbepop:
> @Robbepop I would still already merge it, because atm the `cargo contract` `master` does not work with the ink! `master` instructions. And the `delegator` build script would fail as well. Imho it's more probable for people to run into this mismatch.
> 
> Hmm, I could remove mentions of `.contract` from the PR and we add them back once the UI is updated?

Let's first query @kwingram25 how long implementation of this will take.

ascjones:
> @Robbepop I would still already merge it, because atm the `cargo contract` `master` does not work with the ink! `master` instructions. And the `delegator` build script would fail as well. Imho it's more probable for people to run into this mismatch.

The instructions here should only be updated once we have a new release of `cargo contract`, because the installation instructions in the README here will install the latest release. 

But good that this PR is primed and ready to so that it can be merged as soon as the new `cargo-contract` is released.

Robbepop (commented):
I think this needs an update, right?

cmichi:
> I think this needs an update, right?

I just merged master into it, but besides that I think it's ready.

@ascjones I think it should still be up to date and merge ready, wdyt?

ascjones (approved):
LGTM

PR #583, opened 2020-11-17
Make nested Lazy's clear storage properly

### LINKED ISSUE #580: Leftovers when clearing storage of nested Lazy-like structures
I found a bug whereby we have leftovers in storage for nested Lazy-like structs.

In the example below `Vec.elems: LazyIndexMap` is the offender, though e.g. our `HashMap` is also affected by the bug.

```
#[test]
fn bug() {
    ink_env::test::run_test::<ink_env::DefaultEnvironment, _>(|_| {
        // given
        let root_key = Key::from([0x42; 32]);
        let mut vec: Lazy<StorageVec<u32>> = Lazy::new(StorageVec::new());
        vec.push(13u32);
        SpreadLayout::push_spread(&vec, &mut KeyPtr::from(root_key));
        let _ = <Lazy<StorageVec<u32>> as SpreadLayout>::pull_spread(&mut KeyPtr::from(root_key));

        // when
        SpreadLayout::clear_spread(&vec, &mut KeyPtr::from(root_key));

        // then
        let storage_used = ink_env::test::get_contract_storage_used::<
            ink_env::DefaultEnvironment,
        >(&contract_id()).expect("used storage must be returned");
        assert_eq!(storage_used, 0);

        Ok(())
    }).unwrap()
}
```

Fails with 
```
thread 'collections::vec::tests::bug' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', crates/storage/src/collections/vec/tests.rs:413:9
```

cmichi:
Closes #580.

The bug is that a `Lazy` doesn't clear up it's inner cell for `clear_spread`, if the inner cell has not been loaded. This can easily happen with nested `Lazy`'s ‒ like `Lazy<Lazy<…>>`.

But also we often use `Lazy` in storage structs like `StorageHashMap<…>` (with `values: LazyHashMap<…>` ‒ if these structs are then wrapped in a `Lazy` there is already a nested `Lazy`. `SmallVec`, `Vec`, `Stash` or other storage structs which have this bug if they are wrapped in a `Lazy`.

The way I fixed this for now will probably lead to some discussion and I'm open for other approaches.

### DIFF
--- a/crates/env/src/engine/off_chain/db/accounts.rs
+++ b/crates/env/src/engine/off_chain/db/accounts.rs
@@ -267,6 +267,12 @@ impl Account {
     pub fn get_storage_rw(&self) -> Result<(usize, usize)> {
         self.contract_or_err().map(|contract| contract.get_rw())
     }
+
+    /// Returns the amount of used storage entries.
+    pub fn get_storage_used(&self) -> Result<usize> {
+        self.contract_or_err()
+            .map(|contract| contract.get_storage_used())
+    }
 }
 
 /// The kind of the account.
@@ -300,6 +306,11 @@ impl ContractAccount {
     /// Returns the number of reads and writes from and to the contract storage.
     pub fn get_rw(&self) -> (usize, usize) {
         self.storage.get_rw()
+    }
+
+    /// Returns the number of used storage entries.
+    pub fn get_storage_used(&self) -> usize {
+        self.storage.storage_used()
     }
 }
 
@@ -354,5 +365,10 @@ impl ContractStorage {
     pub fn clear_storage(&mut self, at: Key) {
         self.count_writes += 1;
         self.entries.remove(&at);
+    }
+
+    /// Returns the number of used storage entries.
+    pub fn storage_used(&self) -> usize {
+        self.entries.len()
     }
 }
--- a/crates/env/src/engine/off_chain/db/accounts.rs
+++ b/crates/env/src/engine/off_chain/db/accounts.rs
@@ -271,7 +271,7 @@ impl Account {
     /// Returns the amount of used storage entries.
     pub fn count_used_storage_cells(&self) -> Result<usize> {
         self.contract_or_err()
-            .map(|contract| contract.get_storage_used())
+            .map(|contract| contract.count_used_storage_cells())
     }
 }
 
@@ -310,7 +310,7 @@ impl ContractAccount {
 
     /// Returns the number of used storage entries.
     pub fn count_used_storage_cells(&self) -> usize {
-        self.storage.storage_used()
+        self.storage.count_used_storage_cells()
     }
 }
 
--- a/crates/env/src/engine/off_chain/db/accounts.rs
+++ b/crates/env/src/engine/off_chain/db/accounts.rs
@@ -269,7 +269,7 @@ impl Account {
     }
 
     /// Returns the amount of used storage entries.
-    pub fn get_storage_used(&self) -> Result<usize> {
+    pub fn count_used_storage_cells(&self) -> Result<usize> {
         self.contract_or_err()
             .map(|contract| contract.get_storage_used())
     }
@@ -309,7 +309,7 @@ impl ContractAccount {
     }
 
     /// Returns the number of used storage entries.
-    pub fn get_storage_used(&self) -> usize {
+    pub fn count_used_storage_cells(&self) -> usize {
         self.storage.storage_used()
     }
 }
@@ -368,7 +368,7 @@ impl ContractStorage {
     }
 
     /// Returns the number of used storage entries.
-    pub fn storage_used(&self) -> usize {
+    pub fn count_used_storage_cells(&self) -> usize {
         self.entries.len()
     }
 }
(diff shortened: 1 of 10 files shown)


Robbepop:
Good catch!

The fixed `SpreadLayout::clear_spread` impl for `LazyCell` should look like this:

```rust
fn clear_spread(&self, ptr: &mut KeyPtr) {
    let root_key = ExtKeyPtr::next_for::<Self>(ptr);
    match <T as SpreadLayout>::REQUIRES_DEEP_CLEAN_UP {
        true => {
            // Load from storage and then clear:
            clear_spread_root_opt::<T, _>(root_key, || self.get())
        }
        false => {
            // Clear without loading from storage:
            let footprint = <T as SpreadLayout>::FOOTPRINT;
            assert!(footprint <= 16); // magic number
            let mut key_ptr = KeyPtr::from(*root_key);
            for _ in 0..footprint {
                ink_env::clear_contract_storage(key_ptr.advance_by(1));
            }
        }
    }
}
```

crates/env/src/engine/off_chain/db/accounts.rs
```diff
@@ -267,6 +267,12 @@ impl Account {
     pub fn get_storage_rw(&self) -> Result<(usize, usize)> {
         self.contract_or_err().map(|contract| contract.get_rw())
     }
+
+    /// Returns the amount of used storage entries.
+    pub fn get_storage_used(&self) -> Result<usize> {
```
Robbepop, line 272:
```suggestion
    pub fn count_used_storage_cells(&self) -> Result<usize> {
```

crates/env/src/engine/off_chain/db/accounts.rs
```diff
@@ -301,6 +307,11 @@ impl ContractAccount {
     pub fn get_rw(&self) -> (usize, usize) {
         self.storage.get_rw()
     }
+
+    /// Returns the number of used storage entries.
+    pub fn get_storage_used(&self) -> usize {
```
Robbepop, line 312:
```suggestion
    pub fn count_used_storage_cells(&self) -> usize {
```

crates/env/src/engine/off_chain/db/accounts.rs
```diff
@@ -355,4 +366,9 @@ impl ContractStorage {
         self.count_writes += 1;
         self.entries.remove(&at);
     }
+
+    /// Returns the number of used storage entries.
+    pub fn storage_used(&self) -> usize {
```
Robbepop, line 371:
```suggestion
    pub fn count_used_storage_cells(&self) -> usize {
```

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -352,6 +352,41 @@ where
     })
 }
 
+/// Returns the amount of storage entries used by the account `account_id`.
+///
+/// Returns `None` if the account is non-existent.
+pub fn get_storage_used<T>(account_id: &T::AccountId) -> Result<usize>
```
Robbepop, line 358:
```suggestion
pub fn count_used_storage_cells<T>(account_id: &T::AccountId) -> Result<usize>
```

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -352,6 +352,41 @@ where
...
+            .get_account::<T>(account_id)
+            .ok_or_else(|| AccountError::no_account_for_id::<T>(account_id))
+            .map_err(Into::into)
+            .and_then(|account| account.get_storage_used().map_err(Into::into))
+    })
+}
+
+/// Returns the amount of storage entries used by the account `account_id`.
+///
+/// Returns `None` if no current account is existent.
```
Robbepop, line 374: broken docs

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -352,6 +352,41 @@ where
...
+pub fn get_current_contract_storage_used<T>() -> Result<usize>
```
Robbepop, line 375: maybe we should not provide such convenience methods since if we do we will completely pollute the entire API.

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -151,8 +151,22 @@ where
...
+            true => {
+                // The inner cell needs to be cleared, no matter if it has
+                // been loaded or not. Otherwise there might be leftovers.
+                // Load from storage and then clear:
+                clear_spread_root_opt::<T, _>(root_key, || self.get())
+            }
+            false => {
+                // Clear without loading from storage:
+                let footprint = <T as SpreadLayout>::FOOTPRINT;
+                assert!(footprint <= 16); // magic number
```
Robbepop, line 164: maybe we want to unify these magic numbers with some global constant. could be done in a follow up though.

Robbepop (changes requested):

crates/storage/src/lazy/lazy_cell.rs
cmichi: Good point, I'll do it in a follow-up.

PR #594, opened 2020-11-25
Ensure topics are unique

### LINKED ISSUE #580: Leftovers when clearing storage of nested Lazy-like structures

### LINKED ISSUE #589: When set '0' value in contracts event, may cause `Error::ContractTrapped` and panic in contract
**When set '0' value in contracts event, may cause `Error::ContractTrapped` and panic in contract**
Setting 0 to strorage value is ok, but set 0 to event value, would cause this `ContractTrapped`

This is a bug when I try to test contract example like "erc20" or something else.

ink commit: 3803a2662e89dfa97b6f8b17e87c0cce2d873f48
substrate contract module commit: 
* substrate tag 2.0.0
* 69198d1e5735798c8baa1921c6e2091c93260752
(I test this problem in substrate v2.0.0 version and newest substrate version)

**Describe the bug**
This bug happen when I try to test erc20 example contract. First I run canvas-ui and canvas-node, then I compile erc20 contracts, and use canvas-ui to put-code then deploy contract.

But when I try deploy contract, I pass "0" for `initialSupply` parameter（meaning init value is zero）,then, deploy process return an Error `Error::ContractTrapped`.

I use newest substrate node has the same error. Then I try to figure out the reason. I modified the erc20 example, and find that, when I set 0 to the event value, the deploy process must cause this error:
in erc20 contract:
```rust
        #[ink(constructor)]
        pub fn new(initial_supply: Balance) -> Self {
          // ... other code
            Self::env().emit_event(Transfer {
                from: None,
                to: Some(caller),
                value: 0_u128.into(),  // here, I set 0 for value field
            });
            ::ink_env::debug_println("Erc20 contract init finish");
            instance
        }
```
And If set `value` to other value, like set `value: 1_u128.into()`, this contract could run normally. 

Then I trace in substrate contract module, and find out the error `Error::ContractTrapped` is from:
`primitives/sandbox/with_std.rs` ( in Native execution mode)
```rust
impl<'a, T> Externals for GuestExternals<'a, T> {
	fn invoke_index(
		&mut self,
		index: usize,
		args: RuntimeArgs,
	) -> Result<Option<RuntimeValue>, Trap> {
// ...
		let result = (self.defined_host_functions.funcs[index])(self.state, &args);
		match result {
			Ok(value) => Ok(match value {
				ReturnValue::Value(v) => Some(v.into()),
				ReturnValue::Unit => None,
			}),
			Err(HostError) => (TrapKind::Host(Box::new(DummyHostError)).into()),  // here return "HostError"
		}
	}
```
this function return `HostError` and convert to `TrapKind::Host` Error.
and then
```rust
	pub fn invoke(
		&mut self,
		name: &str,
		args: &[Value],
		state: &mut T,
	) -> Result<ReturnValue, Error> {
// ...
		match result {
			Ok(None) => Ok(ReturnValue::Unit),
			Ok(Some(val)) => Ok(ReturnValue::Value(val.into())),
			Err(_err) => (Error::Execution),
}
```
convert `TrapKind::Host` Error to `Error::Execution`
and finally in `pallet-contracts`, detail in `frame/contracts/src/wasm/runtime.rs` L200:
```rust
		Err(sp_sandbox::Error::Execution) | Err(sp_sandbox::Error::OutOfBounds) =>
			Err(Error::<E::T>::ContractTrapped)?
```
convert `Error::Execution` to `Error::<E::T>::ContractTrapped` error.

**Expected behavior**
contracts should run normally.

cmichi:
Closes #589.


crates/lang/src/events.rs
```diff
@@ -37,3 +39,24 @@ pub trait BaseEvent {
...
+{
+    #[inline(always)]
+    fn size_hint(&self) -> usize {
+        self.prefix.encode().len() + self.value.encode().len()
+    }
+
+    #[inline]
+    fn encode_to<T: scale::Output>(&self, dest: &mut T) {
+        dest.write(self.prefix.encode().as_slice());
+        dest.write(self.value.encode().as_slice());
```
Robbepop, line 60: using `encode` like this will allocate dynamic heap memory. please  use `encode_to` instead.

Robbepop (changes requested):

crates/lang/src/events.rs
Robbepop, line 60:
```suggestion
        dest.write(self.prefix.encode_to(dest));
        dest.write(self.value.encode_to(dest));
```

crates/lang/src/events.rs
```diff
@@ -37,3 +39,24 @@ pub trait BaseEvent {
...
+    pub value: &'b T,
+}
+
+impl<X> scale::Encode for PrefixedValue<'_, '_, X>
+where
+    X: scale::Encode,
+{
+    #[inline(always)]
+    fn size_hint(&self) -> usize {
+        self.prefix.encode().len() + self.value.encode().len()
```
Robbepop, line 54:
```suggestion
        self.prefix.size_hint() + self.value.size_hint()
```

crates/lang/src/lib.rs
```diff
@@ -58,6 +58,7 @@ pub use self::{
     events::{
         BaseEvent,
         EmitEvent,
+        PrefixedValue,
```
Robbepop, line 61: We should not publicly make this available since I consider it to be kind of an implementation detail.

Robbepop (changes requested):

crates/env/src/topics.rs
```diff
@@ -201,3 +201,25 @@ pub trait Topics {
         E: Environment,
         B: TopicsBuilderBackend<E>;
 }
+
+#[doc(hidden)]
+pub struct PrefixedValue<'a, 'b, T> {
+    pub prefix: &'a [u8],
+    pub value: &'b T,
+}
```
Robbepop, line 209: Even though this has hidden docs I'd love it to have some docs explaining the rationals behind its existence to a future ink! developer. ;)

crates/lang/codegen/src/generator/events.rs
```diff
@@ -236,16 +235,19 @@ impl<'a> Events<'a> {
...
                         .map(quote::ToTokens::into_token_stream)
                         .unwrap_or_else(|| quote_spanned!(span => #n));
-                    let field_type = topic_field.ty();
+                    let signature = syn::LitByteStr::new(
+                        format!("{}::{}::{}", contract_ident, event_ident,
+                            field_ident
+                        ).as_bytes(), span);
                     quote_spanned!(span =>
-                        .push_topic::<#field_type>(&self.#field_ident)
+                        .push_topic(&::ink_env::topics::PrefixedValue { value: &self.#field_ident, prefix: #signature })
```
Robbepop, line 243: I know it sounds strange but in macro generated code always try to be as explicit as possible. This includes re-introducing the exact type of the `PrefixedValue<T>`. ;)

crates/lang/codegen/src/generator/events.rs
```diff
@@ -236,16 +235,19 @@ impl<'a> Events<'a> {
...
                     )
                 });
             // Only include topic for event signature in case of non-anonymous event.
             let event_signature_topic = match event.anonymous {
                 true => None,
                 false => Some(quote_spanned!(span=>
-                    .push_topic::<[u8; #len_event_signature]>(#event_signature)
+                    .push_topic(&::ink_env::topics::PrefixedValue { value: #event_signature, prefix: b"" })
```
Robbepop, line 250: Please correct me if I am wrong but maybe we could have the `#event_signature` as `prefix` since it actually is its prefix while having no real value. Maybe this is confusing though but it would unify the approach taken for fields and their values where basically the same information is the `prefix`.

crates/lang/macro/tests/unique_topics.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[ink(event)]
+    pub struct MyEvent {
+        #[ink(topic)]
+        v0: Option<AccountId>,
+        #[ink(topic)]
+        v1: Balance,
+        #[ink(topic)]
+        v2: u128,
+        #[ink(topic)]
+        v3: bool,
```
Robbepop, line 34: Maybe introduce some other field with literally the same type and emit value to be extra sure.

crates/lang/codegen/src/generator/events.rs
cmichi: It's a good idea. I tried it out though and stumbled upon that we have `pub prefix: &'a [u8]` and `pub value: &'b T`, so with `value: b""` we then run into `the trait `Encode` is not implemented for `PrefixedValue<'_, '_, [u8; 0]>``.

crates/env/src/topics.rs
```diff
@@ -201,3 +201,30 @@ pub trait Topics {
         E: Environment,
         B: TopicsBuilderBackend<E>;
 }
+
+#[doc(hidden)]
+/// For each topic a hash is generated. This hash must be unique
+/// for a field and its value. The `prefix` is concatenated
+/// with the `value` and this result is then hashed.
+/// The `prefix` is typically set to the path a field has in
+/// an event struct + the identifier of the event struct.
```
Robbepop, line 210:
```suggestion
/// For each topic a hash is generated. This hash must be unique
/// for a field and its value. The `prefix` is concatenated
/// with the `value` and this result is then hashed.
/// The `prefix` is typically set to the path a field has in
/// an event struct + the identifier of the event struct.
#[doc(hidden)]
```

Robbepop: Maybe some example would be nice here.

Robbepop (commented):
All in all looks good to me!
Just some minor fixes.

crates/env/src/topics.rs
cmichi: I added it :+1: .

Robbepop (approved):
LGTM

PR #595, opened 2020-11-25
Migrate magic number to global const

cmichi:
I tried all kind of things to transition the runtime assertion to a compiler error, but with no success. The hold up is always that we need to compare our magic number against `<T as SpreadLayout>::FOOTPRINT`, which is not available at compile time in Rust at the moment.

I had tried via:
* asserting in a `const fn`
* `static_assertions` crate
* `const_fn_assert` crate
* the `const_generics` feature

### DIFF
--- a/crates/storage/src/lazy/lazy_cell.rs
+++ b/crates/storage/src/lazy/lazy_cell.rs
@@ -161,7 +161,13 @@ where
             false => {
                 // Clear without loading from storage:
                 let footprint = <T as SpreadLayout>::FOOTPRINT;
-                assert!(footprint <= 16); // magic number
+                assert!(
+                    footprint <= crate::traits::MAX_FOOTPRINT,
+                    format!(
+                        "The footprint is limited to {}.",
+                        crate::traits::MAX_FOOTPRINT
+                    )
+                );
                 let mut key_ptr = KeyPtr::from(*root_key);
                 for _ in 0..footprint {
                     ink_env::clear_contract_storage(key_ptr.advance_by(1));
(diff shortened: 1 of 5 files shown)


crates/storage/src/traits/spread.rs
```diff
@@ -14,6 +14,11 @@
 
 use super::KeyPtr;
 
+/// The maximum number of adjunctive storage cells a type may require in
+/// order to be stored in the contract storage with `SpreadLayout`.
+/// The number is arbitrarily chosen. Might need adjustments later.
+pub const MAX_FOOTPRINT: u64 = 16;
```
Robbepop, line 20: This is not actually true. Also this needs renaming for the same reason. The constant is just used by some types in order to make sure that cleaning up behind them won't become way too expensive. Since we are missing Substrate's storage bulk removal feature we cannot do better than this, yet with our current architecture of `ink_storage`.

A better name would probably be `FOOTPRINT_CLEANUP_THRESHOLD`. And maybe we should set it to 32 because that's kinda Rust's magic number in general.

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -161,7 +161,10 @@ where
             false => {
                 // Clear without loading from storage:
                 let footprint = <T as SpreadLayout>::FOOTPRINT;
-                assert!(footprint <= 16); // magic number
+                assert!(
+                    footprint <= crate::traits::FOOTPRINT_CLEANUP_THRESHOLD,
+                    "The footprint limit is violated for this type!"
```
Robbepop, line 166:
```suggestion
                    "cannot clean-up a storage entity with a footprint of {}. maximum threshold for clean-up is {}.",
                    footprint,
                    footprint_threshold,
```

Robbepop: also make `footprint_threshold` a binding to `crate::traits::FOOTPRINT_CLEANUP_THRESHOLD`

Robbepop (approved):
LGTM

PR #596, opened 2020-11-25
Fix dead code

cmichi:


crates/storage/src/traits/layout/impls.rs
```diff
@@ -202,39 +202,3 @@ where
         )))
     }
 }
-
-// impl<T> StorageLayout for BinaryHeap<T>
```
cmichi, line 206: This implementation already exists here:
https://github.com/paritytech/ink/blob/master/crates/storage/src/collections/binary_heap/storage.rs#L37

Robbepop: I think you are confusing the `std` collections with the `ink_storage` collections here.

cmichi: Right. What do we do about the commented out code though? From my pov this PR is still valid, since atm it's rather confusing if it's there intentionally commented out, unintentionally commented out, or was just forgotten.

Robbepop: yeah it is good to remove this commented out code.

Robbepop: we can decide what to do with `std` collections at a later point. It isn't pressing for them.

Robbepop (approved):

PR #597, opened 2020-11-26
Make `Lazy::Drop` always clear up all its storage

cmichi:


crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -109,8 +112,18 @@ where
...
+                Some(entry) => {
+                    clear_spread_root_opt::<T, _>(key, || entry.value().into())
+                }
+                None => {
+                    // the value is not yet in the cache. we need it in there
+                    // though in order to properly clean up.
+                    crate::assert_footprint_threshold!(<T as SpreadLayout>::FOOTPRINT);
+                    let _ = self.get();
+                    let entry = self.entry().expect("entry must exist after prior get");
+                    clear_spread_root_opt::<T, _>(key, || entry.value().into())
```
Robbepop, line 125: Here, as with other clean up code we can make a differentiation using `SpreadLayout::REQUIRE_DEEP_CLEANING` constant to avoid loading from storage if not needed for clean-up purposes.

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -120,9 +120,16 @@ where
...
                     // though in order to properly clean up.
                     crate::assert_footprint_threshold!(<T as SpreadLayout>::FOOTPRINT);
-                    let _ = self.get();
-                    let entry = self.entry().expect("entry must exist after prior get");
-                    clear_spread_root_opt::<T, _>(key, || entry.value().into())
+                    if <T as SpreadLayout>::REQUIRES_DEEP_CLEAN_UP {
+                        let _ = self.get();
+                        let entry =
+                            self.entry().expect("entry must exist after prior get");
+                        clear_spread_root_opt::<T, _>(key, || entry.value().into())
```
Robbepop, line 127: Why not directly use the return of `self.get()`? ^^

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -120,9 +120,16 @@ where
...
+                    } else {
+                        // The type does not require deep clean-up so we can simply clean-up
+                        // its associated storage cell and be done without having to load it first.
+                        ink_env::clear_contract_storage(&key);
```
Robbepop, line 131: Here we need to clean up all `SpreadLayout::FOOTPRINT` neighbouring cells starting from `root_key` as done in other clean-up implementations.

Robbepop: This iterative clean-up that depends on the size of `FOOTPRINT` is the exact reason why we have the `assert_footprint_threshold` in the first place.

cmichi: Thanks, I also added a test which would have found this.

crates/storage/src/lazy/lazy_cell.rs
```diff
@@ -361,6 +381,20 @@ where
...
+        let footprint_threshold = crate::traits::FOOTPRINT_CLEANUP_THRESHOLD;
+        assert!(
+            $footprint <= footprint_threshold,
+            "cannot clean-up a storage entity with a footprint of {}. maximum threshold for clean-up is {}.",
+            $footprint,
+            footprint_threshold,
+        );
+    }
+}
+
```
Robbepop, line 397: why is this a macro btw? ^^ couldn't it be a normal function instead?

Robbepop (approved):
LGTM!

Let's wait for CI to approve.

PR #598, opened 2020-11-26
Evaluate migration of `SmallVec` to `min_const_generics`

### LINKED ISSUE #588: Evaluate usage of `min_const_generics` for SmallVec
`min_const_generics` is about to be stabilized. It allows us to have `const N: usize` as a generic parameter in a type.

We should evaluate if we can get rid of our `generic-array` and `typenum` dependencies for `ink_storage::SmallVec`.

cmichi:
Closes #588.


crates/storage/src/lazy/lazy_array.rs
```diff
@@ -254,30 +216,24 @@ impl<'a, T> DoubleEndedIterator for EntriesIter<'a, T> {
...
-impl<T, N> EntryArray<T, N>
-where
-    N: LazyArrayLength<T>,
+impl<T, const N: usize> EntryArray<T, N>
 {
     /// Creates a new entry array cache.
     pub fn new() -> Self {
         Self {
-            entries: Default::default(),
+            entries:  [(); N].map(|_|Default::default())
```
cmichi, line 223: `#![feature(array_map)]` is required to make this work. The issue is that we need to fill the array initially and the array value `CacheCell` does not support `Copy`. So `[Default::default(); N]` does not work.

I tried using our `array-init` dependency, but from my understanding it doesn't support `min_const_generics` yet.

crates/storage/src/lazy/lazy_array.rs
```diff
@@ -289,7 +245,7 @@ where
     /// returns the old value if any.
     fn put(&self, at: Index, new_value: Option<T>) -> Option<T> {
         mem::replace(
-            unsafe { self.entries.as_slice()[at as usize].get_ptr().as_mut() },
+            unsafe { self.entries[at as usize].get_ptr().as_mut() },
```
cmichi, line 248: Not sure about if I can make this change. Hmm, at least everything builds and tests run.

If I include the `.as_slice()` I need `#![feature(array_methods)]`, otherwise I get build errors.

crates/storage/src/lazy/cache_cell.rs
```diff
@@ -36,11 +36,6 @@ impl<T> CacheCell<T> {
             inner: UnsafeCell::new(value),
         }
     }
-
-    /// Returns the inner value.
-    pub fn into_inner(self) -> T {
-        self.inner.into_inner()
-    }
```
Robbepop, line 43: Hm, ... so I guess this is dead code now after these changes. However, I am not sure if removing this actually makes sense since `CacheCell` is a low-level API intended for general use, not only ink! internal use.

crates/storage/src/lib.rs
```diff
@@ -20,6 +20,7 @@
 //! FFI to interface with SRML contracts and a primitive blockchain
 //! emulator for simple off-chain testing.
 
+#![feature(min_const_generics)]
```
Robbepop, line 23: This PR is mergable as soon as this feature is stabilizied.

crates/storage/src/lazy/lazy_array.rs
```diff
@@ -254,31 +214,28 @@ impl<'a, T> DoubleEndedIterator for EntriesIter<'a, T> {
...
-        }
+        let entries = iter::repeat_with(Default::default)
+            .take(N)
+            .collect::<Vec<CacheCell<Option<StorageEntry<T>>>>>()
+            .try_into();
+        let entries = match entries {
+            Ok(entries) => entries,
+            Err(_) => unreachable!("try_into must work"),
+        };
+        Self { entries }
```
Robbepop, line 228: Whoat? Looks way more complicated than before and also seems to be allocating heap memory which is against the entire point of small vec and lazy array data structures. Please elaborate why.

Robbepop (changes requested):

crates/storage/src/lazy/cache_cell.rs
cmichi: Brought it back.

crates/storage/src/lazy/lazy_array.rs
cmichi: Yeah… I pushed a different approach now, that one requires `#![feature(array_map)]` though 🙈.

Robbepop: Nightly Rust features are a show stopper for us. Please look if the `array_init` crate can help you here. We are already depending on it anyway.

cmichi:
@Robbepop Today I noticed that there already is a feature `const-generics` in the code for `array-init`. It does exactly what we need and I temporarily added it to this PR and migrated the code to use it. I talked to the maintainer and he plans to make it available by default as soon as `min_const_generics` are stabilized.

Yeah so overall we just need to wait until it's stabilized, remove the feature flags from this PR and are good to merge then.

I've marked the PR as blocked until then.

crates/storage/src/lib.rs
```diff
@@ -20,6 +20,8 @@
...
+#![feature(array_map)]
```
Robbepop, line 24: no longer needed ;)

crates/storage/Cargo.toml
```diff
@@ -25,8 +25,7 @@ scale = { package = "parity-scale-codec", version = "2.0", default-features = fa
 derive_more = { version = "0.99", default-features = false, features = ["from", "display"] }
 scale-info = { version = "0.5", default-features = false, features = ["derive"], optional = true }
 cfg-if = "1.0"
-array-init = "1.0"
-generic-array = "0.14.1"
+array-init = { version = "1.0", default-features = false, features = ["const-generics"] }
```
cmichi, line 28: We still have to pass this feature, since this dependency will only make it available be default once `min_const_generics` is in a stable release.

Robbepop: hmmm good point.
best would be to sync `array-init` with us or vice versa.

crates/storage/src/lazy/lazy_array.rs
```diff
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![cfg(not(feature = "ink-unstable"))]
-#![allow(dead_code)]
```
Robbepop: oh, good we removed that one

Robbepop:
We concluded to push this PR forward even though `min_const_generics` has not yet been stabilized.
Since we do not have a guarantee that `min_const_generics` is actually shipping with Rust version 1.51 in 25th of march we will put the `LazyArray` as well as everything that depends on it (namely `ink_storage::collections::SmallVec`) behind the new `ink-unstable` crate feature for the `ink_storage` crate.
This way users can decide themselves if they want to depend on this feature (as opt-in) and therefore risk being forced to nightly Rust for a longer time.

The big advantage is that as soon as `min_const_generics` is stabilized we can simply release the new `SmallVec` and `LazyArray` data structures without breaking user code.

crates/storage/src/lazy/cache_cell.rs
```diff
@@ -38,6 +38,7 @@ impl<T> CacheCell<T> {
     }
 
     /// Returns the inner value.
+    #[allow(dead_code)]
```
Robbepop, line 41: what needed for?

Robbepop (approved):
LGTM :rocket:

PR #600, opened 2020-11-30
[storage] Implement `Drop` for `Pack`

### LINKED ISSUE #581: Drop does not clear up properly for BinaryHeap/HashMap/Stash
I created/adapted `Drop` tests for some of our datastructures ([implementation here](https://gist.github.com/cmichi/da96bb34b5f3391ce13c3647748bb248)).

Those fail though:

```
failures:
    collections::binary_heap::tests::drop_works
    collections::hashmap::tests::drop_works
    collections::stash::tests::drop_works
```

It seems the issue is that `Pack` does not implement `Drop`, which from my understanding can result in corrupted storage since e.g. for `Stash` its `Drop` implementation currently only clears ` Stash::elems: LazyIndexMap<T>`, but not `Stash::header: Pack<Header>`.

cmichi:
Closes #581.

### DIFF
--- a/crates/primitives/src/key_ptr.rs
+++ b/crates/primitives/src/key_ptr.rs
@@ -41,4 +41,9 @@ impl KeyPtr {
         self.key += old_shift;
         &self.key
     }
+
+    /// Returns the underlying offset key.
+    pub fn key(&self) -> Key {
+        self.key
+    }
 }
--- a/crates/primitives/src/key_ptr.rs
+++ b/crates/primitives/src/key_ptr.rs
@@ -43,7 +43,7 @@ impl KeyPtr {
     }
 
     /// Returns the underlying offset key.
-    pub fn key(&self) -> Key {
-        self.key
+    pub fn key(&self) -> &Key {
+        &self.key
     }
 }
--- a/crates/storage/src/collections/binary_heap/mod.rs
+++ b/crates/storage/src/collections/binary_heap/mod.rs
@@ -33,6 +33,7 @@ use crate::{
 };
 
 pub use children_vec::Iter;
+use ink_primitives::Key;
 pub use reverse::Reverse;
 
 /// A priority queue implemented with a binary heap.
@@ -50,6 +51,14 @@ where
 {
     /// The individual elements of the heap.
     elements: ChildrenVec<T>,
+    /// The offset key for the N cells.
+    ///
+    /// If the lazy chunk has been initialized during contract initialization
+    /// the key will be `None` since there won't be a storage region associated
+    /// to the lazy chunk which prevents it from lazily loading elements. This,
+    /// however, is only checked at contract runtime. We might incorporate
+    /// compile-time checks for this particular use case later on.
+    key: Option<Key>,
 }
 
 impl<T> BinaryHeap<T>
@@ -60,6 +69,7 @@ where
     pub fn new() -> Self {
         Self {
             elements: ChildrenVec::new(),
+            key: None,
         }
     }
 
--- a/crates/storage/src/collections/binary_heap/mod.rs
+++ b/crates/storage/src/collections/binary_heap/mod.rs
@@ -33,7 +33,6 @@ use crate::{
 };
 
 pub use children_vec::Iter;
-use ink_primitives::Key;
 pub use reverse::Reverse;
 
 /// A priority queue implemented with a binary heap.
@@ -51,14 +50,6 @@ where
 {
     /// The individual elements of the heap.
     elements: ChildrenVec<T>,
-    /// The offset key for the N cells.
-    ///
-    /// If the lazy chunk has been initialized during contract initialization
-    /// the key will be `None` since there won't be a storage region associated
-    /// to the lazy chunk which prevents it from lazily loading elements. This,
-    /// however, is only checked at contract runtime. We might incorporate
-    /// compile-time checks for this particular use case later on.
-    key: Option<Key>,
 }
 
 impl<T> BinaryHeap<T>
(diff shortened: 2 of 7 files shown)


crates/primitives/src/key_ptr.rs
```diff
@@ -41,4 +41,9 @@ impl KeyPtr {
         self.key += old_shift;
         &self.key
     }
+
+    /// Returns the underlying offset key.
+    pub fn key(&self) -> Key {
+        self.key
```
Robbepop, line 47:
```suggestion
    pub fn key(&self) -> &Key {
        &self.key
```

crates/storage/src/collections/binary_heap/mod.rs
```diff
@@ -50,6 +51,14 @@ where
...
     /// The individual elements of the heap.
     elements: ChildrenVec<T>,
+    /// The offset key for the N cells.
+    ///
+    /// If the lazy chunk has been initialized during contract initialization
+    /// the key will be `None` since there won't be a storage region associated
+    /// to the lazy chunk which prevents it from lazily loading elements. This,
+    /// however, is only checked at contract runtime. We might incorporate
+    /// compile-time checks for this particular use case later on.
+    key: Option<Key>,
```
Robbepop, line 61: What exactly has this to do with the PR? :sweat_smile:

crates/storage/src/pack.rs
```diff
@@ -79,6 +126,23 @@ impl<T> Pack<T> {
     pub fn as_inner_mut(pack: &mut Pack<T>) -> &mut T {
         &mut pack.inner
     }
+
+    /// Sets the key.
+    fn set_key(mut self, value: Option<Key>) -> Self {
+        self.key = value;
+        self
+    }
```
Robbepop, line 134: Why would you need this super low-level and fragile API?

crates/storage/src/pack.rs
```diff
@@ -79,6 +126,23 @@ impl<T> Pack<T> {
...
+}
+
+impl<T> Drop for Pack<T>
+where
+    T: PackedLayout,
+{
+    fn drop(&mut self) {
+        if let Some(key) = self.key {
+            clear_spread_root_opt::<T, _>(&key, || Some(&self.inner))
```
Robbepop, line 143: Not sure it makes sense to use `clear_spread_root_opt` if the `&self.inner` isn't actually an optional.

Robbepop (changes requested):
All in all nice fix and the impl mostly looks good, however, I really wonder about the `BinaryHeap` change.

crates/storage/src/pack.rs
cmichi: I switched to an alternative constructor now:
```
-        Pack::from(forward_pull_packed::<T>(ptr)).set_key(Some(ptr.key()))
+        let inner = forward_pull_packed::<T>(ptr);
+        Pack::new_with_key(inner, *ptr.key())
```
I mean, it's still an API function now, but at least the object can't be modified in hindsight.

crates/storage/src/collections/binary_heap/mod.rs
cmichi: Argh, it was a survivor from a different approach. Removed it now.

crates/storage/src/pack.rs
```diff
@@ -53,21 +55,74 @@ use ink_primitives::Key;
...
     pub fn new(value: T) -> Self {
-        Self { inner: value }
+        Self {
+            inner: value,
+            key: None,
+        }
     }
 
-    /// Returns the packed value.
-    pub fn into_inner(pack: Self) -> T {
```
cmichi, line 69: I removed this function, since it is only used in testing and keeping it would require adding `T: Copy`, since `move occurs because pack.inner has type T, which does not implement the Copy trait`.

I mean we could add the trait only for tests, but it will be quite some overhead code just to support three test functions, which can also access the field directly.

Robbepop: ah okay I see! :+1:

crates/storage/src/pack.rs
```diff
@@ -117,6 +117,14 @@ where
         }
     }
 
+    /// Creates a new packed value.
+    pub fn new_with_key(value: T, key: Key) -> Self {
+        Self {
+            inner: value,
+            key: Some(key),
+        }
+    }
```
Robbepop: Still super low-level API. Please think twice if you really need something like this and if you do think twice again if it really should be made public or even gated with something a la `#[cfg(test)]`.

Robbepop: It is not much better than the previous `set_key` method from an API perspective. Keep in mind that the `Pack` data structure is a high-level, user targeting data structure.

Robbepop (changes requested):

crates/storage/src/pack.rs
cmichi: So I've made the method non-public. But please not: this is no test feature, hence we can't gate it. This method is called from `Pack::SpreadLayout::pull_spread` and used to set the `key` behind which this object is on `Pack.key`. This key is needed for `Drop` to know what to clear.

crates/storage/src/pack.rs
```diff
@@ -79,6 +126,25 @@ impl<T> Pack<T> {
     pub fn as_inner_mut(pack: &mut Pack<T>) -> &mut T {
         &mut pack.inner
     }
+
+    /// Creates a new packed value with a `key`.
```
Robbepop, line 130: Please add comment to describe that this is a low-level API specifically designed for `spread_pull`. However, I still kinda don't really see its point besides perhaps some code dedupe.

Robbepop (approved):

PR #601, opened 2020-11-30
Add generalized collection fuzz testing

cmichi:
This PR adds a trait which enables generalized fuzz testing of any collection which implements it. The fuzz testing tests the collection's interaction with storage.

Tests are created using the macro `fuzz_storage!("test_identifier", StorageHashMap<u32, u32>);`. The second argument identifies the object to be generated by the fuzzer, pushed to storage, pulled from storage, mutated, pushed again, etc.. See `test_utils.rs` for a documentation of the tests which are created.

The only pre-requisite for collections is to implement `Clone` (needed by quickcheck) and `Quickcheck::Arbitrary`. The latter is a trait which describes how a custom data structure can be generated from a generated input.
So instead of the type `StorageHashMap<u32, u32>` you can input any type which fulfills these requirements to `fuzz_storage`. So if you use a more complex type like `StorageHashMap<u32, Option<(u32, bool, Option<u32>)>>`, then as long as each of those types implements `Clone` and `Arbitrary` the fuzzer can generate "random" instances.

My plan is to add the other data structures in a follow-up, but I first want to get some feedback on the general direction before I continue.


crates/storage/src/collections/hashmap/fuzz_tests.rs
```diff
@@ -172,3 +182,84 @@ fn fuzz_defrag(xs: Vec<i32>, inserts_each: u8) {
     })
     .unwrap()
 }
+
+impl<
+        K: Arbitrary + Ord + PackedLayout + Send + Clone + std::hash::Hash + 'static,
+        V: Arbitrary + PackedLayout + Send + Clone + 'static,
+    > Arbitrary for StorageHashMap<K, V>
```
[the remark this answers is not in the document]
cmichi: Strangely this is the result of `cargo fmt`.

crates/storage/src/collections/hashmap/fuzz_tests.rs
```diff
@@ -172,3 +185,80 @@ fn fuzz_defrag(xs: Vec<i32>, inserts_each: u8) {
...
+{
+    fn arbitrary(g: &mut Gen) -> StorageHashMap<K, V> {
+        let hmap = HashMap::<K, V>::arbitrary(g);
+        StorageHashMap::<K, V>::from_iter(hmap.into_iter())
```
Robbepop, line 196:
```suggestion
        StorageHashMap::<K, V>::from_iter(hmap)
```

Robbepop: note that I am not entirely sure with this. please merge my suggestion if it works. but iirc `from_iter` take its input as `T: IntoIter`.

crates/storage/src/collections/hashmap/fuzz_tests.rs
```diff
@@ -172,3 +185,80 @@ fn fuzz_defrag(xs: Vec<i32>, inserts_each: u8) {
...
+{
+    type Collection = StorageHashMap<K, V>;
+    type Item = (&'a K, &'a mut V);
+
+    /// Makes `self` equal to `instance2` by executing a series of operations
+    /// on `self`.
+    fn equalize(&mut self, instance2: &Self::Collection) {
+        let hmap_keys = self.keys().cloned().collect::<Vec<K>>();
+        for k in hmap_keys {
+            if instance2.get(&k).is_none() {
```
Robbepop, line 227:
```suggestion
            if !instance2.contains_key(&k) {
```

crates/storage/src/collections/hashmap/fuzz_tests.rs
```diff
@@ -172,3 +185,80 @@ fn fuzz_defrag(xs: Vec<i32>, inserts_each: u8) {
...
+                let _ = self.take(&k);
+            }
+        }
+
+        let template_keys = instance2.keys().cloned().collect::<Vec<K>>();
```
Robbepop, line 232: Not sure why this `collect` is needed. I guess it isn't?

crates/storage/src/collections/vec/fuzz_tests.rs
```diff
@@ -0,0 +1,84 @@
...
+    Pack,
+};
+
+use quickcheck::{
+    Arbitrary,
+    Gen,
+};
+use std::vec::Vec;
+
+impl<T: Arbitrary + PackedLayout + Send + Clone + 'static> Arbitrary for StorageVec<T>
```
Robbepop, line 32: please move these trait bounds into the `where` clause as done in the rest of the crate.

crates/storage/src/collections/vec/fuzz_tests.rs
```diff
@@ -0,0 +1,84 @@
...
+where
+    T: Clone,
+{
+    fn arbitrary(g: &mut Gen) -> StorageVec<T> {
+        let vec = Vec::<T>::arbitrary(g);
+        let mut svec = StorageVec::<T>::new();
+        vec.iter().for_each(|v| svec.push(v.clone()));
```
Robbepop, line 39: could probably use `from_iter` instead.

Robbepop (approved):
LGTM

PR #602, opened 2020-12-03
[ci] Test workspace docs

### LINKED ISSUE #563: CI: Add a check that `cargo doc --workspace --no-deps` results in no warnings.
We currently do not check if our automated docs are actually working properly.
Some common mistake is improper rendering of getting some intra crate links wrong.
A CI check should defend us against both and other failures in this regard.

cmichi:
Closes #563.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -124,13 +124,13 @@ docs:
   stage:                           workspace
   <<:                              *docker-env
   variables:
-      RUSTDOCFLAGS: -Dwarnings
+    RUSTDOCFLAGS: -Dwarnings
   artifacts:
-      name: "${CI_JOB_NAME}_${CI_COMMIT_REF_NAME}-doc"
-      when: on_success
-      expire_in: 7 days
-      paths:
-        - ./crate-docs/
+    name: "${CI_JOB_NAME}_${CI_COMMIT_REF_NAME}-doc"
+    when: on_success
+    expire_in: 7 days
+    paths:
+      - ./crate-docs/
   script:
     - cargo doc --no-deps --all-features
         -p scale-info -p ink_metadata
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -272,11 +272,6 @@ publish-docs:
   variables:
     GIT_DEPTH:                     0
     GIT_STRATEGY:                  none
-  rules:
-    - if: $CI_PIPELINE_SOURCE == "web"
-    - if: $CI_PIPELINE_SOURCE == "schedule"
-    - if: $CI_COMMIT_REF_NAME == "master"
-    - if: $CI_COMMIT_REF_NAME == "tags"
   script:
     - rm -rf /tmp/*
     - unset CARGO_TARGET_DIR
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -272,6 +272,11 @@ publish-docs:
   variables:
     GIT_DEPTH:                     0
     GIT_STRATEGY:                  none
+  rules:
+    - if: $CI_PIPELINE_SOURCE == "web"
+    - if: $CI_PIPELINE_SOURCE == "schedule"
+    - if: $CI_COMMIT_REF_NAME == "master"
+    - if: $CI_COMMIT_REF_NAME == "tags"
   script:
     - rm -rf /tmp/*
     - unset CARGO_TARGET_DIR
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -128,7 +128,6 @@ docs:
     - 'has_warnings=$(egrep -c "^warning: [1-9]+[0-9]* warnings emitted" <<< "$output" || true)'
     - if [ "$has_warnings" -eq 0 ]; then exit 0; fi
     - echo "$output"
-    - exit 1
 
 codecov:
   stage:                           workspace
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -121,12 +121,23 @@ test:
     - cargo test --verbose --all-features --no-fail-fast --workspace
 
 docs:
-  variables:
-    RUSTDOCFLAGS: -Dwarnings
   stage:                           workspace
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -120,6 +120,17 @@ test:
   script:
     - cargo test --verbose --all-features --no-fail-fast --workspace
 
+
+docs:
+    stage:                           workspace
+    <<:                              *docker-env
+    script:
+        - output=$( cargo doc --workspace --no-deps 2>&1 )
+        - 'has_warnings=$(egrep -c "^warning: [1-9]+[0-9]* warnings emitted" <<< "$output" || true)'
```
cmichi, line 129: I need to wrap the line in `'…'` because the colon is otherwise interpreted by GitLab as the start of some mapping.

.gitlab-ci.yml
```diff
@@ -120,6 +120,17 @@ test:
   script:
     - cargo test --verbose --all-features --no-fail-fast --workspace
 
+
+docs:
+    stage:                           workspace
```
TriplEight, line 125: there are two spaces everywhere

.gitlab-ci.yml
TriplEight: yeah, this is so dumb

.gitlab-ci.yml
```diff
@@ -120,6 +120,17 @@ test:
...
+    <<:                              *docker-env
+    script:
+        - output=$( cargo doc --workspace --no-deps 2>&1 )
+        - 'has_warnings=$(egrep -c "^warning: [1-9]+[0-9]* warnings emitted" <<< "$output" || true)'
+        - if [ "$has_warnings" -eq 0 ]; then exit 0; fi
+        - echo "$output"
+        - exit 1
```
TriplEight, line 132: not needed

.gitlab-ci.yml
```diff
@@ -120,6 +120,17 @@ test:
   script:
     - cargo test --verbose --all-features --no-fail-fast --workspace
 
+
+docs:
```
TriplEight, line 124: why not pass `-D warnings` to have it failed instead?
And why the new job instead of adding this to `publish-docs:` job?

TriplEight (changes requested):

.gitlab-ci.yml
cmichi: > why not pass -D warnings to have it failed instead?

Doesn't work for `cargo doc` unfortunately.

>And why the new job instead of adding this to publish-docs: job?

We want to run this for every PR, so that you know if you e.g. have broken intra-crate links in the doc comments. `publish-docs` is only run after merge.

.gitlab-ci.yml
cmichi: But won't it take the `$?` from the `echo` then as the last exit code, which would be `0` and thus the job would not show as failed?

TriplEight: Did you test if it exits with `$?` I'm not entirely sure
- it shouldn't exit with 1 regardless of what happened, there might be other failures
- it could exit with the other code

.gitlab-ci.yml
TriplEight: it [works](https://github.com/paritytech/substrate/blob/7e94f01af67c498605eb0206b7fca261f12a278c/.gitlab-ci.yml#L477) with `cargo doc`

TriplEight: then it would make sense to move the docks generation from that job to this one, and there just copy the artifacts from this job. To save some CO2 you know.

.gitlab-ci.yml
cmichi: So I tested it by now (the commit and it's revert commit are in this PR). My understanding ist that we need the `exit 1` to mark the job as a failure and if we leave it out it gets marked as a success. I tested it in the mentioned commits by removing `exit 1` and deliberately introducing an error, the job was still marked as success. Yeah so tl;dr is that I think we need it.

.gitlab-ci.yml
cmichi: > it [works](https://github.com/paritytech/substrate/blob/7e94f01af67c498605eb0206b7fca261f12a278c/.gitlab-ci.yml#L477) with `cargo doc`

So I had checked exactly that this morning, before I implemented it the way it is now. I just went back and re-checked to uphold my own mental sanity. tl;dr is that `export RUSTFLAGS="-D warnings"` doesn't have any effect on `rustdoc` (the flags for `rustdoc` are under `RUSTDOCFLAGS`, but it also doesn't have an effect to set it here).

I also created a simple `lib.rs` with broken doc comments:
```
/// Test
/// ```rust
/// [`Foobar`](`foo::does::not::exist`)
pub fn main() {}
```
If then setting the env to `-Dwarnings` and executing `cargo doc --verbose` these warnings are displayed, but the exit code is always `0`.

I also checked the documentation for `rustdoc` and its source code, but couldn't find any way to have the tool exit with 1 when it detects warnings. I hope I didn't miss something, also didn't find a possibility via search engines.

cmichi: > then it would make sense to move the docks generation from that job to this one, and there just copy the artifacts from this job. To save some CO2 you know.

That's a good point! I haven't really worked with artifacts before, my understanding is that they are collected automatically here, right?

I added `--all-features` to the docs generation here, to make the flags for both `cargo doc` invocations compatible. Is this already sufficient?

TriplEight: so you 
1. [set up](https://github.com/paritytech/substrate/blob/master/.gitlab-ci.yml#L480) the collection
2. [take care](https://github.com/paritytech/substrate/blob/master/.gitlab-ci.yml#L490) of putting them there
3. [ask](https://github.com/paritytech/substrate/blob/master/.gitlab-ci.yml#L681) artifacts from this job
4. [skip git](https://github.com/paritytech/substrate/blob/master/.gitlab-ci.yml#L687)
5. draw the rest of the owl

TriplEight (approved):
Perfect

PR #604, opened 2020-12-07
Implement `storage2::Stash::drain_with`

### LINKED ISSUE #459: Implement storage2::Stash::drain_with method
The `storage2::Stash` storage data structure has the weakness of having a pretty costly `take(index)` function since it is required to relink its left-over vacant entries in a doubly-list linked manner. Touching multiple storage entries is common.

Especially when a deep clean-up of some storage stash data structure is needed where we still need access to the cleaned-up values we are doomed to a lot of computation.
We can see usage of this in the existing `storage2::HashMap` implementation that uses `storage2::Stash` underlying for its keys. When `storage2::HashMap::drop` is called it needs exactly this kind of clean-up for the underlying storage stash.

By introducing the so-called `drain_with` API we can eliminate most of the overhead computations and focus on what really needs to be done:

```rust
impl Stash {
    pub fn drain_with<F>(&mut self, f: F)
    where
        F: FnMut(T),
    {
        // call `f` for every moved out `T`
        // at the end the storage Stash is empty and in a valid state
    }
}
```

The name `drain_with` is inferred from the standard library's `Vec::drain` method that has a very similar job.

## ToDo List

- [ ] Implement `storage2::Stash::drain_with` as shown above.
- [ ] Write unit tests to verify the correctness of `storage2::Stash::drain_with` in different use cases.
- [ ] Write benchmarks to compare the new `storage2::Stash::drain_with` against the alternative `into_iter` implementation or (even worse) a coordinated `take(at)`-loop.
- [ ] Make use of the new `storage2::Stash::drain_with` in the above mentioned `storage2::HashMap` implementation where appropriate.

cmichi:
Closes #459.

```
$ cargo bench --bench bench_stash drain_with
    Finished bench [optimized] target(s) in 0.06s
     Running /home/michi/projects/ink/target/release/deps/bench_stash-d1df3feee8eebfd8
Gnuplot not found, using plotters backend
Benchmarking Compare: `drain_with` and `take_all` (populated cache)/drain_with: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.7s, enable flat sampling, or reduce sample count to 60.
Compare: `drain_with` and `take_all` (populated cache)/drain_with                                                                             
                        time:   [1.3534 ms 1.3615 ms 1.3713 ms]
                        change: [-11.781% -10.232% -8.7286%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  2 (2.00%) high mild
  6 (6.00%) high severe
Compare: `drain_with` and `take_all` (populated cache)/take_all                                                                             
                        time:   [2.3021 ms 2.3119 ms 2.3254 ms]
                        change: [-0.6696% +0.0054% +0.7599%] (p = 0.99 > 0.05)
Found 11 outliers among 100 measurements (11.00%)
  6 (6.00%) high mild
  5 (5.00%) high severe

Compare: `drain_with` and `take_all` (empty cache)/drain_with                                                                             
                        time:   [4.7411 ms 4.7619 ms 4.7890 ms]
                        change: [-27.075% -22.755% -18.303%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe
Compare: `drain_with` and `take_all` (empty cache)/take_all                                                                            
                        time:   [7.4695 ms 7.5148 ms 7.5787 ms]
                        change: [-2.8380% -1.8163% -0.6815%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe
```

We currently don't have a consuming `IntoIterator for StorageStash<T>` to compare against, hence I benched against the next best thing ‒ `take_all`.


crates/storage/src/collections/stash/mod.rs
```diff
@@ -485,6 +485,34 @@ where
         }
     }
 
+    /// Drains this stash, calls `f` with each drained `T`.
+    pub fn drain_with<F>(&mut self, mut f: F)
```
Robbepop, line 489: Maybe similar to the `defrag` methods we could use some `Option<usize>` parameter that when `Some(value)` limits the maximum amount of elements that are going to be drained and when `None` simply has no limit on the number of drained elements.

Robbepop: Another option is to turn `F` into `FnMut(T) -> bool` where the return value `bool` states if more elements shall be drained if possible. So `true` means: Please drain more elements whereas `false` means, please stop draining more elements. This would be even more flexible.

crates/storage/src/collections/stash/mod.rs
```diff
@@ -485,6 +485,34 @@ where
...
+            let taken_entry = self.entries.put_get(index, None).expect("entry");
+            if let Entry::Occupied(value) = taken_entry {
+                // Call `f` for every moved out `T`.
+                f(value);
+            }
+
+            // only clear if needed. if we are in a lazy state we don't need to.
+            if clear {
+                self.entries.clear_packed_at(index);
+            }
```
Robbepop, line 505: As always we do not want to clear right away but rather let the cache do its job.

crates/storage/src/collections/stash/mod.rs
```diff
@@ -485,6 +485,34 @@ where
...
+        }
+
+        self.header.len = 0;
+        self.header.len_entries = 0;
+        self.header.last_vacant = 0;
```
Robbepop, line 510: This needs adjustment if we plan to make the above improvements with the limitation on drained elements.

crates/storage/src/collections/hashmap/impls.rs
```diff
@@ -45,7 +45,15 @@ where
     Key: From<<H as HashOutput>::Type>,
 {
     fn drop(&mut self) {
-        self.clear_cells();
+        if self.values.key().is_none() {
+            // We won't clear any storage if we are in lazy state since there
+            // probably has not been any state written to storage, yet.
+            return
+        }
```
Robbepop, line 52: Not connected to this PR really but still a very needed fix. How was this not discovered until now?

crates/storage/src/collections/stash/mod.rs
cmichi: So it's a good idea and I added the latter version.

Imo `HashMap::drop` should always `drop` everything, that's the expectation. There should not be an upper bound where it stops. Hence I still `drain_with` everything. A user can still use `drain_with` with some user-defined upper bound and appropriate handling if it's violated. So the user has some way of controlling sprawling gas costs.

cmichi:
@Robbepop Could you take another look? I updated the benches in the description and addressed your comments.

cmichi:
We decided on closing this issue for now. Our reasoning is that upon closer inspection we found the `HashMap::drop` function not actually being in need for a `drain_with` function ‒ it already performs optimally.

There is currently no requirement by anybody for adding this function to the `Stash` API and since we want to keep the API as minimal as possible we refrain from adding it until a need arises.

I tried to reconstruct how the original issue came to be and couldn't really find code smell. My assumption is that the issue was originally created in order to provide a performant API to clear the entire stash, opposed to a loop of `Stash::take`.

Some notes for if this issue is ever revisited:

(1) The conditional version which I had in this PR at some point is a good idea and could possibly be kept. The function signature hereby was
```pub fn drain_with<F>(&mut self, mut f: F) where F: FnMut(T) -> bool { }```, hence allowing for the closure to control if draining should continue.

(2) We discussed the possibility of having `drain_with` start at the very end of the stash (at `len_entries`). Each entry would be examined, occupied ones would be replace with `None` and the closure would be called for them. Vacant entries would also be replaced with `None` with possible relinking taking place (as in `defrag`).
Building the function this way has the advantage of not requiring a shadow variable of `len_entries` to support `Stash::drop`.

PR #611, opened 2020-12-08
Move hidden images folder as well

cmichi:
`mv /tmp/doc/* .` doesn't match the hidden `.images` folder, hence the images do not appear on https://paritytech.github.io/ink.

The copy of the `.images` folder happens a few lines above the change in this PR.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -298,6 +298,7 @@ publish-docs:
     # remove everything and restore generated docs, README and Jekyll config
     - rm -rf ./*
     - mv /tmp/doc/* .
+    - mv /tmp/doc/.images .
     # Upload files
     - git add --all --force
     - git status


.gitlab-ci.yml
```diff
@@ -299,6 +299,7 @@ publish-docs:
     # remove everything and restore generated docs, README and Jekyll config
     - rm -rf ./*
     - mv /tmp/doc/* .
+    - mv /tmp/doc/.images .
```
ascjones, line 302: What about doing `shopt -s dotglob` as suggested here https://askubuntu.com/a/259386/462538. This way it would copy all hidden files/folders.

cmichi: Yeah I thought about that too, but wasn't to sure about possible side effects, since we e.g. have a `.github` folder as well in the repo.

ascjones (approved):

PR #613, opened 2020-12-10
Trim docs whitespaces for metadata

cmichi:
I noticed that our generated contract metadata for the message/constructor docs always contains a space as a prefix to the actual String. I suppose this is because of the parsing logic for `/// Foobar`.

```bash
$ cat flipper.contract | jq ".. | .docs?  | select(.)" | egrep -v "\[|\]"
  " Creates a new flipper smart contract initialized with the given value."
  " Creates a new flipper smart contract initialized to `false`."
  " Flips the current value of the Flipper's bool."
  " Returns the current value of the Flipper's bool."
```

### DIFF
--- a/crates/metadata/src/tests.rs
+++ b/crates/metadata/src/tests.rs
@@ -169,3 +169,32 @@ fn spec_contract_json() {
         })
     )
 }
+
+#[test]
+fn trim_docs() {
+    // given
+    let name = "foo";
+    let cs = ConstructorSpec::from_name(name)
+        .selector(123_456_789u32.to_be_bytes())
+        .docs(vec![" foobar      "])
+        .done();
+    let mut registry = Registry::new();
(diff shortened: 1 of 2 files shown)


ascjones (approved):

crates/metadata/src/tests.rs
```diff
@@ -169,3 +169,32 @@ fn spec_contract_json() {
...
     )
 }
+
+#[test]
+fn trim_docs() {
+    // given
+    let name = "foo";
+    let cs = ConstructorSpec::from_name(name)
+        .selector(123_456_789u32.to_be_bytes())
+        .docs(vec![" foobar      "])
```
ascjones, line 179: Maybe this would be better if it was "" foo bar      ", because this test would also pass with just removing all spaces

cmichi: That's a good point, I'll include this change in another PR.

PR #616, opened 2020-12-15
Make contract docs visible to rustdoc

### LINKED ISSUE #336: Public contract items should appear in rustdoc
Currently, the public items of a contract (storage items, constructor, messages) dot not appear in the rustdoc because the contract macro puts them in a private module. Only the crate level documentation is visible.

They should appear in rustdoc as a public API to the contract.

cmichi:
Closes #336.
 
I need to make some of our example `mod` public, otherwise there will never be any documentation for the `pub`'s in that module. We should also think about if we want to change the `cargo contract new` template and the workshops.

In terms of how it currently looks I uploaded the output of `cargo doc --no-deps` for two examples:

* [flipper](http://tmp.creal.de/flipper-doc/flipper/index.html)
* [trait-erc20](http://tmp.creal.de/trait-erc20-doc/erc20/index.html)

I expect that we will need to discuss further details of what we want to hide/show in these docs, I created this PR also with the intention of starting this discussion and not with the idea of providing a perfect solution from the start. For example: Do we always want to hide `__ink` in the contract docs? Even if it's [the event base](http://tmp.creal.de/trait-erc20-doc/erc20/erc20/enum.__ink_EventBase.html)?

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Below you can see the code using the `ink_lang` version of ink!.
 use ink_lang as ink;
 
 #[ink::contract]
-mod flipper {
+pub mod flipper {
     /// The storage of the flipper contract.
     #[ink(storage)]
     pub struct Flipper {
--- a/crates/lang/codegen/src/generator/item_impls.rs
+++ b/crates/lang/codegen/src/generator/item_impls.rs
@@ -69,7 +69,7 @@ impl GenerateCode for ItemImpls<'_> {
                 #doc_item_impls
             )*
         };
-        tokens.into()
+        tokens
     }
 }
 
--- a/crates/lang/codegen/src/generator/item_impls.rs
+++ b/crates/lang/codegen/src/generator/item_impls.rs
@@ -48,6 +48,7 @@ impl GenerateCode for ItemImpls<'_> {
             .module()
             .impls()
             .map(|item_impl| self.generate_item_impl(item_impl));
+        let doc_item_impls = item_impls.clone();
         let no_cross_calling_cfg =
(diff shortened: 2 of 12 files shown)


Robbepop:
Actually this is only part of the solution. We also (finally) need dedicated `#[cfg(doc)]` aware ink! codegen that produces code specially for generating documentation using `cargo doc`.

cmichi:
>Actually this is only part of the solution. We also (finally) need dedicated #[cfg(doc)] aware ink! codegen that produces code specially for generating documentation using cargo doc. E.g. ink! will produce different output if it notices that it is compiling for cargo doc.

@Robbepop Did you take a look at the code?  That's exactly what I do e.g. [here](https://github.com/paritytech/ink/pull/616/files#diff-517dd12d245ab4f2f1cdfa3a94f8ef7b71dabd3a2833b94c7247f4eecd75730cR63-R70).

Robbepop (changes requested):
The direction of this PR to solve the underlying issue is to implicitly expose the API of the smart contract to the outside thus enforcing `pub` ink! module.
Unfortunately this forces smart contract authors to have this constraint of a `pub` module.
What I'd like to see instead is to let the users decide what items and definitions they want to expose as with a normal Rust library or application and instead produce very clean and simple contract code with proper documentation everywhere when `#[cfg(doc)]` is enabled.
This requires yet another complete codegen for this case next to `ink-as-dependency` and `std` in the context of metadata and testing.

cmichi:
Closed in favor of #632 as a follow-up.

PR #629, opened 2021-01-13
CI: disable CARGO_HOME caching

TriplEight:
Data race conditions in `CARGO_HOME` were never fixed, so it would be handled some other way.


Robbepop (approved):
Thanks for fixing the CI issues!

Issue #638, opened 2021-01-22
env().block_number() return BlockNumber type u32 instead of u64!


tylerztl:
ink! BlockNumber type is `u64`:
https://github.com/paritytech/ink/blob/master/crates/env/src/types.rs#L131

but, substrate/polkadot BlockNumber type is `u32`:
https://github.com/paritytech/polkadot/blob/master/core-primitives/src/lib.rs#L28

ink! contract call `self.env().block_number()` return error because of type mismatch!

PR #639, opened 2021-01-22
block number type change into u32

tylerztl:
fix [issue#638](https://github.com/paritytech/ink/issues/638)


Robbepop:
Thanks a lot for the fix!
I am still a bit confused since I always thought that `BlockNumber` was `u64` type.
Anyways, thanks to you help it is fixed now.

PR #645, opened 2021-01-27
add example chain extension contract

hammeWang:
add an example contract to show how to interact between the ink smart contract and the substrate chain.

Basically, this contract enables the user to fetch on-chain randomness and update the value inside the smart contract.

### DIFF
--- a/examples/rand-extension/lib.rs
+++ b/examples/rand-extension/lib.rs
@@ -1,0 +1,139 @@
+// Copyright 2018-2021 Parity Technologies (UK) Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


Robbepop:
Wow hey thanks a lot for the PR!
Will take a look briefly.

examples/rand-extension/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+        use super::*;
+        use ink_lang as ink;
+
+        /// We test if the default constructor does its job.
+        #[ink::test]
+        fn default_works() {
+            let rand_extension = RandExtension::default();
+            assert_eq!(rand_extension.get(), [0; 32]);
+        }
+    }
```
Robbepop, line 131: We also provide you with a way to register chain extensions to the off-chain environment for testing purposes. The random function could probably be registered there to return a random sequence and return it as `[u8; 32]` back to the caller.
https://paritytech.github.io/ink/ink_env/test/fn.register_chain_extension.html
It might be that the API does not yet work correctly.

hammeWang: @Robbepop I'll check it.

Robbepop: I will merge this as it seems to be in good shape and create a follow-up issue to add an off-chain test and assign you to it, okay?

Robbepop: https://github.com/paritytech/ink/issues/664

PR #649, opened 2021-01-28
Run dependabot for `examples`

cmichi:
Our `/Cargo.toml` contains
```
exclude = [
    "examples/",
]
```
hence we need so explicitly include `examples` for our dependabot.

We need to list each example individually, since `dependabot` unfortunately does not support wildcards yet (the tracking issue is https://github.com/dependabot/dependabot-core/issues/2178).

For the `delegator` example we don't need to include the `accumulator`, `subber`, etc. specifically, since they are listed in the workspace of `delegator`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,47 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  - package-ecosystem: "cargo"
+    directory: "/examples/contract-terminate"
+    schedule:
+        interval: "daily"
+  - package-ecosystem: "cargo"
+    directory: "/examples/contract-transfer"
+    schedule:
+        interval: "daily"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,3 +4,47 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  - package-ecosystem: "cargo"
```
Robbepop, line 7: Might it be possible to streamline this so that we do not have to touch this file whenever we add another example? I think this will likely be forgotten.

cmichi: See my description of this PR, wildcards are not yet possible.

cmichi: I'll create a follow-up issue so that we don't forget to migrate once wildcards are supported.

PR #663, opened 2021-01-28
Update to new scale-info and codec releases

ascjones:
Related https://github.com/paritytech/cargo-contract/pull/164

### DIFF
--- a/crates/metadata/src/lib.rs
+++ b/crates/metadata/src/lib.rs
@@ -48,12 +48,12 @@ use impl_serde::serialize as serde_hex;
 #[cfg(feature = "derive")]
 use scale_info::{
     form::{
+        FormString,
         PortableForm,
-        FormString
     },
     IntoPortable as _,
-    Registry,
     PortableRegistry,
+    Registry,
 };
 use serde::{
     de::DeserializeOwned,
@@ -91,7 +91,7 @@ impl InkProject {
 
 impl<S> InkProject<S>
 where
-    S: FormString
+    S: FormString,
 {
     /// Returns a read-only registry of types in the contract.
     pub fn registry(&self) -> &PortableRegistry<S> {
--- a/crates/metadata/src/lib.rs
+++ b/crates/metadata/src/lib.rs
@@ -47,25 +47,30 @@ use impl_serde::serialize as serde_hex;
 
 #[cfg(feature = "derive")]
 use scale_info::{
-    form::CompactForm,
-    IntoCompact as _,
+    form::{
+        PortableForm,
+        FormString
+    },
+    IntoPortable as _,
     Registry,
-    RegistryReadOnly,
+    PortableRegistry,
 };
 use serde::{
+    de::DeserializeOwned,
     Deserialize,
     Serialize,
 };
 
 /// An entire ink! project for metadata file generation purposes.
 #[derive(Debug, Serialize, Deserialize)]
-pub struct InkProject {
+#[serde(bound(deserialize = "S: DeserializeOwned"))]
+pub struct InkProject<S: FormString = &'static str> {
     #[serde(flatten)]
-    registry: RegistryReadOnly,
+    registry: PortableRegistry<S>,
     #[serde(rename = "storage")]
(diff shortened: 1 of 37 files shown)


crates/primitives/src/key.rs
```diff
@@ -236,10 +236,12 @@ const _: () = {
     };
 
     impl TypeInfo for Key {
+        type Identity = Self;
+
         fn type_info() -> Type {
             Type::builder()
                 .path(Path::new("Key", "ink_primitives"))
-                .composite(Fields::unnamed().field_of::<[u8; 32]>())
+                .composite(Fields::unnamed().field_of::<[u8; 32]>("[u64; 4]"))
```
Robbepop, line 244: the `[u64; 4]` is just an optimization but we want to serialize it as `[u8; 32]` ideally to avoid problems with endianess.

ascjones: Hm yeah it is still a `[u8; 32]`, I was just mimicking what the derive would put there for the "display name". However the derive would also have generated `field_of::<[u64; 4]>`, which I guess is the point of having this hand-rolled impl. I will change it.

crates/metadata/src/lib.rs
```diff
@@ -47,25 +47,30 @@ use impl_serde::serialize as serde_hex;
 
 #[cfg(feature = "derive")]
 use scale_info::{
-    form::CompactForm,
-    IntoCompact as _,
+    form::{
+        FormString,
```
Robbepop, line 51: What is a `FormString` bringing to the table?

crates/metadata/src/tests.rs
```diff
@@ -28,11 +28,11 @@ fn spec_constructor_selector_must_serialize_to_hex() {
...
         .done();
     let mut registry = Registry::new();
-    let compact_spec = cs.into_compact(&mut registry);
+    let portable_spec = cs.into_portable(&mut registry);
 
     // when
-    let json = serde_json::to_value(&compact_spec).unwrap();
-    let deserialized: ConstructorSpec<CompactForm> =
+    let json = serde_json::to_value(&portable_spec).unwrap();
+    let deserialized: ConstructorSpec<PortableForm<String>> =
```
Robbepop, line 35: Why `PortableForm<String>`? In which way is the `String` Rust type significantly involved with any of the serialization or deserialization? This seems to be a weak spot in the recent design changes where I have had some blind spot.

Robbepop:
All in all the PR LGTM, however I really wonder about some more recent design changes in the `scale-info` crate such as `FormString` and parameterizing `PortableForm` with a `String` type. Both of which do not really make a sense to me from an API design standpoint.

crates/metadata/src/tests.rs
ascjones: Because String implements `DeserializeOwned`, whereas the default is `&'static str`. The driver behind the whole thing is that currently substrate runtimes are not allowed Strings, and to that end `parity-scale-codec` only supports them in `std`.

I am open to alternatives to solve the problem with a better API.

ascjones:
> All in all the PR LGTM, however I really wonder about some more recent design changes in the `scale-info` crate such as `FormString` and parameterizing `PortableForm` with a `String` type. Both of which do not really make a sense to me from an API design standpoint.

See original rationale: https://github.com/paritytech/scale-info/pull/35#issue-533622957

It's because substrate runtimes discourage use of `String` and to that end `parity-scale-codec` only support strings in `std` See https://github.com/paritytech/parity-scale-codec/blob/master/src/codec.rs#L357.

One idea I still need to investigate is the possibility of not storing the metadata in the runtime itself, but doing something similar to `ink!` where we generate the metadata up front using `std`. That would avoid the problem of having to support both `& 'static str` and `String` because we could just use `std` and `String`

crates/metadata/src/tests.rs
Robbepop: conclusion from chat: we will open another issue on the `scale-info` crate to resolve this API design issue and will continue merging this PR.

Issue #631, opened 2021-01-13
[Feature] Dynamic trait based contract calling


Robbepop:
Since ink! 3.0-rc1 it is possible to define special trait definitions for ink! smart contract using the `#[ink::trait_definition]` proc. macro.

# Motivation

Defining and implementing such a trait works as expected with regard to some technical limitations.
Smart contracts today can some limited use of traits by calling or instantiating a contract using just the trait it implements.

However, this process is static and not dynamic, meaning that it is currently not possible to store a contract, e.g. in a `ink_storage::Lazy<dyn MyContractTrait>` or having an input parameter to a contract constructor or message with `contract: &dyn MyContractTrait` and be able to call constructors and messages defined in the trait dynamically on the provided contract instance.

This feature is critical to making trait definitions and implementation actually an integrated user experience for ink! smart contracts.

# Syntax

There are several different ways in which a smart contract might want to interact with another smart contract dynamically.

## As Message Parameter

It should be possible to declare a smart contract message or constructor with an input argument with `contract: &[mut] dyn MyContractTrait` where `MyContractTrait` is a trait definition that was defined using ink!'s `#[ink::trait_definition]` and `contract` is an instance of a smart contract (basically its `AccountId`) that implements this exact trait.

Then the ink! constructor or message receiving this argument can call any messages defined on the trait definition using the given `contract` instance. Note that it would not be possible to call trait definition constructors on `contract` since it represents an already instantiated contract.
Also we have to differentiate between `contract: &dyn MyContractTrait` and `&mut dyn MyContractTrait` where only the latter allows to call `&mut self` defined ink! messages.

### Example

```rust
#[ink(message)]
pub fn call_back(&self, callback: &dyn ContractCallback) {
    if self.condition() {
        callback.do_something();
    }
}
```
Or ...
```rust
#[ink(message)]
pub fn call_back_mut(&self, callback: &mut dyn ContractCallback) {
    if self.condition() {
        callback.mutate_something();
    }
}
```
Where `do_something` is an ink! message defined in the `ContractCallback` trait definition as `&self` message and `mutate_something` is an ink! message defined there as well as `&mut self` message:
```rust
#[ink::trait_definition]
pub trait ContractCallback {
    fn do_something(&self);
    fn mutate_something(&mut self);
}
```

[...]

PR #665, opened 2021-01-29
[WIP] Implement dynamic contract calls

Robbepop:
Implements https://github.com/paritytech/ink/issues/631.

### DIFF
--- a/crates/lang/codegen/src/generator/item_impls.rs
+++ b/crates/lang/codegen/src/generator/item_impls.rs
@@ -157,6 +157,7 @@ impl ItemImpls<'_> {
             #( #attrs )*
             impl #trait_path for #self_type {
                 type __ink_Checksum = [(); #checksum];
+                type __ink_ConcreteImplementer = ::ink_lang::NoConcreteImplementer;
 
                 #( #constructors )*
                 #( #messages )*
--- a/crates/lang/codegen/src/generator/trait_def.rs
+++ b/crates/lang/codegen/src/generator/trait_def.rs
@@ -66,6 +66,85 @@ impl<'a> TraitDefinition<'a> {
             fn #ident(#inputs) -> Self::#output_ident;
         )
     }
+
+    fn generate_for_constructor_never_call(
+        constructor: ir::InkTraitConstructor<'a>,
+    ) -> TokenStream2 {
+        let span = constructor.span();
+        let attrs = constructor.attrs();
+        let sig = constructor.sig();
+        let ident = &sig.ident;
+        let output_ident = format_ident!("{}Out", ident.to_string().to_camel_case());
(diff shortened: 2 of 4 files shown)


hugopeixoto:
Does the scope of this PR include the "As Storage Entity" functionality from #631?
I don't know much about ink internals so it's a bit hard for me to map it to the concepts in the todo list.

Robbepop:
> Does the scope of this PR include the "As Storage Entity" functionality from #631?
> I don't know much about ink internals so it's a bit hard for me to map it to the concepts in the todo list.

Is this a very much needed feature? Actually I would like to try to keep the scope of the initial MVP small.

hugopeixoto:
> Is this a very much needed feature? Actually I would like to try to keep the scope of the initial MVP small.

That's alright, I was sincerely asking if it's part of it or not, not requesting it to be included.

Robbepop:
> > Is this a very much needed feature? Actually I would like to try to keep the scope of the initial MVP small.
> 
> That's alright, I was sincerely asking if it's part of it or not, not requesting it to be included.

The part should open up possibility to add this feature at a later point in time. However, there are still some open questions as to how the storage-related parts should syntactically look like. There are some ways, non of them obviously the best.

xgreenx:
Do you have ETA for this cool change?(=

Robbepop:
> Do you have ETA for this cool change?(=

The latest commit no longer produces compile errors in the new codegen. So the new codegen basically works now. However, plenty of stuff still needs to be implemented. This whole thing is much more than just the new trait system as it will feature quite a bit other stuff, too. Will write about this more soon.

xgreenx:
Could you describe a little bit how static dependency will work, please?

Robbepop:
> Could you describe a little bit how static dependency will work, please?

Best described by this delegator example:
https://gist.github.com/Robbepop/eb32d8b7306681243461520c24b61d0f

Robbepop:
I just updated to `trybuild 1.0.49` which made it necessary to adjuts lots of UI tests.
Due to the fact that I had to touch all UI tests anyways I also moved them to where they actually belong.
Sorry about the huge diff, @ascjones . However, it is really just moves and trivial adjustments so just ignore mostly.

Robbepop:
> @Robbepop There are a number of changes to the way how ink! works in this PR and it would be helpful for the review to have the "Unreleased" section in `RELEASES.md` updated as well. Since we plan on a short-term release candidate as well, we need to do this soon anyway. Could you add this to the PR?

Done. However, I just saw that some of the things that have been implemented as side-PRs have not been included properly into the `RELEASES.md` for the ink! 3.0-rc6. E.g. addition of the new proc. macros `blake2`, `selector_id` and `selector_bytes`.

PR #667, opened 2021-01-31
add rust-toolchain file

honeywest:


Robbepop:
PR missing description and rational, CI is failing.
In the past we were having `rust-toolchain` files but they have since been removed for good reason.

PR #670, opened 2021-02-02
Bump codec and scale-info in rand-extension

hammeWang:


Robbepop (approved):
LGTM

Thanks for the fix!

PR #671, opened 2021-02-02
Debug why no issue is created for failing fuzz test

cmichi:
The CI for `master` currently fails because of 
```
test collections::hashmap::fuzz_tests::fuzz_removes ... FAILED
```

A GitHub issue should have been created automatically, but this hasn't happened.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -316,7 +316,7 @@ fuzz-tests:
     <<:                              *docker-env
     variables:
         # The QUICKCHECK_TESTS default is 100
-        QUICKCHECK_TESTS:            100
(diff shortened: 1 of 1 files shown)


cmichi:
The issue was that the GitLab exits a stage immediately when a commend returns an exit code != 0. Hence the commands following `cargo test` (like the `curl`) were not executed.

The fix follows the recommendation from https://docs.gitlab.com/ee/ci/yaml/#script.

Robbepop (approved):
have you tested if the fix actually fixes the issue?

cmichi:
> have you tested if the fix actually fixes the issue?

Yup, this was the result: https://github.com/paritytech/ink/issues/673.

Issue #662, opened 2021-01-28
Way the selector is calculated by `blake2_512` rather than `blake2_256`


atenjin:
**Questions**
Way the message selector is calculated by `blake2_512` rather than `blake2_256`

Now, the `pallet-contracts` or Substrate host_function for runtime just provide `blake2_256`, do not provide `blake2_512`. Thus in contract self, just could use `blake2_256` to calculate hash. 

However, the message selector id is calculated by `blake2_512` in compilation. Thus if I want to calculate a selector from a string, I can not do it now.

For example, in example/multisig_plain, the contract require user to provide 4 bytes selector, then using build_call to call another contract by this selector. Howevery the 4 bytes selector is not friendly for humen.

And in another case:

If I wanna to hard code  a selector call in contract, I just could set 4 bytes array, cannot write a string and calculate from this string:
```rust
// it's not friendly for maintenance
const selector: [u8;4] = [0x01, 0x02, 0x01, 0x02, ];

// it's friendly for maintenance 
const message_str: &'static [u8] = &b"transfer";
fn call_another(self) {
// do not have `seal_hash_blake2_512` now, thus could not do following thing:
let selector = self.env().seal_hash_blake2_512(message_str)[..4];
// build_call later
}
```

And in Solidity, user could calculate the abi like the following example:
```solidity
function test() {
string call = "f(uint256)";
x.call(bytes4(keccak256(call), a, b)
// or using 
x.call(abi.encodeWithSignature(call, a, b))
}
```

Thus if the selector is calculated by `blake2_256`, we can do the same thing like solidity, and abrove case. But `blake2_512` not.

I have no idea way the selector needs to be calculated by `blake2_512`, I think `blake2_256` is enough.

Robbepop:
Implemented in https://github.com/paritytech/ink/pull/680. Closed.

PR #680, opened 2021-02-04
Use BLAKE2b 256-bit hashes for selectors

Robbepop:
Closes: https://github.com/paritytech/ink/issues/662

Before this PR we accidentally used BLAKE2b 512-bit hashing.
Using 256-bit versions is important for smart contracts so that smart contracts can compute selector hashes on their own through the built-in contracts-pallet hash support which includes BLAKE2b 256-bit but not 512-bit.


cmichi (approved):
LGTM

PR #686, opened 2021-02-12
Fix build warnings + clippy errors from latest nightly

cmichi:
See https://gitlab.parity.io/parity/ink/-/pipelines/123903 for the warnings we get now.

### DIFF
--- a/crates/env/src/engine/off_chain/impls.rs
+++ b/crates/env/src/engine/off_chain/impls.rs
@@ -40,7 +40,6 @@ use crate::{
     TypedEnvBackend,
 };
 use core::convert::TryInto;
-use impl_serde::serialize as serde_hex;
 use ink_primitives::Key;
 use num_traits::Bounded;
 
@@ -278,10 +277,7 @@ impl EnvInstance {
             beneficiary,
             transferred: all,
         };
-        panic!(
-            "0x{}",
-            serde_hex::to_hex(&scale::Encode::encode(&res), false)
-        );
+        panic!(scale::Encode::encode(&res));
     }
 }
 
--- a/crates/env/src/engine/off_chain/impls.rs
+++ b/crates/env/src/engine/off_chain/impls.rs
@@ -278,7 +278,10 @@ impl EnvInstance {
             beneficiary,
             transferred: all,
         };
-        panic!("0x{}", serde_hex::to_hex(&scale::Encode::encode(&res), false));
+        panic!(
+            "0x{}",
+            serde_hex::to_hex(&scale::Encode::encode(&res), false)
+        );
     }
 }
 
--- a/crates/env/src/engine/off_chain/impls.rs
+++ b/crates/env/src/engine/off_chain/impls.rs
@@ -40,8 +40,8 @@ use crate::{
     TypedEnvBackend,
 };
 use core::convert::TryInto;
-use ink_primitives::Key;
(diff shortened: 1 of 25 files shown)


cmichi:
This PR is blocked until https://github.com/myrrlyn/funty/issues/3 is fixed.

crates/env/src/engine/off_chain/impls.rs
```diff
@@ -277,7 +277,7 @@ impl EnvInstance {
             beneficiary,
             transferred: all,
         };
-        panic!(scale::Encode::encode(&res));
+        std::panic::panic_any(scale::Encode::encode(&res))
```
Robbepop, line 280: not yet stabilized: ` This is a nightly-only experimental API. (panic_any #78500)`

Also why do we need it?

Robbepop:
> This PR is blocked until [myrrlyn/funty#3](https://github.com/myrrlyn/funty/issues/3) is fixed.

Might be worth of putting together a temporary workaround instead.

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -420,11 +421,18 @@ pub fn assert_contract_termination<T, F>(
...
-        .downcast_ref::<Vec<u8>>()
+    let encoded_input = value_any
+        .downcast_ref::<String>()
         .expect("panic object can not be cast");
+    let deserialized_vec = encoded_input
+        .replace("[", "")
+        .replace("]", "")
+        .split(", ")
+        .map(|s| u8::from_str(s).expect("u8 cannot be extracted from str"))
+        .collect::<Vec<u8>>();
```
cmichi, line 432: I dislike this hacky way of deserializing, but AFAIK it's the simplest way of getting the inverse of a `panic!("{:?}", Vec<u8>`). We can get rid of this as soon as `panic_any` is stablized, which should be with the next stable release soonish.

TriplEight:
please ignore `ci/gitlab/686`

crates/env/src/engine/off_chain/test_api.rs
Robbepop: Can you create an issue for that so that we do not forget?

crates/env/src/engine/off_chain/impls.rs
```diff
@@ -277,7 +277,7 @@ impl EnvInstance {
             beneficiary,
             transferred: all,
         };
-        panic!(scale::Encode::encode(&res));
+        panic!("{:?}", scale::Encode::encode(&res));
```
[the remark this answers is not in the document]
cmichi: Hmm, so I'm not sure.

I mean, even if we implement setting a tombstone, clearing storage, etc. we would still like to return the result of the contract eviction and it seems that `panic!` is the right way to go.

From my pov your suggestion would just be a semantic change of having something like `unimplemented!("we have not yet implemented everything, but anyway here's the result: …")` instead and changing that back to `panic!(…)` once we have implemented it would then be a breaking change.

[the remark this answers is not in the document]
Robbepop: Well hopefully this discussion is not needed soon due to the works on the upcoming off-chain environment. :P But since contract termination is not really properly implemented in the current off-chain environment I would not dare to actually just mark it as unimplemented. But yeah let's keep it as it is for now.

Robbepop (approved):
LGTM

Issue #705, opened 2021-02-26
the trait bound `__ink_ConstructorDispatchEnum: parity_scale_codec::codec::WrapperTypeDecode` is not satisfied


noahsalvadordenjo:
Running cargo +nightly build throws the following error 
```
the trait bound `__ink_ConstructorDispatchEnum: parity_scale_codec::codec::WrapperTypeDecode` is not satisfied
   --> /home/noone/flipper/flipper/lib.rs:5:1
    |
5   | #[ink::contract]
    | ^^^^^^^^^^^^^^^^ the trait `parity_scale_codec::codec::WrapperTypeDecode` is not implemented for `__ink_ConstructorDispatchEnum`
    | 
   ::: /home/noone/.cargo/registry/src/github.com-1ecc6299db9ec823/ink_env-3.0.0-rc2/src/api.rs:483:8
    |
483 |     T: scale::Decode,
    |        ------------- required by this bound in `decode_input`
    |
    = note: required because of the requirements on the impl of `parity_scale_codec::codec::Decode` for `__ink_ConstructorDispatchEnum`
    = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
```
cargo contract version 0.9.1
````

**Comment by trace-andreason** [comment]
seeing this as well. Here is the error running with the `-Z macro-backtrace` flag:
```
error[E0277]: the trait bound `__ink_ConstructorDispatchEnum: parity_scale_codec::codec::WrapperTypeDecode` is not satisfied
   --> /home/trace/Documents/tmp_contract/lib.rs:5:1
    |
5   | #[ink::contract]
    | ^^^^^^^^^^^^^^^^
    | |
    | the trait `parity_scale_codec::codec::WrapperTypeDecode` is not implemented for `__ink_ConstructorDispatchEnum`
    | in this macro invocation
    | 
   ::: /home/trace/.cargo/registry/src/github.com-1ecc6299db9ec823/ink_env-3.0.0-rc2/src/api.rs:483:8
    |
483 |     T: scale::Decode,
    |        ------------- required by this bound in `decode_input`
    | 
   ::: /home/trace/.cargo/registry/src/github.com-1ecc6299db9ec823/ink_lang_macro-3.0.0-rc2/src/lib.rs:511:1
    |
511 | pub fn contract(attr: TokenStream, item: TokenStream) -> TokenStream {
    | -------------------------------------------------------------------- in this expansion of `#[ink::contract]`
    |
[...]

PR #707, opened 2021-02-27
bump parity-scale-code to 2.0.1

trace-andreason:
trying to address issue #705, not 100% sure it does but probably :shrug:


trace-andreason:
looks like the dependency issue goes further than this, just going to close

PR #719, opened 2021-03-05
Temporarily disable clippy for `rand-extension` example

cmichi:
Our CI currently fails for this example due to an ICE in clippy nightly. 

This blocks us from generating a new docker image, which installs some new packages (`hunspell`, `binaryen`) into the CI. Some PR's ink! and `cargo-contract` are blocked because those packages are not available in the CI.


.gitlab-ci.yml
```diff
@@ -240,16 +240,22 @@ examples-clippy-std:
     - job:                         clippy-std
       artifacts:                   false
   script:
+    # Revert as soon as the clippy nightly ice fix lands
```
Robbepop, line 243: Please also create an issue pointing to this problem so that we do not forget about it.

cmichi: Done! https://github.com/paritytech/ink/issues/721

Robbepop (approved):

PR #723, opened 2021-03-06
Fix execution context parameters in DNS example

austinabell:
- Endowment and gas limit were flipped, which confused me when I used as reference, so figured I'd fix

I automatically format, so if you want me to keep the changes minimal and revert those, let me know :)

### DIFF
--- a/examples/dns/lib.rs
+++ b/examples/dns/lib.rs
@@ -20,7 +20,9 @@ use ink_lang as ink;
 mod dns {
     #[cfg(not(feature = "ink-as-dependency"))]
     use ink_storage::{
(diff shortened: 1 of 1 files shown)


Robbepop:
thank you very much for the fix. please also fix your formatting problems in the PR so we can merge. :)

austinabell:
> thank you very much for the fix. please also fix your formatting problems in the PR so we can merge. :)

Ah, my apologies, didn't notice the custom formatting file and I guess I was formatting the file specifically

Robbepop (approved):
Thanks a lot for the fix! :)

PR #734, opened 2021-03-15
Add support for new `seal_random` API

athei:
This adds support for the new `seal_random` API which will be added by https://github.com/paritytech/substrate/pull/8329 .

It is guarded behind a feature flag to allow supporting older chain versions which do not support the new API.

### DIFF
--- a/crates/env/Cargo.toml
+++ b/crates/env/Cargo.toml
@@ -63,3 +63,6 @@ std = [
     "sha3",
     "blake2",
 ]
+
+# Use the new `seal_random` API. Only activate if this is supported by the target chain.
+seal-random-v1 = []
--- a/crates/env/Cargo.toml
+++ b/crates/env/Cargo.toml
@@ -63,6 +63,3 @@ std = [
     "sha3",
     "blake2",
 ]
-
-# Use the new `seal_random` API. Only activate if this is supported by the target chain.
-seal-random-v1 = []
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -493,7 +493,37 @@ where
 /// # Errors
 ///
 /// If the returned value cannot be properly decoded.
+#[cfg(not(feature = "seal-random-v1"))]
 pub fn random<T>(subject: &[u8]) -> Result<T::Hash>
+where
+    T: Environment,
+{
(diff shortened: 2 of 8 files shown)


crates/env/Cargo.toml
```diff
@@ -63,3 +63,6 @@ std = [
     "sha3",
     "blake2",
 ]
+
+# Use the new `seal_random` API. Only activate if this is supported by the target chain.
+seal-random-v1 = []
```
Robbepop, line 68: I am not sure if this is how we should handle this kind of situation in ink!. It is completely fine for SEAL to provide multiple different versions. However, for ink! isn't it just fine to try to always use the latest version? Especially since we are currently still in release candidate mode where people can and must expect breakage.

Robbepop (commented):
All in all looks good to me. Thanks for the PR!

crates/env/Cargo.toml
athei: For SEAL I started with v3 to be completely backwards compatible. At some point you need to do this for ink!, too. Not sure when you want to do this.

cmichi: Why does it have to be that fine-franular (i.e. per seal method), couldn't the flag be `seal-v3` instead?

athei: This will lead to really messy code when checking for this features. For easier configuration I propose using meta features to bundle them together.

Robbepop: I agree with @athei that things like this are best done on a fine-grained level, however, I would like to be able to break backwards compat for now until we release ink! 3.0 proper as a testament to stability by then.

athei: Sure. This decision is up to you. Do you want me to remove the features and just remove the v0 of `seal_random`?  Please keep in mind that the v0 was already released as pallet_contracts 3.0.

Robbepop: Yes, I want to remove that since it is an API that is hard to use correctly and it has a proper replacement. As soon as we release ink! 3.0 proper we will also communicate the minimum `contracts-pallet` version with which it is compatible.

athei:
I removed the feature.

Robbepop (approved):
LGTM

Thanks a lot @athei !

cmichi (approved):

Issue #721, opened 2021-03-05
Enable clippy for rand-extension example


cmichi:
In https://github.com/paritytech/ink/pull/719 we temporarily disabled running clippy for the `rand-extension` example.

Enable it again as soon as this succeeds for nightly clippy:
```
cargo clippy --verbose --manifest-path examples/rand-extension/Cargo.toml --no-default-features --target wasm32-unknown-unknown -- -D warnings
```

PR #735, opened 2021-03-16
Enable clippy for `rand-extension` example again

cmichi:
Closes https://github.com/paritytech/ink/issues/721.

### DIFF
--- a/crates/env/src/engine/off_chain/mod.rs
+++ b/crates/env/src/engine/off_chain/mod.rs
@@ -127,7 +127,7 @@ impl EnvInstance {
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


crates/env/src/engine/off_chain/mod.rs
```diff
@@ -127,7 +127,7 @@ impl EnvInstance {
         Ok(())
     }
 
-    /// Resets the off-chain environment to unintialized state.
+    /// Resets the off-chain environment to uninitialized state.
```
cmichi, line 130: No idea why this was not detected. I created https://github.com/paritytech/ink/issues/738 for it.

Robbepop (approved):
LGTM

PR #741, opened 2021-03-20
Enum support no struct

yjhmelody:
The enum should support other layouts. What do you think about this?
This PR made incompatible changes to metadata.

BTW, Why use two styles of naming and organization for `storage` and `types` fields?

### DIFF
--- a/crates/metadata/src/layout/mod.rs
+++ b/crates/metadata/src/layout/mod.rs
@@ -17,30 +17,16 @@ mod tests;
 
 use crate::{
     serde_hex,
-    utils::{
-        deserialize_from_byte_str,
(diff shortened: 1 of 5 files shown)


Robbepop:
Hey @yjhmelody ,
thanks for the PR. There are quite a few problems with the current metadata layout information that we need to fix in order to make it actually useful for third party tools such as Canvas UI or Polkadot JS Apps.
Be aware that we have already dig into this but it isn't as simple due to complications with Rust's type system where we still need to find a proper solution.
The current metadata for storage layout information is heavily outdated and was never stabilized unlike other parts of contract metadata. Please do not consider to create expensive support for it.

PR #743, opened 2021-03-22
CI: do not run spellcheck when pipeline is triggered

TriplEight:
To avoid [this kind](https://gitlab.parity.io/parity/ink/-/pipelines/129685) of fails

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -144,6 +144,9 @@ docs:
 spellcheck:
     stage:                           workspace
     <<:                              *docker-env
+    rules:
+      - if: $CI_PIPELINE_SOURCE == "pipeline"
+        when: never
     script:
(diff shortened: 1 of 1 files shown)


cmichi:
@TriplEight Do you have an idea why this happens for pipelines?

TriplEight:
@cmichi why what happens?

cmichi:
> @cmichi why what happens?

You wrote in this PR description "To avoid this kind of fails" and I was wondering if you know why it fails for the pipeline?

I had observed it as well and thought that it might be a bug in `cargo-spellcheck` (I also created a ticket for this strange behavior in their repository: https://github.com/drahnr/cargo-spellcheck/issues/163).

TriplEight:
I just meant that this job is completely unnecessary when testing the image

cmichi (approved):

PR #746, opened 2021-03-25
Remove `ink-unstable` feature

cmichi:
No longer necessary, since Rust 1.51.1 has just been released.


Robbepop:
> No longer necessary, since Rust 1.51.1 has just been released.

Woop Woop

Robbepop (approved):

PR #749, opened 2021-03-30
Upgrade to `array-init` 2.0

cmichi:
The function signaure for `array_init()` changed:

```
- pub fn array_init<Array, F>(mut initializer: F) -> Array
+ pub fn array_init<F, T, const N: usize>(mut initializer: F) -> [T; N]
```

See https://github.com/Manishearth/array-init/commit/723f80e0c2a3c89cf02e8f9bab80babad0ca6d55.


Robbepop (approved):
we could also make use of min_const_generics for ink! in our trait impls.

Issue #539, opened 2020-10-24
Smart Contract Economics


cmichi:
# Problem

Substrate continuously charges a contract for rent. If the contract's balance exhausts these liabilities the contract is evicted and a tombstone (i.e. a hash of the code) is placed in its place.

This requires contracts who want to live for a long time to think about an economic model how they can sustain themselves. Currently it's not clear how they are supposed to do this and some necessary information is not yet exposed to ink! (like how much the costs are per block or the rent projection).

# Open Questions

* In ink!, how should contracts best charge more for calls in order to save up for their rent fee?
  * Mark as payable, check how much the call costs and require an additional amount? How should the "additional amount" be determined?
* What constants need to exposed in ink!?
* Is it possible to expose an estimate of how many blocks until the contract will be evicted in ink!?
* Are there already best practices in related projects?

# ToDo

- [ ] Research how related projects solve this.
- [ ] Implement a solid example demonstrating smart contract economics. Maybe a contract that just tries to live forever? The example should contain off-chain tests.
The other ToDos will probably derive from that. For example, the off-chain environment contains a number of `unimplemented!()` for tombstone/eviction functionality, they need to be implemented.

# Resources
* https://wiki.polkadot.network/docs/en/build-smart-contracts#storage-rent

Swader:
NEAR does this by paying 30% of every TX fee that is a result of a contract interaction to the contract that was used. If the call triggered multiple contracts, or one contract called another, then the 30% is evenly distributed. This is put into the balance of the contract, which uses the proceeds to pay rent or profit, depending on usage and optimization. Perhaps something to consider?

athei:
But that is no guarantee that those keep the contract afloat, right? Because TX fees are not correlated to storage consumption.

Swader:
In NEAR the chain's currency is used as two things: paying for tx fees, and permission slip to a certain amount of storage. Think 1 N = 10kb or something like that. So a contract that stores a lot of data has to have a high balance - they cite the example of Ethereum's USDT contract. I would have to have 10k NEAR tokens to keep the state it has now (i.e. the bigger a contract's state, the more balance it needs to have and that balance is effectively locked). Technically, this is up to the user - they need to pre-pay enough to cover the state they intend to host, and then they rely on the tx fees for auto-top-ups if the contract or lib is popular, or implement functions that are payable so that each call to them is more expensive than the base fee.

[...]

PR #755, opened 2021-04-01
Implement `seal_rent_params`

cmichi:
Needed for https://github.com/paritytech/ink/issues/539.

We haven't yet finalized this API as stable, hence I've put it behind a feature flag for now.

Examples will follow in a separate PR.


Robbepop:
I have had a discussion with @athei where he told me that changes to the API are already in the pipeline about a less expressive but still "good enough" API using a single `seal_rent_deposit` API to query the rent deposit with the current storage used for the contract. Note that this API would be called multiple times during a call. However, the good part is that this would not require manually calculating storage fees on the contract side.

athei:
Yes. I talked to @cmichi about that. I will look into the next iteration of the `seal_rent_deposit` API soon.

crates/env/src/engine/on_chain/ext.rs
```diff
@@ -589,6 +593,19 @@ pub fn set_rent_allowance(value: &[u8]) {
     unsafe { sys::seal_set_rent_allowance(Ptr32::from_slice(value), value.len() as u32) }
 }
 
+pub fn rent_params(output: &mut &mut [u8]) {
```
athei, line 596: Need to be put in a
```rust
#[link(wasm_import_module = "__unstable__")]
extern "C" { }
```
block.

Robbepop: why not just `"unstable"` btw?

athei: I use it to indicate that it has some special semantic and is not only a module name.

cmichi:
@athei The comment is implemented, do you have anything else?

athei:
It was discussed [here](https://github.com/paritytech/ink/pull/792#issuecomment-845818188) that no `ink-unstable` feature should be added. Maybe it should be removed.

cmichi:
> It was discussed [here](https://github.com/paritytech/ink/pull/792#issuecomment-845818188) that no `ink-unstable` feature should be added. Maybe it should be removed.

I get a bit sweaty when thinking about making the current rent API accessible without unstable flag. The reason is that in my mind it's pretty certain that we will have another iteration over the current status and not having it behind `ink-unstable` would mean that we e.g. have to mention it in the next rc changelog. Wdyt @Robbepop?

cmichi:
I thought a bit more about it and came to the conclusion that it's best to remove the unstable feature for rent stuff as well. My reasoning is that we are unstable in the rc state anyway and I also want to add examples soon and making them only available on unstable would make them more complicated. Also we might get community input by not hiding this behind a feature flag.

athei (approved):

Issue #753, opened 2021-03-31
Nightly CI run fails on `publish-docs` job


cmichi:
Happens every night: https://gitlab.parity.io/parity/ink/-/jobs/877640.
```
…
$ ./scripts/pre_cache.sh
/usr/bin/bash: line 141: ./scripts/pre_cache.sh: No such file or directory
Running after_script
00:01
Running after script...
$ rm -rf .git/ ./*
Cleaning up file based variables
00:01
ERROR: Job failed: exit code 1
```

cmichi:
@TriplEight Do you have an idea what could cause this?

PR #756, opened 2021-04-01
CI: override `before_script` so it won't fail on a doc branch

TriplEight:
Closes: https://github.com/paritytech/ink/issues/753


cmichi (approved):
Thanks @TriplEight!

PR #760, opened 2021-04-13
Fix nightly clippy errors

cmichi:


crates/primitives/src/key.rs
```diff
@@ -169,7 +169,7 @@ impl scale::Decode for Key {
 impl Key {
     /// Returns the bytes that are representing the key.
     #[inline]
-    pub fn to_bytes(&self) -> [u8; 32] {
+    pub fn to_bytes(self) -> [u8; 32] {
```
Robbepop, line 172: It probably won't play a role in benchmarks or generated code but we need to keep in mind that even though `Key` is `Copy` it still is fairly big with its 32 bytes in memory footprint. That's equal to copying around 8 `i32` values. It would be nice if we could comfort ourselves by providing some trivial benchmarks to assert that this won't affect performance of the generated code.

cmichi: I've written a benchmark, I've committed it to this PR and reverted it immediately afterwards, since I don't think it makes sense to merge it into `master`.
```
Old (with &self)
Key::to_bytes()         time:   [6.6567 ns 6.7805 ns 6.9185 ns]

New (with self)
Key::to_bytes()         time:   [6.6812 ns 6.8056 ns 6.9412 ns]                              
                        change: [-1.3811% +0.9807% +3.4177%] (p = 0.43 > 0.05) 
```

I've re-run the bench several times, result is always the same ‒ `No change in performance detected.`.

crates/storage/src/collections/bitvec/mod.rs
```diff
@@ -272,17 +272,18 @@ impl Bitvec {
...
-            //         to mutate this value if `value` is `true`.
-            *self.len += 1;
-            if value {
-                self.last_mut()
-                    .expect("must have at least a valid bit in this case")
-                    .set()
-            }
+            return
+        }
+
```
Robbepop, line 277: please remove empty line

Robbepop (approved):

PR #761, opened 2021-04-13
Implement decode for example error types

austinabell:
Fixes build with `ink-as-dependency` feature.


Robbepop (approved):
Thanks a lot for the PR. I just checked locally and you are right and this PR probably fixes the issues.

PR #768, opened 2021-04-16
fix topic error

honeywest:
balance not index

https://substrate.dev/substrate-contracts-workshop/#/2/creating-an-event

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/538b6d21b15733601f9193af5b9f662b94f16ea1/contracts/token/ERC20/IERC20.sol#L70-L76


Robbepop (approved):
thanks! it is good to have better parity with OpenZepelin

PR #769, opened 2021-04-16
Fix prelude

honeywest:
optimize code


crates/prelude/src/lib.rs
```diff
@@ -61,14 +61,14 @@ cfg_if! {
...
+                BinaryHeap,
                 BTreeMap,
                 BTreeSet,
-                BinaryHeap,
                 LinkedList,
                 VecDeque,
             };
-            pub use alloc::collections::*;
             pub use core::ops::Bound;
+            pub use alloc::collections::*;
```
Robbepop, line 71: how does this ordering improve the code? before that it was alphabetically sorted

crates/primitives/src/key.rs
```diff
@@ -202,27 +202,27 @@ impl Add<u64> for Key {
     }
 }
 
-impl<'a> Add<u64> for &'a Key {
+impl Add<u64> for &Key {
```
Robbepop, line 205: Funnily I was not aware that it was nowadays possible to leave away explicit lifetimes in impl blocks like these. Just checked with Rust playground. Good to know ... a bit baffled though.

crates/prelude/src/lib.rs
honeywest: if and else ordering 
They should be consistent
This is a good way to check the code and not easy to make mistakes

crates/primitives/src/key.rs
honeywest: Remove it for better understanding
Especially students who don't know the concept of lifetimes

crates/prelude/src/lib.rs
Robbepop, line 71: actually this is in wrong order. I wonder if our formatting CI actually checks these lines of code since they are conditionally compiled code.

honeywest: why? 
My understanding doesn't need any order
for example 
use A
use B
It's the same
use B
use A

I just need to be consistent

Robbepop: our code guidelines demand alphabetically sorted import statements. the only problem right now is that for whatever reasons `rustfmt` does not automatically check this region of code. We need to inspect why. Probably it is because this part of the codebase is only conditionally compiled.

honeywest: ok, delete it

PR #773, opened 2021-04-21
Fix `Greeter` and run documentation tests in CI

cmichi:
`Greeter` was not working in it's current stage.

Note that even though `cargo test --doc --workspace` improves our testing, it would still not have catched the `Greeter` error which this PR fixes.

I'll create follow-ups for `ink-docs` and the workshop, which are also missing `use ink_prelude::format`.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -376,7 +376,7 @@ fuzz-tests:
     <<:                              *docker-env
     variables:
         # The QUICKCHECK_TESTS default is 100
-        QUICKCHECK_TESTS:            40000
+        QUICKCHECK_TESTS:            20000
     rules:
         - if: $CI_PIPELINE_SOURCE == "schedule"
         - if: $CI_COMMIT_REF_NAME == "master"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -141,6 +141,7 @@ docs:
         -p ink_primitives -p ink_prelude
         -p ink_lang -p ink_lang_macro -p ink_lang_ir -p ink_lang_codegen
     - mv ${CARGO_TARGET_DIR}/doc ./crate-docs
+    - cargo test --doc --workspace
 
 spellcheck:
     stage:                           workspace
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -141,6 +141,7 @@ docs:
         -p ink_primitives -p ink_prelude
         -p ink_lang -p ink_lang_macro -p ink_lang_ir -p ink_lang_codegen
     - mv ${CARGO_TARGET_DIR}/doc ./crate-docs
+    - cargo test --doc --workspace
```
Robbepop, line 144: We should put this under the CI testing section.

.gitlab-ci.yml
```diff
@@ -375,7 +376,7 @@ fuzz-tests:
     <<:                              *docker-env
     variables:
         # The QUICKCHECK_TESTS default is 100
-        QUICKCHECK_TESTS:            40000
+        QUICKCHECK_TESTS:            20000
```
Robbepop, line 379: Is there a way we could define a timeout instead?

crates/lang/macro/Cargo.toml
```diff
@@ -28,6 +28,7 @@ ink_metadata = { version = "3.0.0-rc3", path = "../../metadata/" }
 ink_env = { version = "3.0.0-rc3", path = "../../env/" }
 ink_storage = { version = "3.0.0-rc3", path = "../../storage/" }
 ink_lang = { version = "3.0.0-rc3", path = ".." }
+ink_prelude = { version = "3.0.0-rc3", path = "../../prelude/" }
```
Robbepop, line 31: does `dev-dependency` work here instead?

cmichi: It is already under `[dev-dependencies]`, did you mean that?

.gitlab-ci.yml
cmichi: Updated the PR.

.gitlab-ci.yml
cmichi: So the GitLab CI run times out after one hour, that's a hard timeout. But quickcheck only has this level of control.

crates/lang/macro/Cargo.toml
Robbepop: nvm :facepalm:  I didnt see that in the past view

Robbepop (approved):

PR #792, opened 2021-05-14
Implement `seal_debug_message`

ascjones:
Companion PR for https://github.com/paritytech/substrate/pull/8773, which replaces `seal_println` with `seal_debug_message`.

I've added some convenience macros  `debug_print!` and `debug_println!`, let me know what you think.

# todo

- [x] Test it against latest substrate.
- [x] Find and update any docs relating to legacy println! https://github.com/paritytech/ink-docs/issues/19
- [x] Since `seal_debug_message` is now an `unstable` feature in the contracts pallet, should we hide it behind `ink-unstable`? Or just add a comment that the node needs to compiled with `"pallet-contracts/unstable-interface"` enabled.
- [x] Add `ink-debug` feature which makes `debug_message` and the `debug_print*` macro a noop. Could modify `cargo-contract` to add/remove this feature from the contract/


Robbepop:
> I've added some convenience macros debug_print! and debug_println!, let me know what you think.

Let's be honest: It was about time. :sweat_smile:

crates/env/src/backend.rs
```diff
@@ -115,8 +115,15 @@ pub trait EnvBackend {
...
-    fn println(&mut self, content: &str);
+    /// Emit a custom debug message.
+    ///
+    /// The message is appended to the debug buffer which is then supplied to the calling RPC
+    /// client. This buffer is also printed as a debug message to the node console if the
+    /// `debug` log level is enabled for the `runtime::contracts` target.
+    ///
+    /// This is a no-op if debug message recording is disabled which is always the case
+    /// when the code is executing on-chain.
+    fn debug_message(&mut self, content: &str);
```
Robbepop, line 126: I am not sure but maybe we should rename it to `debug_println` for parity with SEAL API?

crates/env/src/engine/experimental_off_chain/impls.rs
```diff
@@ -225,8 +225,8 @@ impl EnvBackend for EnvInstance {
         )
     }
 
-    fn println(&mut self, content: &str) {
-        self.engine.println(content)
+    fn debug_message(&mut self, message: &str) {
+        self.engine.debug_message(message)
```
Robbepop, line 229: With the new print macros we might even think about getting rid of this since this is not needed and less powerful.

crates/env/src/lib.rs
```diff
@@ -99,3 +99,24 @@ pub use self::{
...
+}
+
+/// Appends a formatted string to the `debug_message` buffer, as per [`debug_print`] but
+/// with a newline appended.
+#[macro_export]
+macro_rules! debug_println {
+    () => ($crate::print!("\n"));
+    ($($arg:tt)*) => ({
+        $crate::debug_print!($($arg)*);
+        $crate::debug_print!("\n");
```
Robbepop, line 120: I'd prefer using a single call to `debug_print` if that's easily possible.

crates/env/src/lib.rs
```diff
@@ -99,3 +99,24 @@ pub use self::{
...
 };
+pub use ink_prelude;
+
+/// Appends a formatted string to the `debug_message` buffer which will be:
+///  - Returned to the caller when the contract is invoked via RPC (*not* via an extrinsic)
+///  - Logged as a `debug!` message on the substrate node, which will be printed to the node
+///    console's `stdout` when the log level is set to `debug`.
+#[macro_export]
+macro_rules! debug_print {
+    ($($arg:tt)*) => ($crate::debug_message(&$crate::ink_prelude::format!($($arg)*)));
```
Robbepop, line 110:
```suggestion
    ($($arg:tt)+) => ($crate::debug_message(&$crate::ink_prelude::format!($($arg)*)));
```

Robbepop: would prevent usage of `debug_print!()`

crates/env/src/lib.rs
Robbepop, line 120:
```suggestion
        $crate::debug_print!('\n');
```

Robbepop (commented):
all in all LGTM

crates/env/src/engine/experimental_off_chain/impls.rs
ascjones: Print macros depend on this  `Backend` method via https://github.com/paritytech/ink/blob/2e6ba6c60a1328996c953d3694a2698c4b26dc75/crates/env/src/api.rs#L515

crates/env/src/lib.rs
ascjones: Simple alternative is something like
```
let msg = $crate::ink_prelude::format!($($arg)*);
$crate::debug_message(&$crate::ink_prelude::format!("{}\n", &msg))
```
Or could be done inline. 

Anything more involved to avoid allocations might be overkill since this should never be executed as part of a transaction.

What do you think?

ascjones: Out of interest Rust's `println!` macro looks like it interns the string in a compiler builtin with a `\n` appended: https://github.com/rust-lang/rust/blob/673d0db5e393e9c64897005b470bfeb6d5aec61b/compiler/rustc_builtin_macros/src/format.rs#L944

crates/env/src/lib.rs
ascjones: This is actually already prevented by the inner call to `format!`, which gives a nice error message:
image

crates/env/src/lib.rs
ascjones: Reverted this because it would have to be `debug_print!("{}". '\n')` to compile

crates/env/src/engine/experimental_off_chain/impls.rs
Robbepop: Ah sorry, I thought this piece of code was actually part of the `ink_lang::EnvAccess` getters that provide syntax for `self.env().println("message")`.

crates/env/src/lib.rs
Robbepop: This is a debug API so I do not care too much about extraneous allocations. However, it would be really confusing to the user if a single call to `debug_message!` always ends up in 2 actual calls to the engine. Imo we definitely need to fix this. I'd accept the allocation solution for now but would mark it with a comment that it is possible to optimize it further using our static 16kb buffer under the hood instead of allocating. If you are into it you can also toy around with this idea.

Robbepop (commented):
I'd love if we could remove the `self.env().println` API with this PR since it is definitely no longer really needed/wanted with this superior macro replacement.

crates/env/src/lib.rs
athei: I agree. Calls into the supervisor should be kept to a minimum. They carry a non trivial overhead even when debugging is disabled (on-chain execution).

crates/env/src/backend.rs
```diff
@@ -115,8 +115,15 @@ pub trait EnvBackend {
...
-    /// Prints the given contents to the console log.
-    fn println(&mut self, content: &str);
+    /// Emit a custom debug message.
+    ///
+    /// The message is appended to the debug buffer which is then supplied to the calling RPC
+    /// client. This buffer is also printed as a debug message to the node console if the
+    /// `debug` log level is enabled for the `runtime::contracts` target.
+    ///
+    /// This is a no-op if debug message recording is disabled which is always the case
+    /// when the code is executing on-chain.
```
athei, line 125: This statement is only true from the viewpoint of pallet-contracts. Contracts will pay for the overhead of calling into the supervisor when a debug message will be emitted during on-chain execution (it will do nothing but call into the supervisor and return). 

I recommend integrating one mitigation into this PR: Evaluate the return value of the `seal_debug_message` and if it is `false` prevent all further calls of this function.

Robbepop: I think we should handle this on a compile time level. So only if we compile using some crate feature (which is enabled by default) the debug message will actually do something useful. Otherwise it will be a no-op. The `cargo-contract` tool should then disable the `debug` crate feature by default upon doing a release build and should leave it when building a debug build for the contract.

athei: This makes sense. I guess you will want different log levels then for message you still want to be in the release build (to help users with wrong inputs).

That said, you can have both. I think for messages that should indeed be left inside a production contract to help real users during production use you will need this runtime check in order to prevent weight costs during on-chain execution.

crates/env/src/engine/on_chain/ext.rs
```diff
@@ -308,7 +308,7 @@ mod sys {
 
         pub fn seal_set_rent_allowance(value_ptr: Ptr32<[u8]>, value_len: u32);
 
-        pub fn seal_println(str_ptr: Ptr32<[u8]>, str_len: u32);
+        pub fn seal_debug_message(str_ptr: Ptr32<[u8]>, str_len: u32);
```
athei, line 311: You are missing the return value here.

crates/env/src/lib.rs
```diff
@@ -114,9 +114,9 @@ macro_rules! debug_print {
 /// with a newline appended.
 #[macro_export]
 macro_rules! debug_println {
-    () => ($crate::print!("\n"));
+    () => ($crate::debug_print!("\n"));
     ($($arg:tt)*) => ({
-        $crate::debug_print!($($arg)*);
-        $crate::debug_print!("\n");
+        let msg = $crate::ink_prelude::format!($($arg)*);
```
Robbepop, line 119: does this compile? doesn't `$crate` stand for the crate? Don't we rather need `::ink_prelude::format!(..)`? Maybe I am missing some Rust rules here.

ascjones: Using `::ink_prelude` depends on the contract importing that crate, which is not guaranteed. `$crate::ink_prelude` resolves to the reexported `::ink::env::ink_prelude`.

crates/env/src/backend.rs
ascjones: > I think we should handle this on a compile time level.

I've done the runtime check but might leave the compile time check for a follow up since it would require some changes to `cargo-contract` which builds by `--release` by default.

crates/env/src/engine/on_chain/ext.rs
```diff
@@ -604,9 +607,30 @@ pub fn random(subject: &[u8], output: &mut &mut [u8]) {
...
+    if unsafe { DEBUG_ENABLED } {
+        let bytes = message.as_bytes();
+        let ret_code = unsafe {
+            sys::seal_debug_message(Ptr32::from_slice(bytes), bytes.len() as u32)
+        };
+        if let Err(Error::LoggingDisabled) = ret_code.into() {
+            unsafe { DEBUG_ENABLED = false }
+        }
+    }
 }
```
Robbepop, line 634: I would love to see more encapsulation of the global static mut variable. E.g. put it and the function under an inline module and just export the function from the module. This way we not only encode via comments that the `DEBUG_ENABLED` static should not be used outside of the `debug_message` function.

Robbepop:
> Since seal_debug_message is now an unstable feature in the contracts pallet, should we hide it behind ink-unstable? Or just add a comment that the node needs to compiled with "pallet-contracts/unstable-interface" enabled.

I personally wouldn't introduce yet another crate feature and instead document the expected behavior when using this SEAL macro/function without respecting the SEAL unstable feature upon compilation.

crates/env/src/engine/on_chain/ext.rs
```diff
@@ -607,28 +609,36 @@ pub fn random(subject: &[u8], output: &mut &mut [u8]) {
...
+
+    /// Call `seal_debug_message` with the supplied UTF-8 encoded message.
+    ///
+    /// If debug message recording is disabled in the contracts pallet, the first call to will return
+    /// a LoggingDisabled error, and further calls will be a no-op to avoid the cost of calling into
+    /// the supervisor.
+    ///
+    /// # Note
+    ///
+    /// This depends on the currently dsiabled
```
Robbepop, line 627: comment is broken

crates/env/src/engine/on_chain/ext.rs
```diff
@@ -607,28 +609,36 @@ pub fn random(subject: &[u8], output: &mut &mut [u8]) {
...
+    ///
+    /// # Safety
+    ///
+    /// Accesses to the static `DEBUG_ENABLED` are safe because we are executing in a single-threaded
+    /// environment.
+    pub fn debug_message(message: &str) {
+        if unsafe { DEBUG_ENABLED } {
```
Robbepop, line 634: don't want to sound pedantic but this requires a small `// SAFETY: ****` comment describing why it is safe to do in this context. (single threaded) (Rust guidelines)

crates/env/src/engine/on_chain/ext.rs
```diff
@@ -607,28 +609,36 @@ pub fn random(subject: &[u8], output: &mut &mut [u8]) {
...
+    /// the supervisor.
+    ///
+    /// # Note
+    ///
+    /// This depends on the currently dsiabled
+    ///
+    /// # Safety
+    ///
+    /// Accesses to the static `DEBUG_ENABLED` are safe because we are executing in a single-threaded
+    /// environment.
```
Robbepop, line 632: This safety comment is not needed since the function is not unsafe but its implementation requires `unsafe` so put the comment where I marked it in the review.

crates/env/src/engine/on_chain/ext.rs
Robbepop: Btw.: What are the cons of putting `DEBUG_ENABLED` into a simple `Cell` dropping all the unsafe code? I think `Cell` does not really have a serious overhead, doesn't it?

crates/env/src/engine/on_chain/ext.rs
```diff
@@ -308,7 +313,7 @@ mod sys {
 
         pub fn seal_set_rent_allowance(value_ptr: Ptr32<[u8]>, value_len: u32);
 
-        pub fn seal_println(str_ptr: Ptr32<[u8]>, str_len: u32);
+        pub fn seal_debug_message(str_ptr: Ptr32<[u8]>, str_len: u32) -> ReturnCode;
```
athei, line 316: Don't forget to put this into the `__unstable__` module.

crates/env/src/engine/on_chain/ext.rs
ascjones: I believe `thread_local!` is only available in `std`

athei: > Btw.: What are the cons of putting DEBUG_ENABLED into a simple Cell dropping all the unsafe code? I think Cell does not really have a serious overhead, doesn't it?

`Cell` is not `Sync` and cannot put into a static. You have the choice between unsafe and using a synchronization primitive like a spinlock. The latter ones would of course introduce overhead not necessary in a singlethreaded environment. You could of course newtype wrap a `Cell` and implement `Sync` for it. That should be safe everywhere in ink!.

> I believe thread_local! is only available in std

Correct. The concept of threads does not exist in a vaccuum. It needs help from the operaring system. Which in this case is pallet_contracts. That said,  who mentioned `thread_local`?

ascjones: > That said, who did mention thread_local!?

Ha, yes usually that is a way to use `Cell` in a `static` to avoid the `Sync` requirement.

ascjones: > Putting the static variable inside the function is a good idea as its only referenced from there.

Hadn't thought of that. Done!

crates/env/src/lib.rs
```diff
@@ -102,3 +102,55 @@ pub use self::{
...
+            () => ($crate::debug_print!("\n"));
+            ($($arg:tt)*) => (
+                $crate::debug_print!("{}\n", $crate::format!($($arg)*));
+            )
+        }
+    } else {
+        #[macro_export]
+        /// Debug messages disabled. Enable the `ink-debug` feature for contract debugging.
+        macro_rules! debug_print {
+            ($($arg:tt)*) => (debug_assert!(true));
```
athei, line 146: What is this `debug_assert!(true)` good for?

ascjones: Nothing really, just to say this is a noop. Maybe should just be `=> ()` instead?

athei: Yeah that is what I was expecting there.

athei (approved):

crates/env/src/engine/on_chain/ext.rs
Robbepop: > That said, who did mention thread_local!?

We use `thread_local!` in ink! for the off-chain environment testing infrastructure.

cmichi (approved):

crates/env/src/engine/on_chain/ext.rs
```diff
@@ -645,11 +649,37 @@ pub fn random(subject: &[u8], output: &mut &mut [u8]) {
...
 }
 
-pub fn println(content: &str) {
-    let bytes = content.as_bytes();
-    unsafe { sys::seal_println(Ptr32::from_slice(bytes), bytes.len() as u32) }
+#[cfg(feature = "ink-debug")]
+/// Call `seal_debug_message` with the supplied UTF-8 encoded message.
+///
+/// If debug message recording is disabled in the contracts pallet, the first call to will
+/// return LoggingDisabled error, and further calls will be a no-op to avoid the cost of calling
```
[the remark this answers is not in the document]
ascjones: Yeah sorry because I merged it already :see_no_evil:

Issue #644, opened 2021-01-27
Add examples how to use `EnvAccess` public methods in the doc comments


cmichi:
Those are some of the most often used methods when users interact with the ink! API. From the questions in our public channels it's clear that we could improve on usage examples. We should make it clearer how to use those methods by adding an `# Example` section to the doc comments.

PR #797, opened 2021-05-26
Add example usage comments to `EnvAccess` methods

cmichi:
Closes https://github.com/paritytech/ink/issues/644.

### DIFF
--- a/crates/lang/Cargo.toml
+++ b/crates/lang/Cargo.toml
@@ -26,6 +26,10 @@ scale = { package = "parity-scale-codec", version = "2.1", default-features = fa
 derive_more = { version = "0.99", default-features = false, features = ["from"] }
 static_assertions = "1.1"
 
+[dev-dependencies]
+ink_lang_ir = { version = "3.0.0-rc3", path = "./ir", default-features = false }
+syn = "1"
+
 [features]
 default = ["std"]
 std = [
--- a/crates/lang/Cargo.toml
+++ b/crates/lang/Cargo.toml
@@ -26,6 +26,10 @@ scale = { package = "parity-scale-codec", version = "2.1", default-features = fa
 derive_more = { version = "0.99", default-features = false, features = ["from"] }
 static_assertions = "1.1"
 
+[dev-dependencies]
+# required for the doctest of `env_access::EnvAccess::instantiate_contract`
+scale-info = { version = "0.6", default-features = false, features = ["derive"] }
+
 [features]
 default = ["std"]
 std = [
--- a/crates/lang/Cargo.toml
+++ b/crates/lang/Cargo.toml
@@ -26,10 +26,6 @@ scale = { package = "parity-scale-codec", version = "2.1", default-features = fa
 derive_more = { version = "0.99", default-features = false, features = ["from"] }
 static_assertions = "1.1"
 
-[dev-dependencies]
-ink_lang_ir = { version = "3.0.0-rc3", path = "./ir", default-features = false }
-syn = "1"
-
 [features]
 default = ["std"]
 std = [
--- a/crates/lang/src/env_access.rs
+++ b/crates/lang/src/env_access.rs
@@ -260,7 +260,6 @@ where
     /// self.env().set_rent_allowance(self.env().balance / 2);
     /// ```
     ///
-    ///
     /// # Note
     ///
     /// For more details visit: [`ink_env::set_rent_allowance`]
--- a/crates/lang/src/env_access.rs
+++ b/crates/lang/src/env_access.rs
@@ -116,8 +116,7 @@ where
     ///
     /// ```
     /// # use ink_lang as ink;
-    /// use ink_prelude;
-    ///
+    /// #
     /// # #[ink::contract]
     /// # pub mod my_contract {
     /// #     #[ink(storage)]
@@ -132,8 +131,7 @@ where
     /// #[ink(message)]
     /// pub fn call_me(&self) {
     ///     let caller = self.env().caller();
-    ///     let message = ink_prelude::format!("got a call from {:?}", caller);
-    ///     ink_env::debug_println(&message);
+    ///     ink_env::debug_println!("got a call from {:?}", &caller);
     /// }
     /// #
     /// #     }
@@ -153,8 +151,7 @@ where
     ///
     /// ```
     /// # use ink_lang as ink;
-    /// use ink_prelude;
-    ///
+    /// #
     /// # #[ink::contract]
     /// # pub mod my_contract {
     /// #     #[ink(storage)]
@@ -171,8 +168,7 @@ where
     /// pub fn fund(&self) {
     ///     let caller = self.env().caller();
     ///     let value = self.env().transferred_balance();
-    ///     let message = ink_prelude::format!("thanks for the funding of {:?} from {:?}", value, caller);
-    ///     ink_env::debug_println(&message);
+    ///     ink_env::debug_println!("thanks for the funding of {:?} from {:?}", value, caller);
     /// }
     /// #
     /// #     }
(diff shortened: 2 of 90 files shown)


Robbepop (commented):
I like this PR very much and would love to have examples for the important `self.env()` API. However, some examples are currently flawed from what I can see and shouldn't/won't compile. Especially the `#[ink(message)]` examples are super useful but they need a lot of hidden context around them in order to make them compile properly. Non-compiling doc tests should be avoided at all cost since they are a nightmare for maintainability.

crates/lang/src/env_access.rs
```diff
@@ -128,6 +156,12 @@ where
 
     /// Returns the price for the specified amount of gas.
     ///
+    /// # Example
+    ///
+    /// ```
+    /// let price = self.env().weight_to_fee(13);
```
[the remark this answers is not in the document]
athei: There is no difference between gas and weight. Please keep in mind that everyone else calls it gas, though. Removing the word "gas" would lead to less confusion in the future but more confusion right now.

crates/lang/src/env_access.rs
```diff
@@ -200,23 +299,101 @@ where
...
+    pub fn instantiate_contract<Args, Salt, C>(
+        self,
+        params: &CreateParams<T, Args, Salt, C>,
+    ) -> Result<T::AccountId>
+    where
+        Args: scale::Encode,
+        Salt: AsRef<[u8]>,
+    {
+        ink_env::instantiate_contract::<T, Args, Salt, C>(params)
     }
```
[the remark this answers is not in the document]
cmichi: I had the same train of thought. But as we briefly discussed via DM there are use cases for this API, so I would opt for leaving it in.

cmichi:
I updated the PR, the code in the doc comment examples is compiled now. So please give it another round of review.

I briefly investigated using a proc. macro to reduce the code duplication, but quickly ran into issues with generating only part of a doc comment.

crates/lang/Cargo.toml
```diff
@@ -26,6 +26,10 @@ scale = { package = "parity-scale-codec", version = "2.1", default-features = fa
 derive_more = { version = "0.99", default-features = false, features = ["from"] }
 static_assertions = "1.1"
 
+[dev-dependencies]
+ink_lang_ir = { version = "3.0.0-rc3", path = "./ir", default-features = false }
+syn = "1"
```
Robbepop, line 31: why exactly do we need these dev. dependencies? For the doc tests? Hmm ... not sure I like this. Are there alternatives to this?

crates/lang/src/env_access.rs
```diff
@@ -111,6 +111,23 @@ where
...
+    /// # <ink_lang_ir::ItemImpl as TryFrom<syn::ItemImpl>>::try_from(syn::parse_quote! {
+    /// # impl MyContract {
+    ///  #[ink(message)]
+    ///  pub fn call_me(&self) {
+    ///     let caller = self.env().caller();
+    ///     let message = ink_prelude::format!("got a call from {:?}", caller);
+    ///     ink_env::debug_println(&message);
+    ///  }
+    /// # }
+    /// # }).unwrap();
```
Robbepop, line 128: While this trick is used somewhere else to test parts of `ink_lang_ir` I would not recommend doing this and instead go with the full blown approach of setting up an ink! smart contract. The problem with this code is that you don't really test anything ink! related but only the syntax analysis which does not cover a lot of the underlying details of the code.

So even though the more optimal approach of setting up an entire ink! smart contract for every single doc test is daunting and will probably be a huge mess without proper library components (which we might want here) the current approach does not really suffice our needs of making sure that our doc tests are going to be battle tested properly.

cmichi:
@Robbepop I updated the PR, could you take another look please?

I experimented with a macro to reduce the code duplication in https://github.com/paritytech/ink/pull/797/commits/f7319126a8c1cdda67fa814f2863b570f3915e6d, but always ran into macro hygiene errors, which from my understanding are not easy to resolve.

Robbepop:
> I experimented with a macro to reduce the code duplication in f731912, but always ran into macro hygiene errors, which from my understanding are not easy to resolve.

Yep, this is exactly what I expected unfortunately. I think for now we cannot do much about this.

crates/lang/src/env_access.rs
```diff
@@ -232,23 +542,148 @@ where
...
+    ///         .transferred_value(10)
+    ///         .exec_input(
+    ///             ExecutionInput::new(Selector::new([0xCA, 0xFE, 0xBA, 0xBE]))
+    ///                 .push_arg(42)
+    ///                 .push_arg(true)
+    ///                 .push_arg(&[0x10u8; 32])
+    ///         )
+    ///         .returns::<()>()
+    ///         .params();
+    ///     self.env().invoke_contract(&call_params).expect("call invocation must succeed");
```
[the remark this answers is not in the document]
cmichi: I don't see what you mean, the line is correctly indented, no?

cmichi:
I updated the doc tests with some more vivid examples.

Also I had to fix https://github.com/paritytech/ink/issues/801 as part of the recent commits. I used your suggestion from that issue:
>Since Weight is not customizable in the contracts pallet we should change the return type to plain u64.

crates/lang/src/env_access.rs
```diff
@@ -276,23 +795,36 @@ where
...
+    /// #         }
+    /// #
+    /// /// Simple resurrection of a contract.
+    /// #[ink(message)]
+    /// pub fn resurrect(&self, contract: AccountId) {
+    ///     self.env().restore_contract(contract,
+    ///         Hash::from([0x42; 32]),
+    ///         1000,
+    ///         &[]
+    ///     )
```
Robbepop, line 822:
```suggestion
    ///     self.env().restore_contract(
    ///         contract,
    ///         Hash::from([0x42; 32]),
    ///         1000,
    ///         &[]
    ///     )
```

crates/lang/src/env_access.rs
```diff
@@ -332,6 +915,32 @@ where
...
+    /// #
+    /// #     impl MyContract {
+    /// #         #[ink(constructor)]
+    /// #         pub fn new() -> Self {
+    /// #             Self {}
+    /// #         }
+    /// #
+    /// #[ink(message)]
+    /// pub fn random_bool(&self) -> bool {
+    ///     let additional_randomness = &[];
```
Robbepop, line 935:
```suggestion
    ///     let additional_randomness = b"seed";
```

Robbepop: shows people that how this for example could actually be used

Robbepop (commented):
Seeing code like
```rust
ink_env::account_id::<T>().expect("couldn't decode contract account ID")
```
all over the place I come to believe we need a follow-up PR that replaces those `expect` calls with `unwrap_or_else(|error| panic!("couldn't decode contract account ID: {}", error))` that provides more information about the underlying error.

crates/lang/src/env_access.rs
```diff
@@ -157,6 +306,33 @@ where
...
+    /// #         pub fn new() -> Self {
+    /// #             Self {}
+    /// #         }
+    /// #
+    /// /// Prints a debug message with the called contract's account id.
+    /// #[ink(message)]
+    /// pub fn call_me(&self) {
+    ///     let account_id = self.env().account_id();
+    ///     ink_env::debug_println!("contract's account id is {:?}", account_id);
+    /// }
```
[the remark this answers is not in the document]
cmichi: Added a simple permissioned contract.

Robbepop (approved):
LGTM

Reminder: We should create an issue for the follow-up PR for https://github.com/paritytech/ink/pull/797#pullrequestreview-695123812.

PR #798, opened 2021-05-27
Implement `seal_rent_status`

cmichi:
Follow-up for https://github.com/paritytech/substrate/pull/8780.

Needed for https://github.com/paritytech/ink/issues/539.

### DIFF
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -186,7 +186,7 @@ where
 /// # Errors
 ///
 /// If the returned value cannot be properly decoded.
-pub fn rent_status<T>(at_refcount: Option<u32>) -> Result<RentStatus<T>>
+pub fn rent_status<T>(at_refcount: Option<core::num::NonZeroU32>) -> Result<RentStatus<T>>
 where
     T: Environment,
 {
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -175,15 +175,19 @@ where
 
 /// Returns information about the required deposit and resulting rent.
 ///
+/// # Parameters
+///
+/// - `at_refcount`: The refcount assumed for the returned `custom_refcount_*` fields.
+///
 /// # Errors
 ///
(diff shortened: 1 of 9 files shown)


crates/env/src/engine/on_chain/ext.rs
```diff
@@ -348,6 +348,11 @@ mod sys {
             output_ptr: Ptr32Mut<[u8]>,
             output_len_ptr: Ptr32Mut<u32>,
         );
+
+        pub fn seal_rent_status(
+            output_ptr: Ptr32Mut<[u8]>,
+            output_len_ptr: Ptr32Mut<u32>,
+        );
```
athei, line 355: This does not match the signature:
https://github.com/paritytech/substrate/blob/1cbf0e16832eca5e0733d65279872ce79ddfe9f2/frame/contracts/src/wasm/runtime.rs#L1612

athei (changes requested):
You are missing one argument from the function.

crates/env/src/api.rs
```diff
@@ -175,15 +175,19 @@ where
 
 /// Returns information about the required deposit and resulting rent.
 ///
+/// # Parameters
+///
+/// - `at_refcount`: The refcount assumed for the returned `custom_refcount_*` fields.
```
Robbepop, line 180: Document what is the semantic of it being `None`?
Also briefly document what is the effect of this parameter.

crates/env/src/engine/on_chain/ext.rs
```diff
@@ -610,6 +616,20 @@ pub fn rent_params(output: &mut &mut [u8]) {
     extract_from_slice(output, output_len as usize);
 }
 
+pub fn rent_status(at_refcount: Option<u32>, output: &mut &mut [u8]) {
```
athei, line 619: Maybe `core::num::NonZeroU32` could be used to express that `0` is not a valid input.

athei (approved):
Maybe

PR #800, opened 2021-06-01
ERC-1155 Example

HCastano:
This is getting big, so I think it's a good time to get some feedback and also get
some of my questions answered.

This PR aims to inkplement ERC-1155 as specified [here](https://eips.ethereum.org/EIPS/eip-1155).

It implements the "core" APIs, the ERC1155 Receivable interface (even though the just
implementation panics and rejects tokens, lol), and also has some additional
functionality such as a way to easily mint tokens.

Note that this contract implementation is missing some extensions which are mentioned in
the ERC-1155 specification document:
- Missing ERC1155 Metadata URI Interface
- No differentiation between fungible and non-fungible tokens
- Tokens are not officially burnable
- No batch minting

I won't be adding these here since a) this PR is already pretty big and b) they could
make for good first contributor issues.

I also think there are some improvements to be made in terms of how "Rusty" some
things are (having events return `None` vs. `0x00`), but I don't want to deviate too
far from the ERC-1155 spec.


examples/erc1155/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+#[ink::contract]
+mod erc1155 {
+    use super::*;
+
+    use ink_prelude::collections::BTreeMap;
+
+    /// An ERC-1155 contract.
+    #[ink(storage)]
+    pub struct Contract {
+        balances: BTreeMap<AccountId, BTreeMap<TokenId, Balance>>,
```
Robbepop, line 66: We usually prefer to use something like `BTreeMap<(AccountId, TokenID), Balance>` instead.

HCastano:
@Robbepop @cmichi can I get some eyes on this 👀

examples/erc1155/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+    "ink_metadata/std",
+    "ink_env/std",
+    "ink_storage/std",
+    "ink_lang/std",
+    "ink_prelude/std",
+    "scale/std",
+    "scale-info",
+    "scale-info/std",
+]
+ink-as-dependency = []
```
Robbepop, line 37: probably for debugging with `debug_message!` you now want to propagate the `ink-debug` feature to `ink_env`.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(test, allow(dead_code))]
+#![cfg_attr(test, allow(unused_imports))]
```
Robbepop, line 17:
```suggestion
#![cfg_attr(test, allow(dead_code, unused_imports))]
```

Robbepop: should work. not tested though

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+use ink_lang as ink;
+use ink_prelude::vec::Vec;
+
+// This is the "magic" return value that we expect if a smart contract supports receiving ERC-1155
+// tokens.
+//
+// It is calculated with
+// `bytes4(keccak256("onERC1155Received(address,address,uint256,uint256,bytes)"))`, and corresponds
+// to 0xf23a6e61.
+const MAGIC_VALUE: [u8; 4] = [242, 58, 110, 97];
```
Robbepop, line 29: I would maybe just call it "selector" then? xD
E.g. `ON_ERC_1155_RECEIVED_SELECTOR` or something like that.
It isn't a magic value since you even explained its value. :D

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+    /// An ERC-1155 contract.
+    #[ink(storage)]
+    #[derive(Default)]
+    pub struct Contract {
+        /// Tracks the balances of accounts across the different tokens that they might be holding.
+        balances: BTreeMap<(AccountId, TokenId), Balance>,
+        /// Which accounts (called operators) have been approved to spend funds on behalf of an owner.
+        ///
+        /// Note that the mapping is Set<(Owner, Operator)>.
+        approvals: BTreeMap<(AccountId, AccountId), ()>,
```
Robbepop, line 220: To make this more readable you could make this mapping strongly typed by introducing a
```rust
struct Approval {
    owner: AccountId,
    operator: AccountId,
}
```
type. Not important though.

Robbepop: Same goes for the `balances` mapping even though it is more obvious there.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+        ///
+        /// Note that as implemented anyone can create tokens. If you were to deploy this to a
+        /// production environment you'd probably want to lock down the addresses that are allowed
+        /// to create tokens.
+        #[ink(message)]
+        pub fn create(&mut self, value: Balance, _data: Vec<u8>) -> TokenId {
+            // Given that TokenId is a `u128` the likelihood of this overflowing is pretty slim.
+            self.token_id_nonce += 1;
+            self.balances
+                .insert((self.env().caller(), self.token_id_nonce), value);
```
Robbepop, line 246: maybe capture `self.env().caller()` into some local binding since you access it 3 times in this function alone.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+
+            // Emit transfer event but with mint semantics
+            self.env().emit_event(TransferSingle {
+                operator: self.env().caller(),
+                from: AccountId::from([0; 32]),
+                to: if value == 0 {
+                    AccountId::from([0; 32])
+                } else {
+                    self.env().caller()
+                },
```
Robbepop, line 256: Is this mandated by the standard? Otherwise I'd maybe use `Option` here.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+        /// Create the initial supply for a token.
+        ///
+        /// The initial supply will be provided to the caller (a.k.a the minter), and the
+        /// `token_id` will be assigned by the smart contract.
+        ///
+        /// Note that as implemented anyone can create tokens. If you were to deploy this to a
+        /// production environment you'd probably want to lock down the addresses that are allowed
+        /// to create tokens.
+        #[ink(message)]
+        pub fn create(&mut self, value: Balance, _data: Vec<u8>) -> TokenId {
```
Robbepop, line 242: why is `_data` unused?

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+            // Given that TokenId is a `u128` the likelihood of this overflowing is pretty slim.
+            self.token_id_nonce += 1;
```
Robbepop, line 244: I see a problem with respect to contract economics even in the case where we restrict the set of people who are able to create new tokens because it seems that there is no way to delete coins altogether and therefore free their occupied space.

Robbepop: We have the `ink_storage::Stash` data structure for these kind of purposes. However, it is a bit more complex that you need here. We might want to add yet another bit-wise data structure for use cases like these.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+        /// Note that as implemented anyone can mint tokens. If you were to deploy this to a
+        /// production environment you'd probably want to lock down the addresses that are allowed
+        /// to mint tokens.
+        #[ink(message)]
+        pub fn mint(&mut self, token_id: TokenId, value: Balance, _data: Vec<u8>) {
+            assert!(
+                token_id <= self.token_id_nonce,
+                "The `token_id` {:?} has not yet been created in this contract.",
+                token_id
+            );
```
Robbepop, line 278: We should be thinking when to return recoverable `Err` instead of unrecoverable panics.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+            data: Vec<u8>,
+        ) {
+            assert!(
+                self.balance_of(from, token_id) >= value,
+                "Insufficent token balance for transfer."
+            );
+
+            if let Some(b) = self.balances.get_mut(&(from, token_id)) {
+                *b -= value
+            }
```
Robbepop, line 311: Maybe you could use the `.entry()` API instead here.
https://paritytech.github.io/ink/ink_storage/collections/hashmap/enum.Entry.html

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+            &mut self,
+            from: AccountId,
+            to: AccountId,
+            token_id: TokenId,
+            value: Balance,
+            data: Vec<u8>,
+        ) {
+            assert!(
+                self.balance_of(from, token_id) >= value,
+                "Insufficent token balance for transfer."
```
Robbepop, line 306: more helpful panic message was nice.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+                token_id,
+                value,
+            });
+
+            // Quick Haxx, otherwise my tests just panic due to the use of eval_contract()
+            #[cfg(not(test))]
+            {
+                // If our recipient is a smart contract we need to see if they accept or
+                // reject this transfer. If they reject it we need to revert the call.
+                let params = build_call::<ink_env::DefaultEnvironment>()
```
Robbepop, line 331: It should be possible to just use `Environment` here as an alias for the used environment.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+                    .callee(to)
+                    .gas_limit(5000) // Q: What's the correct amount to use here?
```
Robbepop, line 333: `5000` magic number? :D It might be too low.

Robbepop: Correct amount? Good question. :S

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+                    .exec_input(
+                        ExecutionInput::new(Selector::new(MAGIC_VALUE))
+                        .push_arg(self.env().caller())
+                        .push_arg(from)
+                        .push_arg(token_id)
+                        .push_arg(value)
+                        .push_arg(data)
```
Robbepop, line 340: I really don't want to be that guy but I wonder why `rustfmt` CI does not want another indentation here.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+                    )
+                    .returns::<ReturnType<Vec<u8>>>()
+                    .params();
+
+                match ink_env::eval_contract(&params) {
+                    Ok(v) => {
+                        assert_eq!(
+                            v,
+                            &MAGIC_VALUE[..],
+                            "Recipient contract does not accept token transfers."
```
Robbepop, line 350: please improve this panic message with the values that caused this

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+                        match e {
+                            ink_env::Error::CodeNotFound => {
+                                // Our recipient wasn't a smart contract, so there's nothing more for
+                                // us to do
+                            }
+                            _ => {
+                                // I have no insight to what's happening here, I can't deploy
+                                // the following:
+                                //
+                                // ink_env::debug_println(&ink_prelude::format!("{:?}", e));
```
Robbepop, line 363: is `ink-debug` crate feature enabled for `ink_env`?

Robbepop: Also please rebase to current `master` branch.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+        #[ink(message)]
+        fn safe_transfer_from(
+            &mut self,
+            from: AccountId,
+            to: AccountId,
+            token_id: TokenId,
+            value: Balance,
+            data: Vec<u8>,
+        ) {
+            // Q: Does the caller change if the function is called from within this smart contract?
```
Robbepop, line 383: A: No

:)

Robbepop: Interesting question though ... never thought about this potential confusion before.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+            _operator: AccountId,
+            _from: AccountId,
+            _token_id: TokenId,
+            _value: Balance,
+            _data: Vec<u8>,
+        ) -> Vec<u8> {
+            unimplemented!("This smart contract does not accept token transfer.")
+        }
+
+        #[ink(message, selector = "0xBC197C81")]
```
Robbepop, line 495: Looking at your constant definitions above we might want to allow for identifiers here if that is possible ...

Robbepop: Could be hard though since ink! proc. macros require selectors at proc. macro compile time unfortunately.

examples/erc1155/lib.rs
HCastano: In the ERC-1155 spec they call it "magic value", but yeah maybe it should have a proper name 😅

examples/erc1155/lib.rs
HCastano: > When minting/creating tokens, the _from argument MUST be set to 0x0 (i.e. zero address).

However, I'm open to taking a more loose interpretation of the standard in order to make it more Rusty

examples/erc1155/lib.rs
HCastano: > there is no way to delete coins altogether and therefore free their occupied space.

I don't know if this is a good "feature" to have tbh. As a token holder I (probably) wouldn't want someone/something to be able to clear my balance in order to save some space on-chain.

Maybe a use case here is that project is transitioning to a new token, in which case you'd want to burn the old token and eventually clear those balance entries. However, this whole mechanism seems very implementation specific

examples/erc1155/lib.rs
HCastano: I still don't quite understand the mechanics of `Err` vs. `panic!` for Ink contracts. 

Let's say I do return an `Err` here, I'm sure a caller smart contract could deal gracefully with the `Err` but what about a normal user? There's no way they can handle anything, so would the contract revert?

examples/erc1155/lib.rs
HCastano: The word of `rustfmt` is law 🤷

examples/erc1155/lib.rs
HCastano: Is `Environment` just a type alias for `ink_env::DefaultEnvironment`?

HCastano:
I didn't get around to fixing all your comments today, will continue tomorrow

examples/erc1155/lib.rs
Robbepop: That's is only the case for `ink_storage` `collections` and `lazy` data structures.
Doing this for things like `self.env().caller()` and others would require a more elaborate process in order to produce optimal code and with our current process we couldn't produce optimal code so we do not even try and instead let the users do their job of not repeating themselves. However, in the future it might be possible to implement lints that help users with these things via `clippy` plugins.

examples/erc1155/lib.rs
Robbepop: Still better than not using `rustfmt` to start with ...

examples/erc1155/lib.rs
Robbepop: This has been under discussion and the last idea we came up was that we differentiate between the caller type.
If the caller is another smart contract then `Err` will not revert so that the caller contract can gracefully handle the failed call.
If the caller is _not_ another smart contract (and therefore the origin in all transactional cases that matter) then the call upon returning an `Err` is going to revert the state change.

examples/erc1155/lib.rs
```diff
@@ -341,7 +350,11 @@ mod erc1155 {
                 value,
             });
 
-            // Quick Haxx, otherwise my tests just panic due to the use of eval_contract()
+            // This is disabled during tests due to the use of `eval_contract()` not being
+            // supported (tests end up panicking).
+            //
+            // We should be able to get rid of this with when the new off-chain testing
+            // environment is available.
```
Robbepop, line 357: The new off-chain environment won't support this either unfortunately. It is pretty hard to support in general.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+use ink_prelude::vec::Vec;
+
+// This is the return value that we expect if a smart contract supports receiving ERC-1155
+// tokens.
+//
+// It is calculated with
+// `bytes4(keccak256("onERC1155Received(address,address,uint256,uint256,bytes)"))`, and corresponds
+// to 0xf23a6e61.
+#[cfg_attr(test, allow(dead_code))]
+const ON_ERC_1155_RECEIVED_SELECTOR: [u8; 4] = [242, 58, 110, 97];
```
cmichi, line 28: I usually prefer `const ON_ERC_1155_RECEIVED_SELECTOR: [u8; 4] = [0xf2, 0x3a, 0x6e, 0x61];`.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+/// The interface is defined here: `<https://eips.ethereum.org/EIPS/eip-1155>`.
+///
+/// The goal of ERC-1155 is to allow a single deployed contract to manage a variety of assets.
+/// These assets can be fungible, non-fungible, or a combination.
+///
+/// By tracking multiple assets the ERC-1155 standard is able to support batch transfers, which
+/// make it easy to transfer a mix of multiple tokens at once.
+#[ink::trait_definition]
+pub trait Erc1155 {
+    /// Transfer the a `value` amount of `token_id` tokens to the `to` account from the `from`
```
cmichi, line 52:
```suggestion
    /// Transfer the `value` amount of `token_id` tokens to the `to` account from the `from`
```

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+                to: Some(caller),
+                token_id,
+                value,
+            });
+        }
+
+        // Helper function for performing single token transfers.
+        //
+        // Should not be used directly since it's missing certain checks which are important to the
+        // ERC-1155 standard (it is expected that the caller has already perfomred these).
```
cmichi, line 318:
```suggestion
        // ERC-1155 standard (it is expected that the caller has already performed these).
```

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+// `bytes4(keccak256("onERC1155BatchReceived(address,address,uint256[],uint256[],bytes)"))`, and
+// corresponds to 0xbc197c81.
+const _ON_ERC_1155_BATCH_RECEIVED_SELECTOR: [u8; 4] = [188, 25, 124, 129];
+
+type TokenId = u128;
+type Balance = <ink_env::DefaultEnvironment as ink_env::Environment>::Balance;
+
+/// The interface for an ERC-1155 compliant contract.
+///
+/// The interface is defined here: `<https://eips.ethereum.org/EIPS/eip-1155>`.
```
cmichi, line 43:
```suggestion
/// The interface is defined here: <https://eips.ethereum.org/EIPS/eip-1155>.
```

The backticks prevent `cargo doc` from rendering the link automatically.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+    fn set_approval_for_all(&mut self, operator: AccountId, approved: bool);
+
+    /// Query if the given `operator` is allowed to control all of `owner`'s tokens.
+    #[ink(message)]
+    fn is_approved_for_all(&self, owner: AccountId, operator: AccountId) -> bool;
+}
+
+/// The interface for an ERC-1155 Token Receiver contract.
+///
+/// The interface is defined here: `<https://eips.ethereum.org/EIPS/eip-1155>`.
```
cmichi, line 116:
```suggestion
/// The interface is defined here: <https://eips.ethereum.org/EIPS/eip-1155>.
```

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+            set_sender(alice());
+            assert_eq!(erc.create(0), 1);
+            assert_eq!(erc.balance_of(alice(), 1), 0);
+
+            erc.mint(1, 123);
+            assert_eq!(erc.balance_of(alice(), 1), 123);
+        }
+
+        #[ink::test]
+        #[should_panic]
```
cmichi, line 732:
```suggestion
        #[should_panic(expected = "The `token_id` 7 has not yet been created in this contract.")
```

cmichi (changes requested):

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+        owner: AccountId,
+        #[ink(topic)]
+        operator: AccountId,
+        approved: bool,
+    }
+
+    /// Indicate that a token's URI has been updated.
+    #[ink(event)]
+    pub struct Uri {
+        value: ink_prelude::string::String,
```
Robbepop, line 211: Why do you prefer using an utf8 interpreted byte sequence instead of a `Vec<u8>` for example?

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+        Ord,
+        PartialOrd,
+        Eq,
+        PartialEq,
+        PackedLayout,
+        SpreadLayout,
+        scale::Encode,
+        scale::Decode,
+    )]
+    #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
```
Robbepop, line 231: Btw, how do you feel about this extensive list of attributes required for some struct like this one in ink! ?

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+        values: Vec<Balance>,
+        data: Vec<u8>,
+    );
+}
+
+#[ink::contract]
+mod erc1155 {
+    use super::*;
+
+    #[allow(unused_imports)]
```
Robbepop, line 169: Why is this needed here? Would it be sufficient to put this under a `cfg_attr` or place those inputs somewhere else?

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+
+// This is the return value that we expect if a smart contract supports batch receiving ERC-1155
+// tokens.
+//
+// It is calculated with
+// `bytes4(keccak256("onERC1155BatchReceived(address,address,uint256[],uint256[],bytes)"))`, and
+// corresponds to 0xbc197c81.
+const _ON_ERC_1155_BATCH_RECEIVED_SELECTOR: [u8; 4] = [188, 25, 124, 129];
+
+type TokenId = u128;
```
Robbepop, line 38: Could you please add documentation for this? If you don't make it public it will show as `u128` in the API of the trait and contract. Maybe it is actually wanted to make it a public item.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+    /// This should be called by a compliant ERC-1155 contract if the intended recipient is a smart
+    /// contract.
+    ///
+    /// If the smart contract implementing this interface accepts token transfers then it must
+    /// return `ON_ERC_1155_RECEIVED_SELECTOR` from this function. To reject a transfer it must revert.
+    ///
+    /// Any callers must revert if they receive anything other than `ON_ERC_1155_RECEIVED_SELECTOR` as a return
+    /// value.
+    #[ink(message)]
+    fn on_erc_1155_received(
```
Robbepop, line 135: Maybe it is the standard naming but I think it is pretty ugly to repeat yourself with `erc_1155` here.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+    use super::*;
+
+    #[allow(unused_imports)]
+    use ink_env::call::{
+        build_call,
+        utils::ReturnType,
+        ExecutionInput,
+        Selector,
+    };
+
```
Robbepop, line 176: Im allergic to empty lines between imports because it disables `rustfmt` automatic ordering of import statements.

examples/erc1155/lib.rs
HCastano: In the spec they use a Solidity `string`, which is a ["Dynamically-sized UTF-8-encoded string"](https://docs.soliditylang.org/en/develop/types.html#dynamically-sized-byte-array).

examples/erc1155/lib.rs
HCastano: I think it's pretty ugly tbh. However, I wouldn't want to hide too much of this behind macro magic if that's what you're thinking

examples/erc1155/lib.rs
HCastano: It's a little awkward, but I understand the naming choice. 

This is a way to handle incoming `ERC-1155` tokens and can be implemented by contracts which themselves are **not** `ERC-1155` contracts.

examples/erc1155/lib.rs
HCastano: I didn't realize it would show up as `u128` in the docs, but that makes sense

examples/erc1155/lib.rs
Robbepop: Hmmm, I see, however in Rust we can also opt-in to have disambiguation by using `Erc1155::on_received` to state that this `on_received` is coming from the `Erc1155` trait. So we usually don't do this in the context of Rust and therefore I'd assume that an ink! example would use Rusty identifiers.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+use ink_prelude::vec::Vec;
+
+// This is the return value that we expect if a smart contract supports receiving ERC-1155
+// tokens.
+//
+// It is calculated with
+// `bytes4(keccak256("onERC1155Received(address,address,uint256,uint256,bytes)"))`, and corresponds
+// to 0xf23a6e61.
+#[cfg_attr(test, allow(dead_code))]
+const ON_ERC_1155_RECEIVED_SELECTOR: [u8; 4] = [0xf2, 0x3a, 0x6e, 0x61];
```
Robbepop, line 28:
```suggestion
const ON_ERC_1155_RECEIVED_SELECTOR: [u8; 4] = [0xF2, 0x3A, 0x6E, 0x61];
```

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+
+// This is the return value that we expect if a smart contract supports batch receiving ERC-1155
+// tokens.
+//
+// It is calculated with
+// `bytes4(keccak256("onERC1155BatchReceived(address,address,uint256[],uint256[],bytes)"))`, and
+// corresponds to 0xbc197c81.
+const _ON_ERC_1155_BATCH_RECEIVED_SELECTOR: [u8; 4] = [0xbc, 0x19, 0x7c, 0x81];
```
Robbepop, line 36:
```suggestion
const _ON_ERC_1155_BATCH_RECEIVED_SELECTOR: [u8; 4] = [0xBC, 0x19, 0x7C, 0x81];
```

examples/erc1155/lib.rs
Robbepop: personal preference but I think big letters in hex-digits are to be preferred everywhere. More readable and same height.

examples/erc1155/lib.rs
Robbepop: personal preference but I think big letters in hex-digits are to be preferred everywhere. More readable and same height.

examples/erc1155/lib.rs
Robbepop: hey, ... it is a serious trade off unfortunately.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+
+            // Given that TokenId is a `u128` the likelihood of this overflowing is pretty slim.
+            self.token_id_nonce += 1;
+            self.balances.insert((caller, self.token_id_nonce), value);
+
+            // Emit transfer event but with mint semantics
+            self.env().emit_event(TransferSingle {
+                operator: Some(caller),
+                from: None,
+                to: if value == 0 { None } else { Some(caller) },
```
Robbepop, line 271:
```suggestion
                to: (value == 0).then(||  caller)
```

Robbepop: was even more beautiful if we had `i32::is_zero` or the `Zero` trait ...

examples/erc1155/lib.rs
Robbepop: in this case, yes. `Environment` within the ink! module is always a type alias for the used environment since every ink! contract so far is concrete over a specific environment. This is because different environments have different properties and therefore the same code could potentially not compile for different environments.

Robbepop: With the planned ink! library contracts a contract will be generic over a given contract (and potentially other stuff).

examples/erc1155/lib.rs
HCastano: I have no strong preference here, so sure, lol

examples/erc1155/lib.rs
HCastano: Haha, I thought about using `bool::then` but I think the `if/else` approach is actually more readable in this case

examples/erc1155/lib.rs
HCastano: Yeah, it is. I would say that if we were to hide something it should be the `*Layout` derives, not the native Rust ones

examples/erc1155/lib.rs
HCastano: That's a good point. I'll make it more Rusty.

FYI, these methods come from the `Erc1155TokenReceiver` trait, not the `Erc1155` trait. So the naming is even more awkward in that sense (`Erc1155TokenReceiver::on_erc_1155_received`)

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+                WALLET.into(),
+                1000000,
+                1000000,
+                ink_env::test::CallData::new(ink_env::call::Selector::new([0x00; 4])), /* dummy */
+            );
+        }
+
+        fn default_accounts(
+        ) -> ink_env::test::DefaultAccounts<ink_env::DefaultEnvironment> {
+            ink_env::test::default_accounts::<ink_env::DefaultEnvironment>()
```
Robbepop, line 580:
```suggestion
        ) -> ink_env::test::DefaultAccounts<Environment> {
            ink_env::test::default_accounts::<Environment>()
```

Robbepop (commented):
The most important thing to deal with in this PR is how we do error handling. Currently the implementation panics a lot which generally is not ideal. Instead `Result` should be returned with a predefined set of potential errors (given an `enum` or so) so that depending contracts can handle those errors if they want to.

examples/erc1155/lib.rs
HCastano: Yep, I forgot about this comment

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,742 @@
...
+            _token_ids: Vec<TokenId>,
+            _values: Vec<Balance>,
+            _data: Vec<u8>,
+        ) {
+            unimplemented!("This smart contract does not accept batch token transfers.")
+        }
+    }
+
+    /// Unit tests.
+    #[cfg(not(feature = "ink-experimental-engine"))]
```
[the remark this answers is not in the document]
HCastano: There are some differences in the APIs.

I've added some conditional compilation flags so that testing works in the experimental environment. If you think this approach is okay we should update other contracts which have this restriction (e.g the `Erc20` example tests)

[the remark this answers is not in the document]
Robbepop: Generally it is okay to just support the new engine since it will replace the old one.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,792 @@
...
+                            .push_arg(caller)
+                            .push_arg(from)
+                            .push_arg(token_id)
+                            .push_arg(value)
+                            .push_arg(data),
+                    )
+                    .returns::<ReturnType<Vec<u8>>>()
+                    .params();
+
+                match ink_env::eval_contract(&params) {
```
HCastano, line 399: I'm still trying to decide if these should be allowed to be gracefully handled by a contract or not. I'm leanings towards the "no" side right now

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,792 @@
...
+            );
+
+            token_ids.iter().zip(values.iter()).for_each(|(&id, &v)| {
+                // If any of the transfers fail we don't want the caller to be able to handle this.
+                // Instead we want to revert the whole call in order to ensure the atomicity of the
+                // batch transfer.
+                assert!(self.perform_transfer(from, to, id, v, data.clone()).is_ok(),
+                    "Failed to transfer {:?} tokens of ID {:?} from {:?} to {:?} as part of a batch transfer.",
+                    v, id, from, to
+                );
```
HCastano, line 489: This might be a little controversial, but this is the scenario I want to avoid:

 - Contract wants to transfer `N` tokens 
 - We're only able to transfer `N/2` tokens
 - Contract gets `Err` but ignores it, so `N/2` transfers are written

HCastano (commented):
So I think the disconnect with the `Err` vs. `panic` thing is that the spec explicitly says things like: _"MUST revert if `_to` is the zero address."_, which in my mind translates directly to `panic` and don't let the caller handle it.

However, after some thinking I agree, for most of these scenarios it's probably OK to let the caller decide how to proceed, and it also makes the code more idiomatic.

examples/erc1155/lib.rs
```diff
@@ -40,6 +40,38 @@ pub type TokenId = u128;
...
+}
+
+// The ERC-1155 result types.
+pub type Result<T> = core::result::Result<T, Error>;
+
+/// Evaluate `$x:expr` and if not true return `Err($y:expr)`.
+///
+/// Used as `ensure!(expression_to_ensure, expression_to_return_on_false)`.
+macro_rules! ensure {
+    ( $x:expr, $y:expr $(,)? ) => {{
```
Robbepop, line 68: please strife for better names also in macros.

examples/erc1155/lib.rs
Robbepop: Maybe do a sanity check for all transfers before doing a single actual transfer instead.
That's how I think contracts should generally be written. Checks. Actions. Signals (events).

examples/erc1155/lib.rs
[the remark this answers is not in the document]
HCastano: I think it's best to leave both for now. If people are meant to use the examples as a reference for how to implement certain things in their own contracts we should show them both the old and new way - at least until the experimental engine is not experimental anymore.

There's also a few other example contracts which support both, such as `contract-terminate`. If we are going to only support a single engine we should remove support from all the example contracts, which would best be done in a follow up PR.

examples/erc1155/lib.rs
Robbepop: `ReturnType` needed to be over `Result<Vec<u8>>` then, right?

examples/erc1155/lib.rs
HCastano: Are you referring to the macro name, or the variable names?

If it's the macro name, I think it's fine. It comes from Substrate's [`frame_support`](https://crates.parity.io/frame_support/macro.ensure.html) and will be immediately familiar to those coming from runtime development.

As for the variables, they're really short lived given the size of the macro, so I think they're also fine as is.

Robbepop: I am referring to `x` and `y`. We should really care for stating of intent with names even in small scopes especially since this piece of code serves as an example. ;)

.config/cargo_spellcheck.dic
```diff
@@ -1,4 +1,5 @@
 80
+0x00
```
Robbepop, line 2: I am very very unsure about this ... better wrap it in `0x00` where you need it.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,827 @@
...
+        if !$x {
+            return Err($y.into())
+        }
```
Robbepop, line 71:
```suggestion
    ( $condition:expr, $error:expr $(,)? ) => {{
        if !$condition {
            return ::core::result::Result::Err(::core::convert::Into::into($error))
        }
```

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,827 @@
...
+            }
+
+            ensure!(to != AccountId::default(), Error::ZeroAddressTransfer);
+            ensure!(!token_ids.is_empty(), Error::BatchTransferMismatch);
+            ensure!(
+                token_ids.len() == values.len(),
+                Error::BatchTransferMismatch,
+            );
+
+            let transfers: Vec<(_, _)> = token_ids.iter().zip(values.iter()).collect();
```
Robbepop, line 498: I would try not to do unnecessary heap memory allocations even though it cleans up the code to some extend. Try to use `clone` instead on the iterator maybe.

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,827 @@
...
+            owners: Vec<AccountId>,
+            token_ids: Vec<TokenId>,
+        ) -> Vec<Balance> {
+            let mut output = Vec::new();
+            for o in &owners {
+                for t in &token_ids {
+                    let amt = self.balance_of(*o, *t);
+                    output.push(amt);
+                }
+            }
```
Robbepop, line 539:
```suggestion
            for o in owners {
                for t in token_ids {
                    let amount = self.balance_of(o, t);
                    output.push(amount);
                }
            }
```

.config/cargo_spellcheck.dic
HCastano: I wasn't sure how else to satisfy the spellchecker 🤷 

I'm open to other suggestions though

examples/erc1155/lib.rs
HCastano: This seems overly explicit to me, however I'd like to hear your reasoning behind it.

Is the concern here that someone would define their own `Err` type and break the macro?

examples/erc1155/lib.rs
HCastano: Nevermind, the compiler wasn't happy about the whole nested `into_iter()` thing

examples/erc1155/lib.rs
HCastano: Sorry, I don't quite understand your suggestion here. How would cloning the vectors be cheaper than allocating a new vector in this case?

Robbepop: You'd clone an iterator over the original `token_ids`. Since we are only operating on shared references I guess `Clone` and maybe even `Copy` should exist for the iterator.

examples/erc1155/lib.rs
HCastano: I've decided to not return a `Result` from `on_received()` - I don't want to risk the caller ignoring the `Err` case and getting user funds stuck

examples/erc1155/lib.rs
HCastano: This doesn't make sense to me 😅 
Can you push some code with what you mean?

HCastano: I'm not convinced this is any cheaper in terms of allocations, but I've gone with your suggestion

Robbepop (approved):
LGTM

examples/erc1155/lib.rs
```diff
@@ -0,0 +1,827 @@
...
+        /// Mint a `value` amount of `token_id` tokens.
+        ///
+        /// It is assumed that the token has already been `create`-ed. The newly minted supply will
+        /// be assigned to the caller (a.k.a the minter).
+        ///
+        /// Note that as implemented anyone can mint tokens. If you were to deploy this to a
+        /// production environment you'd probably want to lock down the addresses that are allowed
+        /// to mint tokens.
+        #[ink(message)]
+        pub fn mint(&mut self, token_id: TokenId, value: Balance) -> Result<()> {
```
cmichi, line 321: Is there a reason why only the ink! messages `mint`, `new` and `create` are public?

PR #810, opened 2021-06-08
Small fixes for case when contract is compiled with compile_as_dependency = true

xgreenx:
Not generate metadata if compile_as_dependency is true
Remove not used Salt

### DIFF
--- a/crates/lang/codegen/src/generator/cross_calling.rs
+++ b/crates/lang/codegen/src/generator/cross_calling.rs
@@ -685,7 +685,7 @@ impl CrossCalling<'_> {
             fn #ident(
                 #( #input_bindings : #input_types ),*
             ) -> Self::#output_ident {
-                ::ink_env::call::build_create::<Environment, Salt, Self>()
+                ::ink_env::call::build_create::<Environment, Self>()
                     .exec_input(
                         ::ink_env::call::ExecutionInput::new(
                             ::ink_env::call::Selector::new([ #( #composed_selector ),* ])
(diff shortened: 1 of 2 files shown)


crates/lang/codegen/src/generator/cross_calling.rs
```diff
@@ -685,7 +685,7 @@ impl CrossCalling<'_> {
             fn #ident(
                 #( #input_bindings : #input_types ),*
             ) -> Self::#output_ident {
-                ::ink_env::call::build_create::<Environment, Salt, Self>()
+                ::ink_env::call::build_create::<Environment, Self>()
```
xgreenx, line 688: It has been added during Salt feature, but it is not used here and causes a compilation error.

Robbepop: It shouldn't case problems with respect to this PR. Just don't remove the `Salt` parameter.

xgreenx: Okey. Do you plan to remove this Salt later? Because it doesn't allow to import contract with `ink-as-dependency` if it implements constructor during the implementation of trait(it cause compilation error).

xgreenx:
Created separate PR without change for Salt: https://github.com/paritytech/ink/pull/811

crates/lang/codegen/src/generator/cross_calling.rs
Robbepop: I don't really understand the problem as stated so far. :S

xgreenx: Okey=) Let's try another description=D
[build_create](https://github.com/paritytech/ink/blob/master/crates/env/src/call/create_builder.rs#L198) method supports only 2 generic parameters. But code here is passing 3 parametrs=) It causes a compilation error. Above I described a case when ink! will generate this code which can't be compiled

PR #812, opened 2021-06-09
Spellchecking for examples + Fix master CI

cmichi:
Some miscellaneous improvements and fixes for the nightly clippy errors which currently make the CI fail on `master`.

### DIFF
--- a/.config/cargo_spellcheck.dic
+++ b/.config/cargo_spellcheck.dic
@@ -35,9 +35,7 @@ scalable
 adjunctive
 invariants
 deserializes
-Dapp
-untyped
-subber
+DApp
 
 hasher/S
 hashmap/S
--- a/.config/cargo_spellcheck.dic
+++ b/.config/cargo_spellcheck.dic
@@ -35,6 +35,9 @@ scalable
 adjunctive
 invariants
 deserializes
+Dapp
+untyped
+subber
 
 hasher/S
 hashmap/S
@@ -48,3 +51,4 @@ defragment/SD
 layout/JG
 endian/P
 accessor/S
+NFT/S
--- a/crates/engine/src/ext.rs
+++ b/crates/engine/src/ext.rs
@@ -343,10 +343,6 @@ impl Engine {
         super::hashing::keccak_256(input, output);
     }
 
-    pub fn now(&self, _output: &mut &mut [u8]) {
-        unimplemented!("off-chain environment does not yet support `now`");
-    }
-
     pub fn block_number(&self, _output: &mut &mut [u8]) {
         unimplemented!("off-chain environment does not yet support `block_number`");
     }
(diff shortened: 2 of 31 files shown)


.config/cargo_spellcheck.dic
```diff
@@ -35,6 +35,9 @@ scalable
 adjunctive
 invariants
 deserializes
+Dapp
+untyped
+subber
```
Robbepop, line 40: Not sure we should include words and names that do not really exist.
Maybe we find a better way here?

crates/engine/src/ext.rs
```diff
@@ -343,10 +343,6 @@ impl Engine {
         super::hashing::keccak_256(input, output);
     }
 
-    pub fn now(&self, _output: &mut &mut [u8]) {
-        unimplemented!("off-chain environment does not yet support `now`");
-    }
-
```
Robbepop, line 349: What happened here?

crates/storage/src/collections/hashmap/mod.rs
```diff
@@ -520,7 +520,7 @@ where
 {
     /// Gets a reference to the key in the entry.
     pub fn key(&self) -> &K {
-        &self.values_entry.key()
+        self.values_entry.key()
```
Robbepop, line 523: It is interesting we had so many cases of superflous `&` ... :thinking:

examples/erc721/lib.rs
```diff
@@ -27,9 +27,9 @@
 //!
 //! ## Error Handling
 //!
-//! Any function that modifies the state returns a Result type and does not changes the state
-//! if the Error occurs.
-//! The errors are defined as an Enum type. Any other error or invariant violation
+//! Any function that modifies the state returns a `Result` type and does not changes the state
+//! if the `Error` occurs.
+//! The errors are defined as an `Enum` type. Any other error or invariant violation
```
Robbepop, line 32:
```suggestion
//! The errors are defined as an `enum` type. Any other error or invariant violation
```

Robbepop (commented):
all in all LGTM.
How did you notice so many of those (hidden) superflous `&`?
Was it due to new warnings or clippy lints?

crates/engine/src/ext.rs
cmichi: Basically I just rename it from `now` to `block_timestamp`. We don't use `now` anywhere, just the latter.

.config/cargo_spellcheck.dic
cmichi: I've added a solution for `untyped` and `subber`, but would leave `DApp` in there, since imho it's an established term.

cmichi:
>How did you notice so many of those (hidden) superflous &?
>Was it due to new warnings or clippy lints?

It was new nightly clippy lints! :rocket:

.config/cargo_spellcheck.dic
Robbepop: `untyped` and `DApp` are fine imo.

Robbepop (approved):

PR #820, opened 2021-06-16
Fix bugs in two examples + Fix `master` CI

cmichi:
* Fixes the broken `master` CI (clippy).
* Fixes two bugs in our examples, which I discovered while adding [`ink-waterfall`](https://github.com/paritytech/ink-waterfall) tests for those examples.
  In the case of `contract-transfer` I added a waterfall test which also checks if the `debug_println!` in that contract example actually appears on the console.

### DIFF
--- a/examples/contract-transfer/Cargo.toml
+++ b/examples/contract-transfer/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2018"
 [dependencies]
 ink_primitives = { version = "3.0.0-rc3", path = "../../crates/primitives", default-features = false }
 ink_metadata = { version = "3.0.0-rc3", path = "../../crates/metadata", default-features = false, features = ["derive"], optional = true }
-ink_env = { version = "3.0.0-rc3", path = "../../crates/env", default-features = false, features = [ "ink-debug" ] }
+ink_env = { version = "3.0.0-rc3", path = "../../crates/env", default-features = false }
 ink_storage = { version = "3.0.0-rc3", path = "../../crates/storage", default-features = false }
 ink_lang = { version = "3.0.0-rc3", path = "../../crates/lang", default-features = false }
 ink_prelude = { version = "3.0.0-rc3", path = "../../crates/prelude", default-features = false }
@@ -21,7 +21,7 @@ path = "lib.rs"
 crate-type = ["cdylib"]
 
 [features]
-default = ["std"]
+default = ["std", "ink-debug"]
 std = [
     "ink_primitives/std",
     "ink_metadata",
@@ -35,3 +35,4 @@ std = [
 ]
 ink-as-dependency = []
 ink-experimental-engine = ["ink_env/ink-experimental-engine"]
+ink-debug = ["ink_env/ink-debug"]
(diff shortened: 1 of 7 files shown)


examples/multisig_plain/lib.rs
```diff
@@ -486,7 +486,7 @@ mod multisig_plain {
         ///
         /// Its return value indicates whether the called transaction was successful.
         /// This can be called by anyone.
-        #[ink(message)]
+        #[ink(message, payable)]
```
cmichi, line 489: Hmm, I just thought a bit more about this:

This covers the case where `Transaction.transferred_value` has been set. The contract currently supports setting this value, but if it is set the contract would pay for it out of its own fund.

This is not really a proper fix, since the contract still doesn't check if the funds transferred with the transaction are `==` to `Transaction.transferred_value`

I would create a follow-up ticket to implement this properly. In a proper implementation the `submit_transaction` would require `transferred_value` to be transferred along with the transaction submission.

examples/multisig_plain/lib.rs
```diff
@@ -484,15 +484,20 @@ mod multisig_plain {
...
+        /// and the transaction would then be reverted.
+        ///
         /// Its return value indicates whether the called transaction was successful.
         /// This can be called by anyone.
-        #[ink(message)]
+        #[ink(message, payable)]
         pub fn invoke_transaction(
             &mut self,
             trans_id: TransactionId,
         ) -> Result<(), Error> {
```
cmichi, line 497: As a side note: There is another issue with this method. It internally uses
```
.map_err(|_| Error::TransactionFailed);
```
which currently would not revert the transaction. This will only happen once we have implemented error forwarding. 
Currently, in this error case, the transaction would be taken (i.e. removed) out of the `StorageStash`, but not be executed due to the error. It would then not be possible to give the execution another try, since the transaction has been removed.

examples/contract-transfer/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2018"
 [dependencies]
 ink_primitives = { version = "3.0.0-rc3", path = "../../crates/primitives", default-features = false }
 ink_metadata = { version = "3.0.0-rc3", path = "../../crates/metadata", default-features = false, features = ["derive"], optional = true }
-ink_env = { version = "3.0.0-rc3", path = "../../crates/env", default-features = false }
+ink_env = { version = "3.0.0-rc3", path = "../../crates/env", default-features = false, features = [ "ink-debug" ] }
```
Robbepop, line 10: A better way was to introduce an `ink-debug` feature and forward that to `ink_env`.

examples/multisig_plain/lib.rs
Robbepop: Yeah this is a very important point made and acts as yet another motivator to implement storage reversal on error for non-`ink-as-dependency` calls. In the meantime we are required to panic in these critical cases.

examples/contract-transfer/Cargo.toml
cmichi: So the problem with this approach is that `cargo-contract` builds with `--no-default-features`. But there is a ticket to forward an `ink-debug` feature: https://github.com/paritytech/cargo-contract/issues/284.

But for now it means that if somebody builds `contract-transfer` with `cargo-contract` right now we need to explicitly enable the feature for `ink_env` like done in this PR, otherwise the user will not see the debug output.

Robbepop: We should raise the priority of the `--debug` flag on `cargo-contract` then.

PR #822, opened 2021-06-17
Enable `spellcheck` CI stage

cmichi:
I just noticed that it was not actually enabled :see_no_evil:.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -152,9 +152,6 @@ docs:
 spellcheck:
     stage:                           workspace
     <<:                              *docker-env
-    rules:
-      - if: $CI_PIPELINE_SOURCE == "pipeline"
-        when: never
     script:
         - cargo spellcheck check --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1
         - for example in examples/*/; do
--- a/crates/lang/codegen/src/generator/dispatch.rs
+++ b/crates/lang/codegen/src/generator/dispatch.rs
@@ -271,7 +271,7 @@ impl Dispatch<'_> {
     /// # ;
     /// ```
     ///
-    /// **Multiple `(>= 2)` inputs:**
+    /// **Multiple (equal to two or more) inputs:**
     /// ```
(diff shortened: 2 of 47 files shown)


Robbepop:
> I just noticed that it was not actually enabled see_no_evil.

Do you know why?

.gitlab-ci.yml
```diff
@@ -152,9 +152,6 @@ docs:
 spellcheck:
     stage:                           workspace
     <<:                              *docker-env
-    rules:
-      - if: $CI_PIPELINE_SOURCE == "pipeline"
-        when: never
```
Robbepop, line 157: What was the effect of this?

crates/lang/codegen/src/generator/dispatch.rs
```diff
@@ -247,7 +247,7 @@ impl Dispatch<'_> {
     /// Returns a tuple of:
     ///
     /// - Vector over the generated identifier bindings (`__ink_binding_N`) for all inputs.
-    /// - `TokenStream` representing the binding identifiers as tuple (for >=2 inputs),
+    /// - `TokenStream` representing the binding identifiers as tuple (for `>= 2` inputs),
```
Robbepop, line 250: "fore equal to or more than 2 inputs" might be better here

crates/storage/src/traits/optspec.rs
```diff
@@ -123,7 +123,7 @@ where
         Some(value) => {
             // Handle the Option<T> as if it was a T.
             //
-            // Sadly this doesn't not work well with `Option<Option<T>>`.
+            // Sadly this does not work well with `Option<Option<T>>`.
```
Robbepop, line 126: Pitfall avoided. :heavy_check_mark:  :D

.gitlab-ci.yml
cmichi: It prevented this CI stage from running if it was run as part of a pipeline (as our CI stages are typically). I think this was introduced mistakenly, since a while ago we had issues with `cargo spellcheck`.

PR #828, opened 2021-06-23
Update Spelling List

HCastano:
This PR adds  `refcount` and `postfix` as real words, fixes a couple of typos, and sorts
the spelling list for your viewing pleasure.


Robbepop (approved):
LGTM although I wonder what the ordering of the config file now is. Some words like `accessor` appear pretty much at the end. Rest seems to be alphabetically sorted.

HCastano:
> LGTM although I wonder what the ordering of the config file now is. Some words like `accessor` appear pretty much at the end. Rest seems to be alphabetically sorted.

Yeah, they're sorted but within one of three independent sections. `assessor/S` is in the last section

PR #830, opened 2021-06-25
Actually add `refcount` as word

HCastano:
Looks like I made a pretty embarassing mistake in #828 since I didn't actually add
`refcount` to the dictionary list.

However, the curious thing is that the CI didn't complain as part of the PR. When the CI
re-ran after having the PR land on `master` it failed. 

Given that the CI jobs are the same between PRs and `master` (as far as I can tell
anyways) this suggests a spurrious failure of `cargo-spellcheck`.
`cargo-spellcheck`.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -153,12 +153,12 @@ spellcheck:
     stage:                           workspace
     <<:                              *docker-env
     script:
-        - cargo spellcheck check --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1
+        - cargo spellcheck check -vvv --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1
         - for example in examples/*/; do
-            cargo spellcheck check --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 ${example};
+            cargo spellcheck check -vvv --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 ${example};
           done
         - for contract in ${DELEGATOR_SUBCONTRACTS}; do
-            cargo spellcheck check --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 examples/delegator/${contract}/;
+            cargo spellcheck check -vvv --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 examples/delegator/${contract}/;
           done
 
 codecov:
--- a/.config/cargo_spellcheck.dic
+++ b/.config/cargo_spellcheck.dic
@@ -43,6 +43,7 @@ invariants
 kB
 layed
 multisig
+multi
 postfix
 prefilled
 recurse
@@ -58,6 +59,7 @@ validator
 variadic
 
 Django/S
+IP/S
 NFT/S
 accessor/S
 allocator/S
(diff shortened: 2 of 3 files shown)


cmichi:
It currently also fails with those on `master`:
```
$ cargo spellcheck check --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1
error: spellcheck(Hunspell)
    --> /builds/parity/ink/crates/lang/ir/src/ir/config.rs:173
     |
 173 |  Asserts that the given input config attribute argument are converted
     |                               ^^^^^^
     | - configure
     |
     |   Possible spelling mistake found.
error: spellcheck(Hunspell)
   --> /builds/parity/ink/crates/storage/src/alloc/mod.rs:42
    |
 42 | ..AK, SHA-2 or BLAKE-2. For technical reasons we should prepend the
    |                                                         ^^^^^^^
    | - pretend
    |
    |   Possible spelling mistake found.
```
https://gitlab.parity.io/parity/ink/-/jobs/986516

cmichi:
@drahnr We are experiencing some issues with `cargo-spellcheck` (see above). Basically our CI doesn't find spellchecking mistakes on one run, but on a different run it suddenly finds new ones. Do you have an idea what could cause this? We are using the latest published crate version.

drahnr:
There are a few things that need to be done, to make this idempotent, I elaborated a bit in https://github.com/drahnr/cargo-spellcheck/issues/181 on them and fixed an outstanding issue.

drahnr (approved):

.gitlab-ci.yml
```diff
@@ -153,12 +153,12 @@ spellcheck:
     stage:                           workspace
     <<:                              *docker-env
     script:
-        - cargo spellcheck check --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1
+        - cargo spellcheck check -vvv --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1
```
drahnr, line 156: It's much much fast if you can pass all files at once, if that does not make the stack explode.

Most time is used to load the builtin rules, if you can make that a onetime offset, it could safe you a couple of `100ms x #files`

.gitlab-ci.yml
```diff
@@ -153,12 +153,12 @@ spellcheck:
...
         - for example in examples/*/; do
-            cargo spellcheck check --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 ${example};
+            cargo spellcheck check -vvv --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 ${example};
           done
         - for contract in ${DELEGATOR_SUBCONTRACTS}; do
-            cargo spellcheck check --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 examples/delegator/${contract}/;
+            cargo spellcheck check -vvv --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 examples/delegator/${contract}/;
```
drahnr, line 161: It's much much fast if you can pass all files at once, if that does not make the stack explode.

Most time is used to load the builtin rules, if you can make that a onetime offset, it could safe you a couple of `100ms x #files`

HCastano: I took a quick look at this, and figured that adding the `--recursive` flag would do the trick, but it doesn't. Maybe an issue for another time though, haha

drahnr: I guess it traverses recursively according to the `mod` defined in all the files given, at least that's what's supposed to happen.

Issue #827, opened 2021-06-22
Add new default bump allocator for heap memory allocations


Robbepop:
Currently we are using `wee_alloc` (https://crates.io/crates/wee_alloc) for allocating heap memory in smart contracts.
This allocator was specifically designed for WebAssembly usage and smart contracts and features relatively low footprint.
However, its footprint is still significant and blows up the complexity and Wasm file size of non-trivial smart contracts measurably.

On Ethereum it is not uncommon to use the most trivial heap memory allocator: The bump allocator which just increases a counter.
The downside of a bump allocator is that it usually cannot deallocate memory and therefore if an executable does lots of allocations during execution the memory will be drained or wasted.
However, since smart contracts execute in s short lived manner this should never really become a problem at all.

Therefore we want to add a bump allocator and set it as the new default allocator for ink! smart contracts. This also includes research into how this unfolds for Wasm binary size for non-trivial smart contracts. We not only expect the bump allocator to take significantly less space but we also expect that common routines such as `alloc` and `dealloc` will take a lot less space. So there are many potential gains for more complex contracts in this regard.

## ToDo

- Implement a very simply bump allocator in the `ink_allocator` crate.
    - Even due to the hint below it might make sense to check how much the `bumpalo` bump allocator already helps.
- Make it possible to switch between `wee_alloc` and the new `bump_alloc` via a crate feature for example.
- Set the new bump allocator as the new default allocator.
- Use our default contracts (or others) to research into how the different allocator unfolds into Wasm binary size gains (or not).

## Hints

I have looked at the `bumpalo` crate on crates.io. While it looks fine, is decently implemented and has respected authors it is not a perfect match for our use case unfortunately since it is design for non embedded use cases mostly where you might have the need to allocate much more than 4GB.
For our use case we could go with a much simpler and more light weight design that resembles some allocator that operates on some fixed size stack space, e.g. another 16kB buffer. Contract executions should rarely have the need to allocate a lot of heap memory during their short execution times.

HCastano:
From what I can tell we aren't able to use `bumpalo`. This is because it has a pretty fundamental assumption baked in, which is that it [allocates and deallocates memory using an already exiting global allocator](https://github.com/fitzgen/bumpalo/issues/100#issuecomment-792972404). This would leave us in the same situation since we're using `wee_alloc` as the global allocator.

As the crate author mentions in that comment though, we could fork `bumpalo` and make those `alloc/dealloc` calls manually.

[...]

PR #831, opened 2021-06-28
Add Bump Allocator

HCastano:
This PR adds a simple bump allocator based off this great tutorial: https://os.phil-opp.com/allocator-designs/#bump-allocator.

The goal here is to see whether or not a simpler allocator design will be able to
significantly reduce the space footprint of contracts which use it. See #827 for more
details.

Some things to note:
- ~~While this PR compiles, contracts compiled with the bump allocator panic on-chain
  (`ContractTrapped`!). I need to look into this~~ Fixed, thanks Alex!
- ~~I haven't been able to initalize the heap in a non-sketchy way - it's been a little
  hard with the requirements from `#[global_allocator]`~~ This isn't relevant anymore

### Early Results
I've built our example contracts with both the Bump allocator and the Wee allocator.

| Contract | Original (Bump Alloc) | Original (Wee Alloc) | Optimized (Bump Alloc) | Optimized (Wee Alloc) |
--- | --- | --- | --- | --- |
| `contract-terminate` | 18.7K | 18.7K | 2.1K | 2.1K |
| `contract-transfer` | 26.3K | 27.4K | 9.3K | 10.3K |
| `dns` | 44.2K | 45.5K | 25.6K | 26.5K |
| `erc1155` | 65.2K | 66.8K | 45.3K | 46.8K |
| `erc20` | 51.8K | 53.2K | 30.5K | 31.8K |
| `erc721` | 68.3K | 69.8K | 41.1K | 42.5K |
| `flipper` | 19.0K | 19.0K | 2.4K | 2.4K |
| `incrementer` | 19.0K | 19.0K | 2.4K | 2.4K |
| `multisig_plain` | 86.2K | 87.3K | 53.6K | 54.7K |
| `rand-extension` | 22.2K | 22.2K | 5.2K | 5.2K |
| `trait-erc20` | 52.3K | 53.7K | 31.0K | 32.2K |
| `trait-flipper` | 19.0K | 19.0K | 2.4K | 2.4K |
| `accumulator` | 25.7K | 25.7K | 8.5K | 8.5K |
| `adder` | 30.3K | 30.3K | 12.8K | 12.8K |
| `subber` | 30.3K | 30.3K | 12.8K | 12.8K |
| `delegator` | 28.3K | 28.3K | 10.8K | 10.8K |

As you can see, the savings are generally around 1K, which really isn't a whole lot given
how much better `wee_alloc` is over the bump allocator.


Robbepop:
> As you can see, the savings are generally around 1K, which really isn't a whole lot given
how much better wee_alloc is over the bump allocator.

1kB saving are already a lot since we are talking about a bottleneck issue here.
Execution times of Wasm smart contracts are less important at the moment than their final Wasm file sizes.
Many of the examples show zero difference since ink! is very good at optimizing away the allocator entirely which is good and expected.
Maybe we can do even better than this, I need to view the code closely.
Also I think the bump allocator should also be able to perform better since it needs less management due to allocated memory not being able to deallocate.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,104 @@
...
+            next: 0,
+            heap_initialized: false,
+        }
+    }
+}
+
+unsafe impl GlobalAlloc for Locked<BumpAllocator> {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        // This should be okay performance wise since we're in a single threaded context anyways
+        let mut bump = self.lock();
```
Robbepop, line 72: Eagerly locking in a smart contract is something you should never do since this help nothing at best and burns gas at worst.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,104 @@
...
+pub struct BumpAllocator {
+    /// Points to the start of the next available allocation
+    next: usize,
+    /// We need some way to initialize our heap. However, I can't figure out how to get the
+    /// initialization working properly with `lazy_static` so this hack is the best I got for now.
+    heap_initialized: bool,
+}
+
+impl BumpAllocator {
+    pub const fn new() -> Self {
```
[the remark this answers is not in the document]
HCastano: It's a draft PR, you can't blame me for not documenting stuff yet 😉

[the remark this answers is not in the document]
Robbepop: I keep forgetting about draft PRs :sweat_smile:

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,104 @@
...
+use core::alloc::{
+    GlobalAlloc,
+    Layout,
+};
+
+/// A page in Wasm is 64KiB
+const PAGE_SIZE: usize = 64 * 1024;
+
+lazy_static::lazy_static! {
+    pub static ref BUMP_ALLOC: Locked<BumpAllocator> = Locked::new(BumpAllocator::new());
```
athei, line 33: This is unused in this PR. I guess some leftover of some experiment?

HCastano: Yeah, it was from some previous attempts

crates/allocator/Cargo.toml
```diff
@@ -15,8 +15,12 @@ categories = ["no-std", "embedded"]
 include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 
 [dependencies]
-wee_alloc = { version = "0.4", default-features = false }
+wee_alloc = { version = "0.4", default-features = false, optional = true }
+spin = { version = "0.9", default-features = false, features = ["mutex", "spin_mutex"], optional = true }
+lazy_static = { version = "1.4", default-features = false, features = ["spin_no_std"], optional = true }
```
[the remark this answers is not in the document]
HCastano: @Robbepop I've updated the code to use a global mut instead, can you take a look and see if that's more or less what you had in mind?

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    unsafe fn dealloc(&self, _ptr: *mut u8, _layout: Layout) {}
+}
+
+struct InnerAlloc {
+    /// Points to the start of the next available allocation.
+    next: usize,
+    /// We need some way to initialize our heap. However, I can't figure out how to get the
+    /// initialization working properly with `lazy_static` so this hack is the best I got for now.
+    heap_initialized: bool,
```
athei, line 49: Maybe just use `next: Option<usize>` in order to indicate the non-initialized state with a `None`. This would be more idiomatic.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,87 @@
...
+        Self {
+            next: 0,
+            heap_initialized: false,
+        }
+    }
+
+    unsafe fn alloc(&mut self, layout: Layout) -> *mut u8 {
+        // TODO: Figure out how to properly initalize the heap
+        if !self.heap_initialized {
+            let ptr = core::arch::wasm32::memory_grow(0, 1);
```
athei, line 63: This does not return a pointer (mentioning this because of the binding's name). It returns the number of pages before the grow instruction was executed.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,87 @@
...
+    next: usize,
+    /// We need some way to initialize our heap. However, I can't figure out how to get the
+    /// initialization working properly with `lazy_static` so this hack is the best I got for now.
+    heap_initialized: bool,
+}
+
+impl InnerAlloc {
+    pub const fn new() -> Self {
+        Self {
+            next: 0,
```
athei, line 55: This is wrong. You will most probably overwrite statically allocated data structures put there by the linker (data, stack). You can't assume that your heap starts at `0`. When I compile the flipper contract I can see that the initial memory size is `2` pages and those are in use:

```wat
(import "env" "memory" (memory (;0;) 2 16))
````


You have two choices now:

1. Use the return value of `memory_grow` to calculate where your heap should start. This is a freshly allocated page and guaranteed to not be in used by the linker. This is how `wee_alloc` deals with it.

2. Try to be clever and get away without even calling `memory.grow` once. AFAIK `lld` used to expose the `__heap_base` symbol which you can reference to determine the address where you can safely put your heap (at this address and any higher address). However, it stopped to to do so by default (you need to put `-Clink-arg=--export=__heap_base` into your rust flags now). Also, one need to make sure to dimension the initial memory large enough so that there is enough space for a heap. In any case this needs new flags to be supplied to the linker (therefore changes to cargo contract). I would recommend using the first approach first and leave this as a later optimization. By the way this is how substrate runtimes handle their heap. With the little difference that the allocator is placed inside the client and not the runtime.

**Review by athei — CHANGES_REQUESTED** [review]
Requesting changes because the contains a bug if I am not mistaken.

File: crates/allocator/src/bump.rs (hunk 7 above)
Comment (HCastano) [reply crates/allocator/src/bump.rs:55]: Oh wow yeah, good point! I think I took the whole "each contract has its own Wasm sandbox" thing too seriously and forgot about everything else 😅

> When I compile the flipper contract I can see that the initial memory size is 2 pages and those are in use

Where/how did you get this? I looked through the "normal" build artifacts (`*.contract`, `*.wasm`, `metadata.json`) and can't find it

> Use the return value of memory_grow to calculate where your heap should start. This is a freshly allocated page and guaranteed to not be in use by the linker. This is how wee_alloc deals with it.

Yeah, this makes sense.

> Try to be clever and get away without even calling memory.grow once.

I'm gonna skip being clever and get something working first 😄

File: crates/allocator/src/bump.rs (hunk 5 above)
Comment (HCastano) [reply crates/allocator/src/bump.rs:49]: Yeah good point.

I was hoping to not need to track the initialization status and instead assume in `alloc` that the heap had already been allocated (maybe in `new()` or some `init()` function), but I haven't been able to do that yet

File: crates/allocator/src/bump.rs (hunk 7 above)
Comment (athei) [reply crates/allocator/src/bump.rs]: `Where/how did you get this? I looked through the "normal" build artifacts (*.contract, *.wasm, metadata.json) and can't find it`

Just disassemble the `.wasm` file. You can use `wasm2wat` for that. Is it part of the `wabt` package.

File: crates/allocator/src/bump.rs (hunk 5 above)
Comment (Robbepop) [reply crates/allocator/src/bump.rs]: ink! codegen could do this

File: crates/allocator/src/bump.rs [hunk 8]
```diff
@@ -0,0 +1,94 @@
...
+
+use core::alloc::{
+    GlobalAlloc,
+    Layout,
+};
+
+/// A page in Wasm is 64KiB
+const PAGE_SIZE: usize = 64 * 1024;
+
+static mut INNER: InnerAlloc = InnerAlloc::new();
```
athei, line 32: Maybe use an `Option<InnerAlloc>` here in order to not have those two Options which represent the same state.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    unsafe fn alloc(&mut self, layout: Layout) -> *mut u8 {
+        // TODO: Figure out how to properly initalize the heap
+        let alloc_start = if let Some(start) = self.next {
+            start
+        } else {
+            let prev_page = core::arch::wasm32::memory_grow(0, 1);
+            if prev_page == usize::max_value() {
+                panic!("OOM")
```
athei, line 71: You won't be able to see this message anywhere I am afraid. We are compiling the core lib so that it just dies silently on panic in order to prevent all the fmt stuff to be pulled it. Additionally, it is recommended to not `panic` here but rather return a nullptr or call `handle_alloc_error`:  https://doc.rust-lang.org/stable/core/alloc/trait.GlobalAlloc.html#errors

What I would recommend is the following: Call `ink_env::debug_message` with your error string and then return a null pointer. This way you can have the error messages when enabling `ink_env/ink-debug`.

The same es true for all the places you call `expect`.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            let start = prev_page.checked_mul(PAGE_SIZE).expect("OOM");
+            self.upper_limit = Some(start + PAGE_SIZE);
```
athei, line 75: Why no checked math here?

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,94 @@
...
+        };
+
+        let aligned_layout = layout.pad_to_align();
+        let alloc_end = match alloc_start.checked_add(aligned_layout.size()) {
+            Some(end) => end,
+            None => return core::ptr::null_mut(),
+        };
+
+        let upper_limit = self.upper_limit
+            .expect("Since we're here the heap must've been initialized, therefore this must exist.");
```
athei, line 86: This `expect` only exist because the state is not modeled optimally. See my comment about using `Option<InnerAlloc>`.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl InnerAlloc {
+    pub const fn new() -> Self {
+        Self {
+            next: None,
+            upper_limit: None,
+        }
+    }
+
+    unsafe fn alloc(&mut self, layout: Layout) -> *mut u8 {
```
athei, line 64: Can you remove this `unsafe` here? I think no unsafe code is actually used in this function. And even if it is we should put blocks around those instead of making the whole function unsafe.

crates/allocator/Cargo.toml
```diff
@@ -15,8 +15,10 @@ categories = ["no-std", "embedded"]
...
 
 [dependencies]
-wee_alloc = { version = "0.4", default-features = false }
+wee_alloc = { version = "0.4", default-features = false, optional = true }
 
 [features]
 default = ["std"]
 std = []
+bump-allocator = []
+wee-allocator = ["wee_alloc"]
```
Robbepop, line 24: We actually don't need 2 crate features if we declare one as default. If this experiment is successful we might declare the `bump-allocator` as the default and have `wee-allocator` as opt-in feature to use that allocator instead for contracts that actually need it.

crates/allocator/src/bump.rs
Robbepop: I very much agree with this! Also we really should not have a check in `alloc` that checks all the time if the allocator has already been initialized and instead enforce initialization via ink! (or other means) so that we can clearly assume that the state has been properly initialized before the first `alloc` call.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,94 @@
...
+static mut INNER: InnerAlloc = InnerAlloc::new();
+
+pub struct BumpAllocator;
+
+unsafe impl GlobalAlloc for BumpAllocator {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        INNER.alloc(layout)
+    }
+
+    unsafe fn dealloc(&self, _ptr: *mut u8, _layout: Layout) {}
```
Robbepop, line 41: we might mark those as `#[inline]`.

crates/allocator/src/bump.rs
Robbepop: I want to add further that I consider `expect` to be a bad use for proper error information forwarding since `expect` can only forward a `&str` and nothing more. Instead what I usually do is to call `.unwrap_or_else(|error| panic!("something bad happened: {}", error))`.

crates/allocator/src/bump.rs
Robbepop: We should make sure to make the `alloc` call as efficient as possible, however, I agree that we want to have all necessary checked arithmetic, however, none more than that.

crates/allocator/src/lib.rs
```diff
@@ -23,8 +23,21 @@
...
+#[cfg(not(feature = "wee-allocator"))]
+#[global_allocator]
+static mut ALLOC: bump::BumpAllocator = bump::BumpAllocator {};
+
 #[cfg(not(feature = "std"))]
 mod handlers;
+
+#[cfg(not(feature = "std"))]
+#[cfg(feature = "bump-allocator")]
+#[cfg(not(feature = "wee-allocator"))]
```
Robbepop, line 42: We really should remove the `bump-allocator` feature and make it our default and make `wee-alloc` an opt-in instead of this where we would always require one of the features of be enabled.

Robbepop (changes requested):

crates/allocator/Cargo.toml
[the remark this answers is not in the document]
Robbepop: Yep, this is what I had in mind.

crates/allocator/src/bump.rs
Robbepop: I agree, keeping unsafe blocks and scopes as small as possible always has advantages.

crates/allocator/src/bump.rs
HCastano: > You won't be able to see this message anywhere I am afraid. We are compiling the core lib so that it just dies silently on panic in order to prevent all the fmt stuff to be pulled it.

Gotcha, is that because of [`cargo-contract`'s use of `panic_immediate_abort`](https://github.com/paritytech/cargo-contract/blob/50d27922d6947cf7f8630dfe0ee7b9bb5b4de249/src/cmd/build.rs#L189)?

> Additionally, it is recommended to not `panic` here but rather return a nullptr or call `handle_alloc_error`: https://doc.rust-lang.org/stable/core/alloc/trait.GlobalAlloc.html#errors

Ah, I didn't read that in the docs, thanks!

crates/allocator/Cargo.toml
HCastano: Yeah, I figured this would something to simplify later if we decide that the bump allocator is worth having

crates/allocator/src/bump.rs
HCastano: It was leftover from when this was the main `GlobalAlloc::alloc()` implementation

crates/allocator/src/bump.rs
HCastano: >  Also we really should not have a check in alloc that checks all the time if the allocator has already been initialized

I'm sure a half decent branch predictor will be able to figure out that the allocator has already initialized and not even do the check 😛 

> instead enforce initialization via ink! (or other means)

Can you be more specific here? I don't know where in ink! we could initialize this. As Alex suggested, we could get ride of the call to `wasm32::grow_memory` by fiddling with the linker flags, but I'd like to avoid this for now.

crates/allocator/src/bump.rs
HCastano: > What I would recommend is the following: Call ink_env::debug_message with your error string and then return a null pointer. This way you can have the error messages when enabling ink_env/ink-debug.

We can't use `ink_env` in the `ink_allocator` crate or else we end up with cyclic dependencies. I think we could move the debugging code out of `ink_env`, but that's probably better left for another PR

crates/allocator/Cargo.toml
Robbepop: Even the first iteration reduced Wasm file sizes by 1.2-1.3kB each so it is already a good win there. I am very sure we will see the `bump_allocator` being our default allocator soon.

crates/allocator/src/bump.rs
```diff
@@ -69,29 +69,33 @@ impl InnerAlloc {
...
                 return core::ptr::null_mut()
             }
 
-            let start = prev_page
-                .checked_mul(PAGE_SIZE)
-                .unwrap_or_else(|| return core::ptr::null_mut());
+            let start = match prev_page.checked_mul(PAGE_SIZE) {
+                Some(s) => s,
+                None => return core::ptr::null_mut(),
+            };
```
Robbepop, line 79: To help with code like this you could define another implementation function for this function with a return type of `Option<*mut u8>` instead of `*mut u8` and convert the `None` case to a `core::ptr::null_mut()`. Then you could use `?` syntax in the implementation function. Then the `alloc` function would simply act as a thin wrapper around this. This might bring other downsides of which I am unaware of, yet.

HCastano:
@Robbepop I've updated the number in the table. They stayed the same for the most part, with the exception of two places where they contract size went up by `0.2K` and `0.3K`.

crates/allocator/src/bump.rs
athei: > Gotcha, is that because of [`cargo-contract`'s use of `panic_immediate_abort`](https://github.com/paritytech/cargo-contract/blob/50d27922d6947cf7f8630dfe0ee7b9bb5b4de249/src/cmd/build.rs#L189)?

Yes and also because the panic messages are not really forwarded from the execution engine to the client console. You will see this messages when executing in inks off-chain environment, though.

crates/allocator/src/bump.rs
HCastano: I've added an initialization function, but my original question still stands. Where in ink! can we initialize this before it's used? My guess is that it's somewhere in `crates/storage/src/alloc`, but I can't figure out where.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Initialize the heap which backs the bump allocator.
+    ///
+    /// Our heap is a single page of Wasm memory (64KiB) and will not grow beyond that.
+    ///
+    /// Note that this function must be called before any allocations can take place, otherwise any
+    /// attempts to perform an allocation will fail.
+    fn init(&mut self) {
+        let prev_page = core::arch::wasm32::memory_grow(0, 1);
+        if prev_page == usize::MAX {
+            todo!()
```
HCastano, line 85: What we do instead of these `todo!`'s will depend on the context in which `init()` is called

crates/allocator/src/bump.rs
Robbepop: Ideally ink! codegen handles this somehow. My question on the matter of optimal codegen still stands too.

crates/allocator/src/bump.rs
Robbepop: The `init` function could return a `enum InitState { AlreadyInitialized, NowInitialized }` value to handle those `todo!` cases.

Robbepop: In the case of `prev_page == usize::MAX` we should panic since that's really an exceptional case where we can do nothing.

Robbepop (commented):
The bump allocator is still in need of some tests even though its implementation is rather simple. However, we definitely want to optimize it at a later point and then tests will be very important for us.

crates/allocator/src/bump.rs
Robbepop: It seems with the new `init` call we no longer are in need of this "uninitialized" state.

crates/allocator/src/bump.rs
```diff
@@ -117,3 +170,33 @@ impl InnerAlloc {
...
+    fn can_alloc_a_vec() {
+        let mut v = Vec::new();
+        v.push(1)
+    }
+
+    #[test]
+    fn can_alloc_a_big_vec() {
+        let mut v = Vec::with_capacity(PAGE_SIZE);
+        v.push(true)
+    }
```
Robbepop, line 193: I would write the tests in terms of unit tests, using the `BumpAllocator` directly instead of the `#[global_alloc]`.

Robbepop: Then you can also specify specific testing APIs which you can call (if you think it is needed) to grasp more information about what happens behind the scene.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,202 @@
...
+            /// Our heap is 64KiB of memory (to match the size of a Wasm page), and will not grow
+            /// beyond that.
+            ///
+            /// Note that this function must be called before any allocations can take place, otherwise any
+            /// attempts to perform an allocation will fail.
+            ///
+            /// This implementation is only meant to be used for testing, since we cannot (easily)
+            /// test the `wasm32` implementation.
+            fn init(&mut self) {
+                let start = unsafe {
```
athei, line 115: This unsafe block can be smaller. You can nest the blocks if necessary.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,202 @@
...
+                    let protection_bits = libc::PROT_WRITE | libc::PROT_READ;
+                    let flags = libc::MAP_ANONYMOUS | libc::MAP_PRIVATE;
+                    let fd = -1;
+                    let offset = 0;
+
+                    // _Technically_ the `PAGE_SIZE` here will more than likely *not* match the page
+                    // size of our non-wasm32 architecture, but it's fine to request that many bytes
+                    // from `mmap`.
+                    libc::mmap(
```
athei, line 124: You could just use `libc::malloc` here which also exists on windows.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,202 @@
...
+    /// allocations are made.
+    #[inline]
+    pub fn init(&self) {
+        // SAFETY: We are in a single threaded context, so we don't have to worry about this being
+        // concurrently mutated by multiple threads.
+        unsafe { INNER.init() }
+    }
+}
+
+unsafe impl GlobalAlloc for BumpAllocator {
```
athei, line 50: You can override `alloc_zeroed` to just call `alloc`. We can save the zero memory operation because the new page is guaranteed to be zero initialized (on wasm).

HCastano:
I didn't get around to addressing all your comments today, will continue tomorrow

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,152 @@
...
+            /// Returns `None` if a page isn't available.
+            ///
+            /// This implementation is only meant to be used for testing, since we cannot (easily)
+            /// test the `wasm32` implementation.
+            fn request_page(&mut self) -> Option<usize> {
+                // _Technically_ the `PAGE_SIZE` here will more than likely *not* match the page
+                // size of our non-wasm32 architecture, but it's fine to request that many bytes
+                // from `malloc`.
+                let start = unsafe {
+                    libc::malloc(PAGE_SIZE)
```
HCastano, line 92: @athei Hey, I'm stuck in the switch from `mmap` to `malloc`. I keep segfaulting when trying to run the tests and I'm not sure why.

I know that I'm not getting  a null pointer back from `malloc`, and I know the `malloc_usable_size` for `start` is equal to `PAGE_SIZE`. 

When running the code through `valgrind` I get the following: 

```
==147101== Process terminating with default action of signal 11 (SIGSEGV)
==147101==  General Protection Fault
==147101==    at 0x11A835: _mm_load_si128 (sse2.rs:1182)
==147101==    by 0x11A835: load_aligned (sse2.rs:59)
==147101==    by 0x11A835: new<(alloc::string::String, bool)> (mod.rs:1726
...
```

Maybe there's some memory segment permissions I need to set?

I've also run the tests with `rust-gdb` and get the following with a `PAGE_SIZE` sized heap.

```
<ink_allocator::bump::BumpAllocator as core::alloc::global::GlobalAlloc>::alloc (self=0x5555556110c8 <std::thread::ThreadId::new::GUARD>, layout=...)
    at crates/allocator/src/bump.rs:41
41	            Some(start) => start as *mut u8,
(gdb) n
44	    }
(gdb) n
alloc::alloc::Global::alloc_impl () at /rustc/798baebde1fe77e5a660490ec64e727a5d79970d/library/alloc/src/alloc.rs:167
167	/rustc/798baebde1fe77e5a660490ec64e727a5d79970d/library/alloc/src/alloc.rs: No such file or directory.
(gdb) n
alloc::raw_vec::RawVec<T,A>::allocate_in () at /rustc/798baebde1fe77e5a660490ec64e727a5d79970d/library/alloc/src/raw_vec.rs:207
207	/rustc/798baebde1fe77e5a660490ec64e727a5d79970d/library/alloc/src/raw_vec.rs: No such file or directory.
(gdb) n
<T as alloc::slice::hack::ConvertVec>::to_vec () at /rustc/798baebde1fe77e5a660490ec64e727a5d79970d/library/alloc/src/slice.rs:228
228	/rustc/798baebde1fe77e5a660490ec64e727a5d79970d/library/alloc/src/slice.rs: No such file or directory.
(gdb) n
std::sys_common::rt::init () at library/std/src/sys_common/rt.rs:23
23	library/std/src/sys_common/rt.rs: No such file or directory.
(gdb) n
24	in library/std/src/sys_common/rt.rs
(gdb) n
std::rt::lang_start_internal () at library/std/src/rt.rs:45
45	library/std/src/rt.rs: No such file or directory.
(gdb) n

Program received signal SIGSEGV, Segmentation fault.
hashbrown::raw::RawIterRange<T>::new () at /cargo/registry/src/github.com-1ecc6299db9ec823/hashbrown-0.11.0/src/raw/mod.rs:1726
1726	/cargo/registry/src/github.com-1ecc6299db9ec823/hashbrown-0.11.0/src/raw/mod.rs: No such file or directory.
(gdb) n
std::sys::unix::stack_overflow::imp::signal_handler () at library/std/src/sys/unix/stack_overflow.rs:94
94	library/std/src/sys/unix/stack_overflow.rs: No such file or directory.
(gdb) n
99	in library/std/src/sys/unix/stack_overflow.rs
(gdb) n
100	in library/std/src/sys/unix/stack_overflow.rs
(gdb) n
104	in library/std/src/sys/unix/stack_overflow.rs
(gdb) n
113	in library/std/src/sys/unix/stack_overflow.rs
(gdb) n
114	in library/std/src/sys/unix/stack_overflow.rs
(gdb) n
118	in library/std/src/sys/unix/stack_overflow.rs
(gdb) n
hashbrown::raw::RawIterRange<T>::new () at /cargo/registry/src/github.com-1ecc6299db9ec823/hashbrown-0.11.0/src/raw/mod.rs:1726
1726	/cargo/registry/src/github.com-1ecc6299db9ec823/hashbrown-0.11.0/src/raw/mod.rs: No such file or directory.
(gdb) n

Program received signal SIGSEGV, Segmentation fault.
hashbrown::raw::RawIterRange<T>::new () at /cargo/registry/src/github.com-1ecc6299db9ec823/hashbrown-0.11.0/src/raw/mod.rs:1726
1726	in /cargo/registry/src/github.com-1ecc6299db9ec823/hashbrown-0.11.0/src/raw/mod.rs
(gdb) n

Program terminated with signal SIGSEGV, Segmentation fault.
The program no longer exists.
```

If I use `4KiB` for `PAGE_SIZE` I still end up getting killed, but there's no mention of a stack overflow in the stack trace.

Is there anything obvious that I'm doing wrong?

crates/allocator/src/bump.rs
HCastano: This was a really helpful tip actually. After banging my head trying to hunt down segfaults today this allowed me to sidestep it all

crates/allocator/src/bump.rs
HCastano: I've ended up changing the `request_page` test implementation so that it doesn't request any memory from the OS, so this comment isn't relevant anymore. With that said though, if you have any knowledge to share here I'd still appreciate it 😄

athei: > I've ended up changing the request_page test implementation so that it doesn't request any memory from the OS

I guess I am missing something but how is this supposed to work? The code looks like you are just returning addresses starting at `0x0` which are certainly invalid.

HCastano: > > I've ended up changing the request_page test implementation so that it doesn't request any memory from the OS
> 
> I guess I am missing something but how is this supposed to work? The code looks like you are just returning addresses starting at `0x0` which are certainly invalid.

It's not a "real" implementation, it's just for checking that the internal state of the allocator is what we expect

crates/allocator/src/bump.rs
HCastano: @athei I made the change since I believe you, but do you have any docs that mention this? I couldn't find any

crates/allocator/src/bump.rs
athei: Ahh k you are not really using the returned memory. Just for unit tests in the same file.  That should be safe. Regarding your initial problem: The one thing comes to mind that the memory did not conform to the requested alignment. Some manual work might have been necessary to pad (increase) the returned pointer from `malloc`.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,243 @@
...
+            fn request_pages(&mut self, pages: usize) -> Option<usize> {
+                let prev_page = core::arch::wasm32::memory_grow(0, pages);
+                if prev_page == usize::MAX {
+                    return None;
+                }
+
+                prev_page.checked_mul(PAGE_SIZE)
+            }
+
+        } else if #[cfg(feature = "std")] {
```
athei, line 96: I would rather see `[cfg(test)]` being used here. It is simpler and expressed the intent of the code. Also because the implementation is unsound for everything but inspecting the internal state via the unit tests.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,243 @@
...
+    next: usize,
+
+    /// The address of the upper limit of our heap.
+    upper_limit: usize,
+
+    /// The number of page requests made so far.
+    ///
+    /// This is meant to mimic the behavior exhibited by `core::arch::wasm32::memory_grow`
+    #[cfg(feature = "std")]
+    page_requests: usize,
```
athei, line 69: I would love if you could manage to get rid of this field. My feeling is that it is possible to use a shared field both for the testing and the production code. Isn't `page_requested = upper_limit / PAGE_SIZE` in the test case where the memory starts at `0`?

crates/allocator/src/bump.rs
athei: We already rely on this because cargo contracts supplies the `--zero-filled-memory` to `wasm-opt`. Without it we have a lot of zero data segments in our contract modules which bloats the binaries size.

These are the semantics for the memory as defined in the wasm module:
https://webassembly.github.io/spec/core/exec/modules.html#memories

Grow semantics:
https://webassembly.github.io/spec/core/exec/modules.html#growing-memories

crates/allocator/src/bump.rs
HCastano: Based off [the docs](https://linux.die.net/man/3/malloc) it doesn't look like that's the case:

> The malloc() and calloc() functions return a pointer to the allocated memory that is suitably aligned for any kind of variable.

crates/allocator/src/bump.rs
HCastano: So we can't _just_ use `cfg[test]` here, otherwise it won't compile. I've changed it to be `#[cfg(any(test, feature = "std"))]` so that at least there's an in-code reference that this is meant to be used for testing

crates/allocator/src/bump.rs
athei: This essential means 8 byte aligned. I can imagine that Rust will request larger alignments in some cases. It is unlikely, though.

crates/allocator/src/bump.rs
athei: > So we can't just use cfg[test] here, otherwise it won't compile.

Any idea why that is? I only see this used in tests outside of `no_std`.

cmichi:
Haven't looked at the code yet, but it would be great to have fuzz tests for this allocator (could be in a follow-up). We have a number of fuzz tests for data structures, etc.. Those are run after each merge to `master` for some time.

crates/allocator/src/bump.rs
HCastano: When the `cargo test` runs it does two passes. One with `--features default std`, and another with `<previous flags> --test` (you can verify this by running with `--verbose`).

This means that on the one pass there's no function called `request_pages` unless the `std` feature is enabled, which leads to compilation failures.

HCastano:
> Haven't looked at the code yet, but it would be great to have fuzz tests for this allocator (could be in a follow-up). We have a number of fuzz tests for data structures, etc.. Those are run after each merge to `master` for some time.

Yeah, that would be a good idea!

crates/env/Cargo.toml
```diff
@@ -65,3 +65,4 @@ std = [
 # Enable contract debug messages via `debug_print!` and `debug_println!`.
 ink-debug = []
 ink-experimental-engine = ["ink_engine"]
+wee-alloc = ["ink_allocator/wee-alloc"]
```
cmichi, line 68: What's the plan for `wee-alloc`? I mean, why even keep it in here? If there is some valuable trade-off between both of them we should add some info on it to the docs: https://paritytech.github.io/ink-docs/datastructures/dynamic-allocation.

cmichi (approved):
LGTM! I like the simplicity :-).

Could you create a follow-up issue for adding fuzz tests please?

crates/allocator/src/bump.rs
athei: But that compilation pass with `std` enabled and `test` disabled should never be executed. So you could add a branch for that where you just put `unreachable!("This code should only be ever called by tests")`.

crates/env/Cargo.toml
HCastano: The tradeoff here is `.wasm` size vs. efficient use of memory.`wee_alloc` is able to re-use freed memory, while the `bump` allocator is not. 

I agree though, if we do keep both it would be valuable to document the trade-offs. I've opened https://github.com/paritytech/ink-docs/issues/24.

HCastano:
> LGTM! I like the simplicity :-).
> 
> Could you create a follow-up issue for adding fuzz tests please?

Sure, I opened https://github.com/paritytech/ink/issues/860.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,241 @@
...
+impl InnerAlloc {
+    const fn new() -> Self {
+        Self {
+            next: 0,
+            upper_limit: 0,
+        }
+    }
+
+    cfg_if::cfg_if! {
+        if #[cfg(all(not(feature = "std"), target_arch = "wasm32"))] {
```
athei, line 77: I think you don't need to check for `std` here as in is always `no_std` when build for wasm.

crates/allocator/src/bump.rs
```diff
@@ -0,0 +1,241 @@
...
+            /// Request a `pages` number of page sized sections of Wasm memory. Each page is `64KiB` in size.
+            ///
+            /// Returns `None` if a page is not available.
+            ///
+            /// This implementation is only meant to be used for testing, since we cannot (easily)
+            /// test the `wasm32` implementation.
+            fn request_pages(&mut self, _pages: usize) -> Option<usize> {
+                Some(self.upper_limit)
+            }
+        } else if #[cfg(all(not(test), feature = "std"))] {
```
athei, line 99: You do not need to check for `not(test)` here anymore because it is checked by the previous branch.

athei (approved):
Just some suggestions for simplification of the ifdef expressions.

Issue #834, opened 2021-06-30
Implement binary_search for ink_storage::Vec


KaiserKarel:
It seems that ink_storage::Vec is missing some useful methods, such as binary search. Is such a feature wanted?

HCastano:
Having `binary_search` seems reasonable to me. Feel free to open a PR

Robbepop:
I would love to see that!

KaiserKarel:
I think that for the search we will need to access storage for the comparison, but no more than that.

HCastano:
Closed by #836.

PR #836, opened 2021-07-01
Implement binary_search for `collections::Vec`

KaiserKarel:
This ports the binary_search implementation from `core`. I've removed the use of `get_unchecked` as that is not present in `LazyIndexMap`.  There's 2 tests added to verify that it works with a full vector and an empty vector. 

related to #834

### DIFF
--- a/crates/storage/src/collections/vec/mod.rs
+++ b/crates/storage/src/collections/vec/mod.rs
@@ -285,6 +285,8 @@ where
     /// let r = s.binary_search_by(|probe| probe.cmp(&seek));
     /// assert!(match r { Ok(1..=4) => true, _ => false, });
     /// ```
+    // The binary_search implementation is ported from [core::slice] and attempts to remain as close
+    // to the source as possible.
     #[inline]
     pub fn binary_search_by<'a, F>(&'a self, mut f: F) -> Result<u32, u32>
     where
--- a/crates/storage/src/collections/vec/mod.rs
+++ b/crates/storage/src/collections/vec/mod.rs
@@ -223,7 +223,10 @@ where
     /// ```
     /// use ink_storage::Vec as StorageVec;
     ///
-    /// let s: StorageVec<i32> = [0, 1, 1, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55].into_iter().copied().collect();
+    /// let s: StorageVec<i32> = [0, 1, 1, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55]
+    ///     .into_iter()
+    ///     .copied()
+    ///     .collect();
     ///
     /// assert_eq!(s.binary_search(&13),  Ok(9));
     /// assert_eq!(s.binary_search(&4),   Err(7));
@@ -266,7 +269,10 @@ where
     /// ```
     /// use ink_storage::Vec as StorageVec;
     ///
-    /// let s: StorageVec<i32> = [0, 1, 1, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55].into_iter().copied().collect();
+    /// let s: StorageVec<i32> = [0, 1, 1, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55]
+    ///     .into_iter()
+    ///     .copied()
+    ///     .collect();
     ///
     /// let seek = 13;
     /// assert_eq!(s.binary_search_by(|probe| probe.cmp(&seek)), Ok(9));
@@ -337,9 +343,24 @@ where
     /// ```
     /// use ink_storage::Vec as StorageVec;
     ///
-    /// let s: StorageVec<(i32, i32)> = [(0, 0), (2, 1), (4, 1), (5, 1), (3, 1),
-    ///          (1, 2), (2, 3), (4, 5), (5, 8), (3, 13),
-    ///          (1, 21), (2, 34), (4, 55)].into_iter().copied().collect();
+    /// let s: StorageVec<(i32, i32)> = [
+    ///     (0, 0),
+    ///     (2, 1),
+    ///     (4, 1),
+    ///     (5, 1),
+    ///     (3, 1),
+    ///     (1, 2),
+    ///     (2, 3),
+    ///     (4, 5),
+    ///     (5, 8),
+    ///     (3, 13),
+    ///     (1, 21),
+    ///     (2, 34),
+    ///     (4, 55),
+    /// ]
+    /// .into_iter()
+    /// .copied()
+    /// .collect();
     ///
     /// assert_eq!(s.binary_search_by_key(&13, |&(a, b)| b),  Ok(9));
     /// assert_eq!(s.binary_search_by_key(&4, |&(a, b)| b),   Err(7));
--- a/crates/storage/src/collections/vec/mod.rs
+++ b/crates/storage/src/collections/vec/mod.rs
@@ -201,7 +201,7 @@ where
         self.elems.put(last_index, Some(value));
     }
 
-    /// Binary searches this sorted slice for a given element.
+    /// Binary searches this sorted vector for a given element.
     ///
     /// If the value is found then [`Result::Ok`] is returned, containing the
     /// index of the matching element. If there are multiple matches, then any
@@ -209,11 +209,10 @@ where
     /// [`Result::Err`] is returned, containing the index where a matching
     /// element could be inserted while maintaining sorted order.
     ///
-    /// See also [`binary_search_by`], [`binary_search_by_key`], and [`partition_point`].
+    /// See also [`binary_search_by`], [`binary_search_by_key`].
     ///
-    /// [`binary_search_by`]: slice::binary_search_by
-    /// [`binary_search_by_key`]: slice::binary_search_by_key
-    /// [`partition_point`]: slice::partition_point
+    /// [`binary_search_by`]: Vec::binary_search_by
+    /// [`binary_search_by_key`]: Vec::binary_search_by_key
     ///
     /// # Examples
     ///
@@ -222,7 +221,7 @@ where
     /// found; the fourth could match any position in `[1, 4]`.
     ///
     /// ```
-    /// let s = [0, 1, 1, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55];
+    /// let s = Vec::from([0, 1, 1, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55]);
     ///
     /// assert_eq!(s.binary_search(&13),  Ok(9));
     /// assert_eq!(s.binary_search(&4),   Err(7));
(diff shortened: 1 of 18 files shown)


KaiserKarel:
I think that the spellcheck failure is a false alarm, as the documentation is taken from core.

crates/storage/src/collections/vec/mod.rs
```diff
@@ -200,6 +200,169 @@ where
...
+        let mut size = self.len();
+        let mut left = 0;
+        let mut right = size;
+        while left < right {
+            let mid = left + size / 2;
+
+            // SAFETY: the call is made safe by the following invariants:
+            // - `mid >= 0`
+            // - `mid < size`: `mid` is limited by `[left; right)` bound.
+            let cmp = f(self.elems.get(mid).unwrap());
```
Robbepop, line 305:
```suggestion
            let cmp = f(&self.elems[mid]);
```

crates/storage/src/collections/vec/mod.rs
```diff
@@ -200,6 +200,169 @@ where
...
+    {
+        use core::cmp::Ordering::*;
+
+        let mut size = self.len();
+        let mut left = 0;
+        let mut right = size;
+        while left < right {
+            let mid = left + size / 2;
+
+            // SAFETY: the call is made safe by the following invariants:
```
Robbepop, line 302: `// SAFETY` comments are only needed for `unsafe` code blocks, although I do appreciate the comment in general here.

crates/storage/src/collections/vec/mod.rs
Robbepop: Both are doing the same, the one is just shorter and a bit more idiomatic in this case of immediate `unwrap`.

Robbepop (commented):
All in all LGTM so far. Will take another more detailed look into the actual implementation. We could use a bit more tests for this to cover some edge cases. Maybe pull some tests over from the `core` implementation?

crates/storage/src/collections/vec/mod.rs
KaiserKarel: agree!

crates/storage/src/collections/vec/mod.rs
KaiserKarel: That's a leftover of the actual vec implementation having a get_unchecked method for perf reasons. Perhaps I should remove the `SAFETY:`, but leave the comment?

KaiserKarel:
> All in all LGTM so far. Will take another more detailed look into the actual implementation. We could use a bit more tests for this to cover some edge cases. Maybe pull some tests over from the `core` implementation?

I originally was unable to find them, but now realize they resided in a different crate. I've added the core tests as well now, and removed my own tests.

KaiserKarel:
Also @Robbepop, sinc binary_search and partition are quite closely related, should I add that one too? Or perhaps create another issue and another PR?

Robbepop:
> Also @Robbepop, sinc binary_search and partition are quite closely related, should I add that one too? Or perhaps create another issue and another PR?

Another API should be done in another PR. :)

Robbepop (commented):
Doc comments for the functions need to be adjusted to fit ink! instead of Rust standard library. It currently contains dead links and invalid docs or docs that just does not fit. Please rework docs in this regard and we are good to go. Also we might be in need of a `sort` function for `storage::Vec` in the near future. Although sorting efficiently will be a major endavour and might probably be a bad idea.

KaiserKarel:
> Doc comments for the functions need to be adjusted to fit ink! instead of Rust standard library. It currently contains dead links and invalid docs or docs that just does not fit. Please rework docs in this regard and we are good to go. Also we might be in need of a `sort` function for `storage::Vec` in the near future. Although sorting efficiently will be a major endavour and might probably be a bad idea.

For sorting it would probably be for efficient to read the storage vec into a normal vec, sort that, and write the result back? That should minimize the amount of gas fees I think.

Robbepop:
> For sorting it would probably be for efficient to read the storage vec into a normal vec, sort that, and write the result back? That should minimize the amount of gas fees I think.

Maybe yes, however due to caching in the `LazyIndexMap` layer this might not be needed. Benchmarks are required here instead of assumptions about performance.

KaiserKarel:
Alright, so a PR involving adding sorting would require benchmarks of the two competing implementations?

Robbepop (approved):
LGTM
@cmichi please review/approve too.

We need to check the spellcheck CI so that our full pipeline is run on this PR before we merge.

crates/storage/src/collections/vec/tests.rs
```diff
@@ -422,11 +423,88 @@ fn clear_works_on_filled_vec() {
...
+        b.binary_search_by(|_| Ordering::Equal),
+        Ok(u8::MAX as u32 / 2)
+    );
+    assert_eq!(b.binary_search_by(|_| Ordering::Greater), Err(0));
+    assert_eq!(b.binary_search_by(|_| Ordering::Less), Err(u8::MAX as u32));
+}
+
+#[test]
+// Test implementation specific behavior when finding equivalent elements.
+// It is ok to break this test but when you do a crater run is highly advisable.
```
HCastano, line 494: This comment references `crater`, which isn't relevant in our case. We should probably remove that part of the comment.

crates/storage/src/collections/vec/mod.rs
```diff
@@ -200,6 +200,163 @@ where
         *self.len += 1;
         self.elems.put(last_index, Some(value));
     }
+
+    /// Binary searches this sorted vector for a given element.
```
HCastano, line 204: Can you add a comment explaining where this comes from and a link to the original source code? It doesn't need to be part of the doc comments since it would mostly be of interest to devs working on the internals

HCastano (approved):

crates/storage/src/collections/vec/mod.rs
```diff
@@ -200,6 +200,165 @@ where
...
+    /// let seek = 4;
+    /// assert_eq!(s.binary_search_by(|probe| probe.cmp(&seek)), Err(7));
+    /// let seek = 100;
+    /// assert_eq!(s.binary_search_by(|probe| probe.cmp(&seek)), Err(13));
+    /// let seek = 1;
+    /// let r = s.binary_search_by(|probe| probe.cmp(&seek));
+    /// assert!(match r { Ok(1..=4) => true, _ => false, });
+    /// ```
+    // The binary_search implementation is ported from [core::slice] and attempts to remain as close
+    // to the source as possible.
```
HCastano, line 289: Do you mind adding a URL to the source code? One of those "permanent" GitHub URLs would be best

crates/storage/src/collections/vec/tests.rs
HCastano: We should leave the first part of the comment ("it is ok to break...") here, it's just the `crater` part that's not relevant

HCastano (commented):
Some small nit picks, but after they're resolved we can merge

KaiserKarel:
> We should leave the first part of the comment ("it is ok to break...") here, it's just the crater part that's not relevant

Actually now that I think about it, wasn't binary_search implementation specific behaviour the reason that polkadot had an outage? Perhaps it is not okay for this test to break, as contracts may rely on the specific behaviour?

HCastano:
> Actually now that I think about it, wasn't binary_search implementation specific behaviour the reason that polkadot had an outage? Perhaps it is not okay for this test to break, as contracts may rely on the specific behaviour?

Umm kinda. The problem with Polkadot was that the Wasm runtime and native runtime had different sorting behaviours which caused a failure in consensus. You can read more about it [here](https://polkadot.network/a-polkadot-postmortem-24-05-2021/). 

In our case once a contract is deployed that's the only instance of it. There is no need for a network of people with potentially different versions of the contract to come to consensus on the behaviour - they only need to run what's already on chain.

Whether or not we want to make any guarantees around sorting behaviour is an interesting topic. I'm not sure what the answer is here though. @cmichi @Robbepop what do you guys think?

Robbepop:
> Whether or not we want to make any guarantees around sorting behaviour is an interesting topic. I'm not sure what the answer is here though. @cmichi @Robbepop what do you guys think?

We should do the same as Rust and provide no sorting behavior since this could in theory severely reduce the efficiency of the algorithm. Furthermore, the API clearly says this and the problem for Polkadot existed because Polkadot did not obey the API calling guarantees and took something for granted that simply did not exist. So the problem was not caused by the Rust API or implementation but purely on the user side (Polkadot).

crates/storage/src/collections/vec/mod.rs
```diff
@@ -200,6 +200,166 @@ where
...
+    ///
+    /// If you want to insert an item to a sorted vector, while maintaining
+    /// sort order:
+    ///
+    /// ```
+    /// let mut s = Vec::from([0, 1, 1, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55]);
+    /// let num = 42;
+    /// let idx = s.binary_search(&num).unwrap_or_else(|x| x);
+    /// s.insert(idx, num);
+    /// assert_eq!(s, [0, 1, 1, 1, 1, 2, 3, 5, 8, 13, 21, 34, 42, 55]);
```
Robbepop, line 241: This example should ideally show `ink_storage::Vec` instead of Rust's `Vec`.

crates/storage/src/collections/vec/mod.rs
```diff
@@ -200,6 +200,166 @@ where
...
+    /// found; the fourth could match any position in `[1, 4]`.
+    ///
+    /// ```
+    /// let s = Vec::from([0, 1, 1, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55]);
+    ///
+    /// assert_eq!(s.binary_search(&13),  Ok(9));
+    /// assert_eq!(s.binary_search(&4),   Err(7));
+    /// assert_eq!(s.binary_search(&100), Err(13));
+    /// let r = s.binary_search(&1);
+    /// assert!(match r { Ok(1..=4) => true, _ => false, });
```
Robbepop, line 230: This example should ideally show ink_storage::Vec instead of Rust's Vec.

crates/storage/src/collections/vec/mod.rs
```diff
@@ -200,6 +200,166 @@ where
...
+    ///
+    /// let seek = 13;
+    /// assert_eq!(s.binary_search_by(|probe| probe.cmp(&seek)), Ok(9));
+    /// let seek = 4;
+    /// assert_eq!(s.binary_search_by(|probe| probe.cmp(&seek)), Err(7));
+    /// let seek = 100;
+    /// assert_eq!(s.binary_search_by(|probe| probe.cmp(&seek)), Err(13));
+    /// let seek = 1;
+    /// let r = s.binary_search_by(|probe| probe.cmp(&seek));
+    /// assert!(match r { Ok(1..=4) => true, _ => false, });
```
Robbepop, line 286: This example should ideally show ink_storage::Vec instead of Rust's Vec.

crates/storage/src/collections/vec/mod.rs
```diff
@@ -200,6 +200,166 @@ where
...
+    /// ```
+    /// let s = Vec::from([(0, 0), (2, 1), (4, 1), (5, 1), (3, 1),
+    ///          (1, 2), (2, 3), (4, 5), (5, 8), (3, 13),
+    ///          (1, 21), (2, 34), (4, 55)]);
+    ///
+    /// assert_eq!(s.binary_search_by_key(&13, |&(a, b)| b),  Ok(9));
+    /// assert_eq!(s.binary_search_by_key(&4, |&(a, b)| b),   Err(7));
+    /// assert_eq!(s.binary_search_by_key(&100, |&(a, b)| b), Err(13));
+    /// let r = s.binary_search_by_key(&1, |&(a, b)| b);
+    /// assert!(match r { Ok(1..=4) => true, _ => false, });
```
Robbepop, line 353: This example should ideally show ink_storage::Vec instead of Rust's Vec.

Robbepop (commented):
Please fix the doc examples to work with the respecitve `ink_storage::Vec` type so that we can merge this.

crates/storage/src/collections/vec/mod.rs
KaiserKarel: Oof you're right, that's missing an import. Do I rename the import to StorageVec, to reuse the same nomenclature as what is generated by the templates?

PR #844, opened 2021-07-07
Add `ink-waterfall`  to CI

cmichi:
[`ink-waterfall`](https://github.com/paritytech/ink-waterfall/) is in a state now where I think we can add it to the ink! CI and see how it goes.

The project now runs complete End-to-End tests against all our ink! examples for the `canvas-ui` as well as the `polkadot-js` UI.

At the moment a complete run of the tests takes between 7-10 minutes, depending on CI workload.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,6 +9,7 @@ stages:
   - check
   - workspace
   - examples
+  - ink-waterfall
   - fuzz
   - publish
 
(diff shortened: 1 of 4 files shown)


Robbepop (approved):
LGTM!
Thanks a lot for this nice upgrade to our CI.

.gitlab-ci.yml
```diff
@@ -9,6 +9,7 @@ stages:
   - check
   - workspace
   - examples
+  - ink-waterfall
```
TriplEight, line 12: adding a stage for just one job makes the rest of your pipeline (fuzz and publish) wait until this job's over. Are you sure you want to make it a blocker?

Issue #818, opened 2021-06-16
Message arguments cannot be destructured 


KaiserKarel:
`pub fn foo(&mut self, Bar { baz, bax }: Bar) `

In regular Rust, function arguments may be restructured like so, but the ink macro will error:

```
error: custom attribute panicked
  --> contracts/pod-media/src/lib.rs:11:1
   |
11 | #[ink::contract]
   | ^^^^^^^^^^^^^^^^
   |
   = help: message: internal error: entered unreachable code: encountered unexpected non identifier in ink! parameter
```

Robbepop:
We might be able to implement this. Thanks for letting us know that support for Rust patterns in message parameters is needed. :)

KaiserKarel:
I reckoned that we want to keep `ink!` as close as possible to real rust. :) Although I don't think that rust patters are used often like this, from time to time it is nice to have.

If one can point me to where this can be implemented, I can try to create a PR myself

Robbepop:
Implementation of ink! analysis is found in `crates/lang/ir` and codegen is in `crates/lang/codegen`. A PR probably needs to touch both given that we have the unwanted behavior for non-identifier parameters in ink! messages which probably resides in `crates/lang/ir`. The problem is that you really need to know where to touch things because in some positions we want to keep generating identifiers instead of patterns.

KaiserKarel:
I assume that there are the steps that need to be implemented:

- Add pattern to AST
- Add definition of pattern to ir

HCastano:
I don't think I've ever seen struct destructuring used for function parameters before, TIL!

Robbepop:
> I assume that there are the steps that need to be implemented:
> 
>     * Allow for patterns in IR: ([section](https://github.com/paritytech/ink/blob/e1486563d9cf3a0ab3154686a989cd88aca579d1/crates/lang/codegen/src/generator/metadata.rs#L171))

Indeed metadata generation will require you to provide an actual identifier for the purpose of serialization in this case.
So if we were to implement destructible message parameters we would also be required to add a new ink! specific parmeter attribute `#[ink(name = "foo")]` in order to tell ink! what name this pattern has which probably undermines the whole point of this feature slightly.

KaiserKarel:
Yeah, the whole point was saving 1 LOC ;). I'll probably change the error message then. Perhaps add an explanation and reference to this issue?

Robbepop:
[...]

PR #846, opened 2021-07-07
[lang] catch illegal struct destructuring pattern in ink! message arguments.

KaiserKarel:
Related to #818


Robbepop (approved):
LGTM

Thank you a lot for the fix and the test!

PR #849, opened 2021-07-08
Speed-up CI

cmichi:
This PR contains a number of speed-ups for our CI:

* We currently execute a `for` loop over all examples and _sequentially_ execute some command for each example (`fmt`, `clippy`, `doc`, …). This PR runs these commands in parallel. I've ignored runs of `cargo test`, since concurrency is already utilized fully by that command.  
The trade-off is that with `xargs`, which I used, it's harder to trace an error, because of overlapping stdout output. But GNU's `parallel` supports grouping output, so we could use that instead.

* Simplified: GitLab runs jobs within a stage in parallel, but different stages sequentially. I've moved a couple of jobs which do some simple checks (`fmt`, `spellcheck`, …) to the `check` stage. There were only two `cargo check` jobs in this stage and they take quite a bit of time; this time can be used to execute other jobs in the meantime concurrently. Also you then don't have the thing where you fix some Rust error, push, and only then notice some spellchecking mistakes after the job then ran through.

* I've removed one redundant job, which executes the `delegator/build-all.sh` script. Since we already build the delegator example with its sub-contracts in another job I think we can remove this job (and save a bit of time).

* I've removed some unnecessary `needs` dependencies between jobs, making those jobs run without requiring the `needs` to finish first.

The runs for this PR take ~7 minutes, while usually a run takes ~12 minutes. This is of course no proper benchmark, as it also depends on overall availability of our CI runners, but it gives a rough idea.


cmichi:
Will revisit at some later point.

PR #851, opened 2021-07-09
Use different mechanism to trigger `ink-waterfall` CI

cmichi:
In the current implementation GitLab is unable to report the CI result back. I think that's also why we didn't see the check show up.

After talking to @TriplEight he pointed me to the way it's done in Polkadot, which is what I am using here as well now.


.gitlab-ci.yml
```diff
@@ -46,7 +46,7 @@ workflow:
     - rustc -vV
     - rustup show
     - bash --version
-    - ./scripts/pre_cache.sh
+    - ./scripts/.ci/pre_cache.sh
```
TriplEight: why hidden?

cmichi: IMHO it's better practice to have stuff that's not part of the core-ink! project itself be hidden (like the stuff in `.github/`). The scripts in this `scripts` folder are meant to be executed by developers contributing to ink!, I feel that having unrelated CI stuff visibly in there is confusing.

TriplEight: ok whatever, I'm going to standardize it a bit later

TriplEight (approved):

PR #852, opened 2021-07-10
Implement storage::vec::Slice

KaiserKarel:
Initial possible design for slices in storage::vec::Vec. If this design is safe and agreed upon, I'll add useful traits such as Iterator and co.

Due to the index trait returning a reference, we cannot slice like so:

```
let v = StorageVec::new()
let x = v[1..10];
```
as the implementation creates owned slices. I've not found a way around this yet.


KaiserKarel:
@Robbepop do you have some preliminary feedback?

HCastano (commented):
So far I think it's looking alright. 

I know it's early, but I still want to make sure it's said: make sure errors are correctly instead of `unwrap()`-ing 😄

crates/storage/src/collections/vec/slice.rs
```diff
@@ -0,0 +1,132 @@
...
+    Range,
+};
+
+/// A view into a storage `Vec`.
+#[derive(Clone, Debug)]
+pub struct Slice<'a, T> {
+    /// The start and end indices inside the `index_map`. Indexing the slice using `n` means that we
+    /// access `n + range.start`.
+    range: Range<u32>,
+    index_map: &'a LazyIndexMap<T>,
```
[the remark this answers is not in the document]
KaiserKarel: Is the `LazyIndexMap` cheaply clonable? It does need to point to the same backing storage as the vector. If so, then an owning one is better.

crates/storage/src/collections/vec/slice.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    pub fn get_mut(&self, index: u32) -> Option<&mut T> {
+        let index = self.range.start + index;
+        if !self.range.contains(&index) {
+            return None
+        }
+        // SAFETY:
+        //  - lazily_load requires that there is exclusive access to the T. Vec::split_at_mut
+        //    guarantees this by ensuring that slices do not overlap.
```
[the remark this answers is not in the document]
KaiserKarel: Agreed. It should state that the caller has ensured the invariants in `new_unchecked` I think

KaiserKarel:
> So far I think it's looking alright.
> 
> I know it's early, but I still want to make sure it's said: make sure errors are correctly instead of `unwrap()`-ing

Totally agree :)

crates/storage/src/collections/vec/slice.rs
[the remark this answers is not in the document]
Robbepop: No it is not cheap. If it was cheap it would implement `Copy`.

crates/storage/src/collections/vec/slice.rs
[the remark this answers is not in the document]
Robbepop: `ink_storage::SmallVec` (array like data structure) could profit from it as well.

crates/storage/src/collections/vec/slice.rs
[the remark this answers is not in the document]
KaiserKarel: I thought so. Perhaps it was like Arc which also does not implement Copy.

crates/storage/src/collections/vec/slice.rs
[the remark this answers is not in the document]
KaiserKarel: Would we prefer a separate `Slice` type for that, generic over `LazyIndexMap` and `LazyArray` through some trait, or just duplicate the Slice definition (with a macro to avoid code duplication)?

KaiserKarel:
Still need to write the unit tests and provide examples in the doc comments, but I think the design is good now. Please take a look @Robbepop and @HCastano.

crates/storage/src/collections/vec/mod.rs
```diff
@@ -518,4 +523,27 @@ where
...
         *self.len = 0;
     }
+
+
+    #[inline]
+    pub fn split_at(&self, mid: u32) -> (Slice<T>, Slice<T>) {
+        assert!(mid <= self.len());
+        (
+            Slice::new_unchecked(0..mid, &self.elems),
+            Slice::new_unchecked(mid..self.len(), &self.elems),
```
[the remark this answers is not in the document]
KaiserKarel: Those are actually different types, for the Slices we need Ranges, not RangeTo :)

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+impl<T, const N: usize> ContiguousStorage for LazyArray<T, N>
+where
+    T: PackedLayout,
+{
+    type Item = T;
+
+    unsafe fn get_mut(&self, index: u32) -> Option<&mut T> {
+        self.cached_entries
```
HCastano, line 375: You should justify your use of `unsafe` here as well

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.range.end == self.range.start
+    }
+
+    /// Returns the first element of the slice, or None if it is empty.
+    pub fn first(&self) -> Option<&T::Item> {
+        self.get(0)
+    }
+
+    /// Returns the first element of the slice, or None if it is empty.
```
HCastano, line 62:
```suggestion
    /// Returns the last element of the slice, or None if it is empty.
```

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    /// Returns the last and all the rest of the elements of the slice, or None if it is empty.
+    pub fn split_last(&self) -> Option<(&T::Item, Slice<&T>)> {
+        let first = self.last()?;
+        Some((
+            first,
+            Slice::new(self.range.start..self.range.end - 1, &self.backing_storage),
+        ))
+    }
```
HCastano, line 83: Not sure if these are really needed if we have a more general `split_at` method

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    /// Creates a new `Slice`.
+    pub fn new(range: Range<u32>, backing_storage: T) -> Slice<T> {
+        Slice {
+            range,
+            backing_storage,
+        }
+    }
```
HCastano, line 91: nit: I'd make this the first method in the `impl` block.

Also, `s/Slice/Self`

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    },
+    traits::PackedLayout,
+};
+use core::ops::Range;
+
+/// A view into contiguous storage.
+#[derive(Clone, Debug)]
+pub struct Slice<T> {
+    /// The start and end indices inside the `storage`. Indexing the slice using `n` means that we
+    /// access `n + range.start`.
```
HCastano, line 37:
```suggestion
    /// access `range.start + n`.
```

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    {
+        assert!(mid <= self.len());
+        (
+            Slice::new(0..mid, &self.backing_storage),
+            Slice::new(mid..self.len(), &self.backing_storage),
+        )
+    }
+}
+
+/// A view into a storage `Vec`.
```
HCastano, line 125: We should update this to not be `Vec` specific

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+#[derive(Clone, Debug)]
+#[must_use = "slices are views into underlying storage and do nothing unless consumed"]
+pub struct SliceMut<T> {
+    /// The start and end indices inside the `index_map`. Indexing the slice using `n` means that we
+    /// access `n + range.start`.
```
HCastano, line 130:
```suggestion
    /// access `range.start + n`.
```

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    range: Range<u32>,
+    backing_storage: T,
```
HCastano, line 132: Missing doc comment

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+impl<T> SliceMut<T>
+where
+    T: ContiguousStorage,
+{
+    /// Returns the number of elements in the slice.
+    #[inline]
```
HCastano, line 140: Why'd you choose to `inline` these methods, but not the ones for `Slice` (and vice versa)?

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+        self.get(0)
+    }
+
+    /// Returns the first element of the slice, or None if it is empty.
+    #[inline]
+    pub fn last(&self) -> Option<&T::Item> {
+        self.get(self.len())
+    }
+
+    /// Returns the first element of the slice, or None if it is empty.
```
HCastano, line 163:
```suggestion
    /// Returns a mutable reference to the first element of the slice, or None if it is empty.
```

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    #[inline]
+    pub fn last_mut(&mut self) -> Option<&mut T::Item> {
+        self.get_mut(self.len())
+    }
+
+    /// Returns the first and all the rest of the elements of the slice, or None if it is empty.
```
HCastano, line 169:
```suggestion
    /// Returns a mutable reference to the first and all the rest of the elements of the slice, or None if it is empty.
```

crates/storage/src/collections/slice/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+        ))
+    }
+
+    /// Creates a new `SliceMut`.
+    ///
+    /// # Safety:
+    ///
+    /// The caller must ensure that mutable slices do not overlap.
+    #[inline]
+    pub(crate) unsafe fn new(range: Range<u32>, backing_storage: T) -> SliceMut<T> {
```
HCastano, line 241: Same nitpick about the location. 

Also, why'd this `pub(crate)` instead of just `pub`?

HCastano (commented):
I haven't been able to look at everything just yet, mainly went through the `Slice`/`SlideMut` impls today

HCastano:
Switching reviewers from Robin to Michi since Robin is on vacation

crates/storage/src/collections/slice/mod.rs
KaiserKarel: To be honest, I forgot to inline these. I mainly inline based on `core`'s implementation, as that seems reasonable.

crates/storage/src/collections/slice/mod.rs
KaiserKarel: They're present in the standard library, which I've mirrored as much as possible so far.

crates/storage/src/collections/slice/mod.rs
KaiserKarel: That should indeed be just `pub`, unless we do not want library authors to be able to create slices.

crates/storage/src/collections/slice/mod.rs
KaiserKarel: > Also, s/Slice/Self
The `use-self` clippy-lint catches that. Should that perhaps be enabled if this is the recommended style?

KaiserKarel:
Most of the tests have been ported from `core`. I still need to add tests to ensure that `split_at_xxx` methods are correct.

Aside from that, spell-check seems to trip over some symbols. These doc comments mirror standard, so do I add those to the config?

HCastano:
> Aside from that, spell-check seems to trip over some symbols. These doc comments mirror standard, so do I add those to the config?

Yeah, just add the words which you think are OK to the `.config/cargo_spellcheck.dic` file. Otherwise you can wrap words in backticks. `cargo-spellcheck` doesn't complain about words formatted like that.

crates/storage/src/lazy/lazy_imap.rs
```diff
@@ -323,7 +323,7 @@ where
     /// a `*mut Entry<T>` pointer that allows for exclusive access. This is safe
     /// within internal use only and should never be given outside the lazy entity
     /// for public `&self` methods.
-    unsafe fn lazily_load(&self, index: Index) -> NonNull<StorageEntry<V>> {
+    pub(crate) unsafe fn lazily_load(&self, index: Index) -> NonNull<StorageEntry<V>> {
```
HCastano, line 326: So based off the doc comments it seems like we shouldn't be exposing this. You'll probably need to move the `ContiguousStorage` implementation to this module as a workaround

crates/storage/src/lazy/lazy_array.rs
```diff
@@ -260,7 +260,7 @@ impl<T, const N: usize> EntryArray<T, N> {
     }
 
     /// Returns an exclusive reference to the entry at the given index if any.
-    unsafe fn get_entry_mut(&self, at: Index) -> Option<&mut StorageEntry<T>> {
+    pub(crate) unsafe fn get_entry_mut(&self, at: Index) -> Option<&mut StorageEntry<T>> {
```
HCastano, line 263: Same comment about visibility here

HCastano:
> I need to implement some tests which verify that compilation fails. Is there a policy on including (test) dependencies? I was thinking of using https://github.com/Manishearth/compiletest-rs.

@KaiserKarel nope, no policies here. However, if you try and introduce a `left-pad`, or a fishy looking crate we'll definitely have something to say about it.

KaiserKarel:
> > I need to implement some tests which verify that compilation fails. Is there a policy on including (test) dependencies? I was thinking of using https://github.com/Manishearth/compiletest-rs.
> 
> @KaiserKarel nope, no policies here. However, if you try and introduce a `left-pad`, or a fishy looking crate we'll definitely have something to say about it.

But then how will we know if a 'bool-is-true'  😄

Robbepop (commented):
Sorry for the late reply.
I have trouble with this PR since it looks to me as if the author simply copy'n'pasted a huge amount of code from the Rust standard library and fiddled around with it until it compiled again without making proper ink! specific design considerations whether the Rust standard library design is actually applicable to ink! itself.
One example for this is the `ContiguousStorage` trait which totally makes sense in Rust standard library means but I do not currently fully understand what the needs and reasoning for it is with respect to ink! storage model.

KaiserKarel:
> simply copy'n'pasted a huge amount of code from the Rust standard library and fiddled around with it until it compiled again without making proper ink! specific design considerations whether the Rust standard library design is actually applicable to ink! itself.

I find this comment a bit attacking, as I went through quite some iterations, attempting to deduplicate a lot of code to abstract over mutable iterators etc. Spent a few evenings looking through this, and I believe that `ContiguousStorage` actually fits well, as ink! has at least two storage types which are semantically contiguous, but cannot be proven to be so by the compiler, such as proving that the mutable references obtained from the backing BtreeMap of StorageVec do not overlap.


This PR is of course a proposal implementation (although I have been a bit too busy lately to finish it up). 

I'd love some feedback on why it does not work for Ink's storage models. I feel like the trait is a bit verbose and a lot of code gets added to the codebase, but couldn't find a simpler implementation.

Robbepop:
Sorry, I did not want to sound attacking towards you since you obviously put a lot of effort into the PR.
My critique is more seen from 10k feet above point of view since the whole code in the PR looks a lot just like copy'n'pasted (which it is) and there is a natural aversion against copy'n'pasted code for me since there are often just way too many things you can oversee while doing so, even if done very cautiously.
The `ContiguousMemory` properties are often used in languages such as C++ or Rust to indicate that it is possible to interact with elements of the data structure using plain old pointer arithmetics to some limited extend.
In our storage model pointer arithmetic does not really make sense since we are actually operating on storage cells instead which are not at all contiguous on hardware level but contiguous with the same respect with which languages such as C++ and Rust are using the `Index` operators for. So they are just contiguous with respect to their abstraction.

However, if you view this on a more high-level point of view where you view storage cells as memory cells you could definitely apply contiguous storage notation to them and instead of pointer arithmetic you could use storage key arithmetic. So you actually got a point there!

KaiserKarel:
> Sorry, I did not want to sound attacking towards you since you obviously put a lot of effort into the PR.
No problem :).

With the copying, I attempted to find a balance with mimicking the stdlib as much as possible, as it seems to me that ink attempts to make the storage types mimick the regular stdlib collections with a drop in replacement. I can of course remove some/all of the helper functions if that is desirable.

In this case, ContiguousStorage might be half-appropriately named, if your first indication is pointer-arithmetic. Naming is also the hardest problem.

However there is a class of collections which can safely provide multiple mutable references/slices, in the case of `ink` those are StorageVec and StorageArray. Rust is a bit of a hassle with abstracting over mutability, that is why the trait is needed. Alternatively we can use a macro to generate concrete impl of iterators for vec and array. Perhaps renaming it to `SliceLike` would be more appropriate.

Robbepop:
> In this case, ContiguousStorage might be half-appropriately named, if your first indication is pointer-arithmetic. Naming is of course the hardest problem. Perhaps renaming it to SliceLike would be more appropriate.

I think `ContiguousStorage` is the better name if documentation makes very clear what is meant by it.

> However there is a class of collections which can safely provide multiple mutable references/slices, in the case of ink those are StorageVec and StorageArray. Rust is a bit of a hassle with abstracting over mutability, that is why the trait is needed. Alternatively we can use a macro to generate concrete impl of iterators for vec and array.

Yeah that bridge between `StorageVec` and `StorageArray` is pretty neat.

KaiserKarel:
Aside from the feedback, I feel uncomfortable merging this without https://github.com/paritytech/ink/pull/853. Not intending to be the one to introduce UB in `ink!`. :fearful:

HCastano:
@KaiserKarel We've decided to close this PR. The reasoning is that we're looking to migrate all our data structures away from caching based to something more lightweight (see #945).

While it would be nice to have a `Mapping` based `Slice`, we don't think it's a high priority (which is why nobody's circled back to this). If you disagree with our decision let us know and we can talk through the best way to move forward.

Thanks for spending time working on this, we appreciate it!

Robbepop:
> While it would be nice to have a Mapping based Slice, we don't think it's a high priority (which is why nobody's circled back to this). If you disagree with our decision let us know and we can talk through the best way to move forward.

I cannot follow how a `Mapping` based `Slice` is anything we actually would ever want. Can you elaborate on this please? Currently we are using `LazyIndexMap` which works entirely different from `Mapping`. As soon as we have a single-caching (aka unchached) `LazyIndexMap` replacement where our `ink_storage::collections::Vec` can be based upon we are good to go further with this experiement again. Until then I would like to keep this closed since it adds tons of very performance critical code that requires very depth-in review.

HCastano:
> I cannot follow how a `Mapping` based `Slice` is anything we actually would ever want. Can you elaborate on this please? Currently we are using `LazyIndexMap` which works entirely different from `Mapping`. 

Sure, maybe I didn't think through all the technical nuances here, but point is we _are_ looking to move to a more lightweight data-structure based world. What specific building blocks we use is kinda irrelevant, but either way I think we'd need to have big changes to this PR (or other PRs, like to introduce a non-caching `LazyIndexMap`) to support this and it's not high priority.

PR #855, opened 2021-07-12
Fix `cargo spellcheck` failure in CI

cmichi:
See https://gitlab.parity.io/parity/ink/-/jobs/1010416.

### DIFF
--- a/.config/cargo_spellcheck.dic
+++ b/.config/cargo_spellcheck.dic
@@ -72,7 +72,7 @@ accessor/S
 allocator/S
 backend/S
 blockchain/S
-callable/S
+callable/MS
 defragment/SD
 encoding/S
 endian/P
--- a/.config/cargo_spellcheck.dic
+++ b/.config/cargo_spellcheck.dic
@@ -1,4 +1,4 @@
-80
+90
 
 ABI
 AST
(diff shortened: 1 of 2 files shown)


.config/cargo_spellcheck.dic
```diff
@@ -1,4 +1,4 @@
-80
+90
```
HCastano, line 1: Why's this changing to `90`?

.config/cargo_spellcheck.dic
```diff
@@ -65,11 +65,14 @@ prepend
...
 Django/S
 IP/S
+ink!/S
 NFT/S
+SCALE/S
 accessor/S
 allocator/S
 backend/S
 blockchain/S
+callable/S
```
HCastano, line 75: Wouldn't the `/S` match `callables`/`callable's`? I don't think these are words...

HCastano (approved):

.config/cargo_spellcheck.dic
cmichi: I switched to `/MS`, for matching only `callable's`.

Yeah I agree that it's not an ideal word, we use it quite often in the codebase though. Do you have a better suggestion?

.config/cargo_spellcheck.dic
HCastano: Oh lol, I had no clue that's what the number was for

PR #842, opened 2021-07-05
Remove `Salt` in code generation during build_create

xgreenx:
[build_create](https://github.com/paritytech/ink/blob/master/crates/env/src/call/create_builder.rs#L198) method supports only 2 generic parameters. But code [here](https://github.com/paritytech/ink/blob/master/crates/lang/codegen/src/generator/cross_calling.rs#L688) is passing 3 parametrs.
It causes a compilation error.

PR #856, opened 2021-07-13
Prepare rc4 release notes

cmichi:
[Preview here](https://github.com/paritytech/ink/blob/cmichi-prepare-rc4-release-notes-2/RELEASES.md)


Robbepop (approved):
LGTM

Good list of changes.

xgreenx:
Good release. It will be cool to integrate [this](https://github.com/paritytech/ink/pull/842) change too. Maybe we can integrate it without unit test, because CI fails but all is working locally.

PR #857, opened 2021-07-13
Temporarily remove `codecov` CI check

cmichi:
* There is some strange linker error in the `master` CI.
* We can't reproduce it locally.
* We suspect that it might have to do with an outdated docker image.
* The failing `master` CI blocks publishing a new `ink-ci-linux:production` image.
* The `codecov` check is broken since a few days anyway (due to some token issue). I created this ticket at the time: https://github.com/paritytech/ci_cd/issues/160.
* We can bring the check back as soon as we sorted the linker error out in a PR.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -186,51 +186,6 @@ spellcheck:
         cargo spellcheck check -vvv --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 examples/delegator/${contract}/;
       done
 
-codecov:
-  stage:                           workspace
-  <<:                              *docker-env
-  <<:                              *test-refs
-  needs:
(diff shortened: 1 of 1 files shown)


HCastano (approved):
> We suspect that it might have to do with an outdated docker image.

I was testing out the Docker images locally and I haven't found this to be the case. Running:

```bash
# On host machine
sudo docker run -it paritytech/ink-ci-linux:production

# In Docker container
git clone https://github.com/paritytech/ink/
cd ink && cargo test --verbose --all-features --no-fail-fast --workspace
```

Works just fine.

I'm thinking it's either something to do with the `RUSTFLAGS`, or maybe come caching issues on the runners.

Either way, Codecov is clearly broken right now, so yeah let's get this merged.

PR #866, opened 2021-07-20
Fix `clippy::self-named-constructor`

cmichi:
The `master` CI currently fails because of this:
https://gitlab.parity.io/parity/ink/-/jobs/1021277


TriplEight (approved):
LGTM

PR #871, opened 2021-07-20
Add Fuzz Testing to Bump Allocator

### LINKED ISSUE #860: Add Fuzz Testing to Bump Allocator
This would be a follow up to https://github.com/paritytech/ink/pull/831 in which a new dynamic memory allocator was
added.

To ensure the correct working behaviour of the allocator it would be benefitial to
have some fuzz testing. 

It looks like the ink! data structure fuzzing uses [quickcheck](https://github.com/BurntSushi/quickcheck), so that could be a
good starting point for this as well.

HCastano:
This closes #860, which is a follow up to #831.

~~Still a draft since I'm trying to figure out my way around `quickcheck` and the right
allocator properties to test.~~

### DIFF
--- a/crates/allocator/src/bump.rs
+++ b/crates/allocator/src/bump.rs
@@ -122,7 +122,6 @@ impl InnerAlloc {
         let alloc_end = alloc_end?;
 
         if alloc_end > self.upper_limit {
-            dbg!(aligned_size);
             let required_pages = dbg!(required_pages(aligned_size));
             let required_pages = required_pages?;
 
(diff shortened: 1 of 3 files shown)


crates/allocator/src/bump.rs
```diff
@@ -264,6 +264,51 @@ mod tests {
         let expected_alloc_start = 2 * PAGE_SIZE + std::mem::size_of::<u8>();
         assert_eq!(inner.next, expected_alloc_start);
     }
+
+    // TODO: What I want to end up doing is turning this into a `quickcheck` test such that the
+    // random sized bytes and the number of allocations comes from `quickcheck`
```
cmichi, line 269: You could have a test a la
```
#[quickcheck]
fn can_alloc_many_different_sized_chunks(chunks: Vec<Vec<u8>>) {
  // …
}
```
…and the iterate over the `Vec` whilst allocating the inner chunk.

crates/allocator/src/bump.rs
```diff
@@ -137,9 +137,31 @@ impl InnerAlloc {
     }
 }
 
+#[inline]
+fn required_pages(size: usize) -> Option<usize> {
```
cmichi, line 141: Could you add a doc comment for this function please :-)?

cmichi (changes requested):
Super-nice that you are adding those tests!

cmichi:
Ah btw.: I really like that you have proper commits with proper messages describing the intention behind the change!

crates/allocator/src/bump.rs
```diff
@@ -235,7 +262,224 @@ mod tests {
...
+        // case in another test)
+        if n.checked_add(PAGE_SIZE - 1).is_none() {
+            return TestResult::discard()
+        }
+
+        let mut inner = InnerAlloc::new();
+
+        let layout = Layout::from_size_align(n, size_of::<usize>()).unwrap();
+        let size = layout.pad_to_align().size();
+        assert_eq!(inner.alloc(layout), Some(0));
```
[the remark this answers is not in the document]
HCastano: > Generally I'm a fan of the Given-When-Then test structure, though I see that this is not always ideal.

In the case of these tests they're _kinda sorta_ structure in this way, but I prefer to have relevant events close to the `assert` statements. Is there any test in particular which you think could benefit from being re-jigged into this structure?

crates/allocator/src/bump.rs
```diff
@@ -137,9 +137,31 @@ impl InnerAlloc {
...
+
+    #[test]
+    fn can_alloc_no_bytes() {
+        let mut inner = InnerAlloc::new();
+
+        let layout = Layout::new::<()>();
+        assert_eq!(inner.alloc(layout), Some(0));
+
+        let expected_limit =
+            PAGE_SIZE * required_pages(layout.pad_to_align().size()).unwrap();
```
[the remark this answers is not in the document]
HCastano: I'm gonna push back a bit on this, since most of the `unwrap()` come after calls to `required_pages()`. In the tests we have preconditions on the size of allocations, ensuring that no individual allocation is going to overflow (unless the test checks for overflows, in which case we don't request pages). This means getting the required number of pages is safe. Adding `expect()`'s and messages would add unnecessary noise to the tests imo.

That said, if there are specific places where you think think we'd benefit from `expect()`, feel free to point them out.

HCastano:
> Ah btw.: I really like that you have proper commits with proper messages describing the intention behind the change!

They're mostly for me, but glad you're enjoying reading them!

crates/allocator/src/bump.rs
```diff
@@ -235,7 +267,251 @@ mod tests {
...
+    fn should_allocate_arbitrary_sized_bytes(n: usize) -> TestResult {
+        // If `n` is going to overflow we don't want to check it here (we'll check the overflow
+        // case in another test)
+        if n.checked_add(PAGE_SIZE - 1).is_none() {
+            return TestResult::discard()
+        }
+
+        let mut inner = InnerAlloc::new();
+
+        let layout = Layout::from_size_align(n, size_of::<usize>()).unwrap();
```
cmichi, line 294: I concur on the `required_pages` unwraps, but would really like to have an `expect` message for the `from_size_align` calls.

cmichi (approved):
Thanks for adding those tests Just a pity that we didn't find any bugs with them (yet) ;-).

PR #872, opened 2021-07-21
Add helper script for updating ink! versions

HCastano:
I've had a little script like this in other projects just to make it dead easy when
updating versions (like in #864). Figured it would be worth proposing.

### DIFF
--- a/scripts/bump-release-version.sh
+++ b/scripts/bump-release-version.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+# One-liner to update between ink! release versions.
+#
+# Usage: ./bump-release-version.sh 3.0.0-rc3 3.0.0-rc4
+
+set -xeu
+
+OLD_VERSION=$1
+NEW_VERSION=$2
+
+find . -type f -name 'Cargo.toml' -exec sed -i -e "s/$OLD_VERSION/$NEW_VERSION/g" {} \;


cmichi:
Thanks for the PR! It's a good idea, but I would rather leave it out for the following reason:
the script will become obsolete as soon as we've reached a non-release-candidate state ‒ then we can upgrade individual crates on their own, i.e. they will have individual versions, making a global search/replace obsolete.

Fine with you?

HCastano:
> the script will become obsolete as soon as we've reached a non-release-candidate state ‒ then we can upgrade individual crates on their own, i.e. they will have individual versions, making a global search/replace obsolete.

Well, it depends on how long that non-`rc` state is going to take to achieve. What do you have in mind?

cmichi:
I'll close the PR, since we should do everything for getting from rc to stable.

HCastano:
@cmichi I think this is relevant again 😉

scripts/bump-release-version.sh
```diff
@@ -0,0 +1,12 @@
...
+# One-liner to update between ink! release versions.
+#
+# Usage: ./bump-release-version.sh 3.0.0-rc3 3.0.0-rc4
+
+set -xeu
+
+OLD_VERSION=$1
+NEW_VERSION=$2
+
+find . -type f -name 'Cargo.toml' -exec sed -i -e "s/$OLD_VERSION/$NEW_VERSION/g" {} \;
```
cmichi, line 12: What about local `Cargo.lock`'s ?

HCastano:
Gonna close this again 😅 

If we go with something like `cargo-release` or a hand rolled helper (like in #1211) we shouldn't need this

PR #874, opened 2021-07-22
Update `cargo-spellcheck` configuration

cmichi:
Follow-up to https://github.com/drahnr/cargo-spellcheck/issues/194.

### DIFF
--- a/.config/cargo_spellcheck.dic
+++ b/.config/cargo_spellcheck.dic
@@ -7,6 +7,7 @@ DApp
 ERC
 FFI
 Gnosis
+GPL
 KECCAK
 Polkadot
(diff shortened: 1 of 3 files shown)


.config/cargo_spellcheck.dic
```diff
@@ -57,8 +58,10 @@ stdout
 tuple
 unordered
 untyped
+v3
```
cmichi, line 61: [This is how it's done in Polkadot as well.](https://github.com/paritytech/polkadot/blob/master/scripts/gitlab/lingua.dic#L241-L243)

HCastano (approved):

PR #878, opened 2021-07-26
Flip the `DEBUG_ENABLED` conditional

athei:
Fixes https://github.com/paritytech/substrate/issues/9334

When implementing the new debug print code we introduced a feature that stops trying to print debug messages when it failed once (when executing as transaction). However, this led to the situation that the code executed on-chain is a tiny bit larger than the off-chain executed code (used for estimation) when only one message was ever tried to be printed. Check my response on the linked issue for more information.

This PR fixes the situation by flipping the conditional so the extra assignment is executed in the "logging enabled" (RPC) case. This requires an additional variable but I think it is still simple enough.

### DIFF
--- a/crates/env/src/engine/on_chain/ext.rs
+++ b/crates/env/src/engine/on_chain/ext.rs
@@ -649,25 +649,6 @@ pub fn random(subject: &[u8], output: &mut &mut [u8]) {
     extract_from_slice(output, output_len as usize);
 }
 
-/// Return the same value as passed in but hide the output value from the optimizer.
-///
-/// # Note
-///
-/// This contains Wasm inline assembly. Since the this module is never compiled
-/// to something else that should be fine.
-#[allow(unused_assignments)]
-#[cfg(feature = "ink-debug")]
-fn is_true(is_true: bool) -> bool {
-    // Inline assembly cannot operate on booleans. Only primitives.
-    let mut dummy: u32 = if is_true { 1 } else { 0 };
-    // This marks the `dummy` variable as written to. LLVM cannot know the value of it.
-    // Safety: We only emit a `nop` instruction which is safe.
-    unsafe {
-        asm!("nop /* {0} */", out(local) dummy);
-    }
-    dummy == 1
-}
-
 #[cfg(feature = "ink-debug")]
 /// Call `seal_debug_message` with the supplied UTF-8 encoded message.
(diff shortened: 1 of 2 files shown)


HCastano:
I don't quite understand how not optimizing away the value of `enabled` fixes this issue, would you be able to clarify that for me? 

As for the unstable feature activation, I'm not a fan. However, I have no suggestions for avoiding it in this case 😅

athei:
> I don't quite understand how not optimizing away the value of enabled fixes this issue, would you be able to clarify that for me?

We are looking basically at this code:
```rust
if DEBUG_ENABLED {
	if let Err(Error::LoggingDisabled) = ret_code.into() {
		DEBUG_ENABLED = false;
	}
}
```

During RPC execution logging is disabled. This means the assignment is not executed. During the real (as a transaction) execution logging is disabled and the assignment is executed. So the code executed during gas estimation (RPC execution) is less and therefore the required gas reported is less.

Even unconditionally assigning a value to `DEBUG_ENABLED` will not work because the as still optimizes the `DEBUG_ENABLED = true` because it knows it is already set to true. Therefore I needed to resort to this trick to stop the analysis of the optimizer and just always do the assignment.

There might be better ways to fix this problem. I just wanted to offer something that fixes the reported bug. I don't recommend merging this but use this as a starting point for a better solution (not needing new unstable features).

HCastano:
Thanks for the clarification! 

As for a solution. My naive approach would be to flip the assignments so that the branch is always run for RPCs and not for real transactions. That way we overestimate gas usage, which shouldn't be a problem. However, I've got no clue what might be involved in making this change (otherwise I suspect you would've already mentioned it).

Out of curiosity, what's the delta between the estimated and actual gas costs?

athei:
> My naive approach would be to flip the assignments so that the branch is always run for RPCs and not for real transactions. That way we overestimate gas usage, which shouldn't be a problem.

That worked! The conditional is now a bit more complicated, though.

> Out of curiosity, what's the delta between the estimated and actual gas costs?

3 megagas

ascjones (approved):
Much better, great idea.

cmichi (approved):
Great that you found the issue, thanks!

HCastano (approved):
Nice, I thought you might have to flip the `Error` type to do this

PR #882, opened 2021-07-29
Add example for mocking chain extensions in off-chain tests

cmichi:


examples/rand-extension/lib.rs
```diff
@@ -130,5 +132,39 @@ mod rand_extension {
...
+            }
+            ink_env::test::register_chain_extension(MockedExtension);
+            let mut rand_extension = RandExtension::default();
+            assert_eq!(rand_extension.get(), [0; 32]);
+
+            // when
+            rand_extension.update().expect("update must work");
+
+            // then
+            assert_ne!(rand_extension.get(), [0; 32]);
```
HCastano, line 166: I don't think this is necessary, it's kinda implied with the `assert_eq!(..., [1; 32])`

examples/rand-extension/lib.rs
```diff
@@ -130,5 +132,39 @@ mod rand_extension {
...
         }
+
+        #[ink::test]
+        fn chain_extension_works() {
+            // given
+            struct MockedExtension;
+            impl ink_env::test::ChainExtension for MockedExtension {
+                /// The static function id of the chain extension.
+                fn func_id(&self) -> u32 {
+                    1101
```
HCastano, line 143: I'd make this a `const` since it's used a couple of times in the example already

HCastano (approved):

examples/rand-extension/lib.rs
```diff
@@ -130,5 +132,39 @@ mod rand_extension {
             let rand_extension = RandExtension::default();
             assert_eq!(rand_extension.get(), [0; 32]);
         }
+
+        #[ink::test]
+        fn chain_extension_works() {
+            // given
+            struct MockedExtension;
+            impl ink_env::test::ChainExtension for MockedExtension {
```
frankli-dev, line 140: Do I need to mock chain extension inside every test function? Is there anyway to define outside and use that in all functions?

frankli-dev:
I'm getting the following error
```
862 |             impl ink_env::test::ChainExtension for MockedExtension {
    |                                 ^^^^^^^^^^^^^^ private trait
    |
note: the trait `ChainExtension` is defined here
   --> /Users/mac/.cargo/registry/src/github.com-1ecc6299db9ec823/ink_env-3.0.0-rc3/src/engine/off_chain/test_api.rs:18:5
    |
18  |     chain_extension::ChainExtension,
```

I think, this PR covers the above issue but does it mean that once the PR is merged, I need to use the latest ink?

examples/rand-extension/lib.rs
cmichi: It's only used in one other place, where I can't replaced it with a `const`. Here:
```
#[ink(extension = FUNC_ID, returns_result = false)]
```

examples/rand-extension/lib.rs
cmichi: You can also define it outside, I just mocked it in here since it's only used in this one test.

cmichi:
> I think, this PR covers the above issue but does it mean that once the PR is merged, I need to use the latest ink?

@frankli-dev Yes, until we release the next release candidate you need to reference the git repository for your `Cargo.toml` dependencies. The error you get is because this PR was not merged yet, after it's merged it will work.

PR #884, opened 2021-08-02
Run clippy with `--all-targets`

cmichi:
We were not running clippy on all targets, this mostly had the effect of test code being excluded.

Edit: I converted the PR to draft to figure out an issue with `trybuild`.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -283,10 +283,10 @@ examples-clippy-wasm:
   <<:                              *test-refs
   script:
     - for example in examples/*/; do
-        cargo clippy --verbose --all-targets --manifest-path ${example}/Cargo.toml --no-default-features --target wasm32-unknown-unknown -- -D warnings ${CLIPPY_FLAGS};
+        cargo clippy --verbose --manifest-path ${example}/Cargo.toml --no-default-features --target wasm32-unknown-unknown -- -D warnings ${CLIPPY_FLAGS};
       done
     - for contract in ${DELEGATOR_SUBCONTRACTS}; do
-        cargo clippy --verbose --all-targets --manifest-path examples/delegator/${contract}/Cargo.toml --no-default-features --target wasm32-unknown-unknown -- -D warnings ${CLIPPY_FLAGS};
+        cargo clippy --verbose --manifest-path examples/delegator/${contract}/Cargo.toml --no-default-features --target wasm32-unknown-unknown -- -D warnings ${CLIPPY_FLAGS};
       done
 
 examples-contract-build:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -24,7 +24,8 @@ variables:
   # this var is changed to "-:staging" when the CI image gets rebuilt
   # read more https://github.com/paritytech/scripts/pull/244
   ALL_CRATES:                      "${PURELY_STD_CRATES} ${ALSO_WASM_CRATES}"
-  DELEGATOR_SUBCONTRACTS:           "accumulator adder subber"
+  DELEGATOR_SUBCONTRACTS:          "accumulator adder subber"
+  CLIPPY_FLAGS:                    "-A clippy::bool_assert_comparison"
 
 workflow:
   rules:
@@ -191,7 +192,7 @@ clippy-std:
       artifacts:                   false
   script:
     - for crate in ${ALL_CRATES}; do
-        cargo clippy --verbose --all-targets --all-features --manifest-path crates/${crate}/Cargo.toml -- -D warnings;
+        cargo clippy ${CLIPPY_FLAGS} --verbose --all-targets --all-features --manifest-path crates/${crate}/Cargo.toml -- -D warnings;
       done
 
 clippy-wasm:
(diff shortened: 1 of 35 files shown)


examples/erc1155/lib.rs
```diff
@@ -792,7 +792,7 @@ mod erc1155 {
             // Note: All of these tests are from the context of the owner who is either allowing or
             // disallowing an operator to control their funds.
             set_sender(owner);
-            assert!(erc.is_approved_for_all(owner, operator) == false);
+            assert!(!erc.is_approved_for_all(owner, operator));
```
HCastano, line 795: I personally don't like this since it's hard to see the two `!` side by side

HCastano (commented):
Looks fine, ping me again once the build is working

examples/erc1155/lib.rs
cmichi: Totally agree, the issue is that it's currently not possible to define global clippy lints (see https://github.com/rust-lang/cargo/issues/5034) and I think fragmentation of linting rules in each example would be a worse trade-off.

cmichi: I added a compromise to this PR: passing the flag `-A clippy::bool_assert_comparison` to all `clippy` invocations. This allows e.g. `assert_eq(…, false);`. 

It's kinda ugly and we should create a follow-up issue to replace this as soon as the clippy issue is fixed.

crates/primitives/benches/bench.rs
```diff
@@ -43,7 +43,7 @@ fn bench_key_add_assign_u64(c: &mut Criterion) {
     c.bench_function("Key2::add_assign(u64)", |b| {
         b.iter(|| {
             let mut copy = black_box(key);
-            let _ = black_box(copy += 1u64);
+            let _ = black_box(|| copy += 1u64);
```
HCastano, line 46: What lint is this from?

crates/storage/src/collections/stash/tests.rs
```diff
@@ -71,7 +71,7 @@ fn from_iterator_works() {
         stash
     });
     assert_eq!(stash.len(), test_values.len() as u32);
-    assert_eq!(stash.is_empty(), false);
+    assert!(!stash.is_empty());
```
HCastano, line 74: I'd revert this to the original state

crates/storage/src/collections/bitvec/bits256.rs
```diff
@@ -310,7 +310,7 @@ mod tests {
         for i in 0..256 {
             let i = i as u8;
             bits256.reset(i);
-            assert_eq!(bits256.get(i), false);
+            assert!(!bits256.get(i));
```
HCastano, line 313: Same thing about using the original here

.gitlab-ci.yml
```diff
@@ -24,7 +24,9 @@ variables:
   # this var is changed to "-:staging" when the CI image gets rebuilt
   # read more https://github.com/paritytech/scripts/pull/244
   ALL_CRATES:                      "${PURELY_STD_CRATES} ${ALSO_WASM_CRATES}"
-  DELEGATOR_SUBCONTRACTS:           "accumulator adder subber"
+  DELEGATOR_SUBCONTRACTS:          "accumulator adder subber"
+  # until https://github.com/rust-lang/cargo/issues/5034 has been implemented
```
HCastano, line 28: I would make an issue in the ink! repo which mentions this + some more context. Then I'd link _that_ issue here

scripts/check-workspace.sh
```diff
@@ -21,6 +21,9 @@ purely_std_crates=("lang/codegen" "metadata")
 also_wasm_crates=("env" "storage" "storage/derive" "allocator" "prelude" "primitives" "lang" "lang/macro" "lang/ir")
 all_crates=("${also_wasm_crates[@]}" "${purely_std_crates[@]}")
 
+# until https://github.com/rust-lang/cargo/issues/5034 has been implemented
```
HCastano, line 24: Same comment here

crates/lang/macro/Cargo.toml
```diff
@@ -44,4 +44,8 @@ std = [
     "ink_lang_ir/std",
     "ink_primitives/std",
 ]
+
+# Due to https://github.com/rust-lang/cargo/issues/6915 features that affect a dev-dependency
+# currently can't be enabled by a parent crate.
+# ink-experimental-engine = ["ink_env/ink-experimental-engine"]
```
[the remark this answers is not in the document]
cmichi: I really want to avoid mixing dev-dependencies with build-dependencies, I've opted for `#[ignore]` now.

scripts/check-workspace.sh
cmichi: I've added a link to our tracking issue and noted it down as a ToDo there.

crates/primitives/benches/bench.rs
cmichi: https://github.com/bheisler/criterion.rs/issues/507. I'm also not happy with this hack, since there is some non-intended overhead with this closure.

examples/erc1155/lib.rs
cmichi: I've removed the above hack now again, after thinking a bit more about it I can't bring myself to hack it this way. Feel free to create a follow-up to switch it back and add global lints, once it's possible.

PR #885, opened 2021-08-03
Track Example Contract Wasm Sizes

HCastano:
Draft since I'm still playing around with stuff.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -292,8 +292,7 @@ examples-contract-build:
   <<:                              *docker-env
   <<:                              *test-refs
   artifacts:
-      paths:
-          - pr-xxx-example-wasm-sizes.csv
+      name: "$CI_MERGE_REQUEST_ID-$CI_COMMIT_SHORT_SHA.csv"
   script:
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -86,356 +86,365 @@ workflow:
...
+    - SIZE_OUT=""
     - for example in examples/*/; do
         pushd $example &&
-        cargo contract build &&
-        popd;
-      done
-    - for contract in ${DELEGATOR_SUBCONTRACTS}; do
-        pushd examples/delegator/$contract &&
-        cargo contract build &&
+        SIZE_OUT=$(cargo contract build | grep "Original" | cut -d " " -f 4,6) &&
```
HCastano, line 305: I think it would be nice to have an `--output-json` option in `cargo-contract` so that a less ugly `jq` query could be used to get the fields we're interested in

HCastano:
As Michi mentioned in https://github.com/paritytech/ink-waterfall/issues/4, having this be so tightly integrated into the ink! CI isn't the best approach

PR #888, opened 2021-08-05
Unlicense examples

cmichi:
Our examples should be licensed as freely as possible.


HCastano (changes requested):
The headers in the example files are still licensed under Apache-2

HCastano (approved):
🇺🇸

PR #890, opened 2021-08-09
Change CI variables source to Vault

sergejparity:
Change behavior of the CI pipeline to use Vault secrets instead of project variables.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -84,6 +84,32 @@ workflow:
   tags:
     - kubernetes-parity-build
 
+#### Vault secrets
+.vault-secrets:                    &vault-secrets
+  variables:
+    VAULT_SERVER_URL:                "https://vault.parity-mgmt-vault.parity.io"
+    VAULT_AUTH_PATH:                 "gitlab-parity-io-jwt"
+    VAULT_AUTH_ROLE:                 "cicd_gitlab_parity_${CI_PROJECT_NAME}"
+  secrets:
+    CODECOV_P_TOKEN:
+      vault:                       cicd/gitlab/$CI_PROJECT_PATH/CODECOV_P_TOKEN@kv
+      file:                        false
+    CODECOV_TOKEN:
+      vault:                       cicd/gitlab/$CI_PROJECT_PATH/CODECOV_TOKEN@kv
+      file:                        false
+    GITHUB_EMAIL:
+      vault:                       cicd/gitlab/$CI_PROJECT_PATH/GITHUB_EMAIL@kv
+      file:                        false
+    GITHUB_USER:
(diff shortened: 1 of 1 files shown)


HCastano:
@sergejparity can you explain to a non-Devops person what this change is about? What are Vault secrets, and why are they better than the current approach?

TriplEight:
@HCastano [this](https://github.com/paritytech/ci_cd/wiki/Handling-CI-CD-secrets-with-Hashicorp-Vault) wiki page should answer your questions. Don't hesitate to ask more and we will update it.

HCastano:
@sergejparity can you fix the secret provider issue with the `ink-waterfall` pipeline?

TriplEight:
interesting! This probably means that we'll have to at least add the prevoder definition
```
  variables:
    VAULT_SERVER_URL:              "https://vault.parity-mgmt-vault.parity.io"
    VAULT_AUTH_PATH:               "gitlab-parity-io-jwt"
    VAULT_AUTH_ROLE:               "cicd_gitlab_parity_${CI_PROJECT_NAME}"
```
to https://github.com/paritytech/ink-waterfall/blob/master/.gitlab-ci.yml

TriplEight (changes requested):
- please first fix the issue with `ink-waterfall`
- I think it's better to mention the needed secrets in every job rather than import them everywhere

sergejparity:
> interesting! This probably means that we'll have to at least add the prevoder definition
> 
> ```
>   variables:
>     VAULT_SERVER_URL:              "https://vault.parity-mgmt-vault.parity.io"
>     VAULT_AUTH_PATH:               "gitlab-parity-io-jwt"
>     VAULT_AUTH_ROLE:               "cicd_gitlab_parity_${CI_PROJECT_NAME}"
> ```
> 
> to https://github.com/paritytech/ink-waterfall/blob/master/.gitlab-ci.yml.
> Even though that project doesn't seem to use the secrets.

Yes, very strange. Looks like we have here classic example of violation of software development principle - "loose coupling, high cohesion". Will investigate it.

TriplEight:
Did you try adding these variables directly to the waterfall CI config?

Another thing with it is that GItLab requires elevated permissions from those who are triggering the downstream pipelines' protected branches, please make sure you have the right to do so, or change the mirroring in waterfall project's UI to allow maintainers to merge into the master or other protected branch.

sergejparity:
Found an issue with pipeline. `variables:` block can be defined just ones per job as multiple definition lead to collision, the latest defined will cancel previous ones. And this was resolved by moving Vault access variables out of the anchor to main pipeline `variables:` block. Leaving only `secrets:` defined inside anchor `.vault-secrets`. Now it correctly starting `ink-waterfall` job.
Also cleaned up history.

TriplEight (approved):

PR #892, opened 2021-08-11
Remove `DispatchRetCode`

athei:
The ink! dispatcher returns an `u32` from its entry points. This `u32` encodes the error that occurred during dispatching (if any).

However, `pallet_contracts` does not interpret any value returned from the entry points. The signature returning a `i32` is only allowed for compatibility reasons. This leads to a bug where supplying invalid inputs to a contract is not detected. This PR fixes this by removing the obsolete `DispatchRetCode` and instead panic on any dispatch error.

Please note that this panicking behavior could be replaced by writing a `Result` to the output buffer while setting the `REVERT` flag in the future. Depending on whether this error should be hand-able by the caller or not.


crates/lang/codegen/src/generator/dispatch.rs
```diff
@@ -95,28 +95,24 @@ impl Dispatch<'_> {
...
-                ::ink_lang::DispatchRetCode::from(
-                    <#storage_ident as ::ink_lang::DispatchUsingMode>::dispatch_using_mode(
-                        ::ink_lang::DispatchMode::Instantiate,
-                    ),
+            fn deploy() {
+                <#storage_ident as ::ink_lang::DispatchUsingMode>::dispatch_using_mode(
+                    ::ink_lang::DispatchMode::Instantiate,
                 )
-                .to_u32()
+                .unwrap()
```
cmichi, line 102: Did you use an `unwrap` here intentionally over an `expect`?

athei: I felt that the error is sufficiently described the `Debug` implementation of `DispatchError`. I am open to write some message here if that is ink! convention, though.

cmichi: Would be great if you could replace the `unwrap`'s with `expects` then. We try to not use `unwrap`'s outside of tests.

athei: Done

cmichi (approved):
Great that you noticed this, and awesome that you put in a PR fixing it!

PR #894, opened 2021-08-12
Print panic message to debug buffer

athei:
We should print panic messages to the debug buffer when logging is enabled. This will print useful messages to the console and UIs (once they implement it). It looks like this:
```
panicked at 'encountered error while querying transferred balance: Decode(Error)', /Users/alexander/Developer/parity/ink/crates/lang/src/dispatcher.rs:145:10
```

Before we consider merging this we must stabilize `seal_debug_message`. 

Depends on: https://github.com/paritytech/cargo-contract/pull/326

### DIFF
--- a/crates/env/src/lib.rs
+++ b/crates/env/src/lib.rs
@@ -44,8 +44,12 @@
 )]
 
 #[cfg(all(not(feature = "std"), target_arch = "wasm32"))]
+#[allow(unused_variables)]
 #[panic_handler]
-fn panic(_info: &core::panic::PanicInfo) -> ! {
+fn panic(info: &core::panic::PanicInfo) -> ! {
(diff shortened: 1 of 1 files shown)


cmichi:
@athei Is there anything left before we can stabilize `seal_debug_message` in the `contracts` pallet? From my POV it seems to be working well, I'm not aware of any issues with it.

athei:
I think it can be stabilized as is. I need to double check whether the weight stuff is sound, though.

cmichi (approved):
We're waiting for the stabilization of `seal_debug_message` before merging. Code looks good.

athei:
https://github.com/paritytech/substrate/pull/9550

HCastano (approved):

PR #863, opened 2021-07-19
Fix Codecov measuring and Publishing

TriplEight:
Apparently, after their security issue, Codecov decided to fix up some security. Now they require tokens even for public projects.
As for failing after the script, it used to before, but it's on their side, so now it does :-1: . This is fixed with `-Z` option.

UPD: in fact, they've released a safer option to upload the test cov results, a binary. https://github.com/paritytech/scripts/pull/318
UPD2: which does not actually work so far.

CC: https://github.com/paritytech/ink/pull/857 
Closes: https://github.com/paritytech/ci_cd/issues/160

PR #895, opened 2021-08-19
CI: fix coverage

TriplEight:
- restores code coverage how it used to measure previously - with an old-style instrumentation (but with branch coverage)
- some chore

CC: https://github.com/paritytech/ink/pull/857
Closes: paritytech/ci_cd#160 https://github.com/paritytech/ink/pull/863 https://github.com/paritytech/ci_cd/issues/13

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -182,6 +182,52 @@ spellcheck:
     - cargo spellcheck check -vvvv --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 -- recursive .
     - cargo spellcheck check -vvvv --cfg=.config/cargo_spellcheck.toml --checkers hunspell --code 1 -- recursive examples/
 
+codecov:
+  stage:                           workspace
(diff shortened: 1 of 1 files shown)


TriplEight:
Note: https://github.com/paritytech/ink/pull/863#pullrequestreview-710866706

TriplEight:
Feel free to merge if you're OK to do without `-Clink-dead-code `.

HCastano:
> Feel free to merge if you're OK to do without `-Clink-dead-code `.

I'm hesitant to merge without this since Robin mentioned that having this flag [was correct](https://github.com/paritytech/ink/pull/863#pullrequestreview-710866706). Does the Codecov job not pass if this flag is enabled, or what happens?

TriplEight:
Shish, I thought it already works for a month. It could have worked at least somehow. The job breaks with this flag.

cmichi:
I'll merge in the current state and create create a follow-up adding the `-Clink-dead-code` flag and a fix for the error it causes.

PR #909, opened 2021-09-08
Release ink! 3.0.0-rc5

cmichi:
I would have liked to wait with the release until we have the state rent stuff removed completely as well, but  a couple users have now run into https://github.com/paritytech/ink/pull/902 and it makes sense to do a release now. Due to semver they will automatically receive the fix.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,3 +1,17 @@
+# Version 3.0-rc5 (2021-09-08)
+
+This is the 5th release candidate for ink! 3.0.
+
+The list below shows the additions, changes and fixes that are visible to users of ink!.
+
+## Added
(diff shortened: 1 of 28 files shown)


ascjones (approved):

forgetso:
For anyone running into the following issue with this release:

```
error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
  --> /home/chris/.cargo/registry/src/github.com-1ecc6299db9ec823/ink_env-3.0.0-rc5/src/lib.rs:56:5
   |
56 |     core::arch::wasm32::unreachable();
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call to unsafe function
```

Try `cargo install --force cargo-contract && rustup update` to make sure everything is up to date.

cmichi:
Thanks @forgetso! Just created a follow-up.

PR #913, opened 2021-09-10
Update release notes

cmichi:
Follow-up to https://github.com/paritytech/ink/pull/909.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -6,6 +6,12 @@ The list below shows the additions, changes and fixes that are visible to users 
 
 ## Compatibility
 
+Make sure to use a recent Rust nightly and `cargo-contract` with the current
+release: 
+```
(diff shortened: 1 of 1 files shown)


RELEASES.md
```diff
@@ -6,6 +6,12 @@ The list below shows the additions, changes and fixes that are visible to users
 
 ## Compatibility
 
+Make sure to use a recent Rust nightly and `cargo-contract` with the current
+release: 
+```
+cargo install --force cargo-contract && rustup update
```
ascjones, line 12: I'm not 100% sure about adding `--force`, it would mean if someone had installed from Github with the same version it would be overwritten. Which may or may not be what they want.

ascjones (approved):

RELEASES.md
cmichi: I've set it to the current fixed version.

PR #917, opened 2021-09-13
Fix code coverage

cmichi:
We need to link dead code for the code coverage. This is because the code coverage must take dead code into account ‒ it's dead code which should actually be tested in some way; if it's not tested it's not covered and this should show up negatively in our coverage results.

For trait cross-calls we have a special case though, which I've described in the code comment of this PR, so for brevity I won't restate it here.

Another way of implementing this would be to place an early return in the methods which insert `__ink_enforce_error` in [`cross_calling.rs`](https://github.com/paritytech/ink/blob/master/crates/lang/codegen/src/generator/cross_calling.rs) (in case `INK_COVERAGE_REPORTING` is set to `true`). I've decided against this, since those code paths would then be excluded for all tests which are run in our `codecov` CI stage, making it impossible to improve coverage for them.


ascjones (commented):
Okay a bit hacky but I can't think of anything better. 

Out of interest, does it measures coverage of the macro code called by the ui tests here?

cmichi:
>Out of interest, does it measures coverage of the macro code called by the ui tests here?

My understanding is that the coverage tool only considers execution branches touched during the actual execution. The UI tests on the other hand just check if the build succeeds (via `trybuild`).

ascjones:
> My understanding is that the coverage tool only considers execution branches touched during the actual execution

If this is indeed the case then we can just exclude all `ui` tests on the code coverage run?

cmichi:
> > My understanding is that the coverage tool only considers execution branches touched during the actual execution
> 
> If this is indeed the case then we can just exclude all `ui` tests on the code coverage run?

@ascjones The results from https://github.com/paritytech/ink/pull/924 show that the UI tests are taken into account for codecov. Hence we can't exclude all of them, but need to limit it to this particular one.

crates/env/src/call/call_builder.rs
```diff
@@ -30,6 +30,7 @@ use core::marker::PhantomData;
 
 /// The final parameters to the cross-contract call.
 #[derive(Debug)]
+#[allow(dead_code)]
```
HCastano, line 33: Can you add a short command explaining why this is okay?

crates/env/src/call/create_builder.rs
```diff
@@ -52,6 +52,7 @@ where
 
 /// Builds up contract instantiations.
 #[derive(Debug)]
+#[allow(dead_code)]
```
HCastano, line 55: Same here

HCastano (commented):
So the PR does what the description says, but I don't quite understand the motivation here.

> We need to link dead code for the code coverage. This is because the code coverage must take dead code into account ‒ it's dead code which should actually be tested in some way; if it's not tested it's not covered and this should show up negatively in our coverage results.

So maybe to re-word what you said: if we don't have test coverage for a piece of code the compiler would consider that "dead" and therefore not include it in the final binary. In this case it wouldn't get counted in the coverage report, thus skewing the numbers (the percentage of code covered would be higher)?

crates/env/src/call/call_builder.rs
cmichi: Ah this was leftover code, thanks for spotting.

cmichi:
>So maybe to re-word what you said: if we don't have test coverage for a piece of code the compiler would consider that "dead" and therefore not include it in the final binary. In this case it wouldn't get counted in the coverage report, thus skewing the numbers (the percentage of code covered would be higher)?

Yeah exactly!

crates/lang/macro/tests/compile_tests.rs
```diff
@@ -23,7 +23,17 @@ fn compile_tests() {
...
+        // The cross-calling implementation for traits provides
+        // an invalid implementation for non-valid message calls
+        // (e.g. cross-calling a `mut` message from a non-`mut` message).
+        // So calling those will result in a compiler or linker error.
+        //
+        // The coverage reporting CI stage though also links dead code,
+        // hence resulting in this invalid implementation being linked
+        // and thus a linker error.
+        t.pass("tests/ui/pass/08-flipper-as-dependency-trait.rs");
+    }
```
Robbepop, line 36: If ink! could find out when it is compiled under coverage reporting then ink! could avoid producing those enforced linker errors. Maybe that's a better solution?

cmichi: I covered this in the description of this PR:

>Another way of implementing this would be to place an early return in the methods which insert `__ink_enforce_error` in [`cross_calling.rs`](https://github.com/paritytech/ink/blob/master/crates/lang/codegen/src/generator/cross_calling.rs) (in case `INK_COVERAGE_REPORTING` is set to `true`). I've decided against this, since those code paths would then be excluded for all tests which are run in our `codecov` CI stage, making it impossible to improve coverage for them.

tl;dr is that by excluding those code paths I think we skew our code coverage results without any possibility of improving on the left out code branches.

crates/lang/codegen/src/generator/cross_calling.rs
```diff
@@ -253,10 +253,18 @@ impl CrossCalling<'_> {
                 & #mut_tok self,
                 #( #input_bindings : #input_types ),*
             ) -> Self::#output_ident {
-                extern {
-                    fn #linker_error_ident() -> !;
+                if option_env!("INK_COVERAGE_REPORTING") != Some("true") {
```
Robbepop, line 256:
```suggestion
                if ::core::option_env!("INK_COVERAGE_REPORTING") != ::core::option::Option::Some("true") {
```

crates/lang/codegen/src/generator/cross_calling.rs
```diff
@@ -253,10 +253,18 @@ impl CrossCalling<'_> {
...
+                    extern {
+                        fn #linker_error_ident() -> !;
+                    }
+                    unsafe { #linker_error_ident() }
+                } else {
+                    // The code coverage reporting CI stage links dead code,
+                    // hence we have to provide an `unreachable!` here. If
+                    // the invalid implementation above is linked this results
+                    // in a linker error.
+                    unreachable!("this is an invalid message call which should never be possible.");
```
Robbepop, line 266:
```suggestion
                    ::core::unreachable!("this is an invalid ink! message call which should never be possible.");
```

crates/lang/codegen/src/generator/cross_calling.rs
```diff
@@ -386,10 +394,18 @@ impl CrossCalling<'_> {
             fn #ident(
                 #( #input_bindings : #input_types ),*
             ) -> Self::#output_ident {
-                extern {
-                    fn #linker_error_ident() -> !;
+                if option_env!("INK_COVERAGE_REPORTING") != Some("true") {
```
Robbepop, line 397:
```suggestion
                if ::core::option_env!("INK_COVERAGE_REPORTING") != ::core::option::Option::Some("true") {
```

crates/lang/codegen/src/generator/cross_calling.rs
```diff
@@ -386,10 +394,18 @@ impl CrossCalling<'_> {
...
+                    extern {
+                        fn #linker_error_ident() -> !;
+                    }
+                    unsafe { #linker_error_ident() }
+                } else {
+                    // The code coverage reporting CI stage links dead code,
+                    // hence we have to provide an `unreachable!` here. If
+                    // the invalid implementation above is linked this results
+                    // in a linker error.
+                    unreachable!("this is an invalid message call which should never be possible.");
```
Robbepop, line 407:
```suggestion
                    ::core::unreachable!("this is an invalid ink! constructor call which should never be possible.");
```

Robbepop (commented):
I like the fix as proposed using our codegen. However, I really hope that the check
```rust
option_env!("INK_COVERAGE_REPORTING") != Some("true")
```
can actually be computed by the Rust compiler at compilation time and kicks out the external function reference. I actually doubt that. Maybe it would be better if we were using some `#cfg` approach similar to `cfg_if` if that can be used with `option_env`.

cmichi:
I've changed the check so that it's computed at compile time.

>Most important question: Have you checked and verified that this codegen really does not introduce the invalid external function declarations?

Yes, the `codecov` stage builds successfully in our CI.

Robbepop:
> I've changed the check so that it's computed at compile time.

I think this reflects a more stable codegen that does not depend on Rust doing certain optimizations.

Robbepop (changes requested):
I like that our own codegen now decides what code to produce, however, there is a lot of duplicate code. The only things that need to be conditionally produced are:
```rust
extern {
    fn #linker_error_ident() -> !;
}
unsafe { #linker_error_ident() }
```
if `option_env!("INK_COVERAGE_REPORTING") != Some("true")` evaluates to `true` and
```rust
// The code coverage reporting CI stage links dead code,
// hence we have to provide an `unreachable!` here. If
                    // the invalid implementation above is linked this results
// in a linker error.
::core::unreachable!("this is an invalid ink! message call which should never be possible.");
```
otherwise.
All the other code in the `quote_spanned` calls is duplicate. Please resolve this.

Robbepop (approved):
LGTM

PR #919, opened 2021-09-14
CI: chown the built docs

TriplEight:
Closes: https://github.com/paritytech/ci_cd/issues/200
cc https://github.com/paritytech/ci_cd/issues/92


.gitlab-ci.yml
```diff
@@ -198,6 +198,8 @@ docs:
         -p ink_primitives -p ink_prelude
         -p ink_lang -p ink_lang_macro -p ink_lang_ir -p ink_lang_codegen
     - mv ${CARGO_TARGET_DIR}/doc ./crate-docs
+    # FIXME: remove me after CI image gets nonroot
```
cmichi, line 201: I guess this is a comment for the CI/CD team? Should we create a follow-up?

TriplEight: it's mentioned both in the issue and description

PR #921, opened 2021-09-16
Fix nightly clippy errors

cmichi:


crates/env/src/call/call_builder.rs
```diff
@@ -30,6 +30,7 @@ use core::marker::PhantomData;
 
 /// The final parameters to the cross-contract call.
 #[derive(Debug)]
+#[allow(dead_code)]
```
cmichi, line 33: Not really happy with the solution of adding the `#[allow(dead_code)]` attributes, any other ideas?

Robbepop: how about some proper `#[cfg(..)]`?

Robbepop: I'd not like to accept an `#[allow(dead_code)]` hack-fix.

cmichi: > how about some proper `#[cfg(..)]`?

The `call_builder` is only used in examples and tests, but not in the crate itself ‒ that's why the fields are never read there. How would a `#[cfg(…)]` fix this? We run `clippy` on all targets.

crates/env/src/call/call_builder.rs
```diff
@@ -41,17 +41,11 @@ where
     /// The transferred value for the call.
     transferred_value: E::Balance,
     /// The expected return type.
-    return_type: ReturnType<R>,
+    _return_type: ReturnType<R>,
```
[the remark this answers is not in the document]
cmichi: I think it's used to preserve the `R`, so kind of like `PhantomData`?

Robbepop (approved):
LGTM

Issue #900, opened 2021-08-25
Rename `name` in metadata to `label`


cmichi:
In our metadata we currently have a `name` Array field for constructors, messages and arguments.

There are some issues with it:

* Contrary to the term it is not actually any kind of identifier. This field does not have any impact on contract execution, it is used purely for display purposes in the UI's.
* We have a field `selector` which actually serves as the identifier. There is some concern that users will mistakenly assume that the `name` has some implication for the contract execution. Hence renaming it to `label` would clarify this.
* For traits, ink! generates a field in the form of `"name": ["BaseErc20", "new"]`. The `BaseErc20` hereby is a trait implemented by the contract. This exposes internal implementation details of the contract without the developer having any influence over it.
* Right now it's not clear how a UI is supposed to handle the different elements in an Array. Concatenate them? How? Having a String would also simplify that.

Recently there was the PR https://github.com/paritytech/ink/pull/859 by @xgreenx where the need for influencing this value in the metadata was also stated.

My proposal is this:
* Rename `name` to `label` for constructors, messages and arguments.
* Make it a String instead of an Array.
* For traits, default to the form `BaseErc20::new` as String.

I'm creating this issue to start a discussion on this. It would be great if especially @seanyoung could give his opinion.

If we decide to do this these would need to be done:

## ToDo

- [ ] Rename `name` to `label` for constructors, messages and arguments.
- [ ] Make it a String instead of an Array.
- [ ] Revive https://github.com/paritytech/ink/pull/859, but with `label` instead of `metadata_name`. I would leave the possibility of modifying the arguments labels out for now.
- [ ] Create follow-up issues for Canvas UI and polkadot-js.

seanyoung:
- Should be there be defined upgrade path for polkadot-js/Canvas UI? Something like: try `label` first, else fall back to `name`. Or will just this change from one version to the next?
- Making this a string rather than an array is a good thing, I think.

cmichi:
>Should be there be defined upgrade path for polkadot-js/Canvas UI? Something like: try label first, else fall back to name. Or will just this change from one version to the next?

Good point and good idea! We should specify this fallback in the follow-up issues for the UI's.

xgreenx:
I agree with everything=) 
The string will be more clear for developers instead of the array. If they want, they can easily split it with `::`.
`label` is more descriptive than `name`.

[...]

PR #923, opened 2021-09-16
Changed the field `name` to `label`

xgreenx:
Updated the code generation of ABI. Instead of "name" in messages, constructors, arguments, events we are using "label".
Changed the behavior of "label" for messages and constructors, now it is a string instead of an array. 
For trait's messages, we will use the "TRAIT_IDENT::MESSAGE_IDENT" format of representation.

This PR is related to https://github.com/paritytech/ink/issues/900.

### DIFF
--- a/crates/lang/codegen/src/generator/metadata.rs
+++ b/crates/lang/codegen/src/generator/metadata.rs
@@ -127,7 +127,7 @@ impl Metadata<'_> {
         let ident = constructor.ident();
         let args = constructor.inputs().map(Self::generate_dispatch_argument);
         quote_spanned!(span=>
-            ::ink_metadata::ConstructorSpec::from_name(::core::stringify!(#ident))
+            ::ink_metadata::ConstructorSpec::from_label(::core::stringify!(#ident))
                 .selector([
                     #( #selector_bytes ),*
                 ])
@@ -216,7 +216,7 @@ impl Metadata<'_> {
                 let args = message.inputs().map(Self::generate_dispatch_argument);
                 let ret_ty = Self::generate_return_type(message.output());
                 quote_spanned!(span =>
-                    ::ink_metadata::MessageSpec::from_name(::core::stringify!(#ident))
+                    ::ink_metadata::MessageSpec::from_label(::core::stringify!(#ident))
                         .selector([
                             #( #selector_bytes ),*
                         ])
@@ -275,11 +275,9 @@ impl Metadata<'_> {
                         as ::ink_lang::reflect::TraitMessageInfo<#local_id>>::SELECTOR
                 }};
                 let ret_ty = Self::generate_return_type(message.output());
+                let label = [trait_ident.to_string(), message_ident.to_string()].join("::");
                 quote_spanned!(message_span=>
-                    ::ink_metadata::MessageSpec::from_trait_and_name(
-                        ::core::stringify!(#trait_ident),
-                        ::core::stringify!(#message_ident)
-                    )
+                    ::ink_metadata::MessageSpec::from_label(#label)
                         .selector(#selector)
                         .args([
                             #( #message_args ),*
--- a/crates/metadata/src/specs.rs
+++ b/crates/metadata/src/specs.rs
@@ -220,10 +220,10 @@ impl ContractSpec {
     deserialize = "F::Type: DeserializeOwned, F::String: DeserializeOwned"
 ))]
 pub struct ConstructorSpec<F: Form = MetaForm> {
-    /// The name of the message.
+    /// The label of the constructor.
     ///
-    /// In case of a trait provided constructor the trait name is prefixed.
-    pub name: Vec<F::String>,
+    /// In case of a trait provided constructor the label is prefixed with the trait label.
+    pub label: F::String,
     /// The selector hash of the message.
     pub selector: Selector,
     /// The parameters of the deployment handler.
@@ -237,7 +237,7 @@ impl IntoPortable for ConstructorSpec {
 
     fn into_portable(self, registry: &mut Registry) -> Self::Output {
         ConstructorSpec {
-            name: registry.map_into_portable(self.name),
+            label: self.label.into_portable(registry),
             selector: self.selector,
             args: self
                 .args
@@ -253,11 +253,11 @@ impl<F> ConstructorSpec<F>
 where
     F: Form,
 {
-    /// Returns the name of the message.
+    /// Returns the label of the constructor.
     ///
-    /// In case of a trait provided constructor the trait name is prefixed.
-    pub fn name(&self) -> &[F::String] {
-        &self.name
+    /// In case of a trait provided constructor the label is prefixed with the trait label.
+    pub fn label(&self) -> &F::String {
+        &self.label
     }
 
     /// Returns the selector hash of the message.
@@ -290,36 +290,18 @@ pub struct ConstructorSpecBuilder<Selector> {
 
 impl ConstructorSpec {
     /// Creates a new constructor spec builder.
-    fn from_name_segments<T>(
-        segments: T,
-    ) -> ConstructorSpecBuilder<Missing<state::Selector>>
-    where
-        T: IntoIterator<Item = &'static str>,
-    {
+    pub fn from_label(
+        label: &'static str,
+    ) -> ConstructorSpecBuilder<Missing<state::Selector>> {
         ConstructorSpecBuilder {
             spec: Self {
-                name: segments.into_iter().collect(),
+                label,
                 selector: Selector::default(),
                 args: Vec::new(),
                 docs: Vec::new(),
             },
             marker: PhantomData,
         }
-    }
-
-    /// Creates a new constructor spec builder.
-    pub fn from_name(
-        name: &'static str,
-    ) -> ConstructorSpecBuilder<Missing<state::Selector>> {
-        Self::from_name_segments([name])
-    }
-
-    /// Creates a new constructor spec builder for a trait provided constructor.
-    pub fn from_trait_and_name(
-        trait_name: &'static str,
-        constructor_name: &'static str,
-    ) -> ConstructorSpecBuilder<Missing<state::Selector>> {
-        Self::from_name_segments([trait_name, constructor_name])
     }
 }
 
@@ -375,11 +357,11 @@ impl ConstructorSpecBuilder<state::Selector> {
 ))]
 #[serde(rename_all = "camelCase")]
 pub struct MessageSpec<F: Form = MetaForm> {
-    /// The name of the message and some optional prefixes.
+    /// The label of the message.
     ///
     /// In case of trait provided messages and constructors the prefix
-    /// by convention in ink! is the name of the trait.
-    name: Vec<F::String>,
+    /// by convention in ink! is the label of the trait.
+    label: F::String,
     /// The selector hash of the message.
     selector: Selector,
     /// If the message is allowed to mutate the contract state.
@@ -413,9 +395,9 @@ mod state {
 }
 
 impl MessageSpec {
-    /// Creates a new message spec from the given name segments.
-    fn from_name_segments(
-        segments: Vec<&'static str>,
+    /// Creates a new message spec builder.
+    pub fn from_label(
+        label: &'static str,
     ) -> MessageSpecBuilder<
         Missing<state::Selector>,
         Missing<state::Mutates>,
@@ -424,7 +406,7 @@ impl MessageSpec {
     > {
         MessageSpecBuilder {
             spec: Self {
-                name: segments,
+                label,
                 selector: Selector::default(),
                 mutates: false,
                 payable: false,
@@ -435,43 +417,18 @@ impl MessageSpec {
             marker: PhantomData,
         }
     }
-
-    /// Creates a new message spec builder.
-    pub fn from_name(
-        name: &'static str,
-    ) -> MessageSpecBuilder<
-        Missing<state::Selector>,
-        Missing<state::Mutates>,
-        Missing<state::IsPayable>,
-        Missing<state::Returns>,
-    > {
-        Self::from_name_segments(vec![name])
-    }
-
-    /// Creates a new message spec builder for a trait provided message.
-    pub fn from_trait_and_name(
-        trait_name: &'static str,
-        message_name: &'static str,
-    ) -> MessageSpecBuilder<
-        Missing<state::Selector>,
-        Missing<state::Mutates>,
-        Missing<state::IsPayable>,
-        Missing<state::Returns>,
-    > {
-        Self::from_name_segments(vec![trait_name, message_name])
-    }
 }
 
 impl<F> MessageSpec<F>
 where
     F: Form,
 {
-    /// Returns the name of the message and some optional prefixes.
+    /// Returns the label of the message.
     ///
     /// In case of trait provided messages and constructors the prefix
-    /// by convention in ink! is the name of the trait.
-    pub fn name(&self) -> &[F::String] {
-        &self.name
+    /// by convention in ink! is the label of the trait.
+    pub fn label(&self) -> &F::String {
+        &self.label
     }
 
     /// Returns the selector hash of the message.
@@ -617,7 +574,7 @@ impl IntoPortable for MessageSpec {
 
     fn into_portable(self, registry: &mut Registry) -> Self::Output {
         MessageSpec {
-            name: registry.map_into_portable(self.name),
+            label: self.label.into_portable(registry),
             selector: self.selector,
             mutates: self.mutates,
             payable: self.payable,
@@ -639,8 +596,8 @@ impl IntoPortable for MessageSpec {
     deserialize = "F::Type: DeserializeOwned, F::String: DeserializeOwned"
 ))]
 pub struct EventSpec<F: Form = MetaForm> {
-    /// The name of the event.
-    name: F::String,
+    /// The label of the event.
+    label: F::String,
     /// The event arguments.
     args: Vec<EventParamSpec<F>>,
     /// The event documentation.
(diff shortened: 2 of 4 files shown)


xgreenx:
@cmichi What do you think about updating `displayName` to be a string instead of an array too?

cmichi:
Hey nice, that you created a PR for this!

> @cmichi What do you think about updating `displayName` to be a string instead of an array too?

:-) I actually had intended to create a ticket for this today! I agree that it's a good idea, but let's keep this to a separate issue. I'll create one where we track this.

README.md
```diff
@@ -221,6 +221,7 @@ In a module annotated with `#[ink::contract]` these attributes are available:
 | `#[ink(constructor)]` | Applicable to method. | Flags a method for the ink! storage struct as constructor making it available to the API for instantiating the contract. |
 | `#[ink(payable)]` | Applicable to ink! messages. | Allows receiving value as part of the call of the ink! message. ink! constructors are implicitly payable. |
 | `#[ink(selector = "..")]` | Applicable to ink! messages and ink! constructors. | Specifies a concrete dispatch selector for the flagged entity. This allows a contract author to precisely control the selectors of their APIs making it possible to rename their API without breakage. |
+| `#[ink(label = "..")]` | Applicable to ink! messages, ink! constructors and ink! implementation blocks. | Specifies a concrete label of method/trait inside of metadata. This allows a contract author to precisely control the labeling of methods inside of ABI making it possible to rename their methods without breakage. |
```
cmichi, line 224:
```suggestion
| `#[ink(label = "..")]` | Applicable to ink! messages, ink! constructors and ink! implementation blocks. | Specifies a concrete label of a method/trait inside the metadata. This allows a contract author to precisely control the labeling of methods inside of ABI making it possible to rename their methods without breakage. |
```

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -372,6 +384,11 @@ pub enum AttributeArg {
     /// Applied on ink! constructors or messages to manually control their
     /// selectors.
     Selector(Selector),
+    /// `#[ink(label = "transfer")]`
+    ///
+    /// Applied on ink! constructors, messages or implementation blocks
+    /// with traits to manually control their labeling inside of metadata.
```
cmichi, line 390:
```suggestion
    /// Applied to ink! constructors, messages or implementation blocks
    /// with traits to manually control their labeling inside the metadata.
```

crates/lang/ir/src/ir/item_impl/message.rs
```diff
@@ -104,6 +104,12 @@ pub struct Message {
     /// This overrides the computed selector, even when using a manual namespace
     /// for the parent implementation block.
     selector: Option<ir::Selector>,
+    /// An optional user provided label of method for metadata.
```
cmichi, line 107:
```suggestion
    /// An optional user provided label of the method for the metadata.
```

crates/lang/ir/src/ir/item_impl/message.rs
```diff
@@ -104,6 +104,12 @@ pub struct Message {
...
+    ///
+    /// # Note
+    ///
+    /// This overrides the default label of method inside of metadata.
```
cmichi, line 111:
```suggestion
    /// This overwrites the default label of the method inside the metadata.
```

crates/lang/ir/src/ir/item_impl/mod.rs
```diff
@@ -80,6 +80,13 @@ pub struct ItemImpl {
     /// names. Generally can be used to change computation of message and
     /// constructor selectors of the implementation block.
     namespace: Option<ir::Namespace>,
+    /// A label of trait used in metadata.
```
cmichi, line 83:
```suggestion
    /// A label of the trait used in the metadata.
```

crates/lang/ir/src/ir/item_impl/mod.rs
```diff
@@ -80,6 +80,13 @@ pub struct ItemImpl {
...
+    ///
+    /// # Note
+    ///
+    /// User can provide label of trait which will be used inside of metadata.
+    /// But he is not obliged to implement this trait.
```
cmichi, line 88:
```suggestion
    /// The user can provide a label of the trait which will be used inside the metadata.
    /// There is no requirement to implement this trait though.
```

crates/lang/ir/src/ir/item_impl/mod.rs
```diff
@@ -356,6 +366,31 @@ impl ItemImpl {
         self.namespace.as_ref()
     }
 
+    /// Returns the trait label of the implementation block if any has been provided.
+    ///
+    /// # Note
+    ///
+    /// If label is `""`, it means that user wants to remove trait labeling in metadata
```
cmichi, line 373:
```suggestion
    /// If the label is `""`, it means that the user wants to remove the trait labeling in the metadata.
```

crates/lang/macro/src/lib.rs
```diff
@@ -377,6 +377,46 @@ use proc_macro::TokenStream;
 ///     # }
 ///     ```
 ///
+///     **Controlling the labeling of messages/constructors in ABI:**
+///
+///     Every ink! message, ink! constructor and ink! implementation section has a label which
+///     is used during generation of metadata(ABI).
```
cmichi, line 383:
```suggestion
///     is used during generation of the metadata (ABI).
```

crates/lang/macro/src/lib.rs
```diff
@@ -377,6 +377,46 @@ use proc_macro::TokenStream;
...
+///
+///     An ink! smart contract author can control the labeling in ABI using the `label` flag. An example is shown below:
```
cmichi, line 385:
```suggestion
///     An ink! smart contract author can control the labeling in the metadata using the `label` flag.
///     An example is shown below:
```

crates/lang/macro/src/lib.rs
```diff
@@ -377,6 +377,46 @@ use proc_macro::TokenStream;
...
+///     impl Flipper {
+///         #[ink(constructor)]
+///         #[ink(label = "new_flipper")] // Works on constructors as well.
+///         pub fn new(initial_value: bool) -> Self {
+///             Flipper { value: false }
+///         }
+///
+///         # /// Flips the current value.
+///         # #[ink(message)]
+///         # #[ink(label = "my_flip")] // You can either specify name out-of-line.
```
cmichi, line 406:
```suggestion
///         # #[ink(label = "my_flip")] // You can also specify the label out-of-line.
```

crates/lang/macro/tests/ui/pass/12-attributes.rs
```diff
@@ -0,0 +1,40 @@
+#[ink_lang::contract]
+mod noop {
+    use ink_lang as ink;
+
+    #[ink::trait_definition]
+    pub trait Flipper {
+        // only `#[ink(message)]` and `#[ink(constructor)]` allowed in trait definition
```
cmichi, line 7:
```suggestion
        // Only `#[ink(message)]` and `#[ink(constructor)]` are allowed in trait definitions
```

crates/metadata/src/specs.rs
```diff
@@ -220,10 +220,10 @@ impl ContractSpec {
     deserialize = "F::Type: DeserializeOwned, F::String: DeserializeOwned"
 ))]
 pub struct ConstructorSpec<F: Form = MetaForm> {
-    /// The name of the message.
+    /// The label of the constructor.
     ///
-    /// In case of a trait provided constructor the trait name is prefixed.
-    pub name: Vec<F::String>,
+    /// In case of a trait provided constructor the label is prefixed with trait label.
```
cmichi, line 225:
```suggestion
    /// In case of a trait provided constructor the label is prefixed with the trait label.
```

crates/metadata/src/specs.rs
```diff
@@ -253,11 +253,11 @@ impl<F> ConstructorSpec<F>
 where
     F: Form,
 {
-    /// Returns the name of the message.
+    /// Returns the label of the constructor.
     ///
-    /// In case of a trait provided constructor the trait name is prefixed.
-    pub fn name(&self) -> &[F::String] {
-        &self.name
+    /// In case of a trait provided constructor the label is prefixed with trait label.
```
cmichi, line 258:
```suggestion
    /// In case of a trait provided constructor the label is prefixed with the trait label.
```

cmichi (changes requested):
Looks good on first glance ‒ great that you implemented this!

Got some language remarks. I guess the PSP-22 should also be adapted at some point in the future.

README.md
Robbepop: > | `#[ink(label = "..")]` | Applicable to ink! messages, ink! constructors and ink! implementation blocks. 

Why do we have the need to label ink! implementation blocks? What is the effect of it?

xgreenx: Marking the implementation block allows you to specify the prefix before the message or constructor. 
The user can use it to remove the prefix of trait, or change the prefix of trait, or add a prefix for the impl section without the trait.

In this [PR](https://github.com/paritytech/ink/pull/859) I described the case when we are using "change the prefix of trait" behavior.

Robbepop:
@xgreenx please rebase this on the current `master` branch so we might be ready for another round of review and merge it.

README.md
Robbepop: I am not sure I like this feature for trait implementation blocks since I see a potential for a lot of confusion. What is an actual concrete use case there?

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -764,6 +788,18 @@ impl TryFrom<syn::NestedMeta> for AttributeFrag {
                             }
                             return Err(format_err!(name_value, "expecteded 4-digit hexcode for `selector` argument, e.g. #[ink(selector = 0xC0FEBABE]"))
                         }
+                        if name_value.path.is_ident("label") {
+                            if let syn::Lit::Str(lit_str) = &name_value.lit {
+                                let name = lit_str.value();
```
Robbepop, line 793: This needs a check to be a valid Rust identifier.
You can check this by using
```rust
syn::parse_str::<syn::Ident>()
    .unwrap_or_else(|error| {
        panic!("encountered invalid non-identifier namespace argument: {}", error)
    })
```

Robbepop: Actually instead of panicking you should return a proper `syn::Error` via `format_err!` instead.

Robbepop: https://github.com/paritytech/ink/pull/931/files#diff-024136e95afaa59cacb0e87a4287e1e425de563db108032ba4b2d40df2c160eaR770

README.md
xgreenx: The actual use case is:
We have a trait with the name `Erc20External`. During generation of labels for methods ink! adds `Erc20External::` prefix. With this prefix it generates `Erc20External::transfer` label. The idea is to label the impl section with the `Erc20` label to generate the `Erc20::transfer` for the method.

We can remove the `label` attribute for the impl section if the `label` attribute also overrides the prefix(in the current implementation it only overrides the label of method and trait prefix affects it).

cmichi: > I am not sure I like this feature for trait implementation blocks since I see a potential for a lot of confusion. What is an actual concrete use case there?

@Robbepop Can you elaborate on where you see potential for confusion here? If you haven't seen yet, the [issue for this PR](https://github.com/paritytech/ink/issues/900) states some more of the motivation:

>For traits, ink! generates a field in the form of `"name": ["BaseErc20", "new"]`. The `BaseErc20` hereby is a trait implemented by the contract. This exposes internal implementation details of the contract without the developer having any influence over it.

crates/lang/ir/src/ir/attrs.rs
xgreenx: @Robbepop  I'm waiting for your decision in the [thread](https://github.com/paritytech/ink/pull/923#discussion_r714503215).
Because it will affect this part too=)

Robbepop: I cannot tell either what the better design is.
Let us summarize what you propose:

1) `#[ink(label = "foo")]` on an ink! implementation block is a prefix for `#[ink(label = "bar")]` on ink! messages in that same block. So both combined you end up with a label `foo::bar`.
2) Any `#[ink(label = "bar")]` on an ink! message or constructor in an ink! implementation block will overwrite the label of the ink! message or constructor entirely. Any `#[ink(label = "foo")]` on the ink! implementation block will be silently ignored.

xgreenx: 1. Yes, the label on ink! the message will overwrite only suffix(doesn't affect the prefix from the label on ink! impl block).
2. Yes, the label on ink! the message will overwrite entirely(affect the prefix from the label on ink! impl block).
3. `#[ink(label = "foo")]` on implementation block will be assumed as an error(it means remove support of the label for impl block).  `#[ink(label = "bar")]` on an ink! message in ink! implementation block will overwrite the default behavior(when the label of ink! message is a combination of trait name and message label).

I implemented the first case.
For me, the second case seems similar to the behavior of the `selector` attribute(if you defined the `selector`, it will ignore `namespace`.
The third case I created after your idea to remove the label for ink! implementation block=)

The idea of the `label` attribute is to add the ability to change the generated ABI. 
For example, in the code, I have two traits: One for internal usage(without ink! messages); another with ink! messages. Both traits have the method with the same name. It requires the developer in the code to explicitly specify the trait which he wants to use. To avoid this, I want to change the name of the method in the trait with ink! messages(I can't change the name in internal trait). But I will change the label of the method in ABI. In this case, the `label` attribute will help me to generate the right ABI.

cmichi:
I've lost a bit track of where we are with this PR and where the discussion went.

@xgreenx Could you maybe quickly summarize what decision needs to be made for you to be able to finish the PR?

xgreenx:
We need to decide do we need the `label` attribute or not. You think that [better to add](https://github.com/paritytech/ink/issues/900#issuecomment-949449714), @Robbepop thinks it is[ better to not add](https://github.com/paritytech/ink/issues/900#issuecomment-949433312) additional attributes=) [My thoughts](https://github.com/paritytech/ink/issues/900#issuecomment-949518025)

Robbepop:
> We need to decide do we need the `label` attribute or not. You think that [better to add](https://github.com/paritytech/ink/issues/900#issuecomment-949449714), @Robbepop thinks it is[ better to not add](https://github.com/paritytech/ink/issues/900#issuecomment-949433312) additional attributes=) [My thoughts](https://github.com/paritytech/ink/issues/900#issuecomment-949518025)

I am strictly against adding the label attribute. the metadata changes are ok. i see no reason in adding a feature where there is no strict requirement and where we are not 100% sure about its design.

cmichi:
>i see no reason in adding a feature where there is no strict requirement and where we are not 100% sure about its design.

@Robbepop Could you please reply to the points brought up in the discussion [here](https://github.com/paritytech/ink/issues/900#issuecomment-949449714)?

cmichi:
@xgreenx I've discussed it with @Robbepop and the outcome is that we'll only include the metadata changes for now and revisit the `#[ink(label)]` discussion at a later point. Would you be so kind to adapt your PR?

xgreenx:
Sure, will adapt it for the master branch and will remove `label` related code=)

xgreenx:
I've updated the PR. In ABI we still use "name" for storage layout and for types. Do I need to update it for storage layout? (because in the case of types we need to update `scale-info`)

crates/lang/codegen/src/generator/metadata.rs
```diff
@@ -275,11 +275,9 @@ impl Metadata<'_> {
                         as ::ink_lang::reflect::TraitMessageInfo<#local_id>>::SELECTOR
                 }};
                 let ret_ty = Self::generate_return_type(message.output());
+                let label = [trait_ident.to_string(), message_ident.to_string()].join("::");
```
xgreenx, line 278: Do you prefer this logic here or in `from_trait_and_label` on "spec" level?

What do you think about the idea that the namespace will affect the label too(not only selector)? If you agree then I will implement it like this https://github.com/paritytech/ink/blob/master/crates/lang/ir/src/ir/selector.rs#L108

cmichi:
@xgreenx Could you merge `master` into this PR?

cmichi:
>In ABI we still use "name" for storage layout and for types. Do I need to update it for storage layout? (because in the case of types we need to update scale-info)

Nope, no need to adapt those.

crates/lang/codegen/src/generator/metadata.rs
xgreenx: I meant, that I removed `from_trait_and_label ` from `spec` and implement the logic of concatenating(joining of trait name and method name by "::") here, on `generator` level.

cmichi: Alright, got it. It's fine here.

Do you want to implement the namespace affecting the label as well? Could also be done as a follow-up.

cmichi: I'll merge this now, I think the PR is fine in its current state and we've been putting you on hold for way too long :-).

Feel free to create a follow-up for including the namespace into the label.

cmichi:
@xgreenx The `ink-waterfall` CI stage is expected to fail, since it builds our example contracts and tries to upload them via polkadot-js. Since that UI doesn't understand the new metadata format yet the upload fails. I've created an issue for it and will merge the PR after that one is solved. Feel free to implement the namespace inclusion in the meantime or as a follow-up, whatever works best for you :-).

crates/lang/codegen/src/generator/metadata.rs
xgreenx: I can implement that in follow-up PR=) @Robbepop What do you think about the namespace affecting the label?

cmichi (approved):

PR #927, opened 2021-09-19
Expand selectors as hex-encoded literals in metadata

Robbepop:
Extracted from work in done in https://github.com/paritytech/ink/pull/665.

### DIFF
--- a/crates/lang/ir/src/literal.rs
+++ b/crates/lang/ir/src/literal.rs
@@ -1,0 +1,137 @@
+// Copyright 2018-2021 Parity Technologies (UK) Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


crates/lang/ir/src/literal.rs
```diff
@@ -0,0 +1,137 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+mod private {
+    /// Used to prevent external direct usage of `HexLiteral::hex_impl_`.
+    pub struct Sealed;
+}
```
HCastano, line 18: Can you describe how this pattern works?

HCastano (approved):
Question about the pattern used, but code/tests look good

crates/lang/ir/src/literal.rs
Robbepop: The `Sealed` type is inaccessible from the outside therefore one cannot provide it (in safe Rust) as parameter for the hidden method and therefore we protect to call it from outside the default trait methods.

PR #928, opened 2021-09-19
Make #[ink(selector = ..)] take a u32 parameter instead of a string

Robbepop:
This PR is the result of work in https://github.com/paritytech/ink/pull/665.

Currently ink! requires a user to input a string parameter for a message or constructor selector like the following:
```
#[ink(selector = "0xC0DECAFE")]
```
This is not necessary and the implementation requires the huge `regex` crate that bloats the compilation times.
The PR changes the `selector` property to take a `u32` integer instead. This allows for inputs like the following:
```
#[ink(selector = C0DECAFE)]
#[ink(selector = 1)]
```
README, example contracts and tests are updated as well.


README.md
```diff
@@ -220,8 +220,8 @@ In a module annotated with `#[ink::contract]` these attributes are available:
...
-| `#[ink(namespace = "..")]` | Applicable to ink! trait implementation blocks. | Changes the resulting selectors of all the ink! messages and ink! constructors within the trait implementation. Allows to disambiguate between trait implementations with overlapping message or constructor names. Use only with great care and consideration! |
+| `#[ink(selector = S:u32)]` | Applicable to ink! messages and ink! constructors. | Specifies a concrete dispatch selector for the flagged entity. This allows a contract author to precisely control the selectors of their APIs making it possible to rename their API without breakage. |
+| `#[ink(namespace = N:string)]` | Applicable to ink! trait implementation blocks. | Changes the resulting selectors of all the ink! messages and ink! constructors within the trait implementation. Allows to disambiguate between trait implementations with overlapping message or constructor names. Use only with great care and consideration! |
```
cmichi, line 224: What do `S` and `N` refer to here? We should add some context above the table to clarify.

cmichi (approved):
Looks good!

Could you add the change to the `Unreleased` section in our release notes as well?

With the next RC we then have to update the workshop and `ink-docs` as well. Maybe already provide PR's for this change there?

Another thing I was just wondering about: Won't there be any issue with the UI's? Because they SCALE-encode the selector as `String` when submitting the transaction, but now we use `u32` internally.

Robbepop:
> Looks good!
> 
> Could you add the change to the `Unreleased` section in our release notes as well?
> 
> With the next RC we then have to update the workshop and `ink-docs` as well. Maybe already provide PR's for this change there?
> 
> Another thing I was just wondering about: Won't there be any issue with the UI's? Because they SCALE-encode the selector as `String` when submitting the transaction, but now we use `u32` internally.

This PR does not really change how metadata is serialized in the end. It is just inspectable from a developer looking into the expanded smart contract code. The https://github.com/paritytech/ink/pull/927 PR actually makes a visible change to the produced metadata that leads to UIs having to adjust their _decode_ (not encoding) of ink! metadata selectors if they do not already support hex-encoded and non hex-encoded integer inputs.

README.md
Robbepop: `S` and `N` are the parameter names with respect to normal Rust syntax respectively.
`S` is the input `selector` of type `u32` and `N` is the input `namespace` of type `string`.

Robbepop:
> Maybe already provide PR's for this change there?

Updating of docs will be needed with this PR. I will update the preliminary `RELEASE` notes.
Although we should update docs once we do an actual release imo. What do you think?

cmichi:
> > Maybe already provide PR's for this change there?
> 
> Updating of docs will be needed with this PR. I will update the preliminary `RELEASE` notes.
> Although we should update docs once we do an actual release imo. What do you think?

Yup, I wrote 

>With the next RC we then have to update the workshop and `ink-docs` as well. Maybe already provide PR's for this change there?

:-), I think it would be good to already have the PR's which do these changes in the pipeline, so that we can merge them right after the next release, so that there'll be no discrepancy.

PR #929, opened 2021-09-21
Fix script which triggers downstream pipelines

cmichi:
The reason why the `ink-waterfall` yesterday mistakenly displayed a successful run for the `scale-info` PR is that the script currently exits with 0 after the timeout. So the script timed out and exited with 0, marking the CI stage as success.

This PR fixes this behavior and increases the timeout from the previous 30 mins to 1h.

### DIFF
--- a/scripts/.ci/trigger_pipeline.sh
+++ b/scripts/.ci/trigger_pipeline.sh
@@ -57,3 +57,5 @@ for i in $(seq 1 360); do
     fi
 sleep 5;
 done
+
+exit 1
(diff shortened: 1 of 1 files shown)


Robbepop (approved):

scripts/.ci/trigger_pipeline.sh
```diff
@@ -57,3 +64,6 @@ for i in $(seq 1 360); do
     fi
 sleep 5;
```
HCastano, line 65: nit: 
```suggestion
sleep POLL_SLEEP;
```

HCastano (approved):

PR #935, opened 2021-09-21
Misc cleanups in ink! codegen

Robbepop:
Work carved out of https://github.com/paritytech/ink/pull/665 in an attempt to make its diff more manageable.

- Some macro hygiene improvements.
- Modernization of some codegen parts.
- Simplification of codegen parts.

### DIFF
--- a/crates/lang/codegen/src/generator/metadata.rs
+++ b/crates/lang/codegen/src/generator/metadata.rs
@@ -69,7 +69,7 @@ impl Metadata<'_> {
         quote! {
             <#contract_ident as ::ink_storage::traits::StorageLayout>::layout(
                 &mut <::ink_primitives::KeyPtr as ::core::convert::From<::ink_primitives::Key>>::from(
-                    <::ink_primitives::Key as ::core::convert::From<[::core::primitive::u8; 32]>>::from([0x00_u8; 32])
+                    <::ink_primitives::Key as ::core::convert::From<[::core::primitive::u8; 32usize]>>::from([0x00_u8; 32usize])
                 )
             )
         }
@@ -84,16 +84,16 @@ impl Metadata<'_> {
         quote! {
             ::ink_metadata::ContractSpec::new()
                 .constructors([
-                    #(#constructors ,)*
+                    #( #constructors ),*
                 ])
                 .messages([
-                    #(#messages ,)*
+                    #( #messages ),*
                 ])
                 .events([
-                    #(#events ,)*
+                    #( #events ),*
                 ])
                 .docs([
-                    #(#docs ,)*
+                    #( #docs ),*
                 ])
                 .done()
         }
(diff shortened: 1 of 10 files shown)


crates/lang/codegen/src/generator/metadata.rs
```diff
@@ -84,16 +84,16 @@ impl Metadata<'_> {
         quote! {
             ::ink_metadata::ContractSpec::new()
                 .constructors([
-                    #(#constructors ,)*
+                    #( #constructors ),*
```
HCastano, line 87: Does this actually make a difference as far as `quote` is concerned, or is it an aesthetic preference?

crates/lang/codegen/src/generator/metadata.rs
```diff
@@ -270,7 +275,7 @@ impl Metadata<'_> {
         match ret_ty {
             None => {
                 quote! {
-                    ::ink_metadata::ReturnTypeSpec::new(None)
+                    ::ink_metadata::ReturnTypeSpec::new(::core::option::Option::None)
```
HCastano, line 278: I feel like we should be able to do all these path clean-ups with a Bash script

crates/lang/codegen/src/generator/mod.rs
```diff
@@ -12,6 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/// Implements `AsRef` for a code generator type.
+///
+/// Code generators always have a shared `contract` reference to the contract.
+/// They need to implement this trait in order to use other code generators.
+macro_rules! impl_as_ref_for_generator {
```
HCastano, line 19: nit: I'd move this to after the imports

crates/lang/codegen/src/generator/storage.rs
```diff
@@ -30,24 +30,18 @@ use syn::spanned::Spanned as _;
 pub struct Storage<'a> {
     contract: &'a ir::Contract,
 }
-
-impl<'a> AsRef<ir::Contract> for Storage<'_> {
-    fn as_ref(&self) -> &ir::Contract {
-        self.contract
-    }
-}
+impl_as_ref_for_generator!(Storage);
```
HCastano, line 33:
```suggestion

impl_as_ref_for_generator!(Storage);
```

crates/lang/codegen/src/generator/storage.rs
```diff
@@ -30,24 +30,18 @@ use syn::spanned::Spanned as _;
...
         let access_env_impls = self.generate_access_env_trait_impls();
         let storage_struct = self.generate_storage_struct();
-        let use_emit_event = if self.contract.module().events().next().is_some() {
-            // Required to allow for `self.env().emit_event(..)` in messages and constructors.
-            Some(quote! { use ::ink_lang::EmitEvent as _; })
-        } else {
-            None
-        };
+        let use_emit_event =
+            self.contract.module().events().next().is_some().then(|| {
```
HCastano, line 41: Nice 😛

HCastano (approved):

crates/lang/codegen/src/generator/metadata.rs
Robbepop: It does make a difference if the list of constructors is not empty.
Before: `.constructors([ first, second, ])`
Now: `.constructors([ first, second ])`
Note the missing comma at the end.

crates/lang/codegen/src/generator/metadata.rs
Robbepop: I hope we won't be trusting a bash script for transformations like these on our code. :sweat_smile:

crates/lang/codegen/src/generator/mod.rs
Robbepop: then it won't work ... Rust's macro system needs this ordering unfortunately.

crates/lang/codegen/src/generator/storage.rs
Robbepop: I really like this new API. ^^
Another nice one is `array::map`.

crates/lang/codegen/src/generator/metadata.rs
HCastano: 🤦‍♂️ didn't notice that while skimming through

crates/lang/codegen/src/generator/mod.rs
HCastano: I feel like I knew that deep down, but TIL

PR #937, opened 2021-09-23
Fix `env::block_timestamp` example

cmichi:
I noticed that the example contains a mistake. The comment says
```
/// The function can be executed at most once every 100 blocks.
```
But the `env` function used there is actually `block_timestamp()`.

I first thought about just modifying the comment to 
```
/// The function can be executed at most once every 100 milliseconds.
```
But since we don't know if the environment `Timestamp` actually counts in milliseconds we can't do that.

### DIFF
--- a/crates/lang/src/env_access.rs
+++ b/crates/lang/src/env_access.rs
@@ -517,25 +517,32 @@ where
     /// # Example
     ///
     /// ```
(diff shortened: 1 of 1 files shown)


Robbepop (approved):
LGTM

Issue #945, opened 2021-09-28
Simple Mapping Storage Primitive


HCastano:
## Motivation

At the moment, `ink!` data structures, such as the `ink_storage::LazyHashMap`, are quite
complicated. This is because they were designed in a very "polite" way. That's to say,
they did a lot of internal bookkeeping to ensure that they cleaned up after themselves,
and that they were only going to the contract storage database when it was absolutely
neccessary.

While this was great from a user's point of view, this approach to data structure in
`ink!` has lead to contract sizes being larger than they probably need to be.

As such, we want to experiment with some simpler data structures, such as a Solidity-like
mapping.

## `ink_storage::Mapping`

The new `ink_storage::Mapping` would be a paper thin type-safe wrapper around some of the
`ink_storage` and `ink_env` storage primitives.

The `Mapping` type would look as follows:

```rust
struct Mapping<K, V>(PhantomData<K>, PhantomData<V>);
```

Note that we may also need some sort of `Key` field to differentiate between different
instances of `Mapping` and avoid collisions. This would be similar to what `LazyHashMap`
does.

The API of the `Mapping` would look (roughly) like this:

```rust
impl<K, V: Default> Mapping<K, V> {
    ...

    /// Write value to `key`.
    fn push(&mut self, key: K, value: V) {
        todo!()
    }

    /// Get the value at `key`.
    ///
    /// If there is no value at the `key` it will return the default value.
    fn get(&self, key: K) -> V {
        todo!()
    }

    ...
}
```

It would also need to implement the `SpreadLayout` and `PackedLayout` traits. This is
what would actually be doing the heavy lifting as far as talking to the contract storage
goes.

Another thing that will have to get done is adding some new low level push/pull
functions which allow the return values to be empty. At the moment these functions assume
that [if you're querying for a key, that it already exists](https://github.com/paritytech/ink/blob/c6f639188cdbe21a768297c015d606b390992d2f/crates/storage/src/traits/mod.rs#L138).

This would not be the case with the new `Mapping`, since you can query any key and if it
doesn't exist you'd get the `Default` value back.

## ERC-20 Benchmarking

After implementing the new `Mapping` type we want to implement a new ERC-20 example
contract with it. The thing to look for here is how much better (or worse) the contract
sizes are compared to our current ERC-20 example.

xgreenx:
Hi, I've already tested the change like this, so I can briefly describe the result and later provide the full report.
use of simple mapping saves 16 KB for the current ERC-20 example.

[...]

PR #946, opened 2021-09-29
Simple Mapping Storage Primitive 

HCastano:
An implementation of what's described in https://github.com/paritytech/ink/issues/945.

~~Some things I'm still unsure about:~~

~~1. The handling of `Mapping`'s `key` field
Not sure if this should be user initialized since it's kinda an implementation detail
which is there to make sure we don't have any collisions in storage.~~

~~2. The `SpreadLayout` implementation
So from what I can tell, when we implement this trait we're supposed to indicate how our
structure should be represented in the contract storage.  In the case of `Mapping`, since
it doesn't actually own any data, its really just a proxy for how we represent `Key` in
storage. Am I thinking about this correctly?~~


crates/storage/src/collections/mapping/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+
+use core::marker::PhantomData;
+
+/// A mapping of key-value pairs directly into contract storage.
+///
+/// If a key does not exist the `Default` value for the `value` will be returned.
+pub struct Mapping<K, V> {
+    key: Key,
+    _phantom_mapping: (PhantomData<K>, PhantomData<V>),
+}
```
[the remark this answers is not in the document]
Robbepop: With `Option<Key>` we'd have a way to store information as to how the `Mapping` got initialized. It would be `None` in an `#[ink(constructor)]` called with `Mapping::new()` and it would be `Some(key)` if constructed with `SpreadLayout::pull_spread`.
The only downside to this approach is that this would make it impossible to use the `Mapping` within a constructor call.
For that exact reason we are caching the stored elements so that we can later push them down the sink.
Maybe we could do the same if we feel that this change is needed.

```rust
pub enum Mapping {
    New {
        ops: Vec<Operation<K, V>>,
    },
    Pulled {
        key: Key,
        _marker: PhantomData<fn() -> (K,V)>,
    }
}

pub enum Operation<K, V> {
    Insert(K, V),
    Erase(K),
}
```
However, this would no longer be as leightweight ... :sweat_smile:

crates/storage/src/collections/mapping/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+
+impl<K, V> Mapping<K, V>
+where
+    K: scale::Encode,
+    V: scale::Encode + scale::Decode + Default,
```
Robbepop, line 44: It is better to use the `PackedLayout` trait here instead of directly using `scale::{Encode, Decode}` as we do in other data structures.

crates/storage/src/collections/mapping/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub fn new(key: Key) -> Self {
+        Self {
+            key,
+            _phantom_mapping: (Default::default(), Default::default()),
+        }
+    }
+
+    /// Insert the given `value` to the contract storage.
+    pub fn insert(&mut self, key: K, value: V) {
+        ink_env::set_contract_storage(&self.key(key), &value);
```
Robbepop, line 58: We should use the `packed_{push, pull}` functions here instead of these super low level ones.

crates/storage/src/collections/mapping/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// Get the `value` at `key` from the contract storage.
+    pub fn get(&self, key: K) -> V {
+        ink_env::get_contract_storage(&self.key(key))
+            .unwrap_or_default()
+            .unwrap_or_default()
```
Robbepop, line 65: We should use the packed_{push, pull} functions here instead of these super low level ones.

Robbepop (changes requested):

crates/storage/src/collections/mapping/mod.rs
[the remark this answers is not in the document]
HCastano: Why are you suggesting `fn() -> (K, V)` over just `(K, V)`?

>  The only downside to this approach is that this would make it impossible to use the Mapping within a constructor call.

Can you elaborate on this?

> For that exact reason we are caching the stored elements so that we can later push them down the sink.
Maybe we could do the same if we feel that this change is needed.

I agree, this wouldn't be lightweight and kinda defeats the point of the `Mapping`

HCastano:
So this PR isn't quite ready yet, but I'm still at a point where I can use `Mapping` in the ERC-20 example and have it compile (this doesn't necessarily means it runs correctly on chain). The size improvements are great:

```
Master:  Original wasm size: 68.0K, Optimized: 28.8K
This PR: Original wasm size: 37.8K, Optimized: 12.6K
```

They're basically the same as @xgreenx's `SimpleHashMap` implementation, although this isn't surprising because our implementations are really similar. Gonna need to do some more thinking about where we can lose some extra bytes 😉

xgreenx:
You are not supporting `len` and your storage key is not optional, so your implementation takes less space=) From ERC-20 example you can remove usage of `Lazy` and it will save 1.3 KB=)

I think it is better to return an `Option<V>` from the `get` method because in this case, the developer will be available to check if the object exists or not. Because in Solidity in some cases it was a headache.

Robbepop:
Compiling the adjusted ERC-20 with `cargo contract build` yields 12.6kB.
Exchanging the `Balance` type from `u128` to `u64` or `u32` we get down to 12.1kB.
Exchanging the `AccountId` from `[u8; 32]` to `[u8; 8]` we get down to 9.7kB.

This means that a majority of the cluft comes from copying the huge `AccountId` type around.
I see that especially in calls to [the allowancehttps://github.com/paritytech/ink/blob/hc-simple-mapping-primitive/examples/erc20/lib.rs#L97) function that takes 2 parameters by value and turns them into a tuple-packed reference.

Ideally we can introduce some traits in order to allow for use cases like these without requiring value types anywhere.
E.g. making `&(U, V)` equal to `(&U, &V)` in those cases so we could avoid a copy there.

Robbepop (commented):
The `Mapping` is missing some trivial `Debug` implementation:

```rust
impl<K, V> core::fmt::Debug for Mapping<K, V> {
    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
        f.debug_struct("Mapping").finish()
    }
}
```

crates/storage/src/collections/mapping/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+        Self {
+            key,
+            _marker: Default::default(),
+        }
+    }
+
+    /// Insert the given `value` to the contract storage.
+    pub fn insert(&mut self, key: K, value: V) {
+        push_packed_root(&value, &self.key(&key));
+    }
```
Robbepop, line 64:
```suggestion
    /// Insert the given `value` to the contract storage.
    pub fn insert<Q>(&mut self, key: &Q, value: V)
    where
        K: Borrow<Q>,
        Q: Encode,
    {
        push_packed_root(&value, &self.key(key));
    }
```

crates/storage/src/collections/mapping/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    /// Get the `value` at `key` from the contract storage.
+    ///
+    /// Returns `None` if no `value` exists at the given `key`.
+    pub fn get(&self, key: &K) -> Option<V> {
+        pull_packed_root_opt(&self.key(key))
+    }
```
Robbepop, line 71:
```suggestion
    pub fn get<Q>(&self, key: &Q) -> Option<V>
    where
        K: Borrow<Q>,
        Q: Encode,
    {
         pull_packed_root_opt(&self.key(key))
    }
```

crates/storage/src/collections/mapping/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    fn key(&self, key: &K) -> Key {
+        let encodedable_key = (self.key, key);
+        let mut output = <Blake2x256 as HashOutput>::Type::default();
+        ink_env::hash_encoded::<Blake2x256, _>(&encodedable_key, &mut output);
+        output.into()
+    }
```
Robbepop, line 78:
```suggestion
    fn key<Q>(&self, key: &Q) -> Key
    where
        K: Borrow<Q>,
        Q: Encode,
    {
        let encodedable_key = (&self.key, key);
        let mut output = <Blake2x256 as HashOutput>::Type::default();
        ink_env::hash_encoded::<Blake2x256, _>(&encodedable_key, &mut output);
        output.into()
    }
```

HCastano:
I'm trying to deploy the modified `ERC-20` built with `cargo-contract` `v15.0` and Substrate `master`, but I keep getting `ContractTrapped` errors on deployment. @athei is there anything obviously wrong with the `Mapping`? If not, I'll have to do some more digging tomorrow.

Probably not that useful, but just in case here is the error. 

<details>

```
signAndSend: status :: {"dispatchError":{"module":{"index":18,"error":12}},"dispatchInfo":{"weight":5037897809,"class":"Normal","paysFee":"Yes"},"events":[{"phase":{"applyExtrinsic":1},"event":{"index":"0x0608","data":["5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",127073230084491]},"topics":[]},{"phase":{"applyExtrinsic":1},"event":{"index":"0x0607","data":["5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",198067438571]},"topics":[]},{"phase":{"applyExtrinsic":1},"event":{"index":"0x0607","data":["5EYCAe5ijiYfyeZ2JJCGq56LmPyNRAKzpG4QkoQkkQNB5e6Z",101500130116736]},"topics":[]},{"phase":{"applyExtrinsic":1},"event":{"index":"0x1106","data":[101500130116736]},"topics":[]},{"phase":{"applyExtrinsic":1},"event":{"index":"0x0607","data":["5GNJqTPyNqANBkUVMN1LPPrxXnFouWXoe2wNSmmEoLctxiZY",25375032529184]},"topics":[]},{"phase":{"applyExtrinsic":1},"event":{"index":"0x0001","data":[{"module":{"index":18,"error":12}},{"weight":5037897809,"class":"Normal","paysFee":"Yes"}]},"topics":[]}],"status":{"inBlock":"0x58f82e155f62f4efed91615163d5cd0e106592edde844542b6916301de1b9774"}}
```

</details>

athei:
Looked through the code but too many abstractions. Can't really tell what is happening. But I am confident that you will narrow it down pretty quickly by using printf debugging.

Keep in mind that `ContractTrapped` means you hit an `unreachable` instruction. This means a panic. So look for those. I think the panick message should be emitted to the debug buffer. However, as long as contract UIs don't execute instantiate as RPC you won't see those when the error happens on instantiation.

crates/lang/codegen/src/generator/storage.rs
```diff
@@ -97,6 +97,7 @@ impl Storage<'_> {
                 derive(::ink_storage::traits::StorageLayout)
             )]
             #[derive(::ink_storage::traits::SpreadLayout)]
+            #[derive(::ink_storage::traits::SpreadAllocate)]
```
Robbepop, line 100: First: You cannot do this since not all `#[ink(storage)]` structs should implement `SpreadAllocate` which is why it is separate from `SpreadLayout` and has not been integrated into the already existing traits.
Second: If this was actually useful (which it isn't) it should reside in another PR.

Robbepop: For ink! codegen to automatically implement `SpreadAllocate` optionally we'd need support for Rust trivial trait bounds which currently are unstable.

Robbepop (changes requested):

crates/lang/codegen/src/generator/storage.rs
HCastano: My bad. I thought that the idea behind merging #995 before the `Mapping` was so we could derive `SpreadAllocate` for storage types in the codegen instead of manually deriving it

Robbepop: If Rust had trivial trait bounds we actually could do exactly that ... :(

examples/erc20/lib.rs
```diff
@@ -59,20 +61,19 @@ mod erc20 {
         /// Creates a new ERC-20 contract with the specified initial supply.
         #[ink(constructor)]
         pub fn new(initial_supply: Balance) -> Self {
+            ink_lang::codegen::initialize_contract(|contract| Self::new_init(contract, initial_supply))
```
HCastano, line 64: @Robbepop using the `codegen` module here feels wrong, is there a better way to expose this to users?

HCastano:
FYI, with the additional bloat from https://github.com/paritytech/ink/pull/998 we're currently sitting at 12.1K.

examples/erc20/lib.rs
Robbepop: The actual plan is still that we provide some useful and well designed syntactic sugar so that this line of code won't be necessary but maybe that's not well thought through and maybe writing out this line of code actually isn't so bad. Then we could put this into some other module or into root.

Robbepop:
> FYI, with the additional bloat from https://github.com/paritytech/ink/pull/998 we're currently sitting at 12.1K.

The bloat can easily be fixed with another PR and more intelligent codegen.

xgreenx:
> FYI, with the additional bloat from #998 we're currently sitting at 12.1K.

Can you try the commit from this https://github.com/paritytech/ink/pull/999 PR? it should reduce the size=)

xgreenx:
I tried to build this change and https://github.com/paritytech/ink/pull/999.
Erc20 with `Lazy` for `total_supply`:
Original wasm size: 34.1K, Optimized: 10.9K
Erc20 without `Lazy`:
Original wasm size: 33.6K, Optimized: 10.5K

xgreenx:
Some other tests:

This + https://github.com/paritytech/cargo-contract/pull/358: Original wasm size: 35.4K, Optimized: 11.7K
This + https://github.com/paritytech/cargo-contract/pull/358 + https://github.com/paritytech/ink/pull/1012: Original wasm size: 34.3K, Optimized: 10.6K
This + https://github.com/paritytech/ink/pull/1012: Original wasm size: 35.2K, Optimized: 11.0K
This + https://github.com/paritytech/ink/pull/999 + https://github.com/paritytech/cargo-contract/pull/358: Original wasm size: 33.2K, Optimized: 10.4K. In this case, https://github.com/paritytech/ink/pull/1012 doesn't cause any effect on `Erc20`.

HCastano:
@xgreenx I haven't had time to loop back into this, so thanks for your experiments!

HCastano:
I think we can probably merge this as-is, and fix anything in follow ups (like not using the `codegen` module directly). Having this in `master` would allow us to start doing more experiments with the `Mapping`

Robbepop:
I would like to see the new Mapping type under ink_storage::lazy instead of ink_storage::collections module before we merge this.

crates/storage/src/lazy/mapping.rs
```diff
@@ -0,0 +1,189 @@
...
+//!
+//! This mapping doesn't actually "own" any data.
+//! Instead it is just a simple wrapper around the contract storage facilities.
+
+use crate::traits::{
+    pull_packed_root_opt,
+    push_packed_root,
+    ExtKeyPtr,
+    KeyPtr,
+    PackedLayout,
```
Robbepop, line 27: LGTM

Robbepop (approved):

PR #947, opened 2021-10-02
Implement selector_id!, selector_bytes! and blake2x256! macros

Robbepop:
Inspired by the METIS library utility macros this PR adds those macros (different implementation and API) to ink! as built-ins.

Link: https://github.com/patractlabs/metis/blob/main/crates/lang/macro/src/lib.rs

Adds the following proc. macros to `ink_lang_macro` crate that are exposed from `ink_lang`:

- `selector_id!`: Takes a string or byte string literal as input and computes the selector ID of type `u32`.
- `selector_bytes!`: Takes a string or byte string literal as input and computes the selector bytes of type `[u8; 4]`.
- `blake2x256!`: Takes a string or byte string literal as input and computes the BLAKE2b 256-bit hash of type `[u8; 32]`.

# ToDo List

- [x] Write UI tests for ...
    - [x] `selector_id!`
    - [x] `selector_bytes!`
    - [x] `blake2x256!`


xgreenx:
- [ ] Maybe will be useful to add a [method](https://github.com/Supercolony-net/openbrush-contracts/blob/cdad79159356a9753c340c8c231570406d4515bb/utils/brush/proc_macros/lib.rs#L371) that will return 'u32' from the hash. It is possible to get from the selector by 'into_be_u32' method. But it is not clear for a new developer:)

Robbepop:
> Maybe will be useful to add a [method](https://github.com/Supercolony-net/openbrush-contracts/blob/cdad79159356a9753c340c8c231570406d4515bb/utils/brush/proc_macros/lib.rs#L371) that will return 'u32' from the hash. It is possible to get from the selector by 'into_be_u32' method. But it is not clear for a new developer:)

From what I understood this is equal to `selector_id!` and therefore duplicate API. I would like to keep the API minimal.

xgreenx:
Oh, I see, you've already returned a 'u32'. I thought that it is 'Selector', but it is 'u32' or '[u8; 4]. Then cool!:)

xgreenx:
Maybe `selector_u32` is more clear than `selector_id`?:)

Robbepop:
> Maybe `selector_u32` is more clear than `selector_id`?:)

Maybe but I thought that `selector_id` is a high level concept as is `Selector` in itself.
So calling it `selector_u32!` would be a half baked solution.
The opposite would be to name it `blake2x256_first4_as_u32` but that is way too concrete.

Robbepop:
Added all the UI tests for the new macros.
Ready for review! :)

xgreenx:
What do you think about `selector_id_u32` and `selector_id_bytes`?:D - Joke:)

crates/lang/codegen/src/generator/selector.rs
```diff
@@ -0,0 +1,55 @@
...
+use derive_more::From;
+use ir::HexLiteral;
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote_spanned;
+
+/// Generates code for the `selector_id!` macro.
+#[derive(From)]
+pub struct SelectorId<'a> {
+    /// The contract to generate code for.
+    macro_input: &'a ir::SelectorMacro<ir::marker::SelectorId>,
```
ascjones, line 25: Can you explain why we require the marker type parameter for `SelectorMacro`?

crates/lang/codegen/src/lib.rs
```diff
@@ -43,6 +43,18 @@ impl<'a> CodeGenerator for &'a ir::ChainExtension {
     type Generator = generator::ChainExtension<'a>;
 }
 
+impl<'a> CodeGenerator for &'a ir::SelectorMacro<ir::marker::SelectorId> {
```
ascjones, line 46: ^^ Oh I see.

crates/lang/macro/tests/ui/selector_id/pass/no_implicit_prelude.rs
```diff
@@ -0,0 +1,5 @@
+#![no_implicit_prelude]
```
ascjones, line 1: TIL

ascjones (approved):
LGTM

crates/lang/codegen/src/generator/selector.rs
Robbepop: this is just to avoid some code dupe because both `SelectorId` and `SelectorBytes` are using the same underlying mechanics.

PR #967, opened 2021-10-19
Upgrade `libsecp256k1` from 0.3.5 ➔ 0.7.0

cmichi:
This is required to fix the failing `cargo-contract` CI (specifically the Windows run).

We are currently using a way outdated version of `libsecp256k1`, upgrading it turned out harder than just switching to the other `no_std` compatible crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 members = [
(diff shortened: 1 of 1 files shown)


Robbepop:
What is the reason for using `libsecp256k1` crate again instead of `secp256k1`?

Issue #949, opened 2021-10-06
Wasm file size reduction: Library Components


Robbepop:
Below is a list of isolated actionable items that may reduce the Wasm file size of certain ink! smart contracts if applied correctly and thoroughly throughout the ink! codebase and codegen.

# Todo List

    - [ ] Retire `ToAccountId` and replace with `AsRef<AccountId>`.
        - Returning a reference instead of a value type is important for `AccountId` types which are usually 256-bits in size
          and therefore too big to be efficiently computed by a Wasm VM.
    - [ ] Make `EnvAccess` API use references as inputs to big environmental types such as `AccountId`.
    - [ ] Make `ink_env::CallBuilder` API use references as inputs to big environmental types such as `AccountId`.

Robbepop:
I began working on:

> Make ink_env::CallBuilder API use references as inputs to big environmental types such as AccountId.

However, it is a bit unfortunate to work on it before #665 has been merged since most of the work for the ink! codegen has to be redone there. I am not done entirely but already see some improvements for the Delegator example contract.

PR #970, opened 2021-10-20
[WIP] Use references instead of values in CallBuilder API

Robbepop:
- [x] `CallBuilder` API
- [ ] `CreateBuilder` API

Seen improvements so far with the Delegator example contract:

| Contract | Before PR | After PR | Reason |
|:--|--:|--:|:--|
| Accumulator | 8.8 kB | 8.8 kB | Expected since it does not use the `CallBuilder` API |
| Adder | 13.0 kB | 12.9 kB | Just minor improvement, need to find out why not more. |
| Subber | 13.0 kB | 12.9 kB | Same as with Adder. |
| Delegator | 10.8 kB | 10.5 kB | Minor improvements in WIP state. Expected to improve further. |

Note, that the numbers do not yet include the full change so they are WIP.
Also I am sure we can do better in some places.

A problem will be how this will work out after #665 has been merged since the ink! codegen will be changed significantly through it and won't allow as easily to be adjusted than with the current codegen (which is unfortunate in this case, however, in general the new codegen is far superior).

### DIFF
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -22,6 +22,7 @@ use crate::{
     },
     call::{
         utils::ReturnType,
+        BalanceEncoder,
         CallParams,
         CreateParams,
     },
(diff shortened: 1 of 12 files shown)


HCastano:
This is pretty stale (it was opened before the new codegen was merged), gonna close it

xgreenx:
But that change still should be implemented because it will decrease significantly the size of contracts and performance with cross-contact calls

HCastano:
It should, but right now there's nobody to implement it. It's still being tracked in #949, so we're not totally forgetting about it

Robbepop:
To give a little bit of background why this staled: The implementation of this experimental PR just got waaaay too ugly which is never a good sign to further follow a road. So I stopped with this road and thought about another way of solving this instead. However, as it happens there were more important tasks to do at that time.

PR #979, opened 2021-10-23
Simple Mapping type improvements

Robbepop:
Based on @HCastano 's https://github.com/paritytech/ink/pull/946 PR.

This fork of the above PR improves the `Mapping` type even further which reduces the ERC-20 example smart contract Wasm file size to 10.8kB so far.

# Implementation Notes

- The ink! codegen does not automatically generate `SpreadAllocate` implementations of its `#[ink(storage)]` struct and at this point it is even undecided if that's a good idea. Therefore we currently are required to implement `SpreadAllocate` manually in the ERC-20 example contract.
- The ERC-20 `new` constructor as well as the `new_init` method seem a bit verbose but we are working on some improvements to make this less of a deal.
- The `EncodeLike` trait usage in the `Mapping` type allows us to use `&(&K1, &K2)` in the `Mapping::{get, insert}` methods instead of just `&(K1, K2)` which saves us a lot of Wasm file size overhead with respect to big value types such as `Balance` and especially `AccountId`.
- To avoid dealing with big value types helper methods have been implemented in the ERC-20 example contract such as `allowance_impl` and `balance_of_impl` that take and forward their arguments by reference instead of by value.

### DIFF
--- a/crates/storage/src/collections/mapping/mod.rs
+++ b/crates/storage/src/collections/mapping/mod.rs
@@ -14,9 +14,61 @@
 
 //! A simple mapping to contract storage.
 
+use crate::traits::{
+    ExtKeyPtr,
+    KeyPtr,
+    PackedLayout,
+    SpreadLayout,
+};
+use ink_env::hash::{
+    Blake2x256,
+    HashOutput,
+};
+use ink_primitives::Key;
+
 use core::marker::PhantomData;
 
 /// A mapping of key-value pairs directly into contract storage.
 ///
 /// If a key does not exist the `Default` value for the `value` will be returned.
-pub struct Mapping<K, V>(PhantomData<K>, PhantomData<V>);
+pub struct Mapping<K, V> {
+    key: Key,
+    mapping: (PhantomData<K>, PhantomData<V>),
+}
+
+impl<K, V> Mapping<K, V>
+where
+    K: scale::Encode,
+    V: scale::Encode + scale::Decode + Default,
+{
+    /// Insert the given `value` to the contract storage.
+    pub fn insert(&mut self, key: K, value: V) {
+        ink_env::set_contract_storage(&self.key(key), &value);
+    }
+
+    /// Get the `value` at `key` from the contract storage.
+    pub fn get(&self, key: K) -> V {
+        ink_env::get_contract_storage(&self.key(key))
+            .unwrap_or_default()
+            .unwrap_or_default()
+    }
+
+    fn key(&self, key: K) -> Key {
+        let encodedable_key = (self.key, key);
+        let mut output = <Blake2x256 as HashOutput>::Type::default();
+        ink_env::hash_encoded::<Blake2x256, _>(&encodedable_key, &mut output);
+        output.into()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn write_to_storage_works() {
+        let mut m1 = Mapping::new();
+        m1.insert("Hello", "World");
+        assert_eq!(m1.get("Hello"), "World");
+    }
+}
--- a/crates/storage/src/collections/mapping/mod.rs
+++ b/crates/storage/src/collections/mapping/mod.rs
@@ -15,15 +15,15 @@
 //! A simple mapping to contract storage.
 
 use crate::traits::{
-    push_packed_root,
-    pull_packed_root_opt,
-    push_spread_root,
-    pull_spread_root,
     clear_spread_root,
+    pull_packed_root_opt,
+    pull_spread_root,
+    push_packed_root,
+    push_spread_root,
     ExtKeyPtr,
     KeyPtr,
-    SpreadLayout,
     PackedLayout,
+    SpreadLayout,
 };
 use ink_env::hash::{
     Blake2x256,
@@ -109,7 +109,8 @@ mod tests {
             assert_eq!(mapping.get(1), 2);
 
             Ok(())
-        }).unwrap()
+        })
+        .unwrap()
     }
 
     #[test]
@@ -119,6 +120,7 @@ mod tests {
             assert_eq!(mapping.get(1), u8::default());
 
             Ok(())
-        }).unwrap()
+        })
+        .unwrap()
     }
 }
--- a/crates/storage/src/collections/mapping/mod.rs
+++ b/crates/storage/src/collections/mapping/mod.rs
@@ -29,6 +29,7 @@ use crate::traits::{
     SpreadLayout,
 };
 use core::marker::PhantomData;
+
 use ink_env::hash::{
     Blake2x256,
     HashOutput,
--- a/crates/storage/src/collections/mapping/mod.rs
+++ b/crates/storage/src/collections/mapping/mod.rs
@@ -70,10 +70,8 @@ where
     #[inline]
     pub fn insert<Q, R>(&mut self, key: &Q, value: &R)
     where
-        K: core::borrow::Borrow<Q>,
-        Q: scale::Encode,
-        V: core::borrow::Borrow<R>,
-        R: PackedLayout,
+        Q: scale::EncodeLike<K>,
+        R: scale::EncodeLike<V> + PackedLayout,
     {
         push_packed_root(value, &self.storage_key(key));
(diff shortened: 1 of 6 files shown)


crates/storage/src/collections/mapping/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+        Q: scale::EncodeLike<K>,
+    {
+        pull_packed_root_opt(&self.storage_key(key))
+    }
+
+    /// Returns a `Key` pointer used internally by the storage API.
+    ///
+    /// This key is a combination of the `Mapping`'s internal `offset_key`
+    /// and the user provided `key`.
+    fn storage_key<Q>(&self, key: Q) -> Key
```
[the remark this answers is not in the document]
xgreenx: I think we don't need to specify the `inline` attribute. Because the compiler will decide it based on the optimization flag(in our case it reduces the size of the code). So compiler better knows what better to inline to reduce the size of the contract=)
https://nnethercote.github.io/perf-book/inlining.html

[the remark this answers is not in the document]
HCastano: Just tried now, and there are no changes to the Wasm size. However, there are also no changes if I remove the `inline`'s from `insert()` and `get()`, so maybe we should get rid of those as well?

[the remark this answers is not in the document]
athei: Not sure if the compiler also takes into account that passing large types (> 64bit) is very expensive in wasm. Ideally, `Q` would be a reference.

crates/storage/src/collections/mapping/mod.rs
```diff
@@ -53,52 +52,49 @@ impl<K, V> core::fmt::Debug for Mapping<K, V> {
...
     /// Insert the given `value` to the contract storage.
-    pub fn insert<Q, R>(&mut self, key: &Q, value: &R)
+    #[inline]
+    pub fn insert<Q, R>(&mut self, key: Q, value: &R)
     where
-        K: core::borrow::Borrow<Q>,
-        Q: scale::Encode,
-        V: core::borrow::Borrow<R>,
-        R: PackedLayout,
+        Q: scale::EncodeLike<K>,
```
HCastano, line 74: Can I think of `EncodeLike` sort-of like `EncodeLike = Borrow + Encode`?

examples/erc20/lib.rs
```diff
@@ -55,24 +60,37 @@ mod erc20 {
     /// The ERC-20 result type.
     pub type Result<T> = core::result::Result<T, Error>;
 
+    impl SpreadAllocate for Erc20 {
```
HCastano, line 63: Should we add a note here that this is intended to only be temporary and will (hopefully) be replaced by #995?

examples/erc20/lib.rs
```diff
@@ -86,15 +104,41 @@ mod erc20 {
...
+        ///
+        /// Returns `0` if no allowance has been set `0`.
+        ///
+        /// # Note
+        ///
+        /// Prefer to call this method over `allowance` since this
+        /// works using references which are more efficient in Wasm.
+        #[inline]
+        fn allowance_impl(&self, owner: &AccountId, spender: &AccountId) -> Balance {
+            self.allowances.get((owner, spender)).unwrap_or_default()
```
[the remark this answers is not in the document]
HCastano: After reading the code more carefully I get why 👍

HCastano (approved):

crates/storage/src/collections/mapping/mod.rs
Robbepop: EncodeLike is a purely marker trait coming from the parity-scale-codec crate. Please go to the crates docs to read more about its general use.

examples/erc20/lib.rs
Robbepop: I think it is better to merge this PR after merging the one PR that improves the SpreadLayout traits.

crates/storage/src/collections/mapping/mod.rs
HCastano: I read the docs, and I think my question is still valid 😄

I think the semantics of a type being able to be SCALE encoded as another type (`EncodeLike`) vs. being able to be borrowed as another type (`Borrow`) are slightly different, which is why I asked

forgetso:
Is there a plan for a function like `get_mut` as part of `Mapping`? How do we update a key value pair - just insert again?

xgreenx:
> Is there a plan for a function like `get_mut` as part of `Mapping`? How do we update a key value pair - just insert again?

Yes, insert again=)

Robbepop:
> Is there a plan for a function like `get_mut` as part of `Mapping`? How do we update a key value pair - just insert again?

A `get_mut` function won't be part of the simple `Mapping` type but it will be available on types that build upon the `Mapping` type internally and provide a more high-level API.

forgetso:
Thanks for the info. Is `iter` deliberately not implemented for `Mapping`? I started converting my contract to `Mapping` but I have methods that previously used `hashmap.keys()`. Now I can't seem to create an iterator for `Mapping` to implement  a `keys` style function.
```
    impl ProviderMapping {
        fn keys(&self) -> Vec<AccountId> {
            let mut keys: Vec<AccountId> = Vec::new();
            for (key, value) in &self.providers.iter() {
                keys.push(key);
            }
            keys
        }
```



```
for (key, value) in &self.providers.iter() {
   |                                                 ^^^^ method not found in `Mapping<ink_env::AccountId, Provider>`
```

Robbepop:
If you come from the Solidity world then `Mapping` is literally just what Solidity provides to you with its own builtin mapping types which are ... just mappings and no real collections that do have knowledge about their contained elements.
So it is not the job of the `Mapping` type to provide collection APIs such as `iter`. This is a very common misunderstanding unfortunately.
If what you need is a collection then you can either build one yourself using the `Mapping` type internally OR use ink!'s `ink_storage::HashMap`.

forgetso:
Cool this is my first time coding smart contracts and I'm coming from a python/JS background, never written in Rust/C. I'll try creating an iterable `Mapping` as per [this example](https://solidity-by-example.org/app/iterable-mapping/) and compare contract size to `Hashmap`. Thanks for the clarification, you probably saved me a few more hours of days of searching.

xgreenx:
> Cool this is my first time coding smart contracts and I'm coming from a python/JS background, never written in Rust/C. I'll try creating an iterable `Mapping` as per [this example](https://solidity-by-example.org/app/iterable-mapping/) and compare contract size to `Hashmap`. Thanks for the clarification, you probably saved me a few more hours or days of searching.

Better to avoid iteration through the map in the smart contracts because it can contain a lot of entries. If you need to resolve some local issue(for example, in your algorithm, you need properties of the `map` to solve some issues. But it is needed only during function execution and after that, you don't need to store that `map` into storage), you can use `ink_prelude::collections::HashMap`. Or if you know that count of entries is not big, then you can store `ink_prelude::collections::HashMap` into the storage(when you will load/save that map you will load/save the whole map with all entries).

xgreenx:
> I have reworked my contract using `Mapping` instead of `Hashmap`, which has reduced the size from `59.8K` to `45K` when building on Supercolony's [reworked-common-part-of-message](https://github.com/Supercolony-net/ink/tree/optimizations/reworked-common-part-of-message) branch. However, I also removed any get functions and replaced them with `Vectors` that contain known keys for `Mappings` - so that will have helped.
> 
> One of the issues I ran into during this process is that `enum` is not supported by `SpreadAllocate`, which seems to be [required as part of the new contract instantiation process](https://github.com/paritytech/ink/pull/995#discussion_r755031713).
> 
> I had to [modify](https://github.com/prosopo-io/ink/commit/bdc676d61804a3fd88f76cb8045238250b3db166) `ink` to allow `enums` to implement `SpreadAllocate`. Hopefully this will be allowed in `ink` in future.
> 
> I also tried and failed to generalise my types that build on `Mapping`, as mentioned [here](https://github.com/paritytech/ink/pull/995#discussion_r756155984).

I tried your contract with [a new version](https://github.com/paritytech/ink/pull/1017) and it takes `Original wasm size: 81.7K, Optimized: 32.4K(32355)`.

Issue #982, opened 2021-10-25
New `trait_definition` fails if to use two traits.


xgreenx:
A new `trait_definition` [generates](https://github.com/paritytech/ink/blob/e8d4739649293a458c3958bec802d2d750067d98/crates/lang/codegen/src/generator/as_dependency/call_builder.rs#L100) `CallBuilder`. But the builder is not unique, so the usage of two traits will generate two same builders. It causes an error during compilation.

What do you think to create one general `CallBuilder` on the ink! level and apply implementation of each trait to this general builder? It allows generating implementation for `CallBuilder` during trait definition instead of implementation block(this feature allows us to support default implementation in traits in the future).

forgetso:
Hi, just chiming in to say this bug fix is necessary for us to use `rc8` `ink!` with OpenBrush so that we can more easily manage our rapidly growing contract. It's reached 2000 lines, including tests, and it would be nice to break it up into smaller chunks!

HCastano:
@forgetso Yep, gonna look into it this week. Can you post a link to your contract, I'm curious to see it

forgetso:
Great! Here is the contract - https://github.com/prosopo-io/protocol/blob/master/contracts/lib.rs. Critique welcome!

cmichi:
Closed with https://github.com/paritytech/ink/pull/1141.

PR #983, opened 2021-10-25
Refactor implementation of call builders for traits

xgreenx:
Before that pull request `#[ink_lang::contract]` generated the implementation of call builders for the contract. That generated code used `ink_lang::codegen::TraitCallForwarderFor<TraitInfoType>` to implement traits defined via `#[ink_lang::trait_definition]`. But if traits are foreign(defined in the another crate) it caused a compilation error that two implementation of `ink_lang::codegen::TraitCallForwarderFor<TraitInfoType1>` and `ink_lang::codegen::TraitCallForwarderFor<TraitInfoType2>` are conflicting. 

The reason the types were conflicting (despite being different) is because of https://github.com/rust-lang/rust/issues/51445. 

The idea of the change is to generate the blanket implementation of the traits for call builders during the definition of the traits(In the same crate where the trait is defined). It will prevent the compilation error because `__ink_TraitInfo` is not foreign.

Closes https://github.com/paritytech/ink/issues/982.


xgreenx:
I implemented my idea from [the comment](https://github.com/paritytech/ink/issues/982#issuecomment-951237096). To show that it works=) It is not a final change, I need validation from @Robbepop and @cmichi. If they agree, I will implement it fully. Also maybe you want to use this approach in several places.

I tested `delegator` example - it works. Also tests with two traits defined in separate crate works.

examples/trait-incrementer/traits/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "increment-traits"
```
HCastano, line 2: Small thing here, but can you make sure that the name of the package and folder are the same? It makes it more intuitive when building contracts with `--manifest-path`.

examples/trait-incrementer/traits/lib.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright 2018-2021 Parity Technologies (UK) Ltd.
```
HCastano, line 1:
```suggestion
// Copyright 2018-2022 Parity Technologies (UK) Ltd.
```

examples/trait-incrementer/traits/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+version = "3.0.0-rc6"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2021"
+
+[dependencies]
+ink_primitives = { version = "3.0.0-rc6", path = "../../../crates/primitives", default-features = false }
```
HCastano, line 8: We should remember to update these to `rc8`

HCastano (commented):
Hey @xgreenx! 

I'm gonna try and get up to speed to this so we can move forward with it. So from what I understand the problem is that right now `ink!` does not support contracts which implement multiple external traits. However, it _does_ support contracts with:
- Multiple internal traits (so defined by the contract itself)
- A single external trait

From the compiler error it looks like this has to do with the structure of the codegen
around the `CallBuilder`. 

I still need to dig into the code, but can you describe (at a high level) how the solution you propose [here](https://github.com/paritytech/ink/issues/982#issuecomment-951237096) with the extra trait implementation fixes this problem?

Also, can you either merge or rebase `master`?

xgreenx:
> I still need to dig into the code, but can you describe (at a high level) how the solution you propose here with the extra trait implementation fixes this problem?

At the moment the code generates the implementation of traits for `CallBuilder` in the body of the contract(`#[ink::contract]`). 

The idea of the suggested fix is to define a generic implementation for `T: ::ink_lang::reflect::ContractEnv + ::ink_lang::ToAccountId<<T as ::ink_lang::reflect::ContractEnv>::Env> + ::ink_lang::reflect::InkStruct` during the trait definition. All these traits will be implemented for `CallBuilder` in the body of the contract -> The `CallBuilder` will use a generic implementation.

So in two words, I suggest implementing traits for `CallBuilder` on the level of trait definition instead of the body of the contract.

> Also, can you either merge or rebase master?

I can do that, but if you are okay with the idea of the change, I can implement it fully(clean up the code, update tests). I did a draft implementation to show that it fixes the problem and still creates a right `CallBuidler`.

HCastano:
> I can do that, but if you are okay with the idea of the change, I can implement it fully(clean up the code, update tests). I did a draft implementation to show that it fixes the problem and still creates a right CallBuidler.

I need to understand the original code before I can assess if your solution should go beyond a draft, so hang tight for now

HCastano:
Okay, so I'm still trying to wrap my head around the root of the problem, maybe you can help me understand.

So again, the compiler complains about _"conflicting implementations of trait `ink_lang::codegen::TraitCallForwarderFor<_>` for type `incrementer::_::CallBuilder`"_.

From the expanded example contract you provided, here are the two places where we try and implement `TraitCallForwarderFor`
- [Increment](https://gist.github.com/HCastano/158cb012dd4c25544320061ddc146f16#file-trait-inc-rs-L1396)
- [Reset](https://gist.github.com/HCastano/158cb012dd4c25544320061ddc146f16#file-trait-inc-rs-L1481)

Since the compiler is complaining, this implies that:

```rust
let T1 = <::ink_lang::reflect::TraitDefinitionRegistry<Environment> as Increment>::__ink_TraitInfo;
let T2 = <::ink_lang::reflect::TraitDefinitionRegistry<Environment> as Reset>::__ink_TraitInfo;
T1 == T2
```

Otherwise the trait implementations would be different.

The definitions for `<TraitDefinitionRegistry as $trait>::__ink_TraitInfo` can be found here:
- [Increment](https://gist.github.com/HCastano/158cb012dd4c25544320061ddc146f16#file-traits-rs-L22)
- [Reset](https://gist.github.com/HCastano/158cb012dd4c25544320061ddc146f16#file-traits-rs-L655)

This leaves us with two different structs, [`__ink_TraitInfoIncrement<E>`](https://gist.github.com/HCastano/158cb012dd4c25544320061ddc146f16#file-traits-rs-L60) and [`__ink_TraitInfoReset<E>`](https://gist.github.com/HCastano/158cb012dd4c25544320061ddc146f16#file-traits-rs-L678).

Since these are different types, the trait implementations should be different. What am I missing here?

xgreenx:
Everything is right. The code generates the implementation for the `Callbuilder`:
```rust
impl ::ink_lang::codegen::TraitCallForwarderFor<#trait_info> for #cb_ident
```

`#trait_info` here is `<::ink_lang::reflect::TraitDefinitionRegistry<Environment> as #trait_path>::__ink_TraitInfo` and it is a problem. If the trait is foreign(defined in external create) then the compiler doesn't know that type exactly in `__ink_TraitInfo`. It can `__ink_TraitInfoIncrement<E>` and `__ink_TraitInfoReset<E>`, because it is external. So it is why these two implementation conflicts with each other. The `__ink_TraitInfo` is foreign and it is limitation of the rust.

When two traits are defined in the crate of the contract then `__ink_TraitInfo` is a local type and compilers know all possible implementations.

HCastano:
Okay, so I was pretty stumped here and luckily @ascjones came to the rescue identifying this is https://github.com/rust-lang/rust/issues/51445.

So the first fix that comes to mind for me is to have the macro extract the concrete type behind `__ink_TraitInfo` and use that instead. This should be a lot less code. Andrew do you have any other suggestions here?

xgreenx:
> So the first fix that comes to mind for me is to have the macro extract the concrete type behind __ink_TraitInfo and use that instead

Could you elaborate more on it? Because I didn't get it=)

HCastano:
> > So the first fix that comes to mind for me is to have the macro extract the concrete type behind __ink_TraitInfo and use that instead
> 
> Could you elaborate more on it? Because I didn't get it=)

So instead of the macro generating `<Foo as Bar>::__ink_TraitInfo` we get it to generate `__ink_TraitInfoIncrement<E>` instead (the issue is around the use of the full syntax). It's kinda like we're hardcoding the type there, haha 😅

xgreenx:
> So instead of the macro generating <Foo as Bar>::__ink_TraitInfo we get it to generate __ink_TraitInfoIncrement<E> instead (the issue is around the use of the full syntax). It's kinda like we're hardcoding the type there, haha 😅

The problem is that `__ink_TraitInfoIncrement<E>` is a **new** generated type and we don't know the exact path to that type -> The user should import that type somehow or we need to get the path to that type, to be able to use it in the `#[ink_lang::contract]`.

It is why it was initially moved to `<Foo as Bar>::__ink_TraitInfo`. During the generation of the contract, we have access to the trait(because the user imported it) -> we have access to the type alias `__ink_TraitInfo`.

Your idea requires the user to import an additional type or explicitly write a path to the trait(in that case we can use the same prefix for the type like `::path::Increment` -> `::path::__ink_TraitInfoIncrement`).

crates/lang/src/reflect/trait_def/info.rs
```diff
@@ -152,3 +152,6 @@ pub trait TraitModulePath {
     /// This is just for convenience.
     const NAME: &'static str;
 }
+
+/// Used to identify ink! structs.
+pub trait InkStruct {}
```
HCastano, line 157: So I get what you're trying to do with this, but it really isn't used to identify ink! structs. It's really just a marker for `CallBuilder`.

If we're going to stick with the blanket impl approach I'd change this to be something like:

```rust
pub trait CallBuilder: ContractEnv + ToAccountId {}
```

HCastano (commented):
Okay, so I get how you're trying to solve this. I'm not a huge fan of the blanket impl, but I get why you're doing it. If possible, I'd look for ways to tighten the scope of the implementation up a bit.

Another thing I noticed, there's no more implementation of `TraitCallForwarderFor` for `CallBuilder`. I wonder if that's going to have any unintended consequences.

Anyways, I'm okay moving forward with this approach, so feel free to merge master and clean it up once you get a chance

crates/lang/src/reflect/trait_def/info.rs
xgreenx: Yea, that is good idea=)

paritytech-cicd-pr:

| `contract-terminate` |                  | +14_926    | 1.31 K               | 215_704        |
| `contract-transfer`  |                  | +14_305    | 7.98 K               | 15_704         |
| `delegator`          |                  | +41_047    | 7.63 K               | 51_343         |
| `dns`                |                  | +31_959    | 10.20 K              | 47_112         |
| `erc1155`            |                  | +46_413    | 27.50 K              | 94_224         |
| `erc20`              |                  | +37_098    | 9.80 K               | 47_112         |
| `erc721`             |                  | +90_930    | 13.70 K              | 125_632        |
| `flipper`            |                  | +15_357    | 1.76 K               | 15_704         |
| `incrementer`        |                  | +15_364    | 1.64 K               | 15_704         |
| `multisig`           |                  | +46_824    | 26.79 K              | 103_580        |
| `proxy`              |                  | +29_614    | 3.81 K               | 32_243         |
| `rand-extension`     |                  | +9_050     | 5.12 K               | 15_704         |
| `trait-erc20`        |                  | +36_453    | 10.10 K              | 47_112         |
| `trait-flipper`      |                  | +15_362    | 1.53 K               | 15_704         |
| `trait-incrementer`  |                  | +30_728    | 1.62 K               | 31_408         |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/178327) | Last update: Tue Feb  8 17:13:00 CET 2022

crates/lang/src/reflect/trait_def/registry.rs
```diff
@@ -61,3 +67,30 @@ where
...
+{
+    #[cold]
+    fn to_account_id(&self) -> E::AccountId {
+        /// We enforce linking errors in case this is ever actually called.
+        /// These linker errors are properly resolved by the cargo-contract tool.
+        extern "C" {
+            fn _ink_compilation_error() -> !;
+        }
+        unsafe { _ink_compilation_error() }
+    }
```
HCastano, line 83: Is this still needed?

crates/lang/src/reflect/trait_def/info.rs
```diff
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::{
+    reflect::ContractEnv,
+    ToAccountId,
+};
+
```
HCastano, line 19: These are only used once, might as well just inline them at the call site

HCastano (commented):
Couple of questions, but I think it looks alright

crates/lang/codegen/src/enforced_error.rs
```diff
@@ -1,68 +0,0 @@
...
-use impl_serde::serialize as serde_hex;
-use quote::format_ident;
-
-/// Errors which may occur when forwarding a call is not allowed.
-///
-/// We insert markers for these errors in the generated contract code.
-/// This is necessary since we can't check these errors at compile time
-/// of the contract.
-/// `cargo-contract` checks the contract code for these error markers
-/// when building a contract and fails if it finds markers.
```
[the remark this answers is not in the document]
xgreenx: It is used in the implementation of the `to+account_id` for `TraitDefinitionRegistry`. So I moved that comment there=)

crates/lang/src/reflect/trait_def/registry.rs
xgreenx: Yep, if someone will try to call `to_account_id` from `TraitDefinitionRegistry` we want to throw an error during compilation

crates/lang/src/reflect/trait_def/info.rs
xgreenx: It will looks ugly=)

```rust

pub trait CallBuilder:
    crate::reflect::ContractEnv
    + crate::ToAccountId<<Self as crate::reflect::ContractEnv>::Env>
{
}
```

crates/lang/src/reflect/trait_def/registry.rs
```diff
@@ -61,3 +67,37 @@ where
...
+    /// when building a contract and fails if it finds markers.
+    #[cold]
+    fn to_account_id(&self) -> E::AccountId {
+        /// We enforce linking errors in case this is ever actually called.
+        /// These linker errors are properly resolved by the cargo-contract tool.
+        extern "C" {
+            fn __ink_enforce_error_to_account_id() -> !;
+        }
+        unsafe { __ink_enforce_error_to_account_id() }
+    }
```
HCastano, line 90: So I'm still trying to understand why this is needed.

Previously these were only in the actual method code, not the `ToAccountId` implementation. E.g, `<TraitDefinitionRegistry as $trait>::$trait_method()`. This was to uphold the [following guarantee from the trait](https://github.com/paritytech/ink/blob/1188ee71284dc1e6b315311561f916fabd668b09/crates/lang/src/reflect/trait_def/registry.rs#L27):

> ///   `TraitDefinitionRegistry` with stub implementations for all methods that
> ///   guarantee that they are never called.

Looking at your changes it looks like we're now going through a `...to_account_id().fire()` sequence in the trait method implementations, so is this here as a way to continue to uphold that?

crates/lang/src/reflect/trait_def/info.rs
```diff
@@ -152,3 +157,11 @@ pub trait TraitModulePath {
     /// This is just for convenience.
     const NAME: &'static str;
 }
+
+/// Implemented by call builders of smart contracts and by `TraitDefinitionRegistry`
+/// during codegen to apply the blanket implementation.
+///
+/// The blanket implementation is generated by `#[ink_lang::trait_definition]` and
+/// defined in the same place with the trait. It generates the body for methods of the trait.
+/// These methods contain cross-contract calls inside and define the `__ink_TraitInfo`.
```
HCastano, line 166: This has too many low level details about the implementation. I'd make the doc comments higher level and maybe write about the implementation details at a more relevant location

HCastano (commented):
Still trying to clarify things around the `__ink_enforce_error` thing

HCastano:
Oh, also a couple of thoughs.

- Do you mind updating the PR name and description to explain what this change is and why its important?
- Can you split out your commits a bit more, makes reviewing a bit easier
- Can you avoid force-pushes if possible? Slightly streamlines pulling new changes down. We also squash the PRs anyways, so the commit history doesn't have to be super clean

examples/delegator/accumulator/lib.rs
```diff
@@ -3,12 +3,25 @@
...
+    pub trait AccumulatorTrait {
+        /// Mutates the internal value.
+        #[ink(message)]
+        fn inc(&mut self, by: i32);
+
+        /// Returns the current state.
+        #[ink(message)]
+        fn value(&self) -> i32;
+    }
+
```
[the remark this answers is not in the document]
xgreenx: I moved those methods into the trait to test the call builders are generated correctly and it didn't affect the size.

crates/lang/src/reflect/trait_def/registry.rs
xgreenx: Previously we generated implementation of trait for `TraitDefinitionRegistry` only to store `__ink_TraitInfo`. It is why methods were stubbed with that error. The idea is that no one calls these methods.

In the current change that implementation is inherited automatically for `TraitDefinitionRegistry` because we added the blanket implementation. But methods are not empty(stubbed), those contain code for cross-contract calls that are using `to_account_id`.

I added an error into `to_account_id` to prevent the calling of methods from `TraitDefinitionRegistry`. 
`<TraitDefinitionRegistry as $trait>::$trait_method()` is using `...to_account_id().fire()` inside and it will cause compilation error like before.

crates/lang/codegen/src/generator/trait_def/trait_registry.rs
```diff
@@ -100,142 +95,31 @@ impl TraitRegistry<'_> {
...
-        let name = self.trait_ident();
+        let trait_ident = self.trait_ident();
         let trait_info_ident = self.trait_def.trait_info_ident();
-        let messages = self.generate_registry_messages();
+        let message_impls = self.generate_ink_trait_impl_messages();
         quote_spanned!(span=>
-            impl<E> #name for ::ink_lang::reflect::TraitDefinitionRegistry<E>
+            /// The blanket implementation of the forwarder to do cross-contract
+            /// calls without customization. It only fires the call via SEAL host function
+            /// and checks that dispatching result is not error.
```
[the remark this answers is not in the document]
xgreenx: It is an internal comment to describe what is going on and why we need blanket implementation. So it is why contains some information related to implementation. We have the same comments in other places where it is important

For example [here](https://github.com/paritytech/ink/blob/master/crates/lang/codegen/src/generator/trait_def/call_forwarder.rs#L314) or [here](https://github.com/paritytech/ink/blob/master/crates/lang/codegen/src/generator/trait_def/call_builder.rs#L306).

crates/lang/src/reflect/trait_def/info.rs
xgreenx: The same idea here. It is a trait for codegen and it described how it works inside. If we will remove the blanket implementation we also will remove that trait.

xgreenx:
> Oh, also a couple of thoughs.
> 
> * Do you mind updating the PR name and description to explain what this change is and why its important?
> * Can you split out your commits a bit more, makes reviewing a bit easier
> * Can you avoid force-pushes if possible? Slightly streamlines pulling new changes down. We also squash the PRs anyways, so the commit history doesn't have to be super clean

- Done
- It is hard to do because I initially did it like one commit. If something is not clear you can ask=)
- I'm force-pushing because I'm rebasing it to be actual with `master`=)

crates/lang/codegen/src/generator/trait_def/trait_registry.rs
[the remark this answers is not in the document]
HCastano: Alright fair enough.

```suggestion
            /// The blanket implementation of the forwarder to do cross-contract
            /// calls without customization. It only fires the call via a SEAL host function
            /// and checks that dispatch result is not an error.
```

crates/lang/src/reflect/trait_def/registry.rs
HCastano: Okay yeah, we're on the same page then.

crates/lang/src/reflect/trait_def/registry.rs
```diff
@@ -61,3 +67,37 @@ where
...
+where
+    E: ink_env::Environment,
+{
+    /// `to_account_id` is not allowed for `TraitDefinitionRegistry`.
+    ///
+    /// We insert markers for these errors in the generated contract code.
+    /// This is necessary since we can't check these errors at compile time
+    /// of the contract.
+    /// `cargo-contract` checks the contract code for these error markers
+    /// when building a contract and fails if it finds markers.
```
HCastano, line 81:
```suggestion
    /// The trait methods which are implemented on the `TraitDefinitionRegistry` should not be
    /// called directly. Instead, users should opt to go through the `__ink_TraitInfo` associated
    /// type instead.
    ///
    /// In order to enforce this we insert these error markings in the generated contract code. The
    /// `cargo-contract` tool will check for these markers during its build step and fail if they
    /// are found.
    ///
    /// We do this since we can't check for these errors at compile time. 
```

crates/lang/src/reflect/trait_def/registry.rs
```diff
@@ -61,3 +67,37 @@ where
...
+    #[cold]
+    fn to_account_id(&self) -> E::AccountId {
+        /// We enforce linking errors in case this is ever actually called.
+        /// These linker errors are properly resolved by the cargo-contract tool.
```
HCastano, line 85:
```suggestion
```

examples/trait-incrementer/traits/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#![cfg_attr(not(feature = "std"), no_std)]
+
+/// Traits are extracted into a separate crate to show how the user can import
+/// several foreign traits and implement those for the contract.
+use ink_lang as ink;
```
HCastano, line 18:
```suggestion
//! Traits are extracted into a separate crate to show how the user can import
//! several foreign traits and implement those for the contract.

use ink_lang as ink;
```

crates/lang/src/reflect/trait_def/registry.rs
```diff
@@ -61,3 +67,37 @@ where
...
+/// Generates the global trait registry implementation for the ink! trait.
+///
+/// This makes it possible to refer back to the global call forwarder and
+/// call builder specific to this ink! trait from anywhere with just the Rust
+/// trait identifier which allows for type safe access.
+///
+/// # Note
+///
+/// The actual implementation of the registration is done in the blanket implementation
+/// during definition of the trait.
```
HCastano, line 102:
```suggestion
/// during definition of the ink! trait.
```

crates/lang/src/reflect/trait_def/registry.rs
HCastano: Can you also update this doc comment on the `TraitDefinitionRegistry` struct?

> /// - Furthermore the ink! codegen implements the ink! trait definition for the
> ///   `TraitDefinitionRegistry` with stub implementations for all methods that
> /// guarantee that they are never called.

HCastano (commented):
Some small doc fixes, but I think this is good to go 😄

crates/lang/src/reflect/trait_def/registry.rs
xgreenx: I think that comment is still true because the implementation of these methods is really stubbed. Calling these methods will cause a compilation error.

HCastano: Maybe it's pedantic, but they're not stubs in my opinion. The only stub is `to_account_id()`

HCastano (approved):
Looks good, thanks!

Gonna see if we can get another quick look (@ascjones 😃?) before merging

ascjones:
> Looks good, thanks!
> 
> Gonna see if we can get another quick look (@ascjones smiley?) before merging

Yes I am looking at this today!

ascjones (commented):
The main issue I have is that this is quite a significant refactoring to solve this particular issue. To this end I explored a few alternatives to see whether it is possible to fix it with a much smaller change. The one I came up with is https://github.com/paritytech/ink/pull/1136. Please have a look and let me know what you think of that. There may be some issues with it that I have not considered.

TriplEight:
@xgreenx how CI team can help in this review?

cmichi:
@TriplEight The CI team was requested automatically for review, since you added yourself as code owners a while ago. I think there were some temporary changes to the GitLab config to comment some stages out, but now no longer. I'm removing the request for you to review.

ascjones:
@xgreenx  #1141 is merged now, which fixes the conflicting impls issue which I understand was the main motivator for this change.

However you did [say](https://github.com/paritytech/ink/pull/1136#issuecomment-1043110853) that there were other motivations for this refactor which aren't mentioned here:

> The idea of my change also was to reduce the redundant of the code. We have 3 places with the same code that can be covered with blanked implementation. + We don't need an introduction of the utility trait TraitCallForwarderFor.
Also, it was preparation for future change that we want to suggest, but we can refactor it there=)
Also, we planned to use that blanked implementation in OpenBrush to not generate the same code at the fourth time.

So up to you how you want to proceed - I suggest to close this and open a fresh PR when the blanket implementation is actually required for the future change, and we can review it in the context of that and removing code redundancy.

cmichi:
@xgreenx Thanks a lot for finding the bug in the first place, providing a solution and partaking in all the discussions on how to fix it! Even though we ended up closing your PR with the first approach to fix the bug, we wouldn't have gotten to the final solution without you and the discussions we had with you on possible solutions.

We want to give you a tip for all of that. Thank you!

```
```
or
```
```

PR #984, opened 2021-10-26
Fixed `is_result_type` and `is_result_err` for generics

xgreenx:
I tried different construction to resolve the issue https://github.com/paritytech/ink/issues/985 but seems it is not possible to resolve it with default rust. Rust doesn't use inherent implementation. Maybe it is a bug in the rust and we need to create an issue.

As a solution, we can use `specialization` feature. It resolves the problem and it is clear how it works=)


xgreenx:
https://github.com/paritytech/ink/pull/998 fixes the problem, so I close this PR.

Issue #985, opened 2021-10-26
Feature `revert_on_error` doesn't work


xgreenx:
`is_result_type` returns `false` for `Output`. So during execution, the contract will be not reverted. https://github.com/paritytech/ink/pull/984 adds two tests to reproduce the problem and also provides a possible solution.

PR #998, opened 2021-11-01
Fix ink! message result return reverts

Robbepop:
Fixes https://github.com/paritytech/ink/issues/985.

@xgreenx Please check if this fixes the problems for you.

Not yet optimized. Will do once we know for sure that this fix actually works.

### DIFF
--- a/crates/lang/codegen/src/generator/dispatch.rs
+++ b/crates/lang/codegen/src/generator/dispatch.rs
@@ -677,7 +677,10 @@ impl Dispatch<'_> {
                         mutates: #mutates_storage,
                         dynamic_storage_alloc: #is_dynamic_storage_allocation_enabled,
                     };
-                    let mut contract: #storage_ident = ::ink_lang::codegen::initiate_message::<#storage_ident>(config)?;
+                    let mut contract: ::core::mem::ManuallyDrop<#storage_ident> =
+                        ::core::mem::ManuallyDrop::new(
+                            ::ink_lang::codegen::initiate_message::<#storage_ident>(config)?
+                        );
                     let result: #message_output = #message_callable(&mut contract, input);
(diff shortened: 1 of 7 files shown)


crates/lang/codegen/src/generator/dispatch.rs
```diff
@@ -672,18 +672,25 @@ impl Dispatch<'_> {
...
+                        mutates: #mutates_storage,
+                        dynamic_storage_alloc: #is_dynamic_storage_allocation_enabled,
+                    };
+                    let mut contract: ::core::mem::ManuallyDrop<#storage_ident> =
+                        ::core::mem::ManuallyDrop::new(
+                            ::ink_lang::codegen::initiate_message::<#storage_ident>(config)?
+                        );
+                    let result: #message_output = #message_callable(&mut contract, input);
+                    let failure = ::ink_lang::is_result_type!(#message_output)
+                        && ::ink_lang::is_result_err!(&result);
```
xgreenx, line 686:
```suggestion
                        && ::ink_lang::is_result_err!(result);
```

xgreenx (commented):
It works=)

xgreenx:
I tested with Europa and you can see that `flags: 1`. So it is reverted
image

crates/lang/codegen/src/generator/dispatch.rs
Robbepop: Thank you for the debugging help!

Robbepop:
On the downside of this PR is that it currently slightly increases the Wasm file sizes of some ink! smart contracts. E.g. ERC-20 raised from 28.9kB to 31.4kB in size. We might be able to find some optimizations to reduce this additional bloat.

cmichi (approved):

PR #1003, opened 2021-11-04
Wrap Comments

HCastano:
While looking at #979 I noticed that Robin (kinda) wrapped my module comments. I wondered
if we had any `RustFmt` config around that, and it turns out that while we did
(`comment_width = 80`) we were not actual using it since we had `wrap_comments = false`.
According to the [`comment_width` docs](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#comment_width) we need to set it to `true` in order for it to have an effect.


HCastano:
Looks like we'll have to manually fix some code examples and tables with this change, so maybe I'll leave this for another time...

Robbepop:
I am against enabling this rustfmt rule since it uglifies a lot of comments imo. i tried it and it severely did not satisfy me.

PR #1010, opened 2021-11-09
Specify minor version for `parity-scale-codec` in TOML files

HCastano:
I just spent some time trying to figure out why my UI tests were failing locally but
passing on the CI. I realized that it was because my local repo was using
`parity-scale-codec` `v2.1.1`. This was totally file as far as Cargo was concerned since
we only specified `2` in our TOML files. However, for our test this was not fine, and
`v2.3.1` was required. Since `ink!` doesn't have a Lockfile I figured I'd manually bump
the required version ([`cargo-upgrade`](https://github.com/killercup/cargo-edit#cargo-upgrade) made this a breeze FYI).

### DIFF
--- a/crates/engine/Cargo.toml
+++ b/crates/engine/Cargo.toml
@@ -15,7 +15,7 @@ categories = ["no-std", "embedded"]
 include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 
 [dependencies]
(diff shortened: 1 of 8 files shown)


cmichi (approved):
Ah yeah, that's an unfortunate thing about this ui test fixture.

PR #1012, opened 2021-11-10
Remove `always` from `inline` to allow compiler decide that to do

xgreenx:
It's funny, but removing `always` with combination https://github.com/paritytech/cargo-contract/pull/358, reduces the size of `Erc20` contract in `master` from  31.4K to 27.4K

### DIFF
--- a/crates/env/src/chain_extension.rs
+++ b/crates/env/src/chain_extension.rs
@@ -83,7 +83,7 @@ pub struct ChainExtensionMethod<I, O, ErrorCode> {
 
 impl ChainExtensionMethod<(), (), ()> {
     /// Creates a new chain extension method instance.
-    #[inline(always)]
+    #[inline]
     pub fn build(func_id: u32) -> Self {
         Self {
             func_id,
@@ -100,7 +100,7 @@ impl<O, ErrorCode> ChainExtensionMethod<(), O, ErrorCode> {
     /// `I` represents the input type of the chain extension method.
     /// All tuple types that may act as input parameters for the chain extension method are valid.
     /// Examples include `()`, `i32`, `(u8, [u8; 5], i32)`, etc.
-    #[inline(always)]
+    #[inline]
     pub fn input<I>(self) -> ChainExtensionMethod<I, O, ErrorCode>
     where
         I: scale::Encode,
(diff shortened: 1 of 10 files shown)


HCastano:
This looks promising! We should wait until the `ink-waterfall` pipeline is fixed though, since there we can see how this PR changes the sizes of all the example contracts instead of just the ERC-20 one.

cmichi (approved):
Very nice discovery!

HCastano (approved):
So here are the changes in size of our example contracts compared to `master`. Note that all sizes are in KB, and `-` means that it is smaller with this PR applied.

```
contract-terminate,0.0,0.0
rand-extension,-0.055000305,-0.12400007
incrementer,0.035999298,-0.06700015
erc1155,-63.791008,-62.012
trait-erc20,-2.6740036,-2.7460003
dns,-0.10300064,-0.22099876
delegator,0.38000107,0.15100002
multisig,-14.578995,-13.453999
subber,0.20700073,0.14900017
flipper,0.069000244,-0.03699994
erc20,-2.675003,-2.7460003
trait-flipper,0.024999619,-0.06700003
contract-transfer,0.0,0.0
trait-incrementer,-0.20399857,-0.32500005
erc721,-3.8440018,-3.9720001
adder,0.20700073,0.14900017
accumulator,0.20699883,0.14900017
```

Notice the dramatic size reduction to the ERC-1155 example 🤯. We should really do some digging and figure out why this "simple" change has produced such a great improvement there.

cmichi:
@xgreenx We want to give you a tip for all the great stuff you have contributed now and in the past. We're happy to have you as part of our ecosystem!

```
```
or
```
```

xgreenx:
> @xgreenx We want to give you a tip for all the great stuff you have contributed now and in the past. We're happy to have you as part of our ecosystem!

Glad to hear it=) Thank you, it is [part of our grant](https://github.com/w3f/Grants-Program/blob/master/applications/openbrush-follow-up.md#milestone-3-reduce-the-size-of-erc20-contract), so you don't need to worry about it(=

> So here are the changes in size of our example contracts compared to master. Note that all sizes are in KB, and - means that it is smaller with this PR applied.

Hmm, contracts that use `builder` have a bad impact(`delegator,0.38000107,0.15100002`), maybe we should keep inlining for them. I will try it.

xgreenx:
> Hmm, contracts that use `builder` have a bad impact(`delegator,0.38000107,0.15100002`), maybe we should keep inlining for them. I will try it.

I tried to return `#[inline(always)]` back in some places to reduce the size of the `delegator` back. Seems that https://github.com/paritytech/cargo-contract/pull/358 should cover that case. Because the size is the same with `#[inline(always)]` and `#[inline]` if I build with https://github.com/paritytech/cargo-contract/pull/358. So I think we can merge it=)

PR #1016, opened 2021-11-15
Optimized decode of AccountId

xgreenx:
Default decode takes ~300 bytes

### DIFF
--- a/crates/env/src/types.rs
+++ b/crates/env/src/types.rs
@@ -268,21 +268,18 @@ impl<'a> TryFrom<&'a [u8]> for AccountId {
 /// This is a mirror of the `Hash` type used in the default configuration
 /// of PALLET contracts.
 #[derive(
-    Debug,
-    Copy,
-    Clone,
-    PartialEq,
-    Eq,
-    Ord,
-    PartialOrd,
-    Hash,
-    Encode,
-    Decode,
-    From,
-    Default,
+    Debug, Copy, Clone, PartialEq, Eq, Ord, PartialOrd, Hash, Encode, From, Default,
 )]
 #[cfg_attr(feature = "std", derive(TypeInfo))]
 pub struct Hash([u8; 32]);
+
+impl Decode for Hash {
+    fn decode<I: scale::Input>(input: &mut I) -> Result<Self, scale::Error> {
+        let mut hash = Hash { 0: [0; 32] };
+        input.read(hash.0.as_mut_slice())?;
+        Ok(hash)
+    }
+}
 
 impl<'a> TryFrom<&'a [u8]> for Hash {
     type Error = TryFromSliceError;
--- a/crates/env/src/types.rs
+++ b/crates/env/src/types.rs
@@ -211,21 +211,18 @@ pub type RentFraction = Perbill;
 /// This is a mirror of the `AccountId` type used in the default configuration
 /// of PALLET contracts.
 #[derive(
-    Debug,
-    Copy,
-    Clone,
-    PartialEq,
-    Eq,
-    Ord,
-    PartialOrd,
-    Hash,
-    Encode,
-    Decode,
-    From,
-    Default,
+    Debug, Copy, Clone, PartialEq, Eq, Ord, PartialOrd, Hash, Encode, From, Default,
 )]
 #[cfg_attr(feature = "std", derive(TypeInfo))]
 pub struct AccountId([u8; 32]);
+
+impl Decode for AccountId {
+    fn decode<I: scale::Input>(input: &mut I) -> Result<Self, scale::Error> {
+        let mut account_id = AccountId { 0: [0; 32] };
+        input.read(account_id.0.as_mut_slice())?;
+        Ok(account_id)
+    }
+}
 
 impl AsRef<[u8; 32]> for AccountId {
     #[inline]
--- a/crates/primitives/src/key.rs
+++ b/crates/primitives/src/key.rs
@@ -179,8 +179,9 @@ impl scale::Decode for Key {
     where
         I: scale::Input,
     {
-        let bytes = <[u8; 32] as scale::Decode>::decode(input)?;
-        Ok(Self::from(bytes))
+        let mut key = Key { 0: [0; 32] };
+        input.read(key.0.as_mut_slice())?;
+        Ok(key)
     }
 
     #[inline]


Robbepop:
> Default decode takes ~300 bytes

Can you tell us the effect on some examples for this? It would be interesting to know why the default is so bad here. Also we could apply this to Hash and Key types as well.

xgreenx:
Yea, sorry. thought that `ink-waterfall` will show that=)

Before change:
Original wasm size: 34.3K, Optimized: 10.6K
-rw-r--r--  1 501  20  10632 Nov 15 22:18 target/ink/erc20.wasm

After change:
Original wasm size: 33.6K, Optimized: 10.2K
-rw-r--r--  1 501  20  10242 Nov 15 22:16 target/ink/erc20.wasm

cmichi:
>Yea, sorry. thought that `ink-waterfall` will show that=)

We'll merge https://github.com/paritytech/ink-waterfall/pull/16 hopefully today, then it will :-).

xgreenx:
> > Yea, sorry. thought that `ink-waterfall` will show that=)
> 
> We'll merge [paritytech/ink-waterfall#16](https://github.com/paritytech/ink-waterfall/pull/16) hopefully today, then it will :-).

How to trigger the report?=)

cmichi:
> How to trigger the report?=)

We just merged the PR, it's triggered automatically now with every run of the `ink-waterfall` CI stage, so every ink! PR where the CI manages to get there (it's one of the last steps).

cmichi:
@xgreenx Btw. the above report already includes the changes you made in https://github.com/paritytech/cargo-contract/pull/358.

xgreenx:
> > How to trigger the report?=)
> 
> We just merged the PR, it's triggered automatically now with every run of the `ink-waterfall` CI stage in ink! ‒ so every ink! PR where the CI manages to get there (it's one of the last stages).

Last stage... Maybe it is possible to run manually for some PR(for example https://github.com/paritytech/ink/pull/1017)?=D

cmichi:
> Last stage... Maybe it is possible to run manually for some PR(for example #1017 )?=D

Not trivially, I guess it's the reward you get once you've fixed all the CI stages before it ;-). But is there anything preventing you from fixing the CI stages at that PR?

xgreenx:
> > Last stage... Maybe it is possible to run manually for some PR(for example #1017 )?=D
> 
> Not trivially, I guess it's the reward you get once you've fixed all the CI stages before it ;-). But is there anything preventing you from fixing the CI stages at that PR?

I'm not sure about the final state of the code, so I'm waiting for some input from you and @Robbepop=) Because it requires removing some logic added before and maybe you have your vision of the code. I described the main idea and results that I got=)

crates/env/src/types.rs
```diff
@@ -211,22 +211,19 @@ pub type RentFraction = Perbill;
...
-    Decode,
-    From,
-    Default,
+    Debug, Copy, Clone, PartialEq, Eq, Ord, PartialOrd, Hash, Encode, From, Default,
 )]
 #[cfg_attr(feature = "std", derive(TypeInfo))]
 pub struct AccountId([u8; 32]);
 
+impl Decode for AccountId {
+    fn decode<I: scale::Input>(input: &mut I) -> Result<Self, scale::Error> {
```
Robbepop, line 220: what happens to the benchmarks if you add #[inline] here?

Robbepop (commented):
I find it strange that the Decode derive impl and the impls provided by this PR are not optimized to same or similar IR. It would be nice to know why. Some guesses: Special treatment for Copy types, no guaranteed RVO, no guaranteed copy elision etc.

xgreenx:
> I find it strange that the Decode derive impl and the impls provided by this PR are not optimized to same or similar IR. It would be nice to know why. Some guesses: Special treatment for Copy types, no guaranteed RVO, no guaranteed copy elision etc.

The default implementation of `Decode` for the generic array is complex. In wasm, it takes ~360 bytes. I found it during analyzing of objdumped contract and decide to create this change=)
image

crates/env/src/types.rs
xgreenx: I didn't get any effect=) If you want I can try to put it here, but how we will see the impact?

Funny moment, with this PR https://github.com/paritytech/ink/pull/1017, this part of code requires `#[inline(never)]`.
Without inline `Erc20` takes 9.2KB, with `#[inline(never)]` it takes 8.8KB

Robbepop:
> > I find it strange that the Decode derive impl and the impls provided by this PR are not optimized to same or similar IR. It would be nice to know why. Some guesses: Special treatment for Copy types, no guaranteed RVO, no guaranteed copy elision etc.
> 
> The default implementation of `Decode` for the generic array is complex. In wasm, it takes ~360 bytes. I found it during analyzing of objdumped contract and decide to create this change=)
> image
> 

Looks like we should rather fix and optimise the generic array implementation than all of its dependents.

xgreenx:
> Looks like we should rather fix and optimise the generic array implementation than all of its dependents. We could use MaybeUninit or even array::map for this. Shouldnt be too hard.

I agree that better to do a fix for the general case. But it means that general implementation will be not so general because it requires some restrictions like `Copy` trait, or `Default` or `Size` or another=) And it can affect someone who uses `parity-scale-codec`. So this decision is up to you, I only highlighted that can be optimized=D

Robbepop:
> > Looks like we should rather fix and optimise the generic array implementation than all of its dependents. We could use MaybeUninit or even array::map for this. Shouldnt be too hard.
> 
> I agree that better to do a fix for the general case. But it means that general implementation will be not so general because it requires some restrictions like `Copy` trait, or `Default` or `Size` or another=) And it can affect someone who uses `parity-scale-codec`. So this decision is up to you(Parity team), I only highlighted that can be optimized=D

If MaybeUninit is used no further restrictions need to be put in place. There are some crates for safe and efficient array initialization.

xgreenx:
> I will try it. If everything is okay, then I will create a pull request=)
> 
> BTW, if it is not hard then it's strange why it is not implemented from the beginning)

I [implemented idea](https://github.com/paritytech/parity-scale-codec/compare/master...Supercolony-net:optimization/not-complex-array?expand=1) with `MaybeUninit`. After that change the size of `Erc20` decreased:
Original wasm size: 34.0K, Optimized: 10.4K(10389)

But a native implementation(from this PR) still is better=) Original wasm size: 33.5K, Optimized: 10.2K(10163)

If you agree to use usable feature(like [negative-impls](https://doc.rust-lang.org/beta/unstable-book/language-features/negative-impls.html) or [specialization](https://rust-lang.github.io/rfcs/1210-impl-specialization.html)), we can specify the implementation for `u8`=D

xgreenx:
This change is implemented on `parity-scale-codec` level=) https://github.com/paritytech/parity-scale-codec/pull/299

Issue #989, opened 2021-10-29
Support ink! messages and constructors that take references as parameters


Robbepop:
Currently ink! messages and constructors are required to be `Sized` and `scale::Decode`.
This means that ink! messages and constructors that take simple references are not allowed, e.g.:
```rust
#[ink(message)]
fn transfer(&mut self, from: &AccountId, to: &AccountId, amount: &Balance) -> bool;
```
This is a downside with respect to Wasm file sizes since in Wasm references are 32-bit sizes while types such as `Balance` and `AccountId` have massive sizes that force Wasm to costly put them on the shadow stack. This produces badly performing Wasm code as well as bloats the resulting Wasm binary.

In order to counteract this slightly users currently can add private methods that take their arguments by references where the ink! messages and constructors immediately forward to. This has the downside of mostly repeating the APIs over and over again in those common cases.

For ink! we should do better and allow ink! messages and constructors to take their arguments as references.
This could work by making ink! aware of input reference parameters and adjust ink! codegen in a way that the ink! message and constructor decoders forward their arguments in `ExecuteMessage` and `ExecuteConstructor` by reference using the `ref` Rust keyword.

This requires to adjust the `ink_lang_ir` and `ink_lang_codegen` crates.

PR #1017, opened 2021-11-15
Reworked message dispatch logic

xgreenx:
In this PR I moved common parts of each message to `execute_dispatchable`(pull of the contract). In the future, we can optimize it to do one push of the contract.

Introduced an additional static buffer to store input. Now the user has access to the input buffer at any time during the execution of the contract(it is loaded only one time). This buffer is needed to rework the decoding process.

Moved the decode logic to the `CALLABLE` part. It allows the compiler to use value after decoding and do some optimization.

Call `unwrap` instead of passing error upper during `execute dispatchable` and decoding. It allows the compiler to optimize the code better.

Optimized `return_value` function to simply use buffer for encoding the result. We don't need complex methods from `ScoppedBuffer` because after the call of that method the program will exit.

All these optimizations allow reducing the code of optimized contracts well. 

`Erc20` takes 7862 bytes instead of 10282(saved 2420).

The more methods in the contract, the stronger the effect of the change.
[That contract](https://github.com/prosopo-io/protocol/blob/master/contracts/lib.rs) contains a lot of methods and the chagne saved 6489 bytes: 
`Original wasm size: 71.8K, Optimized: 38.8K(38844)` -> `Original wasm size: 81.7K, Optimized: 32.4K(32355)`


crates/lang/codegen/src/generator/dispatch.rs
```diff
@@ -256,17 +256,22 @@ impl Dispatch<'_> {
                 let constructor_ident = constructor.ident();
                 let selector_id = constructor.composed_selector().into_be_u32().hex_padded_suffixed();
                 let selector_bytes = constructor.composed_selector().hex_lits();
-                let input_bindings = generator::input_bindings(constructor.inputs());
+                let input_bindings = constructor.inputs()
+                    .map(|_| quote! {
+                        ::scale::Decode::decode(&mut _input).map_err(|_| ::ink_lang::reflect::DispatchError::CouldNotReadInput)?
```
xgreenx, line 261: Also removing of `.map_err` and usage of  `.unwrap()` reduce size for ~500 bytes. So we can create two analogs of `CALLABLE`(for `debug` and `release` mode).

xgreenx: `Erc20`: Original wasm size: 31.5K, Optimized: 9.1K(9067) - saved 571
`Erc1155`: Original wasm size: 83.9K, Optimized: 46.2K(46168) - saved 600
`multisig`: Original wasm size: 103.8K, Optimized: 46.8K(46798) - saved 359

cmichi:
@xgreenx You can get the `ink-waterfall` stage to run by removing the failing CI checks altogether from `.gitlab-ci.yml`. So just remove everything in `clippy-std`, `clippy-wasm`, etc. and push these changes to this PR.

xgreenx:
> @xgreenx You can get the `ink-waterfall` stage to run by removing the failing CI checks altogether from `.gitlab-ci.yml`. So just remove all failing stages (`clippy-std`, `clippy-wasm`, etc.) and push these changes to this PR.

`ink-waterfall` failed=( `tests::delegator::delegator_works`. I tried it on local node and it works(manually, didn't try to run a test). Maybe it is not enough gas?

xgreenx:
> > @xgreenx You can get the `ink-waterfall` stage to run by removing the failing CI checks altogether from `.gitlab-ci.yml`. So just remove all failing stages (`clippy-std`, `clippy-wasm`, etc.) and push these changes to this PR.
> 
> `ink-waterfall` failed=( `tests::delegator::delegator_works`. I tried it on local node and it works(manually, didn't try to run a test). Maybe it is not enough gas?

Oh, after my change the contract requires 53k additional gas=D Because I'm allocating 16MB in stack=D But I think we can decide in this PR how to optimize it(for example use static buffer form instance)

crates/lang/src/reflect/dispatch.rs
```diff
@@ -337,172 +340,22 @@ pub trait DispatchableConstructorInfo<const ID: u32> {
...
-///             <<Contract as ContractMessageDecoder>::Type as Decode>::decode(
-///                 &mut &input_bytes[..]).is_err()
-///         );
-///     }
-/// }
-/// ```
-pub trait ContractMessageDecoder {
-    /// The ink! smart contract message decoder type.
-    type Type: scale::Decode + ExecuteDispatchable;
+pub trait ContractMessageExecutor {
```
Robbepop, line 351: Why remove all the documentation and usage examples? Please re-add or write new.

crates/lang/src/reflect/dispatch.rs
```diff
@@ -337,172 +340,22 @@ pub trait DispatchableConstructorInfo<const ID: u32> {
...
-///             <<Contract as ContractConstructorDecoder>::Type as Decode>::decode(
-///                 &mut &input_bytes[..]).is_err()
-///         );
-///     }
-/// }
-/// ```
-pub trait ContractConstructorDecoder {
-    /// The ink! smart contract constructor decoder type.
-    type Type: DecodeDispatch + ExecuteDispatchable;
+pub trait ContractConstructorExecutor {
```
Robbepop, line 356: Please re-add or write new docs with usage examples.

crates/lang/tests/ui/contract/pass/dispatch-decoder-works.rs
```diff
@@ -1,132 +1,132 @@
...
-        assert_eq!(
-            <<Contract as ContractMessageDecoder>::Type
-                as DecodeDispatch>::decode_dispatch(&mut &input_bytes[..])
-                .map(|_| ())
-                .unwrap_err(),
-            DispatchError::InvalidParameters,
-        );
-    }
-}
+// use ink_lang as ink;
```
Robbepop, line 1: Please re-enable this UI test if possible. We should always try to have testable code.

crates/env/src/api.rs
```diff
@@ -526,6 +526,28 @@ where
     })
 }
 
+/// Returns the buffer back to the caller of the executed contract.
+///
+/// # Note
+///
+/// This function  stops the execution of the contract immediately.
+pub fn return_value_scoped(return_flags: ReturnFlags, return_value: &[u8]) -> ! {
```
Robbepop, line 534: very confusing name imo. The scoped buffer concept is an internal low-level concept that should not be exposed to the user. I'd rename this to `return_bytes` since that's what it is.

crates/env/src/api.rs
```diff
@@ -526,6 +526,28 @@ where
...
+        EnvBackend::return_value_scoped(instance, return_flags, return_value)
+    })
+}
+
+/// Returns the buffer back to the caller of the executed contract.
+///
+/// # Note
+///
+/// This function  stops the execution of the contract immediately.
+pub fn input_scoped(buffer: &mut [u8]) {
```
Robbepop, line 545: Very confusing name for the reason already described in another comment. I would rename this to `input_bytes` since that is what is actually is.

crates/env/src/backend.rs
```diff
@@ -207,6 +207,8 @@ pub trait EnvBackend {
     where
         T: scale::Decode;
 
+    fn input_scoped(&mut self, buffer: &mut [u8]);
```
Robbepop, line 210: Missing documentation.

crates/env/src/backend.rs
```diff
@@ -220,6 +222,17 @@ pub trait EnvBackend {
     where
         R: scale::Encode;
 
+    /// Returns the buffer to the caller of the executed contract.
```
Robbepop, line 225:
```suggestion
    /// Returns the bytes to the caller of the executed contract.
```

crates/env/src/backend.rs
```diff
@@ -220,6 +222,17 @@ pub trait EnvBackend {
...
+    ///
+    /// # Note
+    ///
+    /// Calling this method will end contract execution immediately.
+    /// It will return the given return value back to its caller.
+    ///
+    /// The `flags` parameter can be used to revert the state changes of the
+    /// entire execution if necessary.
+    fn return_value_scoped(&mut self, flags: ReturnFlags, return_value: &[u8]) -> !;
```
Robbepop, line 234:
```suggestion
    fn return_value_scoped(&mut self, flags: ReturnFlags, returned_bytes: &[u8]) -> !;
```

crates/env/src/api.rs
Robbepop, line 534:
```suggestion
pub fn return_bytes(return_flags: ReturnFlags, returned_bytes: &[u8]) -> ! {
```

crates/env/src/api.rs
Robbepop, line 545:
```suggestion
pub fn input_bytes(buffer: &mut [u8]) -> Result<()> {
```

Robbepop: Another problem I have with this API is that it does not cover the erraneous case where `buffer` is too small to hold the entire given input.

crates/lang/codegen/src/generator/dispatch.rs
```diff
@@ -523,46 +482,27 @@ impl Dispatch<'_> {
...
-                    fn decode_dispatch<I>(input: &mut I)
-                        -> ::core::result::Result<Self, ::ink_lang::reflect::DispatchError>
-                    where
-                        I: ::scale::Input,
-                    {
-                        match <[::core::primitive::u8; 4usize] as ::scale::Decode>::decode(input)
+                impl ::ink_lang::reflect::ExecuteDispatchable for __ink_ConstructorDecoder {
+                    fn execute_dispatchable() -> ::core::result::Result<(), ::ink_lang::reflect::DispatchError> {
+                        const CAPACITY: usize = 1 << 14;
+                        let mut local_buffer: [u8; CAPACITY] = [0; CAPACITY];
```
Robbepop, line 490: Why not use the already existing static buffer that each ink! smart contract already has? I do not see a technical reason why using the already existing static buffer should be less efficient than using a local one.

Robbepop (changes requested):
I have not yet reviewed the whole thing but I am very skeptical of the introduction of the local static buffer. Imo we should make sure to use the already existing global static buffer in those cases.

xgreenx:
I want to clarify, that it is draft PR to show the idea=) I will add documentation and tests when you will agree with the idea(=

crates/lang/codegen/src/generator/dispatch.rs
xgreenx: We should use a static buffer because it allows us to not allocate the memory in runtime -> less gas.

I will try to rework it to use a static buffer. The problem is that `off_chain` and `off_chain_experemental` don't use a static buffer. So I can't return `&[u8]`. But if you are okey with that idea I will add a static buffer to that engines.

The local buffer is used to fastly create that PR and check the impact=)

crates/env/src/api.rs
xgreenx: But we used the same schema in the previous implementation. We decoded the whole ink's message variant

.gitlab-ci.yml
```diff
@@ -6,11 +6,7 @@
 
 
 stages:
-  - check
```
Robbepop, line 9: Why does this PR change the CI?

crates/env/src/api.rs
```diff
@@ -491,24 +491,14 @@ where
...
-    T: scale::Decode,
-{
-    <EnvInstance as OnInstance>::on_instance(|instance| {
-        EnvBackend::decode_input::<T>(instance)
+/// When using ink! users do not have to decode input themselves as they are
+/// normally decoded by the ink! code generation automatically.
+pub fn input_bytes<'a>() -> &'a [u8] {
+    <EnvInstance as OnInstance<'a>>::on_instance(|instance: &'a mut EnvInstance| {
+        EnvBackend::input_bytes(instance)
     })
```
Robbepop, line 502: This API as stated is extremely unsafe to use by design. This is exactly why the `decode_input` API exists as is. Just notice how with this you could safely (!!) do the following:
```rust
let input = ink_env::input_bytes(); // Returns a slice refering to ink!'s static buffer
let caller = ink_env::caller(); // Mutates ink!'s static buffer to hold an AccountId
// at this point the above input binding no longer holds the input but the AccountId
```
I think I do not have to point out how extremely unsafe this API design is and therefore I would never accept a PR introducing it, even if it was marked unsafe since we do not want to introduce unsafe APIs if we can do better.

Robbepop (changes requested):

.gitlab-ci.yml
xgreenx: I will revert it back after merging with master

crates/env/src/api.rs
xgreenx: The idea was that borrow check will prevent you from the usage of `ink_env::caller()` if you will use `input` after that. But I tested it now and... It allows that=D So yes, it should be changed.

Do you have idea how better to provide the usage of static buffer to the user in safety way?

Robbepop: > Do you have idea how better to provide the usage of static buffer to the user in safety way?

Well I designed the `decode_input` API exactly this way because of exactly this issue. So my proposal is to keep `decode_input` API as is.

xgreenx: But the idea of the change is to decode directly in the place where we want to use arguments. Instead of decoding them in another place and then moving them to the place where we want to use them=)

Robbepop: I think a better idea is to still use `decode_input` but not move the arguments and instead use references where the user wants to use references. Let me refer to [this issue](https://github.com/paritytech/ink/issues/989) for this use case.

Robbepop: What I experienced while rewriting the entire codegen is that the differentiation between decoding the messages and executing the messages caused a minor Wasm size bloat that could probably be fixed without changing the `decode_input` API at all. I haven't tried yet myself though.

Robbepop: Concerning `return_bytes` I'd also like to keep the previous `return_value` with some built-in optimizations that can be used to avoid ink!'s static buffer. We could use `parity_scale_codec::Encode::encoded_fixed_size` method for that for example.

xgreenx: > I think a better idea is to still use decode_input but not move the arguments and instead use references where the user wants to use references. Let me refer to this issue for this use case.


Apart from optimizing the size, I am also pursuing another idea: Decode input arguments directly in the message to be agnostic to other messages. At the moment you need to know the list of all messages from all traits and inherent implementation to create decodable enum.

This restriction doesn't allow to support default implementation in traits, because users should enumerate all methods in the implementation section(it breaks the idea of default implementation).

```rust
// This line allows the user to inherit default implementation of the trait
#[ink(impl)]
impl Erc20 for Ers20Struct {}
```

If methods decode the input by self. We can add a method to the trait:
```rust
pub trait Erc20: ContractEnv {
    ...
    fn execute_input(selector: [u8; 4], input: &[u8]) {
        match selector {
        ...    
        }
    }
}
```

xgreenx: > Concerning return_bytes: I'd also like to keep the previous return_value with some built-in optimizations that can be used to avoid ink!'s static buffer. We could use parity_scale_codec::Encode::encoded_fixed_size method for that for example. This way we keep a high-level API and have the same performance enhancements as with your proposed low-level API.

I've already removed the usage of `return_bytes`. But with access to a static buffer(or to another buffer where we can store the encoded result), [it](https://github.com/Supercolony-net/ink/blob/d2b7a5275854d627100ae52c0d8b671e6b27c218/crates/lang/codegen/src/generator/dispatch.rs#L587) can be optimized more. It is a common part for all messages, and we can do only one `::ink_storage::traits::push_spread_root` instead of several=)

Robbepop: > But with access to a static buffer(or to another buffer where we can store the encoded result), it can be optimized more.

I highly doubt this statement but cannot prove it at the moment. Using a static buffer should not at all be different performance wise from a local buffer in Wasm where we operate in a single threaded environment.

Robbepop: > If methods decode the input by self. We can add a method to the trait:

Adding artificial methods to ink! trait definitions is a no-go. We must make sure that ink! modifies trait definitions (and others) as minimally as possible.

xgreenx: > I highly doubt this statement but cannot prove it at the moment. Using a static buffer should not at all be different performance wise from a local buffer in Wasm where we operate in a single threaded environment.

I meant that we can change code-generation and optimize it do one `::ink_storage::traits::push_spread_root` at the end of `execute_dispatchable` instead of the end of each message section.

BTW, It should have the same performance. But in the case of a local buffer you need to allocate that memory in the stack - it is additional operations in Runtime.

Robbepop: > I meant that we can change code-generation and optimize it do one ::ink_storage::traits::push_spread_root at the end of execute_dispatchable instead of the end of each message section.

Yes, there is a lot of optimization potential in the way we currently dispatch ink! constructors and messages but none of which requires changes in the `ink_env` APIs imo. That's what I meant.

xgreenx: > Adding artificial methods to ink! trait definitions is a no-go. We must make sure that ink! modifies trait definitions (and others) as minimally as possible.

I agree with it. But I think that this is the only possible solution at the moment if we want to support all the features of traits in rust. Because if the user defines type alias, or constant, or super trait, you will not able to do a strick as you did for `__ink_TraitInfo`=)

Robbepop: I think we should not discuss all at once. Please let us focus on the improvements to the dispatch logic in this PR.

xgreenx: I mentioned trait because it requires the next logic:
First I need to get 4 bytes to find a selector. After I know which method I need to call that method will decode the remaining bytes into arguments.

So it is why I'm trying to decode them agnoticly=)

What do you think about two static buffers? One will be always read-only and will contain input bytes. Another will be used for encoding/decoding. In this case, the user will have the access to the input at any time. And he can decode what he wants by himself.

xgreenx: We always get the whole input for the user(even if he wants only 4 bytes we will read all 16 MB). And if he wants to have a simple signature of method(with 0 arguments) but later he wants to decode some parts of the input, he again should get the whole data. And it can be expensive to call `ext::seal_input` each time. So I think idea with second array only for input is not bad(=

xgreenx:
The change reduces the size of each contract except subcontracts of the `delegator` example. I will check why and how we can improve it too.

The open question is: Are we okay to have a separate static buffer to store input of the contract?

xgreenx:
I found that all contracts from [workspace](https://github.com/paritytech/ink/blob/141673b578c74e4382753b80faf2c1594a23fe91/examples/delegator/Cargo.toml#L45) contain debug information(stuff related to `Debug` and `Display`). It increases the size of all contracts there.

Removing `[workspace]` decreases the size of each contract.  I think it is better to remove it to see the actual size of the contracts.

xgreenx:
The change is ready, so I'm waiting for your comments=)

HCastano:
@xgreenx can you update this for the latest master? I'm curious as to how the size improvements look. Since this is a big change to the dispatch logic I think we should only move forward with it if the savings look good

xgreenx:
`ink-waterfall` didn't update the sizes=)

HCastano:
> `ink-waterfall` didn't update the sizes=)

Yep, looks like there's a problem (https://github.com/paritytech/ink-waterfall/issues/20).

The job did succeed though, and here are the results ([link to run](https://gitlab.parity.io/parity/ink-waterfall/-/jobs/1352865)):

```
|----------------------|------------------|------------|----------------------|----------------|
| `accumulator`        | +0.19 K          |            | 1.62 K               |                |
| `adder`              | -0.12 K          |            | 2.82 K               |                |
| `contract-terminate` | -0.12 K          | -1         | 1.19 K               | 200_777        |
| `contract-transfer`  | -0.05 K          | +16        | 7.94 K               | 1_415          |
| `delegator`          | -0.86 K          | -1_063     | 9.20 K               | 8_447          |
| `dns`                | -1.14 K          | -1_816     | 9.33 K               | 12_472         |
| `erc1155`            | -2.88 K          | -3_718     | 24.62 K              | 44_093         |
| `erc20`              | -2.06 K          | -1_929     | 8.16 K               | 7_828          |
| `erc721`             | -2.44 K          | +1_264     | 11.26 K              | 35_966         |
| `flipper`            | +0.04 K          | +21        | 1.81 K               | 368            |
| `incrementer`        | +0.05 K          | +23        | 1.69 K               | 363            |
| `multisig`           | -1.20 K          | -6_313     | 27.36 K              | 48_583         |
| `proxy`              | -0.23 K          | -80        | 3.58 K               | 2_549          |
| `rand-extension`     | -1.07 K          | -92        | 4.04 K               | 6_562          |
| `subber`             | -0.13 K          |            | 2.83 K               |                |
| `trait-erc20`        | -2.08 K          | -1_928     | 8.44 K               | 8_716          |
| `trait-flipper`      | +0.09 K          | +22        | 1.62 K               | 364            |
| `trait-incrementer`  | +0.11 K          | +47        | 1.73 K               | 727            |
```

So the savings are alright, but they're not mindblowing either. I'm on the fence about this given the nature of the changes, but give me some time to try and better understand the new dispatch logic you're proposing before we decide what to do

xgreenx:
> So the savings are alright, but they're not mindblowing either

The more methods in the contract, the stronger the effect of the change.
[That contract](https://github.com/prosopo-io/protocol/blob/master/contracts/lib.rs) contains a lot of methods and the change saved 6489 bytes

HCastano:
@xgreenx This has been neglected for a while (I guess mostly by us, sorry).

If you want to revive it please open a new PR so we can a fresh perspective and set of sizes

PR #1019, opened 2021-11-16
Remove all instance of `inline`

HCastano:
A naive follow up to #1012. I just ran `fd -e rs -x sed -i '' -e '/\#\[inline\]/d'` and
want to see how contract sizes get affected.

### DIFF
--- a/crates/allocator/src/bump.rs
+++ b/crates/allocator/src/bump.rs
@@ -35,7 +35,6 @@ static mut INNER: InnerAlloc = InnerAlloc::new();
 pub struct BumpAllocator;
 
 unsafe impl GlobalAlloc for BumpAllocator {
-    #[inline]
     unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
         match INNER.alloc(layout) {
             Some(start) => start as *mut u8,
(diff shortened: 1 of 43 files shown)


xgreenx:
It also includes https://github.com/paritytech/cargo-contract/pull/358, I think you also need to check the impact only of https://github.com/paritytech/cargo-contract/pull/358.

Robbepop:
While the numbers of our example contracts looks okayish with respect to this PR I would not suggest merging this since removal of all `#[inline]` is not very thought through. Instead we should try to identify which `#[inline]` annotated functions should actually not be `#[inline]`. Note that with respect to inline annotations there is a ton of noise involved. Removing `#[inline]` for a function `foo` might end up improving example A but worsen example B at the same time.

HCastano:
> While the numbers of our example contracts looks okayish with respect to this PR I would not suggest merging this since removal of all `#[inline]` is not very thought through. Instead we should try to identify which `#[inline]` annotated functions should actually not be `#[inline]`. Note that with respect to inline annotations there is a ton of noise involved. Removing `#[inline]` for a function `foo` might end up improving example A but worsen example B at the same time.

Yeah I agree, this was a brute-force follow up more out of curiosity than anything. I've opened #1035 so that we remember to try a more targeted approach in the future.

Issue #739, opened 2021-03-19
Allow fallback message and constructor handlers


Robbepop:
Towards a future with upgradeable ink! smart contracts.

## Motivation & Related Work

In order to create an upgradeable contract on Ethereum contract authors usually have the following architecture of 2 or 3 different contracts interacting with each other:

1. **Forwarder:** This contract receives all messages from outside and simply forwards them blindly to the contract at the address `A` which it stores on its own contract storage. The contract at address `A` normally is the logic layer of the set of contracts where all the interesting things are happening. It is also the part that is usually upgraded. Upgrading in this scheme relinks the address `A` from the `Forwarder` contract to a new version of the logic layer contract. If there is a `Data` contract the `Forwarder` contract also needs to store its address so that it can notify the `Data` layer about an upgrade for the `Logic` layer. This is required so that the `Data` layer can protect accesses from outside.
2. **Logic:** This contract contains all the logic of the contract. It has several constructors and messages to expose its API to the outside world. Interactions should only ever happen through the harded `Forwarder` contract and the logic layer necessarily needs to protect itself from accesses besides that one. Optionally a contract author can decide to put their data into yet another `data` contract. If this is the case the `Logic` contract needs to invoke messages of the `Data` contract in order to interact with any persisting storage data and it will have absolutely no storage to itself. When upgrading a contract only the `Logic` layer is going to be swapped out while the `Forwarder` and `Data` layers are kept. The `Forwarder` simply needs to relink to the new `Logic` layer contract address though and the new `Logic` contract needs to keep the original links to the old `Data` contract so that all storage is kept as is.
3. **Data:** The data layer is only concerned about storing and loading persisted storage data required by the logic layer. It should offer a proper API for the `Logic` layer contract and must protect itself from accesses other than the `Logic` layer contract or the `Forwarder` contract. Upgrading a contract should never swap out the `Data` layer contract since this would be an extraordinarily costly operation to do. Keeping and persisting the state between upgrades is critical.

## Upgradeability in ink!

[...]

PR #1020, opened 2021-11-17
Implement wildcard selector, add `proxy` example

cmichi:
Closes https://github.com/paritytech/ink/issues/739.

This PR intentionally
* …does not contain any wildcard selector trait support.
* …does not contain an off-chain test for the new `proxy` example. Will be added later.

I'll add an E2E test for the `ink-waterfall` as a follow-up.

Pictures for the `proxy/readme.md` would also be nice, can also be done as a follow-up.


examples/proxy/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        pub fn forward(&self) -> u32 {
+            ink_env::call::build_call::<ink_env::DefaultEnvironment>()
+                .callee(self.forward_to)
+                .call_flags(
+                    ink_env::CallFlags::default()
+                        .set_clone_input(true)
+                        .set_tail_call(true),
+                )
+                .gas_limit(0)
+                .transferred_value(self.env().transferred_balance())
```
cmichi, line 63: We should unify this API to be just `.transferred_balance` everywhere.

examples/proxy/lib.rs
```diff
@@ -0,0 +1,71 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use ink_lang as ink;
+
+#[ink::contract]
+pub mod proxy {
+    /// A simple proxy contract, enabling upgradable smart contracts.
```
xgreenx, line 7: We need to clarify that it is not a full upgradable contract because this contract only forwards calls to another contract. But the state is stored on a forwarded contract, not a Proxy. When delegated call will be implemented on `contract-pallet`, we should use it here, and then it will be a full upgradable contract.

cmichi: Good point, I clarified it. 

Btw.: The proxy contract is everything we've currently planned in the direction of upgradable contracts. I think the upgradable contracts milestone is out of scope for your current grant, but having proper delegate-call support in the `contracts` pallet as well as ink! would be a great follow-up to your current grant.

examples/proxy/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+
+        /// Fallback message for a contract call that doesn't match any
+        /// of the other message selectors.
+        ///
+        /// *Note:*
+        /// - We allow payable messages here and would forward any optionally supplied
+        ///   value as well.
+        /// - If the self receiver would be `forward(&mut self)` here, this would not
+        ///   imply that the contract to which we forward to does not mutate it's own
+        ///   state.
```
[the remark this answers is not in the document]
cmichi: The contract to which we forward does not care, I'll try to clarify it better.

examples/proxy/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        #[ink(message, payable, selector = "_")]
+        pub fn forward(&self) -> u32 {
+            ink_env::call::build_call::<ink_env::DefaultEnvironment>()
+                .callee(self.forward_to)
+                .call_flags(
+                    ink_env::CallFlags::default()
+                        .set_clone_input(true)
+                        .set_tail_call(true),
```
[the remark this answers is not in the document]
cmichi: See the rustdocs here: https://paritytech.github.io/ink/ink_env/struct.CallFlags.html.

examples/proxy/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+                )
+                .gas_limit(0)
```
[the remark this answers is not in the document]
cmichi: Yeah, that's a good idea, feel free to create an issue.

crates/lang/tests/ui/contract/fail/message-selector-invalid-type-01.stderr
```diff
@@ -1,4 +1,4 @@
-error: #[ink(selector = ..)] attributes with string inputs are deprecated. use an integer instead, e.g. #[ink(selector = 1)] or #[ink(selector = 0xC0DECAFE)].
+error: #[ink(selector = ..)] attributes with string inputs are deprecated (except for the wildcard selector "_"). use an integer instead, e.g. #[ink(selector = 1)] or #[ink(selector = 0xC0DECAFE)].
```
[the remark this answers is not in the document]
cmichi: Yeah, so the issue is [here](https://github.com/paritytech/ink/blob/master/crates/lang/ir/src/ir/attrs.rs#L809). We would need access to `_` there.

But the crate which we use for parsing attributes a la `key = "value"` does not support parsing verbatims such as `_`. Specifically the type `syn::Meta::NameValue` which we use as part of the [`syn::parse_meta`](https://docs.rs/syn/1.0.81/syn/struct.Attribute.html#method.parse_meta) return value does not support this.

I briefly looked into `syn::parse_meta`, but it was nothing trivial. Maybe we could also do some pre-processing, so that `selector = _` is transformed into `selector = "_"` before passing it to the lib, but I haven't looked into that.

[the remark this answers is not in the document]
xgreenx: Or you can implement your own `parse_meta`=) You can copy the main logic from `syn` + add support of `_`=)

examples/proxy/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        ///   value as well.
+        /// - If the self receiver were `forward(&mut self)` here, this would not
+        ///   have any effect whatsover on the contract we forward to.
+        #[ink(message, payable, selector = "_")]
+        pub fn forward(&self) -> u32 {
+            ink_env::call::build_call::<ink_env::DefaultEnvironment>()
+                .callee(self.forward_to)
+                .call_flags(
+                    ink_env::CallFlags::default()
+                        .set_clone_input(true)
```
athei, line 76: The input isn't used after the call. No need to keep it around.

```suggestion
                        .set_forward_input(true)
```

athei (approved):

examples/proxy/lib.rs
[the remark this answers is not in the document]
HCastano: I know I can look at the docs, but given that this is an example contract I think we should be adding explanations inline where possible 😄

Robbepop:
> This PR intentionally
> 
>     * …does not contain any wildcard selector trait support. I will create a follow-up issue for this.

This feature is highly unwanted since it would heavily interfere with trait modularity. We should make sure that traits never allow for wildcard selectors.

crates/lang/ir/src/ir/item_impl/callable.rs
```diff
@@ -116,6 +116,10 @@ where
         <C as Callable>::is_payable(self.callable)
     }
 
+    fn is_wildcard_selector(&self) -> bool {
```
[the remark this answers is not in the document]
Robbepop: this naming makes more sense imo

crates/lang/ir/src/ir/item_impl/callable.rs
```diff
@@ -162,6 +166,9 @@ pub trait Callable {
     /// Flagging as payable is done using the `#[ink(payable)]` attribute.
     fn is_payable(&self) -> bool;
 
+    /// Returns `true` if the ink! callable is flagged as a wildcard selector.
+    fn is_wildcard_selector(&self) -> bool;
```
Robbepop, line 170:
```suggestion
    fn has_wildcard_selector(&self) -> bool;
```

Robbepop: as explained above

crates/lang/ir/src/ir/item_impl/constructor.rs
```diff
@@ -207,6 +213,10 @@ impl Callable for Constructor {
         true
     }
 
+    fn is_wildcard_selector(&self) -> bool {
```
Robbepop, line 216:
```suggestion
    fn has_wildcard_selector(&self) -> bool {
```

crates/lang/ir/src/ir/item_impl/message.rs
```diff
@@ -100,6 +100,8 @@ pub struct Message {
     pub(super) item: syn::ImplItemMethod,
     /// If the ink! message can receive funds.
     is_payable: bool,
+    /// If the ink! message is a wildcard selector.
+    is_wildcard_selector: bool,
```
Robbepop, line 104:
```suggestion
    has_wildcard_selector: bool,
```

crates/lang/ir/src/ir/item_impl/constructor.rs
```diff
@@ -578,4 +588,18 @@ mod tests {
...
     }
+
+    #[test]
+    fn try_from_wildcard_constructor_fails() {
+        let item: syn::ImplItemMethod = syn::parse_quote! {
+            #[ink(constructor, selector = "_")]
+            pub fn my_constructor() -> Self {}
+        };
+        let errs: syn::Error = <ir::Constructor as TryFrom<_>>::try_from(item)
+            .expect_err("must result in Err");
```
HCastano, line 599:
```suggestion
            .expect_err("having a wildcard selector on a constructor must result in an Err");
```

crates/lang/ir/src/ir/item_impl/message.rs
```diff
@@ -205,7 +208,15 @@ impl TryFrom<syn::ImplItemMethod> for Message {
         let (ink_attrs, other_attrs) = Self::sanitize_attributes(&method_item)?;
         let is_payable = ink_attrs.is_payable();
         let selector = ink_attrs.selector();
+        let is_wildcard_selector = ink_attrs.is_wildcard_selector();
+        if is_wildcard_selector && selector.is_some() {
+            return Err(format_err!(
+                method_item.span(),
+                "ink! message cannot contain wildcard selector and specified selector"
```
Robbepop, line 215: Even more correct would be:

> "ink! message cannot have multiple selectors"

crates/lang/ir/src/ir/item_impl/message.rs
```diff
@@ -205,7 +208,15 @@ impl TryFrom<syn::ImplItemMethod> for Message {
...
+            ))
+        }
         Ok(Self {
+            is_wildcard_selector,
```
Robbepop, line 219:
```suggestion
            has_wildcard_selector,
```

Robbepop (commented):
Haven't reviewed the entire thing, yet, however, I have to say that this PR goes a very long way in making sure that a wildcard selector is something completely different from a selector whereas the beauty in the original design actually made a lot of sense from the fact that we can treat both as different kinds of selectors.
So instead of putting `is_wildcard_selector` everywhere it would probably a more elegant design to make `ir::Selector` an enum that can represent a concrete selector or a wildcard selector.

HCastano (approved):
Nice!

crates/lang/ir/src/ir/item_impl/message.rs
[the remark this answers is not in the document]
Robbepop: yes please ... ^^

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -395,6 +403,11 @@ pub enum AttributeArg {
     /// Applied on ink! constructors or messages in order to specify that they
     /// can receive funds from callers.
     Payable,
+    /// `#[ink(selector = "_")]`
+    ///
+    /// Applied on ink! messages to define a fallback messages that is invoked
+    /// if no other ink! message matches a given selector.
+    WildcardSelector,
```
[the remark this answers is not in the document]
Robbepop: this distinctions makes some impls simpler

crates/lang/ir/src/ir/item_impl/message.rs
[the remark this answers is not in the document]
cmichi: Yeah, so I had originally considered this. But it requires this:
```rust
pub struct Selector {
    bytes: [u8; 4],
}
```
to be made into
```rust
pub struct Selector {
    bytes: Option<[u8; 4]>,
}

fn is_wildcard(&self) -> bool {
  self.bytes.is_none()
}
```
This is no trivial change though, as it requires a lot of new checks throughout our codebase and more special logic, whereas with the way I implemented it currently it's way less of a diff.

crates/lang/codegen/src/generator/dispatch.rs
```diff
@@ -114,6 +115,23 @@ impl Dispatch<'_> {
             .count()
     }
 
+    /// Returns the index of the ink! message which has a wildcard selector, if existent.
+    fn query_wildcard_message(&self) -> Option<usize> {
```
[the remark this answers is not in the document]
cmichi: The function signature of `Dispatch::expand_message_input` to which this return value is passed requires an index.

crates/lang/ir/src/ir/item_impl/message.rs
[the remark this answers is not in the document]
Robbepop: I think this will be very easy if we instead make a new type that wraps the `ir::Selector`:
```rust
pub enum SelectorOrWildcard {
    Wildcard,
    Selector(ir::Selector),
}
```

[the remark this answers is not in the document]
cmichi: Done.

crates/lang/ir/src/ir/item_impl/constructor.rs
```diff
@@ -158,6 +158,12 @@ impl Constructor {
                     ir::AttributeArg::Constructor | ir::AttributeArg::Selector(_) => {
                         Ok(())
                     }
+                    ir::AttributeArg::WildcardSelector => {
+                        Err(Some(format_err!(
+                            arg.span(),
+                            "wildcard selectors are not allowed for constructors"
```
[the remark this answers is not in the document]
cmichi: I've added the capability of using wildcard selectors for constructors now.

crates/lang/ir/src/ir/trait_def/item/trait_item.rs
```diff
@@ -83,6 +84,8 @@ impl<'a> InkTraitMessage<'a> {
             &ir::AttributeArgKind::Message,
             |arg| {
                 match arg.kind() {
+                    ir::AttributeArg::Selector(SelectorOrWildcard::Wildcard) =>
+                        Err(Some(format_err!(arg.span(), "wildcard selectors are only supported for plain ink! messages, not for traits."))),
```
Robbepop, line 88:
```suggestion
                        Err(Some(format_err!(arg.span(), "wildcard selectors are only supported for inherent ink! messages or constructors, not for traits."))),
```

crates/lang/ir/src/ir/item_impl/message.rs
```diff
@@ -100,6 +100,8 @@ pub struct Message {
     pub(super) item: syn::ImplItemMethod,
     /// If the ink! message can receive funds.
     is_payable: bool,
+    /// If the ink! message is a wildcard selector.
+    has_wildcard_selector: bool,
```
Robbepop, line 104: We could ues `selector: Option<SelectorOrWildcard>` here instead of the current `has_wildcard_selector` and `selector` fields.

crates/lang/ir/src/ir/item_impl/constructor.rs
```diff
@@ -64,6 +64,8 @@ use syn::spanned::Spanned as _;
 pub struct Constructor {
     /// The underlying Rust method item.
     pub(super) item: syn::ImplItemMethod,
+    /// If the ink! constructor is a wildcard selector.
+    has_wildcard_selector: bool,
```
Robbepop, line 68: Same as for ink! messages: I think it would be cleaner to use `selector: Option<SelectorOrWildcard>` here instead of both current `has_wildcard_selector` and `selector` fields since they can represent invalid state (`(true, Some(_))`) which my proposed field cannot.

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -1116,15 +1213,15 @@ mod tests {
...
+                SelectorOrWildcard::Selector(Selector::from([0, 0, 0, 42])),
             )])),
         );
         assert_attribute_try_from(
             syn::parse_quote! {
                 #[ink(selector = 0xDEADBEEF)]
             },
             Ok(test::Attribute::Ink(vec![AttributeArg::Selector(
-                Selector::from([0xDE, 0xAD, 0xBE, 0xEF]),
+                SelectorOrWildcard::Selector(Selector::from([0xDE, 0xAD, 0xBE, 0xEF])),
```
Robbepop, line 1224: Maybe it would be nice to have a `From<ir::Selector>` impl for `SelectorOrWildcard` at hand.

Robbepop: Or maybe even a `selector` constructor:
```rust
SelectorOrWildcard::selector([0xDE, 0xAD, 0xBE, 0xEF])
```

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -809,6 +896,23 @@ impl TryFrom<syn::NestedMeta> for AttributeFrag {
...
+                                        name_value,
+                                        "#[ink(selector = ..)] attributes with string inputs are deprecated. \
+                                        use an integer instead, e.g. #[ink(selector = 1)] or #[ink(selector = 0xC0DECAFE)]."
+                                    ))
+                                }
+                                return Ok(AttributeFrag {
+                                    ast: meta,
+                                    arg: AttributeArg::Selector(SelectorOrWildcard::Wildcard),
+                                })
+                            }
```
Robbepop, line 914: I'd appreciate if we create a follow-up issue that fixes this work around with a proper implementation that customizes the argument parsing.

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -723,13 +759,64 @@ impl From<InkAttribute> for Attribute {
...
+        let ts: TokenStream2 = attr
+            .tokens
+            .into_iter()
+            .map(|tt| {
+                match tt {
+                    TokenTree2::Group(grp) => transform_wildcard_selector_to_string(grp),
+                    _ => tt,
+                }
+            })
+            .collect();
```
Robbepop, line 817: This is very fragile because it changes syntax without looking at the semantics.
A point where this can fail is any input that includes a `_` token, such as:
```rust
#[compose(Bar, _)]
#[precondition x == Some(_)]
```

Since with this implementation we accept `_` as well as `"_"` we should just remove this transformation and instead use a custom parsing local to this code point that parses the given `attr.tokens` instead of blindfoldedly applying these generic transformations to its input. And if the parsing finds `selector = _` then it can replace it with `selector = "_"`. This way we have a stable implementation without requiring too much efforts into re-implementing the custom parsing for attributes everywhere.

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -511,6 +531,22 @@ impl core::fmt::Display for AttributeArg {
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SelectorOrWildcard {
+    Wildcard,
+    Selector(ir::Selector),
+}
+
+impl core::fmt::Display for SelectorOrWildcard {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> Result<(), core::fmt::Error> {
+        match self {
+            Self::Selector(selector) => write!(f, "{:?}", selector.to_bytes()),
```
Robbepop, line 544: If `ir::Selector` already implements `Display` we should ideally just forward to its `Display` implementation using `core::fmt::Display::fmt(&selector, f)`.

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -511,6 +531,22 @@ impl core::fmt::Display for AttributeArg {
     }
 }
 
+/// Either a wildcard selector or a specified selector.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SelectorOrWildcard {
+    Wildcard,
+    Selector(ir::Selector),
```
Robbepop, line 538: Ideally both variants have some documentation. ;)

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -496,7 +516,7 @@ impl core::fmt::Display for AttributeArg {
             Self::Constructor => write!(f, "constructor"),
             Self::Payable => write!(f, "payable"),
             Self::Selector(selector) => {
-                write!(f, "selector = {:?}", selector.to_bytes())
+                write!(f, "selector = {}", selector)
```
Robbepop, line 519: If ir::Selector already implements Display we should ideally just forward to its Display implementation using core::fmt::Display::fmt(&selector, f).

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -267,7 +270,9 @@ impl InkAttribute {
     /// Returns the selector of the ink! attribute if any.
```
Robbepop, line 270:
```suggestion
    /// Returns the non-wildcard selector of the ink! attribute if any.
```

crates/lang/codegen/src/generator/dispatch.rs
```diff
@@ -114,6 +115,40 @@ impl Dispatch<'_> {
...
+            .impls()
+            .map(|item_impl| item_impl.iter_messages())
+            .flatten()
+            .positions(|item| item.has_wildcard_selector());
+        let next = wildcard_selectors.next();
+        assert!(
+            wildcard_selectors.next().is_none(),
+            "found more than one wildcard selector message, contracts must have at most one."
+        );
+        next
```
Robbepop, line 132:
```suggestion
            .position(|item| item.has_wildcard_selector())
```

Robbepop: We don't need to assert anything in the `ink_codegen` module since it relies on the fact that the whole semantics checking are already done by the `ink_ir` layer. This way `ink_codegen` transformation are kind of infallible ideally.

crates/lang/codegen/src/generator/dispatch.rs
```diff
@@ -114,6 +115,40 @@ impl Dispatch<'_> {
...
+            .impls()
+            .map(|item_impl| item_impl.iter_constructors())
+            .flatten()
+            .positions(|item| item.has_wildcard_selector());
+        let next = wildcard_selectors.next();
+        assert!(
+            wildcard_selectors.next().is_none(),
+            "found more than one wildcard selector constructor, contracts must have at most one."
+        );
+        next
```
Robbepop, line 149:
```suggestion
            .position(|item| item.has_wildcard_selector())
```

Robbepop: We don't need to assert anything in the `ink_codegen` module since it relies on the fact that the whole semantics checking are already done by the `ink_ir` layer. This way `ink_codegen` transformation are kind of infallible ideally.

Robbepop (commented):
I think as an optimization for our codegen we might want to consider specializing for the case when there is only one ink! message with a wildcard selector. This way we can drop parts of the decoding of the selector and input parameters. What do you think?
This should be made as a follow-up to this PR imo with benchmarks showing that this actually improves Wasm file sizes.

examples/proxy/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        /// not match a selector of this contract is forwarded to.
+        #[ink(message)]
+        pub fn change_forward_address(&mut self, new_address: AccountId) {
+            assert_eq!(
+                self.env().caller(),
+                self.admin,
+                "caller {:?} does not have sufficient permissions, only {:?} does",
+                self.env().caller(),
+                self.admin,
+            );
```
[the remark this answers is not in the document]
cmichi: This will remove the debugging output though right?

examples/proxy/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+                .fire()
+                .unwrap_or_else(|err| {
+                    panic!(
+                        "cross-contract call to {:?} failed due to {:?}",
+                        self.forward_to, err
+                    )
+                });
+            unreachable!(
+                "the forwarded call will never return since `tail_call` was set"
+            );
```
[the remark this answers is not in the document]
cmichi: Yeah ideally, but `fire()` returns `Result<(), Err>` and I don't see how we can return `!` without introducing an additional API `fire_without_return` or something like that.

examples/proxy/lib.rs
cmichi: I'd leave it there, but in this case I think it's clear that it should be streamlined.

crates/lang/ir/src/ir/attrs.rs
cmichi: Yeah it's fragile, the case you mentioned is not possible though, since the impl of `transform_wildcard_selector_to_string` already has two preconditions that match on `selector = ` before replacing.

examples/proxy/lib.rs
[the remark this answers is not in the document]
Robbepop: The thing is: In case of `tail_calls` it would not make sense to handle a `Result` anyways so we should directly return a raw `!`. Using compile time type information and generics-based specialization this should be implementable.

examples/proxy/lib.rs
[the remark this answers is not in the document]
Robbepop: Which debugging output do you mean exactly?
If our `Error` properly implements `Display` this should be fine upon `unwrap`-ing the `Err`.
In case you mean on-chain debugging output: We should not be concerned about this problem in this PR. I think we should make should from a language design perspective that proper Rust error handling is not inferior to panicking.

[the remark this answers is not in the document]
cmichi: So you would return `Result<(), String>` then?

HCastano (commented):
I haven't looked at all the new changes just yet, will finish this up tomorrow

crates/lang/ir/src/ir/attrs.rs
```diff
@@ -723,13 +765,66 @@ impl From<InkAttribute> for Attribute {
     }
 }
 
+/// This function replaces occurrences of a `TokenTree::Ident` of the sequence
+/// `selector = _` with the sequence `selector = "_"`.
+///
+/// This is done because `syn::Attribute::parse_meta` does not support parsing a
+/// verbatim like `_`. For this we would need to switch to `syn::Attribute::parse_args`,
+/// which requires a more in-depth rewrite of our IR parsing.
+fn transform_wildcard_selector_to_string(grp: Group2) -> TokenTree2 {
```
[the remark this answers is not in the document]
cmichi: Yeah, no one is happy with this. There'll be a follow-up issue, feel free to implement it. I started but it quickly gets super-involved, since we need to change a lot of stuff for it.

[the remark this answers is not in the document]
cmichi: Yeah, no one is happy with this. There'll be a follow-up issue, it quickly gets super-involved because we can't mix `parse_meta` with the plain `syn::parse` API.

examples/proxy/lib.rs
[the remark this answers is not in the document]
cmichi: Could you elaborate on what return type you would then choose? `Result<(), String>`? In that case `String` would also be included in the wasm32 target.

crates/lang/tests/ui/contract/pass/constructor-wildcard-selector.rs
```diff
@@ -0,0 +1,19 @@
+use ink_lang as ink;
+
+#[ink::contract]
+mod contract {
+    #[ink(storage)]
+    pub struct Contract {}
+
+    impl Contract {
+        #[ink(constructor, selector = _)]
```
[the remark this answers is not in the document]
Robbepop: Can you elaborate on this please? I cannot follow.

examples/proxy/lib.rs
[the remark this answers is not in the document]
Robbepop: Create a simple error type with a `Display` impl that tells you everything like you'd do in a real world production Rust codebase. ;)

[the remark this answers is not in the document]
Robbepop: For Wasm file size improvements you could try making the fields over references to `AccountId` or leaving away `caller` sinc the caller is obviously aware of their own `AccountId`.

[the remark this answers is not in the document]
cmichi: I've updated the PR.

examples/proxy/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        /// # Note:
+        ///
+        /// Returning an `Result::Err` here will result in the transaction
+        /// being reverted, in that sense it is equal to e.g. placing an
+        /// `assert!` here instead.
+        #[ink(message)]
+        pub fn change_forward_address(
+            &mut self,
+            new_address: AccountId,
+        ) -> Result<(), NotAdminErr> {
```
cmichi, line 81: @Robbepop While running the waterfall tests for this example, I found that since I switched from `assert!` to returning `Result::Err` the transaction is no longer reverted ‒ it always results in `ExtrinsicFailed`.

If you don't see anything that I missed I'll consider this a bug, revert to `assert!` and create a follow-up issue.

xgreenx: Hmm, I tested that on Europa 0.3.6 and it is reverted.
image

Another question, how the user can forward some data via polkadot-js?=) Because I can only read that=D
image

cmichi: Hmm, could be that `substrate-contracts-node` is somewhat outdated.

cmichi: Found it, the UI doesn't reflect this yet: https://github.com/polkadot-js/apps/issues/6465.

>Another question, how the user can forward some data via polkadot-js?=) Because I can only read that=D

@xgreenx You have to switch the metadata of this contract to some other contract. This can be done by clicking the "Add an existing contract" button, entering the proxy smart contract address and upload a different `.contract` file there. The UI will then use the ABI of this contract to interact with the proxy.

cmichi:
I've reverted returning an `Err` to the `assert!` for now, since the UI doesn't support displaying this yet and hence `ink-waterfall` can't test for it. Will add it back once the UI exposes it.

Issue #906, opened 2021-09-06
Remove all state rent related complexity


cmichi:
Follow-up to https://github.com/paritytech/substrate/pull/9669.

## ToDo

- [ ] Remove all logic originally introduced for state rent ‒ tombtone, deposit, rent status, etc..

Robbepop:
There is a lot more tied to the decision to entirely disregard state rent for ink!.
For example, we will need a partial rewrite of our storage infrastructure and traits as the `SpreadLayout` and `PackedLayout` clear functionality will no longer really be needed.
Many `storage` facilities have been written in a way to allow for automatic removal of unused storage which also is no longer necessary with this decision and therefore those data structures should be entirely rewritten or even redesigned in order to allow to maximum performance.
We can keep `SpreadLayout` and `PackedLayout` traits and the general infrastructure that is inferred by them but there is definitely a lot of cruft going to appear once we are decided to drop state rent entirely.

cmichi:
>[…] there is definitely a lot of cruft going to appear once we are decided to drop state rent entirely.

State rent was removed entirely from the `contracts` pallet with https://github.com/paritytech/substrate/pull/9669, the reasoning is described further in the PR. Right now the methods `seal_rent_status` et al. still exist, so existing contracts don't break. They are only stubs though and return constant values.

The Substrate issue contains close to ten follow-up issues in different repos, where state rent now has been removed as well (or is about to be removed) as a consequence.

Robbepop:
Okay, so state rent removal is pretty much set into stone from what this reads.
If that's the case we will require some ink! adjustments as well that go well beyond removing those APIs.

Robbepop:
I thought about it for a moment. Maybe it makes sense to create a project to have a platform where we can discuss what needs to be done for ink! in order to adjust the whole codebase for a world without state rent.
Mainly this will affect `ink_storage` and partially `ink_env`.
There is going to be the need to rewrite a lot of `ink_storage` data structures, redefine the base traits `SpreadLayout` and `PackedLayout` (e.g. remove the `clear` parts) and redesign utilities such as `ink_storage::Box`.

HCastano:
We remove state rent et. al a while ago

PR #1023, opened 2021-11-17
Add `remove` method to `Mapping`

HCastano:
While updating one of the example contracts I realized that this would be a useful method
to have.

### DIFF
--- a/crates/storage/src/lazy/mapping.rs
+++ b/crates/storage/src/lazy/mapping.rs
@@ -206,7 +206,7 @@ mod tests {
 
     impl<T> PackedLayout for DeepClean<T>
     where
-        T: PackedLayout + scale::Encode + scale::Decode,
+        T: PackedLayout + scale::EncodeLike<T>,
     {
         fn pull_packed(&mut self, at: &Key) {
             <T as PackedLayout>::pull_packed(&mut self.0, at);
--- a/crates/storage/src/lazy/mapping.rs
+++ b/crates/storage/src/lazy/mapping.rs
@@ -20,7 +20,6 @@
 //! Instead it is just a simple wrapper around the contract storage facilities.
 
 use crate::traits::{
-    clear_packed_root,
     pull_packed_root_opt,
     push_packed_root,
     ExtKeyPtr,
@@ -97,9 +96,17 @@ where
         Q: scale::EncodeLike<K>,
     {
         let storage_key = self.storage_key(&key);
-        let v = self.get(key)?;
-        clear_packed_root(&v, &storage_key);
-        Some(v)
+        let value = self.get(key)?;
+
+        if <V as SpreadLayout>::REQUIRES_DEEP_CLEAN_UP {
+            // There are types which need to perform some action before being cleared
+            // (e.g `storage::Box`), here we indicate to those types that they should
+            // start tidying up.
+            <V as PackedLayout>::clear_packed(&value, &storage_key);
+        }
+        ink_env::clear_contract_storage(&storage_key);
+
+        Some(value)
     }
 
     /// Returns a `Key` pointer used internally by the storage API.
--- a/crates/storage/src/lazy/mapping.rs
+++ b/crates/storage/src/lazy/mapping.rs
@@ -89,24 +89,19 @@ where
     }
 
     /// Clears the value at `key` from storage.
-    ///
-    /// Returns the value at the key if it previously existed, otherwise returns `None`.
-    pub fn remove<Q>(&self, key: Q) -> Option<V>
+    pub fn clear_entry<Q>(&self, key: Q)
     where
         Q: scale::EncodeLike<K>,
     {
         let storage_key = self.storage_key(&key);
-        let value = self.get(key)?;
-
         if <V as SpreadLayout>::REQUIRES_DEEP_CLEAN_UP {
-            // There are types which need to perform some action before being cleared
-            // (e.g `storage::Box`), here we indicate to those types that they should
-            // start tidying up.
-            <V as PackedLayout>::clear_packed(&value, &storage_key);
+            // There are types which need to perform some action before being cleared. Here we
+            // indicate to those types that they should start tidying up.
+            if let Some(value) = self.get(key) {
+                <V as PackedLayout>::clear_packed(&value, &storage_key);
+            }
         }
         ink_env::clear_contract_storage(&storage_key);
-
-        Some(value)
     }
 
     /// Returns a `Key` pointer used internally by the storage API.
@@ -185,6 +180,45 @@ const _: () = {
 mod tests {
     use super::*;
 
+    /// A dummy type which `REQUIRES_DEEP_CLEAN_UP`
+    #[derive(PartialEq, Debug, scale::Encode, scale::Decode)]
+    struct DeepClean<T>(T);
+
+    impl<T> SpreadLayout for DeepClean<T>
+    where
+        T: PackedLayout,
+    {
+        const FOOTPRINT: u64 = 1;
+        const REQUIRES_DEEP_CLEAN_UP: bool = true;
+
+        fn pull_spread(ptr: &mut KeyPtr) -> Self {
+            DeepClean(<T as SpreadLayout>::pull_spread(ptr))
+        }
+
+        fn push_spread(&self, ptr: &mut KeyPtr) {
+            <T as SpreadLayout>::push_spread(&self.0, ptr)
+        }
+
+        fn clear_spread(&self, ptr: &mut KeyPtr) {
+            <T as SpreadLayout>::clear_spread(&self.0, ptr)
+        }
+    }
+
+    impl<T> PackedLayout for DeepClean<T>
+    where
+        T: PackedLayout + scale::Encode + scale::Decode,
+    {
+        fn pull_packed(&mut self, at: &Key) {
+            <T as PackedLayout>::pull_packed(&mut self.0, at);
+        }
+        fn push_packed(&self, at: &Key) {
+            <T as PackedLayout>::push_packed(&self.0, at);
+        }
+        fn clear_packed(&self, at: &Key) {
+            <T as PackedLayout>::clear_packed(&self.0, at);
+        }
+    }
+
     #[test]
     fn insert_and_get_work() {
         ink_env::test::run_test::<ink_env::DefaultEnvironment, _>(|_| {
@@ -209,12 +243,46 @@ mod tests {
     }
 
     #[test]
-    fn can_remove_entries() {
+    fn can_clear_entries() {
         ink_env::test::run_test::<ink_env::DefaultEnvironment, _>(|_| {
+            // Given
             let mut mapping: Mapping<u8, u8> = Mapping::new([0u8; 32].into());
+            let mut deep_mapping: Mapping<u8, DeepClean<u8>> =
+                Mapping::new([1u8; 32].into());
+
             mapping.insert(&1, &2);
-            assert_eq!(mapping.remove(&1), Some(2));
-            assert_eq!(mapping.remove(&1), None);
+            assert_eq!(mapping.get(&1), Some(2));
+
+            deep_mapping.insert(&1, &DeepClean(2));
+            assert_eq!(deep_mapping.get(&1), Some(DeepClean(2)));
+
+            // When
+            mapping.clear_entry(&1);
+            deep_mapping.clear_entry(&1);
+
+            // Then
+            assert_eq!(mapping.get(&1), None);
+            assert_eq!(deep_mapping.get(&1), None);
+
+            Ok(())
+        })
+        .unwrap()
+    }
+
+    #[test]
+    fn can_clear_unexistent_entries() {
+        ink_env::test::run_test::<ink_env::DefaultEnvironment, _>(|_| {
+            // Given
+            let mapping: Mapping<u8, u8> = Mapping::new([0u8; 32].into());
+            let deep_mapping: Mapping<u8, DeepClean<u8>> = Mapping::new([1u8; 32].into());
+
+            // When
+            mapping.clear_entry(&1);
+            deep_mapping.clear_entry(&1);
+
+            // Then
+            assert_eq!(mapping.get(&1), None);
+            assert_eq!(deep_mapping.get(&1), None);
 
             Ok(())
         })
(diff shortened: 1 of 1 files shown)


crates/storage/src/lazy/mapping.rs
```diff
@@ -85,14 +86,27 @@ where
...
+        clear_packed_root(&v, &storage_key);
+        Some(v)
     }
 
     /// Returns a `Key` pointer used internally by the storage API.
     ///
     /// This key is a combination of the `Mapping`'s internal `offset_key`
     /// and the user provided `key`.
-    fn storage_key<Q>(&self, key: Q) -> Key
+    fn storage_key<Q>(&self, key: &Q) -> Key
```
Robbepop, line 109: I dont think this change makes sense since `EncodeLike` already allows to take references.

crates/storage/src/lazy/mapping.rs
```diff
@@ -85,14 +86,27 @@ where
...
+
+    /// Clears the value at `key` from storage.
+    ///
+    /// Returns the value at the key if it previously existed, otherwise returns `None`.
+    pub fn remove<Q>(&self, key: Q) -> Option<V>
+    where
+        Q: scale::EncodeLike<K>,
+    {
+        let storage_key = self.storage_key(&key);
+        let v = self.get(key)?;
```
Robbepop, line 100: This always loads the value from storage before clearing it. This is not always needed for all types and thus adds extra overhead and therefore no longer follows the zero-overhead principles. We should make sure to only load data from storage when actually needed.

crates/storage/src/lazy/mapping.rs
HCastano: The compiler says otherwise 😛

crates/storage/src/lazy/mapping.rs
HCastano: How do you recommend we do this? Is there an easy way to check if a value already exists in storage without loading it?

xgreenx: You don't need to clear packet. You only need to do a `ink_env::clear_contract_storage(root_key)`. It will clear storage cell if it is not empty

Robbepop: > You don't need to clear packet. You only need to do a `ink_env::clear_contract_storage(root_key)`. It will clear storage cell if it is not empty

Unfortunately this is not entirely correct since there are `T` with `T: PackedLayout` where `PackedLayout` actually does something important upon clean-up. An example for this is the `ink_storage::Box` type.

crates/storage/src/lazy/mapping.rs
Robbepop: Then there might be some misunderstanding or something might be wrong that we need to inspect further:
https://docs.rs/parity-scale-codec/2.3.1/parity_scale_codec/trait.EncodeLike.html#impl-EncodeLike%3CT%3E-for-%26T

HCastano: So if I take out the reference from the method signature I get the following error:

```
error[E0277]: the trait bound `&Q: EncodeLike<K>` is not satisfied
  --> crates/storage/src/lazy/mapping.rs:99:44
   |
99 |         let storage_key = self.storage_key(&key);
   |                                ----------- ^^^^ the trait `EncodeLike<K>` is not implemented for `&Q`
   |                                |
   |                                required by a bound introduced by this call
   |
help: consider removing the leading `&`-reference
   |
99 -         let storage_key = self.storage_key(&key);
99 +         let storage_key = self.storage_key(key);
   |
help: consider extending the `where` bound, but there might be an alternative better way to express this requirement
   |
69 |     V: PackedLayout, &Q: EncodeLike<K>
   |                    ~~~~~~~~~~~~~~~~~~~

For more information about this error, try `rustc --explain E0277`.
```

> Then there might be some misunderstanding or something might be wrong that we need to inspect further:
https://docs.rs/parity-scale-codec/2.3.1/parity_scale_codec/trait.EncodeLike.html#impl-EncodeLike%3CT%3E-for-%26T

This doesn't mean that you can put a `&T` when a `T` is expected by a method signature.

```rust
use parity_scale_codec::{EncodeLike, Encode};

fn foo<T: Encode, U: EncodeLike<T>>(t: U) {
    unimplemented!()
}

fn bar<T: Encode, U: EncodeLike<T>>(t: &U) {
    unimplemented!()
}

fn test() {
    foo::<u8, u8>(1u8);
    foo::<u8, u8>(&1u8); // Doesn't compile

    bar::<u8, u8>(&1u8);
    bar::<u8, u8>(1u8); // Doesn't compile
}
```

There was also [a comment by Alex](https://github.com/paritytech/ink/pull/979/files#r743079204) suggesting we turn this into a reference anyways to alleviate some costs in Wasm.

Robbepop: That's because you should not use `let storage_key = self.storage_key(&key);` but simply `let storage_key = self.storage_key(key);` since `key: Q` is already bound to be `EncodeLike`.

crates/storage/src/lazy/mapping.rs
HCastano: So should we be using `ink_env::get_contract_storage()` instead?

Robbepop: https://github.com/paritytech/ink/blob/7432565b7e91cbe41f47be57ee0a344429724cc8/crates/storage/src/traits/mod.rs#L224

Robbepop: The `lazy` storage facilities implement exactly this for single cells, take a look:
https://github.com/paritytech/ink/blob/7432565b7e91cbe41f47be57ee0a344429724cc8/crates/storage/src/lazy/lazy_imap.rs#L293

xgreenx: I think the approach with `SpreadLayout::REQUIRES_DEEP_CLEANUP` is good. Later we can simplify it during https://github.com/paritytech/ink/issues/906

crates/storage/src/lazy/mapping.rs
HCastano: Since I want to call `Mapping::get(key: Q)` after computing the storage key I can't do that since it would move `key`

crates/storage/src/lazy/mapping.rs
```diff
@@ -85,14 +85,35 @@ where
...
+
+    /// Clears the value at `key` from storage.
+    ///
+    /// Returns the value at the key if it previously existed, otherwise returns `None`.
+    pub fn remove<Q>(&self, key: Q) -> Option<V>
+    where
+        Q: scale::EncodeLike<K>,
+    {
+        let storage_key = self.storage_key(&key);
+        let value = self.get(key)?;
```
Robbepop, line 99: The whole point of `REQUIRES_DEEP_CLEAN_UP` is to be able to move this particular line of code inside the if-then clause with the premise that for most common types it evaluates to `false`.

HCastano: Sorry, but this doesn't make sense to me given the API I want to achieve with `remove`. In the non `REQUIRES_DEEP_CLEAN_UP` case how would we return the value back to the user?

Robbepop: I would not return any value in the first place. ;)
The `Mapping` is a fundamental low-level storage facility. Therefore we should only provide APIs with which we can efficiently operate.
The `remove` method therefore should not return a value since then we'd be required to always costly load a value even though we could avoid doing so in the most common cases.
If we later find out that getting a value and then removing it (simulating a `remove` that returns the removed value) is more costly than a built-in `Mapping::remove_get` that is implemented as you currently do it, then we might add that method later on.

Robbepop: Note that for example [`LazyIndexMap::clear_packed_at`](https://github.com/paritytech/ink/blob/master/crates/storage/src/lazy/lazy_imap.rs#L293) also does _not_ return the removed value for the same reason. We might even want to name both methods the same although I am not sure if `clear_packed_at` is actually a decent name.

HCastano: Thanks for clarifying how you think the API of `remove` should be. I was modelling it off `HashMap/BTreeMap::remove()` which return the removed element if it exists. This would help make `Mapping` a drop-in replacement in some of our example contracts

crates/storage/src/lazy/mapping.rs
```diff
@@ -164,6 +180,45 @@ const _: () = {
 mod tests {
     use super::*;
 
+    /// A dummy type which `REQUIRES_DEEP_CLEAN_UP`
+    #[derive(PartialEq, Debug, scale::Encode, scale::Decode)]
+    struct DeepClean<T>(T);
```
HCastano, line 185: I wasn't able to find any of the "default" types that we implement `SpreadLayout` for which had `REQUIRES_DEEP_CLEAN_UP` set to `true`. Let me know if there is such a type so I can use that instead

crates/storage/src/lazy/mapping.rs
```diff
@@ -164,6 +180,45 @@ const _: () = {
...
+        fn pull_packed(&mut self, at: &Key) {
+            <T as PackedLayout>::pull_packed(&mut self.0, at);
+        }
+        fn push_packed(&self, at: &Key) {
+            <T as PackedLayout>::push_packed(&self.0, at);
+        }
+        fn clear_packed(&self, at: &Key) {
+            <T as PackedLayout>::clear_packed(&self.0, at);
+        }
+    }
```
cmichi, line 220: Hmm, I thought we have a derive macro for this somewhere.

cmichi (approved):

crates/storage/src/lazy/mapping.rs
cmichi: So we have `#[derive(SpreadLayout, PackedLayout)]` (see the `multisig` example), can you check if you can get rid of the two impls here with the macro?

crates/storage/src/lazy/mapping.rs
```diff
@@ -85,14 +85,30 @@ where
     where
         Q: scale::EncodeLike<K>,
     {
-        pull_packed_root_opt(&self.storage_key(key))
+        pull_packed_root_opt(&self.storage_key(&key))
+    }
+
+    /// Clears the value at `key` from storage.
+    pub fn clear_entry<Q>(&self, key: Q)
```
cmichi, line 92:
```suggestion
    pub fn remove<Q>(&self, key: Q)
```
I would name it `remove`, since the term entry` doesn't appear anywhere else, there's no `get_entry`, no `insert_entry`. Also it can be misleading, since we also have an entry API for some data structures.

You'll have to adapt the other occurrences as well.

crates/storage/src/lazy/mapping.rs
Robbepop: that won't be possible for the `SpreadLayout` since it heavily diverges from the derive macro automated implementation.
For the `PackedLayout` I am not sure but I would personally keep it without derive because `ink_storage` is a frequently compiled crate and so far we did a good job in not using the derive macros from `ink_storage_derive` within the `ink_storage` crate for the reason of compilation time reduction.

crates/storage/src/lazy/mapping.rs
Robbepop: I too think that `remove` as a name is better.

crates/storage/src/lazy/mapping.rs
Robbepop: `ink_storage::Box` is an example.

Generally `ink_storage::Box` is _the_ interface to work with `SpreadLayout` types from within `PackedLayout` cells.

crates/storage/src/lazy/mapping.rs
```diff
@@ -164,6 +180,45 @@ const _: () = {
...
+
+    impl<T> SpreadLayout for DeepClean<T>
+    where
+        T: PackedLayout,
+    {
+        const FOOTPRINT: u64 = 1;
+        const REQUIRES_DEEP_CLEAN_UP: bool = true;
+
+        fn pull_spread(ptr: &mut KeyPtr) -> Self {
+            DeepClean(<T as SpreadLayout>::pull_spread(ptr))
```
Robbepop, line 195:
```suggestion
            Self(<T as SpreadLayout>::pull_spread(ptr))
```

crates/storage/src/lazy/mapping.rs
HCastano: I want to reserve `remove` in case we want to add an analogue to `HashMap/BTreeMap::remove`. Although I guess we can also use `take` for that. What do you guys think?

crates/storage/src/lazy/mapping.rs
HCastano: Ah I didn't even think of using derive macros for this, I'll give it a go

crates/storage/src/lazy/mapping.rs
HCastano: I tried using a `Box`, but [it uses the `REQUIRES_DEEP_CLEAN_UP` of `T`](https://paritytech.github.io/ink/src/ink_storage/traits/impls/prims.rs.html#237), which are all `false` for primitives types as far as I could tell

crates/storage/src/lazy/mapping.rs
cmichi: I agree. I'm wondering, do you think you'll get away without `take` for the examples?

HCastano: So far I haven't needed it, but I've only gone through `ERC20`, `ERC1155`, and just started `ERC721`. Either way, its behaviour can be replicated by doing `get()` followed by `clear_entry` (or whatever we name it)

crates/storage/src/lazy/mapping.rs
HCastano: Looks like I can't use the `derive` macros within this crate 🤷 Either way, we'd still need at least one manual implementation since we want to set `REQUIRES_DEEP_CLEAN_UP = true`

crates/storage/src/lazy/mapping.rs
Robbepop: You usually use ink_storage::Box with SpreadLayout types that do nit implement PackedLayout. For example other ink_storage::collections types such as ink_storage::Vec.

crates/storage/src/lazy/mapping.rs
Robbepop: The reason is comoilation time and also that most types defined in ink_storage cannot use derives anyways since they are forming the basic building ground. Therefore we do not make use of storage derives in the crate itself.

PR #1024, opened 2021-11-18
Update README for Rococo Deployment

cmichi:
[Preview](https://github.com/paritytech/ink/tree/cmichi-migrate-readme-to-rococo).


HCastano:
This needs a bit more work in terms of how to actually get ROC tokens onto Canvas. I can push some commits with that info later today

HCastano (approved):

Issue #990, opened 2021-10-30
Reduce Wasm code bloat of emitting ink! events


Robbepop:
Currently emitting ink! events is very costly with respect to Wasm file sizes.
The Wasm file size bloat primarily stems from the fact that upon emitting an event we first create an event instances that we immediately encode.
The creation process under Wasm is very costly if an event instance has a large memory footprint.
In the ERC-20 example contract we have several events with a structure like:
```rust
#[ink(event)]
pub struct Transferred {
    from: Option<AccountId>,
    to: Option<AccountId>,
    amount: Balance,
}
```
Where `AccountId` is 256 bit and `Balance` is 128 bit wide. This means that even in packed form the `Transferred` struct has a memory footprint of at least 642 bits. Creating instances of this type in Wasm is extremely costly since Wasm has to use the shadow stack for big types like these. What we should do instead is to use references where possible, however:
```
#[ink(event)]
pub struct Transferred<'a> {
    from: Option<&'a AccountId>,
    to: Option<&'a AccountId>,
    amount: &'a Balance,
}
```
Is currently not an allowed event definition since we require ink! events to be `scale::Encode` _and_ `scale::Decode` where the latter requires `Self: 'static` which is not true for the event definition with references.
Maybe we could drop the latter trait bound since we do not really require events to be decodable - that's just a nice-to-have feature.
OR we may be able to emit event without constructing an event instance in the first place by creating intelligent constructors that work with `EncodeLike<T>` for each event field instead of `T` so it is possible to feed references instead.

ascjones:
As an experiment I crudely removed the topics codegen, which reduced `erc20` based on https://github.com/paritytech/ink/pull/946 from `12.0KB` to `9.9KB`. As a comparison with all calls to `emit_event` commented out, we are at `9.1KB`.

xgreenx:
Maybe we should take into account https://github.com/paritytech/ink/issues/809 during implementation of this issue=)

ascjones:
A promising [experiment](https://github.com/paritytech/ink/compare/aj-events-refs?expand=1): I managed to get `7.9K` (down from `11.0K`) from the latest `master` with making everything a reference:

```
#[ink(event)]
    pub struct Transfer<'a> {
        #[ink(topic)]
        from: Option<&'a AccountId>,
        #[ink(topic)]
        to: Option<&'a AccountId>,
        value: &'a Balance,
    }
```

```
Self::env().emit_event(&Transfer {
    from: None,
[...]

PR #1026, opened 2021-11-19
Extract non-generic part of `push_topic` to reduce code size

ascjones:
Marginal total of `0.3KB` (`10.3` down from `10.6`) on ERC20. 

However it appears to reduce the amount of increase of each additional type of topic (see example below), since the extracted non generic part will not be duplicated each time. e.g. commenting out the `Option<AccountId>` topics

image

`master` `10.2` -> `10.6` = `+0.4`
`this PR` `10.1` -> `10.3` = `+0.2`

Rel #990

### DIFF
--- a/crates/env/src/engine/on_chain/impls.rs
+++ b/crates/env/src/engine/on_chain/impls.rs
@@ -151,6 +151,21 @@ where
     {
         let mut split = self.scoped_buffer.split();
         let encoded = split.take_encoded(topic_value);
+        let result = Self::push_topic_encoded(encoded);
+        self.scoped_buffer.append_encoded(&result);
+    }
+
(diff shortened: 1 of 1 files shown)


Robbepop:
Does this still result in the same gains after moving to the inner function approach?

ascjones:
> Does this still result in the same gains after moving to the inner function approach?

It did for `erc20`, yes. I've just triggered the CI job again so let's see.

ascjones:
Not sure how to get the bot to repost the example contract sizes, this might be the run: https://gitlab.parity.io/parity/ink-waterfall/-/jobs/1232706.

Anyway testing locally revealed the same improvement.

Robbepop (approved):
LGTM

Issue #891, opened 2021-08-11
Adapt `rand-extension` example to pass arguments to the runtime


cmichi:
This is a question that comes up sometimes. We should illustrate how this can be done in our `rand-extension` example.

See this Stack Overflow question for more details: [How do you pass and handle arguments to a `ChainExtension`?](https://stackoverflow.com/questions/68332726/how-do-you-pass-and-handle-arguments-to-a-chainextension).

# ToDo
- [x] Pass an argument from ink! into the runtime's chain extension and do something with it there.
- [x] Answer the Stack Overflow question.

PR #1029, opened 2021-11-22
Update chain extension example to show argument passing

justinFrevert:
This changes the `rand-extension` Ink! example code to pass along an argument from the contract to the chain extension in the runtime. This PR addresses the first ToDo [here](https://github.com/paritytech/ink/issues/891), which is a repeat user question. I verified this works in a Substrate node on my machine.

### DIFF
--- a/examples/rand-extension/lib.rs
+++ b/examples/rand-extension/lib.rs
@@ -86,8 +86,9 @@ mod rand_extension {
             Self::new(Default::default())
         }
 
-        /// Seed a random value by passing some known argument `subject` to the runtime's random
-        /// source. Then, update the current value with the new random value
+        /// Seed a random value by passing some known argument `subject` to the runtime's
+        /// random source. Then, update the current `value` stored in this contract with the
+        /// new random value.
         #[ink(message)]
         pub fn update(&mut self, subject: [u8; 32]) -> Result<(), RandomReadErr> {
             // Get the on-chain random seed
--- a/examples/rand-extension/lib.rs
+++ b/examples/rand-extension/lib.rs
@@ -146,7 +146,7 @@ mod rand_extension {
             assert_eq!(rand_extension.get(), [0; 32]);
 
             // when
-            rand_extension.update().expect("update must work");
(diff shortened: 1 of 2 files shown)


justinFrevert:
The failing CI step seems to be unrelated to this change and prevalent among other current PRs.

examples/rand-extension/lib.rs
```diff
@@ -88,9 +88,9 @@ mod rand_extension {
 
         /// Update the value from the runtimes random source.
```
cmichi, line 89: Could you update this comment here and describe the effect of `subject`?

cmichi:
> The failing CI step seems to be unrelated to this change and prevalent among other current PRs.

It's fixed now, it was an issue with one of the GitHub tokens.

cmichi:
@justinFrevert Could you also merge `master`? We think the contract sizes which were posted might be due to your PR being behind `master`.

justinFrevert:
Thanks for grabbing it, @HCastano

examples/rand-extension/lib.rs
```diff
@@ -86,11 +86,12 @@ mod rand_extension {
             Self::new(Default::default())
         }
 
-        /// Update the value from the runtimes random source.
+        /// Seed a random value by passing some known argument `subject` to the runtime's random
+        /// source. Then, update the current value with the new random value
```
cmichi, line 90:
```suggestion
        /// Seed a random value by passing some known argument `subject` to the runtime's
        /// random source. Then, update the current `value` stored in this contract with the
        /// new random value.
```
```

cmichi (approved):
Thanks @justinFrevert!

cmichi:
@justinFrevert Are you up for answering the StackExchange question as well?

justinFrevert:
> @justinFrevert Are you up for answering the StackExchange question as well?

[I gave it a shot](https://stackoverflow.com/questions/68332726/how-do-you-pass-and-handle-arguments-to-a-chainextension). Let me know how that looks.

cmichi:
Great, thx! I'll also incorporate your answer into our `ink-docs`.

Issue #364, opened 2020-03-22
Uploading wasm file appear function non-existent error


jeremymj:
I want to use the signature verification method in my contract,and add dependency in `cargo.toml` as fellow:
`schnorrkel = { version = "0.9.1",features = ["preaudit_deprecated", "u64_backend"], default-features = false}`  
using `cargo contract build` command  can generate wasm file normally and uploading the wasm file to node,  got some error message:   
```
2020-03-22 22:13:36.028 tokio-blocking-driver DEBUG runtime  DispatchError
2020-03-22 22:13:36.028 tokio-blocking-driver DEBUG runtime  module imports a non-existent function
```
using `cargo-contract` version is 0.5.0 ;
node of substrate version is:  `version 2.0.0-alpha.4-0b3020796-x86_64-linux-gnu`
more about reproduce code, please go to [my repository example](https://github.com/jeremymj/dependcytest)

Robbepop:
Can you please insert your `.wasm` file and tell us how you generate the `.wasm` file?
Did you use `cargo-contract`? If yes, what version? Please inspect with `cargo contract --version`.
Also `preaudit_deprecated` looks like you are using a deprecated feature.

jeremymj:
yes,I use `cargo-contract` generate wasm file and  `cargo-contract`  version is `0.5.0`. 
Based on your suggestions, I updated the test code and removed  `preaudit_deprecated`, but the problem persists.This is the  [wasm file](https://github.com/paritytech/ink/files/4366497/dependcytest.txt) I generated.

Robbepop:
I don't know why it is there but I inspected your `.wasm` file and found the import that is causing this:
```
(import "env" "_ZN4core9panicking5panic17h9ff647c44ac14387E" (func (;6;) (type 1)))
```
To me it looks like a Rust or C++ mangled name.
So some dependency maybe is importing this?

jeremymj:
I don't understand the meaning of this expression `_ZN4core9panicking5panic17h9ff647c44ac14387E`,
This dependency is estimated to be introduced in `schnorrkel`. You can use the github address I provided to check whether the dependency is introduced in the `cargo.lock` file.

athei:
This is rust name mangeling. Something inserted a call to `core::panicking::panic`: https://doc.rust-lang.org/stable/core/panicking

[...]

PR #1034, opened 2021-11-23
Disable overflow checks for examples

cmichi:
Until https://github.com/paritytech/ink/issues/364 is resolved once again.


HCastano (approved):
We need a follow-up PR to `cargo-contract` as well for the `flipper` template

PR #1040, opened 2021-11-24
Text fixes

cmichi:


crates/env/src/api.rs
```diff
@@ -62,12 +62,12 @@ where
 /// # Errors
 ///
 /// If the returned value cannot be properly decoded.
-pub fn transferred_balance<T>() -> T::Balance
+pub fn transferred_value<T>() -> T::Balance
```
HCastano, line 65: Why do you think `value` is better than `balance` here?

crates/env/src/call/call_builder.rs
```diff
@@ -130,7 +130,7 @@ where
 /// ## Example 1: No Return Value
 ///
 /// The below example shows calling of a message of another contract that does
-/// not return any value back to its caller. The called function ...
+/// not return any value back to its caller. The called function…
```
HCastano, line 133: I'd leave the normal set of three periods instead of a single character. Easier to type, and imo easier to read.

If you do want to change them, I'd go with a colon instead.

crates/lang/ir/src/ir/item_impl/callable.rs
```diff
@@ -251,7 +251,7 @@ pub trait Callable {
 /// }
 /// ```
 ///
-/// ... then the selector of `my_message` is composed such as:
+/// …then the selector of `my_message` is composed such as:
```
Robbepop, line 254: somehow I find those unicode triple dots in some positions extremely ugly and not at all an improvement. is it just me?

HCastano (changes requested):
Sorry, but I hate the ellipsis related changes 😅

crates/lang/macro/src/lib.rs
```diff
@@ -483,7 +483,7 @@ pub fn selector_bytes(input: TokenStream) -> TokenStream {
 ///         #[ink(message, payable)]
 ///         pub fn fund(&self) {
 ///             let caller = self.env().caller();
-///             let value = self.env().transferred_balance();
+///             let value = self.env().transferred_value();
```
Robbepop, line 486: isn't `transferred_balance` the better or more concrete name actually? why was `transferred_value` chosen over `transferred_balance` if what we handle with are `Balance` types?

crates/lang/macro/src/lib.rs
```diff
@@ -515,7 +515,7 @@ pub fn selector_bytes(input: TokenStream) -> TokenStream {
 ///     #[ink(storage)]
 ///     pub struct Erc20 {
 ///         total_supply: Balance,
-///         // more fields ...
+///         // more fields…
```
Robbepop, line 518: I'd appreciate if there was a space in those cases just what it has been before. or is this bad style that I have to myself adjust personally?

cmichi:
The reason for the ellipsis change:

* Unicode exists for a reason, I hope nobody uses `->` anymore instead of `➔`.
* The `...` appear in doc comments, which are generated into HTML. By using the proper Unicode symbol we allow the frontend to decide on its own how it wants to depict them. Each font typically has a specific symbol for those three dots, it's just that in monospace they appear as one character. In normal spaced fonts it looks different. Like here: …

I anticipated that this is a hugely controversial change. Let's wait a bit to see if more people chime in.

Regarding the `transferred_balance` vs. `transferred_value`: I'm totally unopinionated here, I just want to make it consistent and the latter term was used more prevalent.

cmichi:
@HCastano @Robbepop I've updated the PR to make it less controversial, I've reverted the replacements in monospace code. If you still have issues with the PR post some rational arguments, "more ugly" is not one of those.

crates/lang/ir/src/ir/item_impl/callable.rs
```diff
@@ -221,7 +221,7 @@ pub trait Callable {
 /// }
 /// ```
 ///
-/// ... then the selector of `my_message` is simply `0xDEADBEEF` since it overrides
+/// …then the selector of `my_message` is simply `0xDEADBEEF` since it overrides
```
HCastano, line 224: I'd still get rid of the ellipsis unicode character. Tbh I don't even know what key combination to press to plop one out. I'm sure I could figure it out (or copy the existing ones in the code), but keeping three periods makes our life easier when writing docs. Secondly, if you're worried about rendering, any clever enough font (e.g [FiraCode](https://github.com/tonsky/FiraCode/)) will be able to produce nice ligatures out three periods (or `->` as you mentioned during the last round of review).

HCastano (commented):
Again, the only thing I don't like are the ellipsis. I gave a better explanation this time around

crates/lang/ir/src/ir/item_impl/callable.rs
cmichi: >makes our life easier when writing docs

This shouldn't be the goal, we should aim for the best user experience, not for our best experience when writing code.

>any clever enough font (e.g FiraCode) will be able to produce nice ligatures out three periods

You linked a mono space font, but the changes I am making are in for the non-monospace rustdoc-generated HTML and there is no font which will modify the generated HTML into some other symbol. Ligatures are about making characters which appear next to each other appear nicely together (e.g. fi), they're not about merging characters into a different symbol.

HCastano: If the developers don't have a good experience while writing docs, docs aren't going to get written, and the users lose either way 😉

HCastano (approved):
I still disagree with the unicode ellipsis, but we've already bike-shed over this enough 😅

PR #1043, opened 2021-11-25
Release ink! 3.0.0-rc7

cmichi:
[Preview here](https://github.com/paritytech/ink/blob/cmichi-release-ink-rc7/RELEASES.md).

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -2,15 +2,28 @@
 
 This is the 7th release candidate for ink! 3.0.
 
+Since our last release candidate we implemented a number of contract size improvements.
+With those improvements the size of our `erc20` example has reduced significantly:
+
+|          |             | Release Build with `cargo-contract` |
+|:---------|:------------|:------------------------------------|
+| `erc20`  | `3.0.0-rc6` | 29.3 K                              |
+| `erc20`  | `3.0.0-rc7` | 10.4 K                              |
+
+The savings apply partly to our other examples; for `erc20` they are most
+significant since it has been migrated to use a new [`Mapping`](https://paritytech.github.io/ink/ink_storage/lazy/struct.Mapping.html)
+data structure, which we introduce with this release candidate.
+The other examples will be migrated to this new data structure as a next step.
+
 ## Added
-- The ink! codegen now heavily relies on static type information based on traits defined in `ink_lang`.
+- Added support for wildcard selectors ‒ [#1020](https://github.com/paritytech/ink/pull/1020).
+    - This enables writing upgradable smart contracts using the proxy/forward pattern.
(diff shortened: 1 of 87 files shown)


RELEASES.md
```diff
@@ -1,53 +1,72 @@
...
       definition and ink! trait message respectively.
+- We switched to Rust edition 2021 ‒ [#977](https://github.com/paritytech/ink/pull/977).
+- Update chain extension example to show argument passing ‒ [#1029](https://github.com/paritytech/ink/pull/1029).
 
 ## Fixed
-- Contracts that are compiled as root (the default) now properly revert the transaction if a message
-  returned `Result::Err`.
-    - This does not apply to ink! smart contracts that are used as dependencies. Therefore it is still possible to match against a result return type for a called dependency.
+- Contracts now revert the transaction if an ink! message returns `Result::Err` ‒ [#975](https://github.com/paritytech/ink/pull/975), [#998](https://github.com/paritytech/ink/pull/998).
+    - **Note:** This does not apply to ink! smart contracts which are used as dependencies of a contract!<br/>
```
Robbepop, line 62: This was a bug and has been fixed. No longer holds true.

Robbepop (approved):
LGTM

RELEASES.md
cmichi: So if the `adder` in our `delegator` example hypothetically would return `Result::Err` the entire transaction is reverted? So no way to match on `Result` in the calling contract?

Robbepop: Not the entire transaction is reverted but the state transition of the Adder contract call.

xgreenx:
I think that [bug](https://github.com/paritytech/ink/issues/982) is critical for that release because a workaround doesn't exist.

Maybe you have time to fix that or check https://github.com/paritytech/ink/pull/983 with my suggestion about the fix?

Robbepop:
> I think that [bug](https://github.com/paritytech/ink/issues/982) is critical for that release because a workaround doesn't exist.
> 
> Maybe you have time to fix that or check https://github.com/paritytech/ink/pull/983 with my suggestion about the fix?

We can fix this bug for another RC. no need to delay ghis release imo.

xgreenx:
> We can fix this bug for another RC. no need to delay ghis release imo.

For OpenBrush it means that we will use `rc6`=( If you agree with the suggested fix I will implement it today and you can release it=)

Robbepop:
I will need more time for that. We will be trying to release another smaller release very soon. A smaller follow-up release makes sense anyways since this one is already quite huge.

xgreenx:
Okey=( Hope the next RC will be soon=) Because that bug is a blocker for us

Robbepop:
> Okey=( Hope the next RC will be soon=) Because that bug is a blocker for us

The issues will be put on top of my priority list. ;)

PR #1049, opened 2021-11-25
Explicitly link against rlibc to get non buggy version of memcpy

### LINKED ISSUE #364: Uploading wasm file appear function non-existent error

athei:
Fixes #364

The "ghost import" is called by the compiler inserted `memcpy` function. This is a workaround by providing a non buggy version of this functions ourselves.

### DIFF
--- a/crates/env/Cargo.toml
+++ b/crates/env/Cargo.toml
@@ -30,6 +30,7 @@ static_assertions = "1.1"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 libsecp256k1 = { version = "0.7.0", default-features = false }
+rlibc = "1"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 ink_engine = { version = "3.0.0-rc6", path = "../engine/", default-features = false, optional = true }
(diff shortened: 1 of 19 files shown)


Robbepop:
I wonder what effects this has on the Wasm file sizes and compilation times but I guess `rlibc` compilation time is rather unsignificant.

athei:
I guess file sites in this PR will get up because I re-enabled overflow-checks. But look at this crate. It is tiny.

cmichi (approved):
:heart:

PR #1052, opened 2021-11-26
Switch from `libsecp256k1` to `secp256k1`

cmichi:
Substrate also wants to switch at some point.

This also fixes the failing `cargo-contract` Windows CI, due to incompatibilities of `libsecp256k1`.

### DIFF
--- a/crates/engine/Cargo.toml
+++ b/crates/engine/Cargo.toml
@@ -23,7 +23,7 @@ sha3 = { version = "0.9" }
 blake2 = { version = "0.9" }
 
 # ECDSA for the off-chain environment.
-libsecp256k1 = { version = "0.7.0" }
+secp256k1 = { version = "0.20.3", features = ["recovery"] }
 
 [features]
 default = ["std"]
--- a/crates/engine/src/ext.rs
+++ b/crates/engine/src/ext.rs
@@ -387,11 +387,13 @@ impl Engine {
         message_hash: &[u8; 32],
         output: &mut [u8; 33],
     ) -> Result {
-        use libsecp256k1::{
-            recover,
+        use secp256k1::{
+            recovery::{
+                RecoverableSignature,
+                RecoveryId,
+            },
             Message,
-            RecoveryId,
-            Signature,
+            Secp256k1,
         };
 
         // In most implementations, the v is just 0 or 1 internally, but 27 was added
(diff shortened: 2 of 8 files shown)


Robbepop:
It seems that the new crate significantly increases the Wasm file sizes. Is there anything we can do to counteract that? Maybe features we can enable/disable for the crate?

cmichi:
> It seems that the new crate significantly increases the Wasm file sizes. Is there anything we can do to counteract that? Maybe features we can enable/disable for the crate? Or obvious/low hanging optimizations on ours or their side with respect to the APIs?

So I've investigated it more by now. It's only `erc1155` for which the size increases by 50 bytes. It's not clear to me why exactly that happens, but it goes back to the function `eth_compatibility::to_eth_address` which imports `secp256k1::PublicKey`. If I comment the implementation of that function out and just return a default value there, the `erc1155` size shrinks. Note that neither this function, nor the crate, is in any way called by `erc1155`. For whatever reason it increases its size though slightly.

I won't investigate further right now, but will create a follow-up issue to move the no-std `secp256kp1` stuff from `eth_compatibility` into the pallet instead.

HCastano (commented):
The code looks fine,  but I don't quite understand the motivation behind the change.

- For the Windows CI failures, can you provide a link to a failed `cargo-contract` CI run for a bit of context?
- Do you know anything about why Substrate wants to switch from `libsecp256k1`? Maybe an issue or a PR you can link to?

Maybe its the Rust-maxi in me, but unless we have a good reason to I'd rather stick to a pure Rust library than switch to a Rust wrapper around a C library 🦀

cmichi:
> The code looks fine, but I don't quite understand the motivation behind the change.
> 
> * For the Windows CI failures, can you provide a link to a failed `cargo-contract` CI run for a bit of context?

Yeah, it's the CI on `master` for `cargo-contract`: https://github.com/paritytech/cargo-contract/runs/4356310675?check_suite_focus=true.

A third-party dep of `libsecp256k1` has an issue with Windows (the `/` as path separator in the log output below):
```
  --> C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\crunchy-0.2.2\src\lib.rs:36:1
   |
36 | include!(concat!(env!("OUT_DIR"), "/lib.rs"));
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: build failed, waiting for other jobs to finish...
error: couldn't read \\?\C:\Users\runneradmin\AppData\Local\Temp\cargo-contract.test.uBebfD\new_project_1\target\ink\release\build\crunchy-3be6a8d59ba5eab4\out/lib.rs: The filename, directory name, or volume label syntax is incorrect. (os error 123)
```

There are ways to work around that ofc, like fixing that dep or pinning a version maybe, but since the goal is the other crate either way we can also just switch now.

cmichi:
>Do you know anything about why Substrate wants to switch from `libsecp256k1`? Maybe an issue or a PR you can link to?

Found the issue: https://github.com/paritytech/substrate/issues/8089.

HCastano (approved):
Thanks for the context!

PR #1054, opened 2021-11-28
Actualizing of the size of the `delegator` example

xgreenx:
Extracted that change from [here](https://github.com/paritytech/ink/pull/1017#issuecomment-980830916). 

Using `workspace` forces the compiler to keep stuff for debugging. It increases the size of `adder`, `subber`, `accumulator`. 

`accumulator`: `Original wasm size: 26.8K, Optimized: 6.2K` -> `Original wasm size: 20.5K, Optimized: 1.7K`
`subber`: `Original wasm size: 28.4K, Optimized: 7.4K` -> `Original wasm size: 21.5K, Optimized: 2.9K`
`adder`: `Original wasm size: 28.4K, Optimized: 7.4K` -> `Original wasm size: 21.5K, Optimized: 2.9K`

I think it is better to remove it to track the actual size of these contracts.
CI fails with that change due to a strange error. I think it is related to caching, so created that PR separately. Hope you will find the reason=)

### DIFF
--- a/examples/delegator/Cargo.toml
+++ b/examples/delegator/Cargo.toml
@@ -12,11 +12,11 @@ ink_storage = { version = "3.0.0-rc7", path = "../../crates/storage", default-fe
 ink_lang = { version = "3.0.0-rc7", path = "../../crates/lang", default-features = false }
 
 scale = { package = "parity-scale-codec", version = "2", default-features = false, features = ["derive"] }
+scale-info = { version = "1", default-features = false, features = ["derive"], optional = true }
 
 adder = { version = "3.0.0-rc7", path = "adder", default-features = false, features = ["ink-as-dependency"] }
 subber = { version = "3.0.0-rc7", path = "subber", default-features = false, features = ["ink-as-dependency"] }
 accumulator = { version = "3.0.0-rc7", path = "accumulator", default-features = false, features = ["ink-as-dependency"] }
-scale-info = { version = "1", default-features = false, features = ["derive"], optional = true }
 
 [lib]
 name = "delegator"
@@ -41,10 +41,3 @@ std = [
     "accumulator/std",
 ]
 ink-as-dependency = []
-
-[workspace]
-members = [
-    "accumulator",
-    "adder",
-    "subber",
-]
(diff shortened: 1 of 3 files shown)


Robbepop:
Thanks for debugging this. I was already wondering why those delegator sub contracts were always a lot bigger than for example the very similar Incrementer example contracts.

@ascjones @cmichi Is there a way me might be able to mitigate this through a pass via cargo-contract?

cmichi:
Hey those are some really nice results, @xgreenx !

I've debugged the CI failure and I think it was just some temporary thing, the CI in https://github.com/paritytech/ink/pull/1055 builds all contracts fine. The results are not yet posted there due to some issue with the posting bot, I'll fix that next.

> @ascjones @cmichi Is there a way me might be able to mitigate this through a pass via cargo-contract?

Yes, I think that's the right approach and I've created https://github.com/paritytech/cargo-contract/issues/377 for it. @xgreenx Are you up for the issue on the `cargo-contract` side?

xgreenx:
> Are you up for the issue on the cargo-contract side?

I don't know how to ignore the `[workspace]` from parent `Cargo.toml`=) The problem is that `Cargo.toml` in the `delegate` folder affects the build process in the `delegatore/adder` folder. I will try to investigate that question but I'm not sure about the result=)

xgreenx:
https://github.com/paritytech/cargo-contract/pull/378 fixes the issue=) So, I think we can close this PR and wait for the fix from `cargo-contract`.

xgreenx:
@cmichi I thought the CI uses the latest `cargo-contract`=)

cmichi:
> @cmichi I thought the CI uses the latest `cargo-contract`=)

Yep, it does, but the CI docker container is only rebuild nightly ‒ so in the case of your `cargo-contract` merge yesterday it took some time for that `carg-contract` `master` version to appear in the ink! CI. If you now look at the GitLab logs you'll see
```
$ cargo-contract --version
cargo-contract 0.16.0-3b5d562-x86_64-linux-gnu
```
(which is the latest `cargo-contract` `HEAD`)

The sizes posted by the bot above are when this PR here is compared against ink! `master` ‒ which now doesn't yield any difference anymore, since `master` already uses your merged `cargo-contract` PR.

The bot above still shows one number, but that's only because there is some small noise in one of the late decimal points, you can ignore that.

I'll close this issue now, since it has been solved in `cargo-contract`.

PR #1057, opened 2021-11-30
Use `Mapping` in ERC-1155 Example

HCastano:


examples/erc1155/lib.rs
```diff
@@ -264,7 +266,9 @@ mod erc1155 {
         /// Initialize a default instance of this ERC-1155 implementation.
         #[ink(constructor)]
         pub fn new() -> Self {
-            Default::default()
+            ink_lang::codegen::initialize_contract(|_| {
+                Contract::default();
+            })
```
HCastano, line 271: @Robbepop I found that this wasn't necessary. I could replace

```suggestion
            Default::default()
```

And I could deploy and use the contract fine. It seems like a coincidence since the default `ROOT_KEY` is `0x00` and that's how a default `Mapping` would end up initialized. Any thoughts on this?

examples/erc1155/lib.rs
```diff
@@ -326,21 +334,27 @@ mod erc1155 {
         //
         // Should not be used directly since it's missing certain checks which are important to the
         // ERC-1155 standard (it is expected that the caller has already performed these).
+        //
+        // # Panics
+        //
+        // If `from` does not hold any amount any `token_id` tokens.
```
cmichi, line 340:
```suggestion
        // If `from` does not hold any `token_id` tokens.
```

examples/erc1155/lib.rs
```diff
@@ -326,21 +334,27 @@ mod erc1155 {
...
         ) {
-            self.balances
-                .entry((from, token_id))
-                .and_modify(|b| *b -= value);
+            let mut sender_balance = self
+                .balances
+                .get(&(from, token_id))
+                .expect("Caller should have ensured that `from` holds `token_id`.");
+            sender_balance -= value;
+            self.balances.insert(&(from, token_id), &sender_balance);
```
cmichi, line 353: Oh boi, how ugly this gets. An `entry` API would be so nice.

cmichi (approved):

PR #1058, opened 2021-11-30
Use `Mapping` in ERC-721 Example

HCastano:


examples/erc721/lib.rs
```diff
@@ -130,12 +129,18 @@ mod erc721 {
...
-            }
+            ink_lang::codegen::initialize_contract(|_| {
+                Self {
+                    // TODO: How should this be properly initialized?
+                    //
+                    // Turns out that `token_owner` and `token_approvals` have the same type
+                    // signature and they're both Default initialized to ROOT_KEY [0u8; 32]. This
+                    // means that they end up operating on the same storage keys :(
+                    // token_approvals: Default::default(),
+                    token_approvals: Mapping::new([1u8; 32].into()),
```
HCastano, line 140: @Robbepop looks like we need another piece of "entropy" when [assigning storage keys](https://github.com/paritytech/ink/blob/78b45330e453500a759a2b53b7ab6f16bfb91caa/crates/storage/src/lazy/mapping.rs#L115) in the `Mapping`. 

I haven't been able to figure out what we could use though, do you have any ideas?

examples/erc721/lib.rs
```diff
@@ -439,8 +449,10 @@ mod erc721 {
                     .expect("Cannot get accounts");
             // Create a new contract instance.
             let mut erc721 = Erc721::new();
+            dbg!(erc721.get_approved(1));
```
HCastano, line 452: I'll get rid of these once the initialize thing is sorted

examples/erc721/lib.rs
HCastano: I noticed this "bug" when before swapping the initialization routine to actually use `ink_lang::codegen::initialize_contract`. By using it the `Mapping`'s get intialized with different `offset_key`s, so false alarm 😅

examples/erc721/lib.rs
```diff
@@ -297,16 +310,26 @@ mod erc721 {
...
-            let entry = owned_tokens_count.entry(*to);
-            increase_counter_of(entry);
-            vacant_token_owner.insert(*to);
+
+            let count = owned_tokens_count
+                .get(to)
+                .map(|mut c| {
+                    c += 1;
+                    c
+                })
```
athei, line 327:
```suggestion
                .map(|c| c + 1)
```

examples/erc721/lib.rs
```diff
@@ -325,19 +348,14 @@ mod erc721 {
...
-                Ok(())
+
+            if approved {
+                self.operator_approvals.insert((&caller, &to), &approved);
             } else {
-                match self.operator_approvals.insert((caller, to), approved) {
-                    Some(_) => Err(Error::CannotInsert),
-                    None => Ok(()),
-                }
+                self.operator_approvals.remove((&caller, &to));
```
athei, line 355: Do we still need the `bool` as the value in the mapping after this change? Couldn't it be just `()` now that we only check for existence?

examples/erc721/lib.rs
```diff
@@ -349,43 +367,40 @@ mod erc721 {
...
                 from: caller,
                 to: *to,
                 id,
             });
+
             Ok(())
         }
 
         /// Removes existing approval from token `id`.
         fn clear_approval(&mut self, id: TokenId) -> Result<(), Error> {
```
athei, line 391: It no longer can return an error.

athei (approved):

PR #1059, opened 2021-11-30
Move linting checks to start of CI run

HCastano:
It's really annoying to wait half an hour while the CI does its job, only to realize that
you forgot to run `rustfmt` and now you have to wait another half an hour for the CI to
pass.

This PR moves our linting steps to the beginning of the CI pipeline so that these
failures can be caught and addressed quickly.

This PR also allows the CI build to continue despite the lint steps failing. This is
useful if you care about a step later in the CI pipeline (e.g contract sizes) but don't
necessarily care if your code is formatted properly.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -109,6 +109,10 @@ workflow:
       file:                        false
 
 #### stage:                        lint
+#
+# Note: For all of these lints we `allow_failure` so that the rest of the build can
+# continue running despite them not passing. Merging is still disallowed since (most) of
+# the lint steps are marked as "Required" in GitHub.
 spellcheck:
   stage:                           lint
   <<:                              *docker-env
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -107,6 +108,91 @@ workflow:
       vault:                       cicd/gitlab/$CI_PROJECT_PATH/GITHUB_SSH_PRIV_KEY@kv
       file:                        false
 
+#### stage:                        lint
+#
+# Note: For all of these lints we `allow_failure` so that the rest of the build can
+# continue running despite them not passing. Merging is still disallowed since (most) of
```
cmichi, line 114: Just as an FYI: I checked and it's only `spellcheck` which is currently not required for merging.

cmichi (approved):
Let's try it out.

PR #1061, opened 2021-12-01
Use `Mapping` in DNS Example

HCastano:


examples/dns/lib.rs
```diff
@@ -5,10 +5,10 @@ use ink_lang as ink;
 #[ink::contract]
 mod dns {
     #[cfg(not(feature = "ink-as-dependency"))]
+    use ink_storage::lazy::Lazy;
```
cmichi, line 8: @HCastano Could you explain why this usage needs to be under the feature flag and the others not?

HCastano: The old code was like that. I just removed the feature flag gate, I don't think it's really needed anyways

cmichi (approved):

Issue #1044, opened 2021-11-25
Reflect the `contracts` pallet's switch to charging a deposit for storage in ink!


cmichi:
Once Alex' deposit PR (https://github.com/paritytech/substrate/pull/10082) is merged we need to adapt ink! to reflect these changes. 

From the top of my head this includes remove the notion of a subsistence threshold and adding `storage_limit` as a parameter to the off-chain env. But whoever takes this ticket should look through the PR again to make sure we don't miss anything.

PR #1062, opened 2021-12-01
Remove `BelowSubsistenceThreshold` and `NewContractNotFunded`

cmichi:
Do not merge yet!

Closes https://github.com/paritytech/ink/issues/1044.

Follow-up to https://github.com/paritytech/substrate/pull/10082.

I haven't adapted the off-chain testing environment for the deposit mechanism, I'll create a follow-up for that. Right now, the off-chain env doesn't take the new deposit mechanism into account in any way.

### DIFF
--- a/examples/contract-transfer/lib.rs
+++ b/examples/contract-transfer/lib.rs
@@ -25,7 +25,7 @@ pub mod give_me {
         ///
         /// - Panics in case the requested transfer exceeds the contract balance.
         /// - Panics in case the requested transfer would have brought this
-        ///   contract's balance below the minimum balance (i.e. the chains
+        ///   contract's balance below the minimum balance (i.e. the chain's
         ///   existential deposit).
         /// - Panics in case the transfer failed for another reason.
         #[ink(message)]
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -156,7 +156,7 @@ where
 }
 
 /// Returns the minimum balance that is required for creating an account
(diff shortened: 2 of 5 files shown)


athei (changes requested):
Return codes didn't change for backwards compatibility. We merely stopped emitting some of them. Please check: https://github.com/paritytech/substrate/blob/450cd6804dd77312b27d72ebdebccc4c6027f579/frame/contracts/src/wasm/runtime.rs#L47-L77

cmichi:
@athei Argh, ofc, I've updated the PR.

athei (approved):

examples/contract-transfer/lib.rs
```diff
@@ -39,8 +24,8 @@ pub mod give_me {
         /// # Errors
         ///
         /// - Panics in case the requested transfer exceeds the contract balance.
-        /// - Panics in case the requested transfer would have brought the
-        ///   contract balance below the subsistence threshold.
+        /// - Panics in case the requested transfer would have brought this
+        ///   contract's balance below the minimum balance.
```
HCastano, line 28: What balance is the `minimum balance` referring to in this case?

HCastano (commented):
What's this PR waiting for?

examples/contract-transfer/lib.rs
cmichi: The chains existential deposit, I clarified it in the comment.

PR #1063, opened 2021-12-01
Name API consistently: `transferred_balance` ➔ `transferred_value`

cmichi:
I honestly don't care what we rename it to, *but we should finally make the naming consistent*. 

I'm not opening up the whole thing with renaming endowment to value that was started in https://github.com/paritytech/substrate/pull/10082 here.

### DIFF
--- a/crates/env/src/engine/on_chain/impls.rs
+++ b/crates/env/src/engine/on_chain/impls.rs
@@ -357,7 +357,7 @@ impl TypedEnvBackend for EnvInstance {
         self.get_property_inplace::<T::AccountId>(ext::caller)
     }
 
-    fn transferred_balance<T: Environment>(&mut self) -> T::Balance {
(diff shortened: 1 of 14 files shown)


ascjones:
> but we should finally make the naming consistent

:point_up: :100:

crates/env/src/engine/on_chain/impls.rs
```diff
@@ -357,7 +357,7 @@ impl TypedEnvBackend for EnvInstance {
         self.get_property_inplace::<T::AccountId>(ext::caller)
     }
 
-    fn transferred_balance<T: Environment>(&mut self) -> T::Balance {
+    fn transferred_value<T: Environment>(&mut self) -> T::Balance {
```
ascjones, line 360: If we really wanted to nitpick on the consistency then the `seal` fn `value_transferred` should also be the same :grinning:

ascjones (approved):
LGTM.

PR #1064, opened 2021-12-02
Rename `EcdsaRecover` ➔ `EcdsaRecovery`

cmichi:
@ascjones I'm also thinking about if we should rename `ink_env::ecdsa_recover(…)` to something like `ink_env::ecdsa_recovery(…)` or `ink_env::recover_ecdsa(…)`. What does the native speaker say?

### DIFF
--- a/crates/engine/src/ext.rs
+++ b/crates/engine/src/ext.rs
@@ -92,7 +92,7 @@ define_error_codes! {
     /// recording was disabled.
     LoggingDisabled = 9,
     /// ECDSA pubkey recovery failed. Most probably wrong recovery id or signature.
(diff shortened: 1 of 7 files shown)


ascjones (approved):

ascjones:
> @ascjones I'm also thinking about if we should rename `ink_env::ecdsa_recover(…)` to something like `ink_env::ecdsa_recovery(…)` or `ink_env::recover_ecdsa(…)`. What does the native speaker say?

On the basis that the fn name should be a verb `recover` and that `ecdsa` is the name of the algorithm rather than referring to the public key which is being recovered, I think the current `ink_env::ecdsa_recover(…)` makes the most sense.

Issue #1027, opened 2021-11-21
Make constructors (non)payable


athei:
When https://github.com/paritytech/substrate/pull/10082 is merged there is no longer a need to pass an endowment larger than `0` except when the contract logic itself requires it. This makes `endowment` no different from `value` in a call extrinsic.

Therefore I think we should extend the `payable` logic to constructors. The UI can then remove the endowment field for non payable constructors.

Robbepop:
Can you elaborate on this a bit?
So will it be possible to have `value` _and_ `endownment` fields upon contract instantiation?
Currently in ink! all constructors are implicitly payable since unpayable constructors make no sense since the contract would be evicted immediately after.

athei:
> So will it be possible to have value and endownment fields upon contract instantiation?

No, there is still only `endowment`.

> Currently in ink! all constructors are implicitly payable since unpayable constructors make no sense since the contract would be evicted immediately after.

And this is exactly the part that is about to change soon. Unpayable constructors make sense then.

Robbepop:
So should we non-confusingly rename this to "Allow ink! constructors to be non-payable" ?

athei:
Sure, if that makes more sense to you. The issue is about allowing this `payable` attribute there which is now set by default.

athei:
Needs to be done before the stable release because it is a breaking change.

PR #1065, opened 2021-12-02
Allow specifying `payable` constructors

HCastano:
Before this change ink! constructors were payable by default. This was because all
contracts were required to place an endowment in order to stay alive.

However, with the changes introduced in https://github.com/paritytech/substrate/pull/10082 this is no
longer necessary. Instead we can allow users to specify whether or not they want their
constructors to be `payable`.

We'll need to wait for the Substrate PR to get merged before merging this one. This is
because if we deploy a contract without also initializing it with some value we get a
`NewContractNotFunded` error.

Closes https://github.com/paritytech/ink/issues/1027.

P.S, we'll also want to change `endowment` to `value` to be consistent with
`pallet-contracts`, but that can be done in a follow-up PR.


athei (approved):

crates/lang/tests/ui/contract/fail/constructor-payable-invalid-1.rs
```diff
@@ -6,7 +6,7 @@ mod contract {
     pub struct Contract {}
 
     impl Contract {
-        #[ink(constructor, payable)]
+        #[ink(constructor, payable = true)]
```
cmichi, line 9: Why change the existing test instead of creating a new one for this case?

HCastano: This test was added as a new one [in this commit](https://github.com/paritytech/ink/pull/1065/commits/fb4f13663de38b1b71b64bc2db13dbdf3f1d597c). There was a `git mv` in this [earlier commit](https://github.com/paritytech/ink/pull/1065/commits/2acf3c56896ab193a730ab88898587703bb683f4) though, maybe that's causing the confusion?

Either way, the [history](https://github.com/paritytech/ink/commits/hc-make-constructors-non-payable/crates/lang/tests/ui/contract/fail/constructor-payable-invalid-1.rs)  for this file shows it was only introduced in this PR, so it should be fine imo.

HCastano:
So UI support for this isn't quite there since https://github.com/polkadot-js/apps/issues/6645 hasn't been tackled, but really the only "gotcha" here for users is that if they specify a `value` if deploying a contract with a non-payable constructor they get a `ContractTrapped` error. If we're okay with dealing with that for now I say we merge this.

athei:
Does this PR also includes this information into the metadata? Currently, constructors don't have a payable field.

ascjones:
It doesn't look like it has the extra metadata field, as Jaco says in that issue there should be a `V3` metadata for this including the addition of this field.

It would go in this: https://github.com/paritytech/ink/blob/master/crates/metadata/src/specs.rs#L223

athei:
So in this PR we need to:
- Add the payable field to metadata
- Bump the metadata version

HCastano:
> It doesn't look like it has the extra metadata field, as Jaco says in that issue there should be a `V3` metadata for this including the addition of this field.
> 
> It would go in this: https://github.com/paritytech/ink/blob/master/crates/metadata/src/specs.rs#L223

Ah, okay I didn't realize Jaco was referring to the metadata on our end. Sure I can add this

crates/lang/src/codegen/dispatch/execution.rs
```diff
@@ -73,6 +73,14 @@ where
 /// Configuration for execution of ink! constructor.
 #[derive(Debug, Copy, Clone)]
 pub struct ExecuteConstructorConfig {
+    /// Yields `true` if the ink! constructor accepts payment.
+    ///
+    /// # Note
+    ///
+    /// If no ink! constructor within the same ink! smart contract
+    /// is payable then this flag will be `true` since the check
+    /// then is moved before the constructor dispatch as an optimization.
```
cmichi, line 82: Hmm, not happy about this. This possibility of `payable` being set to `true` despite no payable constructor seems to be destined for a bug. How about we reflect this information e.g. with an enum here as a wrapper around the `bool`?

HCastano: I think there was already a small bug here actually. If there were indeed no payable messages for example, we'd never hit [this check](https://github.com/paritytech/ink/blob/d70973dcf45f72bd18b705e1f24c46d756133e0e/crates/lang/src/codegen/dispatch/execution.rs#L104). Although to be fair, payment would have been denied in [call()](https://github.com/paritytech/ink/blob/d70973dcf45f72bd18b705e1f24c46d756133e0e/crates/lang/codegen/src/generator/dispatch.rs#L4444) (maybe this was the optimization alluded to in the comment?). 

Either way, this was confusing and I think it's unnecessary so I removed it.

cmichi (approved):

cmichi:
Can you create a follow-up issue for `ink-docs`?

crates/lang/codegen/src/generator/dispatch.rs
```diff
@@ -706,7 +732,6 @@ impl Dispatch<'_> {
             );
             let accepts_payment = quote_spanned!(message_span=>
                 false ||
-                !#any_message_accept_payment ||
```
athei, line 709: Can we have a UI test for this: Multiple constructors which are all not payable?

HCastano:
> Can you create a follow-up issue for `ink-docs`?

Done: https://github.com/paritytech/ink-docs/issues/41

Issue #1068, opened 2021-12-07
Remove `libsecp256k1` dependency of `ink_eth_compatibility`


cmichi:
We need the `libsecp256k1` dependency only in the `to_eth_address` function, for getting the uncompressed 65-byte public key from the ECDSA public key.

Ways to remove the dep would be to either add a new API function in the `contracts` pallet, find another no-std compatible crate, or extract the functionality from either `libsecp256k1` or `secp256k1` into `ink_eth_compatibility`.

cmichi:
@athei @agryaznov 

There is not yet an issue for this in `pallet-contracts`, but I think we should implement it there. This would enable us to get rid of the `libsecp256k1` dependency in ink!, thus making contracts using the `ink_eth_compatibility` smaller.

I can imagine having e.g. sth like a `seal_to_eth_address(ecdsa_compressed_public_key: [u8; 32])`.

That's the code which would need to be moved to this new `pallet-contracts` function: https://github.com/paritytech/ink/blob/master/crates/eth_compatibility/src/lib.rs#L99-L119.

athei:
I think there should be some host function available for the runtime to offload that to. Search for that here: https://github.com/paritytech/substrate/blob/master/primitives/io/src/lib.rs

agryaznov:
> I think there should be some host function available for the runtime to offload that to. Search for that here: https://github.com/paritytech/substrate/blob/master/primitives/io/src/lib.rs

Haven't found appropriate function there, will move current impl dependent on `libsecp256k` into pallet.

agryaznov:
We now have `seal_` function for this in pallet_contract. Do we need to implement also some fn to be used from inside ink! contracts? @cmichi

athei:
Of course. Your linked PR should not just remove the dependency but replace it with a call of the seal function.

agryaznov:
Do we really need this `ink_eth_compatibility` crate after all? It isn't being used anywhere, and only has one more fn aside from `to_eth_address()`, which is `to_default_account_id()` and looks like something that should be put out of ink as well (probably to the same ecdsa trait extension in frame_support).

athei:
Yeah I don't think we need this extra crate anymore. I think we only had it to avoid pulling in the crypto dependency. But better wait for @cmichi to comment.

PR #1069, opened 2021-12-07
Do not include `libsecp256k1` on Windows

cmichi:
Fixes the broken `cargo-contract` Windows CI. See https://github.com/paritytech/ink/issues/1068 for more details.

CI fails due to expected incompatible ink! with `substrate-contracts-node` since the deposit PR merge.

### DIFF
--- a/crates/eth_compatibility/src/lib.rs
+++ b/crates/eth_compatibility/src/lib.rs
@@ -95,7 +95,7 @@ impl ECDSAPublicKey {
     // which is incompatible with Windows.
     // We have https://github.com/paritytech/ink/issues/1068 for removing this
     // dependency altogether.
-    #[cfg(not(target_arch = "windows"))]
+    #[cfg(not(target_os = "windows"))]
     pub fn to_eth_address(&self) -> EthereumAddress {
         use ink_env::hash;
         use libsecp256k1::PublicKey;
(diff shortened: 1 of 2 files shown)


crates/eth_compatibility/src/lib.rs
```diff
@@ -91,6 +91,11 @@ impl ECDSAPublicKey {
     ///
     /// assert_eq!(pub_key.to_eth_address().as_ref(), EXPECTED_ETH_ADDRESS.as_ref());
     /// ```
+    // We do not include this function on Windows, since it depends on `libsecp256k1`,
+    // which is incompatible with Windows.
+    // We have https://github.com/paritytech/ink/issues/1068 for removing this
+    // dependency altogether.
+    #[cfg(not(target_arch = "windows"))]
```
HCastano, line 98:
```suggestion
    #[cfg(not(target_os = "windows"))]
```

HCastano (changes requested):
So this is the error message a user on Windows would get if they tried to use `to_eth_address`:

```
error[E0599]: no method named `to_eth_address` found for struct `ECDSAPublicKey` in the current scope
  --> lib.rs:29:28
   |
29 |             let addr = key.to_eth_address();
   |                            ^^^^^^^^^^^^^^ method not found in `ECDSAPublicKey`

For more information about this error, try `rustc --explain E0599`.
```

I wonder if we can have a `cfg_if` check and panic with a better message instead

HCastano (approved):

PR #1070, opened 2021-12-07
Make Vector usable when initialized with `SpreadAllocate`

HCastano:
At the moment the `ink_storage::Vec` type is unusable when initialized using
`SpreadAllocate::allocate_spread`. This is because `Vec::len` does not account that the
vector can be in a state where the it is initialized but there is nothing in storage.

This creates problems when using `Vec::len`, such as during `Drop` ([see here](https://github.com/paritytech/ink/blob/708717bc7e874f6c5393c922a9bd3b4d10878fc9/crates/storage/src/collections/vec/mod.rs#L123)),
or `Vec::push`.

This PR changes the behaviour of `Vec::len()` to return a `0` length if there is no
length in storage. It also changes all uses of `Lazy::get_mut` for modifying `Vec.len` to
use `Lazy::set` instead since that doesn't panic if no item exists in storage.

Currently a draft since there are some other storage data structures which use `Vec`
whose tests need to be updated and I'm not quite sure what the best way to fix those
tests is just yet.

### DIFF
--- a/crates/storage/src/alloc/tests.rs
+++ b/crates/storage/src/alloc/tests.rs
@@ -158,7 +158,7 @@ fn spread_pull_push_works() {
 }
 
 #[test]
-#[should_panic(expected = "encountered empty storage cell")]
+#[should_panic(expected = "index is out of bounds")]
 fn spread_clear_works() {
     run_default_test(|| {
         let alloc = spread_layout_alloc_setup();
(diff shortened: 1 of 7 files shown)


cmichi:
@HCastano Could you fix the failing tests?

cmichi:
@HCastano I can't think of a better approach, so I'm fine with the PR if you fix the other tests.

@ascjones Do you have an opinion?

HCastano:
We got rid of `ink_storage::Vec` in #1111, so I'm gonna close this

PR #1074, opened 2021-12-08
Fix nightly clippy errors

cmichi:
The `master` CI is currently failing because of this (and because of our incompatible `master` with the UI's).

It's a pity that `ink-waterfall` currently does not work due to the UI's, I would be very interested in seeing the gas/size changes. Let's wait with the merge until we have the UI compatible again.


ascjones (approved):

HCastano (approved):
The changes are pretty small, but either way I sampled a few (ERC20, ERC1155, Multisig, and DNS) contracts just to double check the sizes - and there was no change in any of them. So this is good to merge

PR #1036, opened 2021-11-24
Remove the state rent API

cmichi:
Closes https://github.com/paritytech/ink/issues/997.

Follow-up to https://github.com/paritytech/substrate/pull/9669.

I'll remove the subsistence threshold stuff once we merge the deposit PR into Substrate.

Issue #1000, opened 2021-11-02
In cross-call contract, ink! cannot pass compile for the parent contract contains event


atenjin:
**Questions**
It's a bug in ink!, it may appear after pr  https://github.com/paritytech/ink/pull/665

**FAQ**

ink! version: e8d4739649293a458c3958bec802d2d750067d98
cargo-contract version: cargo-contract 0.15.0-unknown-x86_64-linux-gnu

**Describe the bug**
example repo:
https://github.com/atenjin/ink-event-bug

in this repo, test2 contract crate depends on test contract crate. So the contract `Test2` in test2 will use the `TestRef` from the contract `Test` in test.

and in test/lib.rs:L20, I add a event definition, and do nothing.
```rust
    #[ink(event)]
    pub struct OtherThing;
```
and in test2/lib.rs:L23, I aslo define a event:
```rust
    #[ink(event)]
    pub struct Tmp;
```
and emit this event in following code, line 50
```rust
self.env().emit_event(Tmp {});
```

then in test2 directory, execute compile command, it cannot pass compile for the following error:
```bash
error[E0284]: type annotations needed
  --> /home/king/workspace/patractlabs/contracts/tmp/test2/lib.rs:50:24
   |
50 |             self.env().emit_event(Tmp {});
   |                        ^^^^^^^^^^ cannot infer type for type parameter `C` declared on the trait `EmitEvent`
   |
   = note: cannot satisfy `<_ as ContractEventBase>::Type == _`

```
if remove the `emit_event`, then `Test2` can pass the compilation. But we need the event.

And If I remove the event definition in `Test`, then compiling `Test2` can work normally.

and other people so in element, if I change the emit_event like this, can pass the compilation.
```rust
            // self.env().emit_event(Tmp {});
            ink_lang::codegen::EmitEvent::<Test2>::emit_event(self.env(), Tmp {});
```

Obviously, the type inference is broken when Test2 use the contract TestRef.

I think it can be fixed just changing some small part, right? @Robbepop

DukeCaesar:
Same issue.

cmichi:
@ascjones Do you have an idea if this will be fixed with https://github.com/paritytech/ink/pull/1243 ?

ascjones:
Yes it will

yeousunn:
I am having the same issue. Any idea how to go around it?

I am using ink version 4.2.1.

PR #1075, opened 2021-12-08
Remove `Multisig` example contract

HCastano:
This PR removes the `Multisig` example contract in preparation of the (temporary) removal
of the `ink_storage::collections` crate.

This example is the only one we have which makes use of the `ink_storage::{Vec, Stash}`
data structures, which at the moment don't play nice with the non-caching data structures
which we want smart contract developers to use (see #1070 for a little bit of context).

This is a good example though, so once we re-work the `collections` we should bring it
back.


xgreenx:
You don't need to remove the whole example. You easy can use `ink_storage::lazy::Lazy<ink_prelude::vec::Vec<AccountId>>` instead of `StorageVec` and later update it to use a new vector without caching. That example can be useful for people without a new vector=)

athei:
But I think we don't want people to use that either? Doesn't that explode the file sizes?

xgreenx:
`ink_prelude::vec::Vec` takes much less space than `ink_storage::collections::Vec`. The main idea of `multisig` example is to show how to do a cross-contract call, how to define your own structures, how to manage transactions. It is not production example of the multisig, but it shows well how to work with ink!.

HCastano:
@xgreenx so the other problem here is the use of `Stash`, which uses a `LazyIndexMap` internally. I think it's possible to re-write the contract without using it, but I want to push that down the road a little since we're trying to push for a stable release soon

xgreenx:
> @xgreenx so the other problem here is the use of `Stash`, which uses a `LazyIndexMap` internally. I think it's possible to re-write the contract without using it, but I want to push that down the road a little since we're trying to push for a stable release soon

I see. Yes, I think it should use `Mapping` and use a hash of the transactions as `TransactionId` instead of the index in `Stash`=) But if you want to re-write it later, okay. When do you plan to do a stable release? Because it still contains several critical bugs for users=)

HCastano:
> When do you plan to do a stable release? Because it still contains several critical bugs for users=)

@xgreenx So the main things as far as `ink!` is concerned are
- Non-payable constructors, since this is a breaking change
  - This is mostly just waiting for UI support
- Removal of non-caching collections
  - It's easier to add back code than to remove it later
- Removal of the old off-chain environment
- Storage deposit support
  - There's a PR open for this, but is waiting on some off-chain env stuff 
 
We think that any other issues can be addressed in minor releases later. Out of curiosity, what issues do you deem as critical? I suspect https://github.com/paritytech/ink/issues/982, but anything else?

athei:
Ported the multisig example. However, the off chain tests are failing with an error I don't comprehend. Maybe someone can assist me with debugging: https://github.com/paritytech/ink/pull/1078

xgreenx:
> We think that any other issues can be addressed in minor releases later. Out of curiosity, what issues do you deem as critical? I suspect #982, but anything else?

Yes, https://github.com/paritytech/ink/issues/982 is critical because it doesn't allow to use of traits fully and no workaround exists.

https://github.com/paritytech/ink/issues/1000 has a workaround, but it is related to https://github.com/paritytech/ink/issues/809 and I think that resolving it can change the API, so why it is also is critical. 

https://github.com/paritytech/ink/issues/900 requires additional [follow-up change](https://github.com/paritytech/ink/pull/923#discussion_r755387899) which affects selectors in trait.  

Did you clean up all stuff related to that? https://github.com/paritytech/ink/issues/906

HCastano:
> Yes, #982 is critical because it doesn't allow to use of traits fully and no workaround exists.

On the surface this seems like something that we can add after the 3.0 release without breaking anything for users, but I don't have a lot of insight into this tbh.

> #1000 has a workaround, but it is related to #809 and I think that resolving it can change the API, so why it is also is critical.

Again, on the surface this looks like the API could be backwards compatible, but don't quote me on this.

> #900 requires additional follow-up change which affects selectors in trait.

I'm gonna defer to @cmichi here since he's been chatting with you in that issue

> Did you clean up all stuff related to that? #906

So a bunch of stuff was removed in https://github.com/paritytech/ink/pull/1036, but I think there might still be some more to clean up.

Btw, we probably shouldn't be having this discussion here. @cmichi can you enable "Discussions" for the repo so we can move this conversation there?

HCastano:
Closing in favour of https://github.com/paritytech/ink/pull/1078.

PR #1078, opened 2021-12-10
Port multisig example to low level data structures

athei:
This removes all usages of data structures in `collections`. The new code shows some pain points of ink! and the pallet-contracts when removing those collections. I think this is exactly the reason why we should port this example with sub optimal code: We can come back, iterate and observe the improvements.

These are some pain point of the contract which could be fixed but would change the interface so I did leave them as is:

1. Owners are stored in a packed `Vec`. The complexity of adding and removing owners scales linearly with the amount of owners. We might need to cap them.
2. The transactions id are generated from a monotonic counter which will run out if ids eventually. I don't think this will be an issue in practice but we might want to either provide a "defrag" function or let the user pass in the id. I would prefer the latter but it will change the interface.

It also shows some shortcomings of the pallet-contracts map interface. I think we want the following functionality to paller-contracts:
1. When setting a key in a mapping we want to return whether this key already existed
2. Check for existence without reading the value
3. Allow to get and remove and the same time (`take`)

### DIFF
--- a/crates/storage/src/lazy/mapping.rs
+++ b/crates/storage/src/lazy/mapping.rs
@@ -38,10 +38,19 @@ use ink_primitives::Key;
 
 /// A mapping of key-value pairs directly into contract storage.
 #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
-#[derive(Default)]
 pub struct Mapping<K, V> {
     offset_key: Key,
     _marker: PhantomData<fn() -> (K, V)>,
+}
+
+/// We implement this manually because the derived implementation adds trait bounds.
+impl<K, V> Default for Mapping<K, V> {
+    fn default() -> Self {
+        Self {
+            offset_key: Default::default(),
+            _marker: Default::default(),
+        }
+    }
 }
 
 impl<K, V> core::fmt::Debug for Mapping<K, V> {
--- a/examples/multisig/lib.rs
+++ b/examples/multisig/lib.rs
@@ -266,7 +266,7 @@ mod multisig {
         /// Redundant information to speed up the check whether a caller is an owner.
         is_owner: Mapping<AccountId, ()>,
         /// Minimum number of owners that have to confirm a transaction to be executed.
-        requirement: Lazy<u32>,
+        requirement: u32,
     }
 
     impl Multisig {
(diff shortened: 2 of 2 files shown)


HCastano:
Since we don't have the `ink-waterfall` running I checked the size differences locally:

| Reference         | Original     | Optimized |
|--------------|-----------|------------|
| master| 108.5K     | 48.2K      | 
| PR      | 67.7K,  | 31.2K      |

A gain of 17K, not bad 😎

crates/storage/src/lazy/mapping.rs
```diff
@@ -38,12 +38,21 @@ use ink_primitives::Key;
 
 /// A mapping of key-value pairs directly into contract storage.
 #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
-#[derive(Default)]
 pub struct Mapping<K, V> {
     offset_key: Key,
     _marker: PhantomData<fn() -> (K, V)>,
 }
 
+/// We implement this manually because the derived implementation adds trait bounds.
```
HCastano, line 46: What additional trait bounds are added here? 

I checked the `cargo-expand`-ed contracts using both versions and there was no difference between them.

HCastano (approved):

crates/storage/src/lazy/mapping.rs
athei: The `Default` derive requires `Default` for each type parameter. But for this `Mapping` we don't want these bounds. In substrate we have a `DefaultNoBounds` derive for this.

examples/multisig/lib.rs
```diff
@@ -575,50 +605,46 @@ mod multisig {
...
-                    })
-            {
-                if self.confirmations.take(&(trans_id, *owner)).is_some() {
-                    *self.confirmation_count.entry(trans_id).or_insert(0) += 1;
+            for trans_id in &self.transaction_list.transactions {
+                let key = (*trans_id, *owner);
+                if self.confirmations.get(&key).is_some() {
+                    self.confirmations.remove(&key);
+                    let mut count = self.confirmation_count.get(&trans_id).unwrap_or(0);
+                    count += 1;
```
xgreenx, line 634: Hmm, seems it is a bug, it should be -1=)

xgreenx:
I tried to implement multisig by another way, I think it is more clear=) https://github.com/paritytech/ink/pull/1079

What do you think?

athei:
> I tried to implement multisig by another way, I think it is more clear=) #1079
> 
> What do you think?

I think the idea is interesting. But please let us merge this "port only" PR first once the waterfall is up and running. Your PR adds some further changes to the contract and also breaks the interface. This isn't bad but I would like to do this seperatly (it is based on this PR anyways).

examples/multisig/lib.rs
athei: You are absolutely right. Thanks for catching this.

athei:
Removing lazy from `requirement` reduced size to **28.4K**.

HCastano:
These are the final numbers I get:

```
Original wasm size: 66.4K, Optimized: 29.9K
```

Gonna merge this since it's quite an improvement over what we have.

PR #1079, opened 2021-12-11
[Draft] Another implementation of multisig

xgreenx:
This implementation allows avoiding the storing of the order of transactions. `TransactionId` is a hash of the encoded transaction.

Each transaction has a status. `PartialConfirmed` stores the list of confirmations and id of the owners set when that status was created. If we updated the list of owners, that id will mismatch and it means that we need to validate that each confirmation. 

I think that this example is simpler in understanding.

It is only an idea. But if you think that it is not bad, I will finish it=)


athei:
If you move stuff around at least put it into a seperate PR. This bloats the diff and makes it harder to review than necessary. Apart from that I think the idea is good but would require changes to ink-waterfall because the interface changes.

I will give a more in depth review once the diff is cleaned up.

xgreenx:
Rebased to master=)

athei (commented):
IMHO this changes too much at the same time without any immediate need. Please let us first get my minimal port PR merged once the waterfall is running again and then we can look into these improvements.

HCastano:
@xgreenx do you want to pick this up again, or should I close it?

xgreenx:
The idea of that change is to avoid iterations throw confirmations and don't store a complex structure in the storage.
It changes the internal implementation. If you think that it is not a bad idea, we can continue either you can close. We can replace the current implementation of multisig or we can mark it as "multisig_v2"

HCastano:
If you don't mind I'd move forward with it and see how the `ink-waterfall` benchmarks look.

I also wouldn't keep two Multisig examples around, so "improving" the current example is fine with me

HCastano:
This is pretty stale so I'm gonna close this. If you ever want to pick this back up feel free to do so

PR #1085, opened 2021-12-16
Update ERC-20 and ERC-721 UI Tests

HCastano:
This PR updates the UI tests which compile example `ERC-20` and `ERC-721` contracts to
match our latests examples, which use `Mapping`.


cmichi (approved):
Ah right, good that you thought of this!

PR #1086, opened 2021-12-16
Use `Mapping` in `trait-erc20` Example

HCastano:
Looks like I forgot a contract. This is basically just a copy-pasta of our updated
`ERC-20` example.


cmichi (approved):
Too bad that we don't have the stats bot right now!

HCastano:
> Too bad that we don't have the stats bot right now!

Running locally:

| Reference         | Original     | Optimized |
|--------------|-----------|------------|
| master | 65.6K     | 28.2K     | 
| PR      | 35.8K  | 11.0K      |

PR #1094, opened 2022-01-05
Fix failing clippy runs on `master`

cmichi:


crates/metadata/src/specs.rs
```diff
@@ -138,6 +138,7 @@ impl ContractSpecBuilder<Invalid> {
 
 impl<S> ContractSpecBuilder<S> {
     /// Sets the messages of the contract specification.
+    #[must_use]
```
ascjones, line 141: If this is `return_self_not_must_use` then I the docs say we can just decorate the `ContractSpecBuilder` type itself with that rather than all methods

ascjones (approved):

HCastano (approved):
+1 to Andrew's comment.

I really appreciate the commit messages btw ❤️

PR #1100, opened 2022-01-10
Update metadate to support payable constructors

HCastano:
On top of #1065. I figured I'd split it out to make reviews easier.

### DIFF
--- a/crates/metadata/src/specs.rs
+++ b/crates/metadata/src/specs.rs
@@ -227,6 +227,8 @@ pub struct ConstructorSpec<F: Form = MetaForm> {
     pub label: F::String,
     /// The selector hash of the message.
     pub selector: Selector,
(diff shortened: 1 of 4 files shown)


crates/metadata/src/specs.rs
```diff
@@ -508,7 +536,7 @@ impl<S, P, R> MessageSpecBuilder<S, Missing<state::Mutates>, P, R> {
 }
 
 impl<S, M, R> MessageSpecBuilder<S, M, Missing<state::IsPayable>, R> {
-    /// Sets if the message is mutable, thus taking `&mut self` or not thus taking `&self`.
+    /// Sets if the message is payable, thus accepting value for the caller.
```
cmichi, line 539: Good catch!

cmichi (approved):

PR #1107, opened 2022-01-21
Allow `non_camel_case_types` for EventBase ident

HCastano:
Small PR to stop `rust-analyzer` from complaining when it expands the macros in the
example contracts.

### DIFF
--- a/crates/lang/codegen/src/generator/events.rs
+++ b/crates/lang/codegen/src/generator/events.rs
@@ -88,6 +88,7 @@ impl<'a> Events<'a> {
         let base_event_ident =
             proc_macro2::Ident::new("__ink_EventBase", Span::call_site());
         quote! {
+            #[allow(non_camel_case_types)]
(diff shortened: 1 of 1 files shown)


Robbepop:
> Small PR to stop `rust-analyzer` from complaining when it expands the macros in the
> example contracts.
> 

Maybe instead we could just generate proper type names? ^^

HCastano:
> > Small PR to stop `rust-analyzer` from complaining when it expands the macros in the
> > example contracts.
> 
> Maybe instead we could just generate proper type names? ^^

Basically all of the internal codegen types are already using this convention (and their warnings are being suppressed in the same way).

I'd be happy to change these types to follow Rust convention in a follow up, but I don't really want to do that today tbh.

cmichi (approved):

HCastano:
(Gonna merge, but as I mentioned to Robin I'll follow up with a proper fix)

Issue #996, opened 2021-10-31
Review Existing Storage Type Sizes


HCastano:
We should go through the existing caching types we have and see if they're viable for deployment on a parachain in terms of size. Any that aren't should be considered for deprecation.

HCastano:
Soft-removed in #1111.

PR #1111, opened 2022-01-25
Hide `collection` and `lazy` modules

HCastano:
The goal of this PR is to begin the process of removing all data structures other than
`Mapping` from the public `ink!` API. We're starting by just hiding the `collections` and
`lazy` modules from users. Eventually we'll want to remove the code entirely, but I
wanted to make the minimal amount of changes required to get this compiling.

### Rational

You may be asking: why remove all our great collections? Well, if you've noticed a theme
in `ink!` over the last few months its been that of reducting complexity in order to
reduce the size of contracts. This takes another step towards that.

Until we can provide full-featured collections which don't bloat contract sizes the only
reasonable step forward to (somewhat strongly) encourage developers to use `Mapping` as
their main "data type".

### Eager Loading
One consequence of this PR is that because there is no more `Lazy` type, every single
storage item from your contract will be loaded from storage eagerly. To give a more
concrete example, consider the following contract storage struct:

```rust
#[ink(storage)]
pub struct Contract {
    ids: u32,
    accounts: ink_prelude::vec::Vec<AccountId>,
}
```

In this case every time we execute the contract we'd load up the entire `Vec` from
storage (even if we didn't use the `Vec` during the execution). For large `Vec` this
would be inefficient, but it's something you as a developer will need to be mindful of
going forward.

### Compatibility
Unfortunately, this will be a **breaking change** for any contract using collections
other than `Mapping`. If your contracts were mainly just using `HashMap/BTreeMap` the
migration should be pretty straightforward - although during the migration you will
notice that the API for `Mapping` is more rudamentary (by design) than the latter two
data structures.

### Follow-Ups
- Depending on the reception to this, I'll do the actual code pruning in some follow up
  PRs.
- There are some TODOs with `XYZ` as the issue numbers. Once I make those issues I'll
  update the PR to fill them in.
- Since we don't have collections which need dynamic allocations anymore I believe that
  we can get rid of the dynamic storage allocator code entirely. Someone correct me if
  I'm wrong here.
- There's definitely documentation internal to `ink!` and external (such as on our docs
  portal) that will need to be updated
- More ergonomic way to use `ink_lang::codegen::initialize_contract()` for initializing
  `Mapping`'s

Related issue: https://github.com/paritytech/ink/issues/996


crates/lang/macro/src/lib.rs
```diff
@@ -136,8 +136,14 @@ pub fn selector_bytes(input: TokenStream) -> TokenStream {
 ///
 ///     An example for this is the following type that could potentially be used
 ///     within a contract's storage struct definition:
-///     ```
-///     // A storage vector of storage vectors.
+///
+///
+///     ```ignore
+///     # // TODO [#XYZ]: Right now we've hidden the `StorageVec` from public access so this
```
athei, line 142: Is this PR work in progress? Otherwise please no TODOs in code. Should create an issue for this instead.

crates/storage/src/collections/binary_heap/mod.rs
```diff
@@ -106,7 +106,9 @@ where
     ///
     /// # Example
     ///
-    /// ```
+    /// ```ignore
+    /// # // TODO [#XYZ]: We currently ignore this test since we stopped exposing the `collections`
+    /// # // module publicly.
```
athei, line 111: But from within the `collections` module (here) you should still have access?

crates/storage/src/lib.rs
```diff
@@ -62,9 +68,13 @@ mod test_utils;
 
 #[doc(inline)]
 pub use self::{
-    alloc::Box,
+    lazy::Mapping,
+    memory::Memory,
```
athei, line 72: Do we really need this? It feels like an anti pattern to make use of this. Data should be passed as an argument where possible IMHO. Also, I think supporting this in the future will be annoying.

athei (commented):
I agree that we should be able to hide/remove the storage alocator. I think you need to make the CI happy in order to get our change break down (gas/size), right? Or is this still broken from the storage deposit fallout? @cmichi

crates/lang/macro/src/lib.rs
HCastano: That's what the `#XYZ` is supposed to indicate, the number of the issue. I didn't want to jump the gun on creating the issues :)

crates/storage/src/collections/binary_heap/mod.rs
HCastano: No, the doc tests get treated as if they were in a separate crate (kinda like the `tests` directory), so we don't have access to private members

crates/storage/src/lib.rs
HCastano: Tbh, I'm not sure. I haven't seen any `ink!` contracts using `Memory`, but I know it's used in Solidity contracts. Maybe @cmichi or @ascjones can chime in here?

HCastano:
> I think you need to make the CI happy in order to get our change break down (gas/size), right?

There was an issue with GitHub access tokens that got fixed a couple days ago. I'm just gonna re-run the CI and it should post it

paritytech-cicd-pr:

These are the results when building the `examples/*` contracts from this branch with `cargo-contract 0.17.0-06c5598` and comparing them to ink! `master`: 

| `contract-terminate` |                  |            | 1.31 K               | 200_778        |
| `contract-transfer`  |                  |            | 7.98 K               | 1_399          |
| `delegator`          | -2.43 K          | +786       | 7.63 K               | 10_296         |
| `dns`                | -0.58 K          | +864       | 9.89 K               | 15_152         |
| `erc1155`            |                  |            | 27.50 K              | 47_811         |
| `erc20`              | -0.42 K          | +257       | 9.80 K               | 10_014         |
| `erc721`             |                  |            | 13.70 K              | 34_702         |
| `flipper`            |                  |            | 1.76 K               | 347            |
| `incrementer`        |                  |            | 1.64 K               | 340            |
| `multisig`           | -1.77 K          | +1_860     | 26.79 K              | 56_756         |
| `proxy`              |                  |            | 3.81 K               | 2_629          |
| `rand-extension`     |                  |            | 5.12 K               | 6_654          |
| `trait-erc20`        | -0.42 K          | +15        | 10.10 K              | 10_659         |
| `trait-flipper`      |                  |            | 1.53 K               | 342            |
| `trait-incrementer`  |                  |            | 1.62 K               | 680            |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/177026) | Last update: Sat Jan 29 00:58:03 CET 2022

crates/storage/src/lib.rs
cmichi: No strong opinion here. I have a bad feeling about not having any example using this though. In consequence we don't have any proper on-chain `ink-waterfall` test for it either.

>Data should be passed as an argument where possible IMHO.

Agree, but I don't know enough about why it was introduced in Solidity in the first place.

Does @ascjones know more?

crates/lang/macro/src/lib.rs
```diff
@@ -136,8 +136,14 @@ pub fn selector_bytes(input: TokenStream) -> TokenStream {
...
+///     # // doesn't compile.
+///     # //
+///     # // On a sidenote, I think that with the removal of the `collections` module we'll be able
+///     # // to get rid of all the dynamic storage allocation infrastructure.
```
cmichi, line 146: Just put all of this into an issue.

cmichi:
Why is it necessary to remove `lazy`? E.g. the example with a `Vec<T>` in storage and trying to avoid it being loaded eagerly seems like a super-plausible case for it.

athei:
Because of the interaction with `initialize_contract`: If you have a `Lazy` and don't overwrite it after calling `initialize_contract` you are fucked.

cmichi (approved):
If you move the comments into individual GitHub issues I'm fine with it. It's super-sad that we have to hide those collections for the moment, but we can always bring them back in later versions.

crates/lang/macro/src/lib.rs
HCastano: Yep made #1120

athei (approved):
Yeah let's keep `Memory` around for now. We can always remove it later since it doesn't touch storage.

PR #1116, opened 2022-01-27
Debug events

forgetso:
When debugging contracts it can be helpful to be able to print events in unit tests. This PR allows events to be debug printed and decorates all instances of event with `Debug` in the code base.

### DIFF
--- a/crates/lang/codegen/src/generator/events.rs
+++ b/crates/lang/codegen/src/generator/events.rs
@@ -88,7 +88,7 @@ impl<'a> Events<'a> {
         let base_event_ident =
             proc_macro2::Ident::new("__ink_EventBase", Span::call_site());
         quote! {
-            #[derive(::scale::Encode, ::scale::Decode)]
+            #[derive(::scale::Encode, ::scale::Decode, Debug)]
             pub enum #base_event_ident {
                 #( #event_idents(#event_idents), )*
             }
--- a/crates/lang/codegen/src/generator/events.rs
+++ b/crates/lang/codegen/src/generator/events.rs
@@ -89,7 +89,7 @@ impl<'a> Events<'a> {
             proc_macro2::Ident::new("__ink_EventBase", Span::call_site());
         quote! {
             #[allow(non_camel_case_types)]
-            #[derive(::scale::Encode, ::scale::Decode, Debug)]
+            #[derive(::scale::Encode, ::scale::Decode, ::core::fmt::Debug)]
             pub enum #base_event_ident {
                 #( #event_idents(#event_idents), )*
             }
--- a/crates/engine/src/test_api.rs
+++ b/crates/engine/src/test_api.rs
@@ -24,7 +24,7 @@ use crate::{
 use std::collections::HashMap;
 
 /// Record for an emitted event.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct EmittedEvent {
     /// Recorded topics of the emitted event.
     pub topics: Vec<Vec<u8>>,
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -38,6 +38,7 @@ use crate::{
     Result,
 };
 use ink_primitives::Key;
+use static_assertions::_core::fmt::Debug;
 
 /// Returns the address of the caller of the executed contract.
 ///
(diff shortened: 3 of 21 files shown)


crates/lang/codegen/src/generator/events.rs
```diff
@@ -89,7 +89,7 @@ impl<'a> Events<'a> {
             proc_macro2::Ident::new("__ink_EventBase", Span::call_site());
         quote! {
             #[allow(non_camel_case_types)]
-            #[derive(::scale::Encode, ::scale::Decode)]
+            #[derive(::scale::Encode, ::scale::Decode, Debug)]
```
HCastano, line 92: As a general tip, inside of macros we aim to use full paths in case the code where the macro is be called has defined their own `Debug` type/trait (in which case this wouldn't work as expected) 

```suggestion
            #[derive(::scale::Encode, ::scale::Decode, ::core::fmt::Debug)]
```

HCastano (changes requested):
Sorry, I don't really understand this PR. From what I can see you're just introducing some extra trait `Debug` bounds to a couple of methods.

Why can't you achieve the same result by manually deriving `Debug` on your own events?

crates/lang/codegen/src/generator/events.rs
forgetso: Thanks for the tip

forgetso:
> Why can't you achieve the same result by manually deriving `Debug` on your own events?

I added `#[derive(Debug)]` to my events, but they won't print in the unit tests because of

```
= help: the trait `Debug` is not implemented for `__ink_EventBase`
     = note: add `#[derive(Debug)]` to `__ink_EventBase` or manually `impl Debug for __ink_EventBase
```

Is this what you mean? My way to solve it was to try to tag `EventBase` with debug and that led to this PR. Is there some other way of achieving this?

HCastano:
@forgetso can you share the code that's causing you trouble? 

I just tagged some events in the ERC-20 example with `Debug` and they compiled fine

```rust
    /// Event emitted when a token transfer occurs.
    #[ink(event)]
    #[derive(Debug)]
    pub struct Transfer {
        #[ink(topic)]
        from: Option<AccountId>,
        #[ink(topic)]
        to: Option<AccountId>,
        value: Balance,
    }
```

So maybe I'm misunderstanding what you're doing here.

forgetso:
> I just tagged some events in the ERC-20 example with `Debug` and they compiled fine

Sure, it's not the tagging that causes the error. It's trying to print the events from within unit tests.

            let emitted_events = ink_env::test::recorded_events().collect::<Vec<_>>();

            // This line causes the error when running unit tests
            ink_env::debug_println!("{:?}", emitted_events);

HCastano:
@forgetso I can't reproduce your issue. Here is [a contract I made](https://gist.github.com/HCastano/3d56463744892c01e055a1dcc498661e) which prints the emitted events with `debug_println` and it runs fine.

Can you upload the full contract somewhere so I can try and reproduce it?

forgetso:
The problem seems to be in the experimental unit tests. I've created a repo that fails when you run 

https://github.com/forgetso/ink-events-debug

Run using:
`cargo +nightly test --no-default-features --features std,ink-experimental-engine --verbose -- --nocapture`


```
error[E0277]: `EmittedEvent` doesn't implement `Debug`
  --> lib.rs:67:45
   |
67 |             ink_env::debug_println!("{:?}", emitted_events);
   |                                             ^^^^^^^^^^^^^^ `EmittedEvent` cannot be formatted using `{:?}` because it doesn't implement `Debug`
   |
```

HCastano (changes requested):
Okay, thanks for the repo. That would've been handy to have right off the bat.

So as far as the changes we want here, from my quick testing on your repo really all we need is to derive `Debug` on the experimental engine's `EmittedEvent` struct.

We shouldn't need to add the `Debug` trait bounds any other methods. As well, we shouldn't derive `Debug` for all the example events unless its actually used. Even then, I'd only derive it under `std` compilation (`#[cfg_attr(feature = "std", derive(Debug))]`)

forgetso:
Sorry, our default testing is all experimental engine a we always need to set the transferred balance etc. It didn't occur to me that this might be engine specific. I'll try modifying the PR based on your suggestion when I get some time.

HCastano:
Hey @forgetso, any updates here?

HCastano:
@forgetso congrats on the baby 👶

No worries. I'm gonna close it, feel free to re-open or make a new PR whenever you're ready

PR #1126, opened 2022-02-09
Add phantom markers for ink! primitives

cmichi:
I'm about to add custom linting into `cargo-contract` using `dylint`. Unfortunately the way we use proc. macros currently the information that a function is e.g. an ink! constructor is not available for lint implementations.

I'm not super proud of this workaround, but `cfg` attributes are retained and accessible within a lint implementation. This hack doesn't generate any additional code and it's hidden to users. Hence I think it's justifiable to do it that way. If we find a better workaround in the future we can still change it.

### DIFF
--- a/crates/lang/codegen/src/generator/events.rs
+++ b/crates/lang/codegen/src/generator/events.rs
@@ -77,6 +77,11 @@ impl<'a> Events<'a> {
     /// Generates the base event enum that comprises all user defined events.
     /// All emitted events are converted into a variant of this enum before being
     /// serialized and emitted to apply their unique event discriminant (ID).
+    ///
+    /// # Developer Note
+    ///
+    /// The `__ink_EventBase` config attribute is used here to convey the information
+    /// that this function is an ink! constructor to `dylint`.
     fn generate_event_base(&self) -> TokenStream2 {
         let storage_ident = &self.contract.module().storage().ident();
         let event_idents = self
--- a/crates/lang/codegen/src/generator/events.rs
+++ b/crates/lang/codegen/src/generator/events.rs
@@ -80,8 +80,8 @@ impl<'a> Events<'a> {
     ///
     /// # Developer Note
     ///
-    /// The `__ink_EventBase` config attribute is used here to convey the information
-    /// that this function is an ink! constructor to `dylint`.
+    /// The `__ink_dylint_EventBase` config attribute is used here to convey the
+    /// information that this function is an ink! constructor to `dylint`.
     fn generate_event_base(&self) -> TokenStream2 {
         let storage_ident = &self.contract.module().storage().ident();
         let event_idents = self
--- a/crates/lang/codegen/src/generator/events.rs
+++ b/crates/lang/codegen/src/generator/events.rs
@@ -90,6 +90,7 @@ impl<'a> Events<'a> {
         quote! {
             #[allow(non_camel_case_types)]
             #[derive(::scale::Encode, ::scale::Decode)]
+            #[cfg(not(target_os = "__ink_EventBase"))]
             pub enum #base_event_ident {
                 #( #event_idents(#event_idents), )*
             }
--- a/crates/lang/codegen/src/generator/events.rs
+++ b/crates/lang/codegen/src/generator/events.rs
@@ -81,7 +81,7 @@ impl<'a> Events<'a> {
     /// # Developer Note
     ///
     /// The `__ink_dylint_EventBase` config attribute is used here to convey the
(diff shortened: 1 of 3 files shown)


paritytech-cicd-pr:

| `contract-terminate` |                  | +14_926    | 1.31 K               | 215_704        |
| `contract-transfer`  |                  | +14_305    | 7.98 K               | 15_704         |
| `delegator`          |                  | +41_047    | 7.63 K               | 51_343         |
| `dns`                |                  | +31_959    | 10.20 K              | 47_112         |
| `erc1155`            |                  | +46_413    | 27.50 K              | 94_224         |
| `erc20`              |                  | +37_098    | 9.80 K               | 47_112         |
| `erc721`             |                  | +90_930    | 13.70 K              | 125_632        |
| `flipper`            |                  | +15_357    | 1.76 K               | 15_704         |
| `incrementer`        |                  | +15_364    | 1.64 K               | 15_704         |
| `multisig`           |                  | +46_824    | 26.79 K              | 103_580        |
| `proxy`              |                  | +29_614    | 3.81 K               | 32_243         |
| `rand-extension`     |                  | +9_050     | 5.12 K               | 15_704         |
| `trait-erc20`        |                  | +36_453    | 10.10 K              | 47_112         |
| `trait-flipper`      |                  | +15_362    | 1.53 K               | 15_704         |
| `trait-incrementer`  |                  | +30_728    | 1.62 K               | 31_408         |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/178496) | Last update: Wed Feb  9 15:03:49 CET 2022

crates/lang/codegen/src/generator/events.rs
```diff
@@ -90,6 +90,7 @@ impl<'a> Events<'a> {
         quote! {
             #[allow(non_camel_case_types)]
             #[derive(::scale::Encode, ::scale::Decode)]
+            #[cfg(not(target_os = "__ink_EventBase"))]
```
ascjones, line 93: What alternatives are there to this, what about `#[cfg(not(feature = "__ink_EventBase"))]`. 

Would be nice to have custom tool attributes e.g. https://rust-lang.github.io/rfcs/2103-tool-attributes.html, but maybe could piggyback on a `rustfmt` or `clippy` one?

cmichi: >What alternatives are there to this, what about `#[cfg(not(feature = "__ink_EventBase"))].`

That would work too, I'll change it.

>maybe could piggyback on a `rustfmt` or `clippy` one?

Makes us dependent on those though. I had the same idea at first and was looking at `clippy` attributes, but AFAICT those are only enums, so we can't encode ink!-specific info.

cmichi: Btw. I hadn't tried `feature` since I had concerns that this might require the feature definition somewhere. But from testing locally it seems to work just fine.

cmichi:
> What about adding in `dylint` or `cargo_contract` to the identifiers?

Are you referring to what is exemplarily described as `#[tool_name::foo]` in https://rust-lang.github.io/rfcs/2103-tool-attributes.html? I haven't found a way to actually define a `tool_name` without introducing code.

ascjones (approved):

crates/lang/codegen/src/generator/events.rs
```diff
@@ -77,6 +77,11 @@ impl<'a> Events<'a> {
     /// Generates the base event enum that comprises all user defined events.
     /// All emitted events are converted into a variant of this enum before being
     /// serialized and emitted to apply their unique event discriminant (ID).
+    ///
+    /// # Developer Note
+    ///
+    /// The `__ink_dylint_EventBase` config attribute is used here to convey the
+    /// information that this function is an ink! constructor to `dylint`.
```
HCastano, line 84: Here it shouldn't be constructor, but I'm not sure what you're getting at.

Should it maybe be `generator` instead?

crates/lang/codegen/src/generator/storage.rs
```diff
@@ -84,6 +84,11 @@ impl Storage<'_> {
     }
 
     /// Generates the storage struct definition.
+    ///
+    /// # Developer Note
+    ///
+    /// The `__ink_dylint_Storage` config attribute is used here to convey the
+    /// information that this function is an ink! constructor to `dylint`.
```
HCastano, line 91: Same here, I think this should be `generator`

HCastano (commented):
I want a little bit of clarification in a comment

PR #1130, opened 2022-02-09
Don't copy attributes to call builders

xgreenx:
The current codegen copy all attributes from the original method. But that attributes were initially added for the original function and may require some restrictions on the struct that uses these attributes.

For example, OpenBrush provides [`#[brush::modifiers]` attribute](https://github.com/Supercolony-net/openbrush-contracts#modifiers) that inserts some code before the function. 
In the case of [`when_paused`](https://github.com/Supercolony-net/openbrush-contracts/blob/main/contracts/security/pausable/mod.rs#L24) modifier, the struct should implement the `PausableStorage` trait. [That contract](https://github.com/Supercolony-net/openbrush-contracts/blob/main/examples/pausable/lib.rs#L9) derives that trait but auto-generated call builders don't derive it(and they can't because it is complicated for them). And it will cause a compilation error because call builders will try to use that modifier.

The idea of that PR is to not insert attributes from the original method for auto-generated call builders. 

Call builders are used only for cross-contract communication so they don't participate in the code logic of the contract and they don't participate in the metadata generation(ABI). So attributes like `#[doc]` is not important. 

BUT, maybe someone wants to apply some functional attributes like `clippy` or `allow` and that change will break it. So maybe we also need to think about how ink! can provide customization of attributes(which attributes should be passed to auto-generated structures and which not).

### DIFF
--- a/crates/lang/codegen/src/generator/as_dependency/call_builder.rs
+++ b/crates/lang/codegen/src/generator/as_dependency/call_builder.rs
@@ -279,7 +279,6 @@ impl CallBuilder<'_> {
             ir::Receiver::Ref => quote! { build },
             ir::Receiver::RefMut => quote! { build_mut },
         };
-        let attrs = message.attrs();
         quote_spanned!(span=>
             type #output_ident = <<<
                 Self
@@ -288,7 +287,6 @@ impl CallBuilder<'_> {
                 as #trait_path>::#output_ident;
 
             #[inline]
-            #( #attrs )*
             fn #message_ident(
                 & #mut_token self
                 #( , #input_bindings: #input_types )*
@@ -356,7 +354,6 @@ impl CallBuilder<'_> {
         let span = message.span();
         let callable = message.callable();
         let message_ident = message.ident();
-        let attrs = message.attrs();
         let selector = message.composed_selector();
         let selector_bytes = selector.hex_lits();
         let input_bindings = generator::input_bindings(callable.inputs());
@@ -380,7 +377,6 @@ impl CallBuilder<'_> {
             >
         );
         quote_spanned!(span=>
-            #( #attrs )*
             #[allow(clippy::type_complexity)]
             #[inline]
             pub fn #message_ident(
--- a/crates/lang/codegen/src/generator/as_dependency/contract_ref.rs
+++ b/crates/lang/codegen/src/generator/as_dependency/contract_ref.rs
@@ -332,7 +332,6 @@ impl ContractRef<'_> {
     ) -> TokenStream2 {
         use ir::Callable as _;
         let span = message.span();
-        let attrs = message.attrs();
         let storage_ident = self.contract.module().storage().ident();
         let message_ident = message.ident();
         let call_operator = match message.receiver() {
@@ -344,7 +343,6 @@ impl ContractRef<'_> {
         let input_types = message.inputs().map(|input| &input.ty).collect::<Vec<_>>();
         let output_type = message.output().map(|ty| quote! { -> #ty });
         quote_spanned!(span=>
-            #( #attrs )*
             #[inline]
             pub fn #message_ident(
                 & #mut_token self
@@ -375,14 +373,12 @@ impl ContractRef<'_> {
         constructor: ir::CallableWithSelector<ir::Constructor>,
     ) -> TokenStream2 {
         let span = constructor.span();
-        let attrs = constructor.attrs();
         let constructor_ident = constructor.ident();
         let selector_bytes = constructor.composed_selector().hex_lits();
         let input_bindings = generator::input_bindings(constructor.inputs());
         let input_types = generator::input_types(constructor.inputs());
         let arg_list = generator::generate_argument_list(input_types.iter().cloned());
         quote_spanned!(span =>
-            #( #attrs )*
             #[inline]
             #[allow(clippy::type_complexity)]
             pub fn #constructor_ident(
--- a/crates/lang/codegen/src/generator/trait_def/call_builder.rs
+++ b/crates/lang/codegen/src/generator/trait_def/call_builder.rs
@@ -357,7 +357,6 @@ impl CallBuilder<'_> {
     ) -> TokenStream2 {
         let span = message.span();
         let message_ident = message.ident();
-        let attrs = message.attrs();
         let output_ident = generator::output_ident(message_ident);
         let output = message.output();
         let output_sig = output.map_or_else(
@@ -380,7 +379,6 @@ impl CallBuilder<'_> {
                 ::ink_env::call::utils::Set<#output_sig>,
             >;
 
-            #( #attrs )*
             #[inline]
             fn #message_ident(
                 & #mut_tok self
--- a/crates/lang/codegen/src/generator/trait_def/call_forwarder.rs
+++ b/crates/lang/codegen/src/generator/trait_def/call_forwarder.rs
@@ -388,7 +388,6 @@ impl CallForwarder<'_> {
         let trait_ident = self.trait_def.trait_def.item().ident();
         let forwarder_ident = self.ident();
         let message_ident = message.ident();
-        let attrs = message.attrs();
         let output_ident = generator::output_ident(message_ident);
         let output_type = message
             .output()
@@ -408,7 +407,6 @@ impl CallForwarder<'_> {
         quote_spanned!(span =>
             type #output_ident = #output_type;
 
-            #( #attrs )*
             #[inline]
             fn #message_ident(
                 & #mut_tok self


paritytech-cicd-pr:

| `contract-terminate` |                  |            | 1.31 K               | 215_704        |
| `contract-transfer`  |                  |            | 7.98 K               | 15_704         |
| `delegator`          |                  |            | 7.63 K               | 51_343         |
| `dns`                |                  |            | 10.20 K              | 47_112         |
| `erc1155`            |                  |            | 27.50 K              | 94_224         |
| `erc20`              |                  |            | 9.80 K               | 47_112         |
| `erc721`             |                  |            | 13.70 K              | 125_632        |
| `flipper`            |                  |            | 1.76 K               | 15_704         |
| `incrementer`        |                  |            | 1.64 K               | 15_704         |
| `multisig`           |                  |            | 26.79 K              | 103_580        |
| `proxy`              |                  |            | 3.81 K               | 32_243         |
| `rand-extension`     |                  |            | 5.12 K               | 15_704         |
| `trait-erc20`        |                  |            | 10.10 K              | 47_112         |
| `trait-flipper`      |                  |            | 1.53 K               | 15_704         |
| `trait-incrementer`  |                  |            | 1.62 K               | 31_408         |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/178591) | Last update: Thu Feb 10 00:06:15 CET 2022

HCastano:
So just from your PR description the change seems okay. However, I tried to build the pausable contract you linked and it built fine.

Can you explain how to reproduce the compilation error you're talking about?

xgreenx:
The pausable contract in the `master` is using ink `rc-6`=) Call builders were introduced in `rc7`. 

We started the [migration to `rc8`](https://github.com/Supercolony-net/openbrush-contracts/tree/ink-rc8) and at the moment we are using our own version of the ink(while we are waiting for the merging of PRs). But we already fixed that issue so you can't reproduce it publickly.

The error is next:
```shell
error[E0277]: the trait bound `my_pausable::_::CallBuilder: brush::contracts::pausable::PausableStorage` is not satisfied
  --> lib.rs:23:28
   |
23 |         #[brush::modifiers(when_not_paused)]
   |                            ^^^^^^^^^^^^^^^ the trait `brush::contracts::pausable::PausableStorage` is not implemented for `my_pausable::_::CallBuilder`
   |
note: required by a bound in `brush::contracts::pausable::when_not_paused`
  --> /Users/green/External/4ire.labs/openbrush-contracts/contracts/security/pausable/mod.rs:40:8
   |
40 |     T: PausableStorage,
   |        ^^^^^^^^^^^^^^^ required by this bound in `brush::contracts::pausable::when_not_paused`

error[E0271]: type mismatch resolving `for<'r> <[closure@lib.rs:23:9: 23:45] as FnOnce<(&'r mut my_pausable::_::CallBuilder,)>>::Output == Result<_, _>`
  --> lib.rs:23:28
   |
23 |         #[brush::modifiers(when_not_paused)]
   |                            ^^^^^^^^^^^^^^^ expected enum `Result`, found struct `ink_env::call::CallBuilder`
   |
   = note: expected enum `Result<_, _>`
            found struct `ink_env::call::CallBuilder<DefaultEnvironment, Set<ink_env::AccountId>, Unset<u64>, Unset<u128>, Set<ExecutionInput<ArgumentList<ArgumentListEnd, ArgumentListEnd>>>, Set<ReturnType<Result<(), brush::contracts::pausable::PausableError>>>>`
note: required by a bound in `brush::contracts::pausable::when_not_paused`
  --> /Users/green/External/4ire.labs/openbrush-contracts/contracts/security/pausable/mod.rs:41:26
   |
41 |     F: FnOnce(&mut T) -> Result<R, E>,
   |                          ^^^^^^^^^^^^ required by this bound in `brush::contracts::pausable::when_not_paused`

error[E0308]: mismatched types
  --> lib.rs:23:9
   |
23 |         #[brush::modifiers(when_not_paused)]
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `ink_env::call::CallBuilder`, found enum `Result`
24 |         pub fn flip(&mut self) -> Result<(), PausableError> {
   |                                   ------------------------- expected `ink_env::call::CallBuilder<DefaultEnvironment, Set<ink_env::AccountId>, Unset<u64>, Unset<u128>, Set<ExecutionInput<ArgumentList<ArgumentListEnd, ArgumentListEnd>>>, Set<ReturnType<Result<(), brush::contracts::pausable::PausableError>>>>` because of return type
   |
   = note: expected struct `ink_env::call::CallBuilder<DefaultEnvironment, Set<ink_env::AccountId>, Unset<u64>, Unset<u128>, Set<ExecutionInput<ArgumentList<ArgumentListEnd, ArgumentListEnd>>>, Set<ReturnType<Result<(), brush::contracts::pausable::PausableError>>>>`
                found enum `Result<_, _>`
   = note: this error originates in the attribute macro `brush::modifiers` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `MyFlipperRef: brush::contracts::pausable::PausableStorage` is not satisfied
  --> lib.rs:23:28
   |
23 |         #[brush::modifiers(when_not_paused)]
   |                            ^^^^^^^^^^^^^^^ the trait `brush::contracts::pausable::PausableStorage` is not implemented for `MyFlipperRef`
   |
note: required by a bound in `brush::contracts::pausable::when_not_paused`
  --> /Users/green/External/4ire.labs/openbrush-contracts/contracts/security/pausable/mod.rs:40:8
   |
40 |     T: PausableStorage,
   |        ^^^^^^^^^^^^^^^ required by this bound in `brush::contracts::pausable::when_not_paused`

```

ascjones (dismissed):
LGTM

cmichi:
@xgreenx I'm not so sure about this PR and would like to ask for some more clarification.

>The idea of that PR is to not insert attributes from the original method for auto-generated call builders.
>
>Call builders are used only for cross-contract communication so they don't participate in the code logic of the contract and they don't participate in the metadata generation(ABI). So attributes like #[doc] is not important.

If I run `cargo doc --document-private-items` on your branch with `flipper` the doc comments/attributes for e.g. the methods disappear. 

Generally I feel like the current behavior is what I as a developer would expect and it's also closer to Rust then to have those attributes disappear. I can imagine some clippy use cases (e.g. complex types) where it's justified to have those attributes be retained. 

Can you explain the use case you have some more? Maybe we find another solution. Maybe you can elaborate some on how you mean this:
>For example, OpenBrush provides [#[brush::modifiers] attribute](https://github.com/Supercolony-net/openbrush-contracts#modifiers) that inserts some code before the function.

cmichi:
>As a solution, we can filter attributes […]
> […] provide an ability to mark those attributes that should be not propagated.

This is the solution I would prefer. So to have the default of attributes like `clippy` and so on to be propagated and have a possibility to filter out ones which shouldn't.

xgreenx:
Then we need to decide how it can be marked for filtering=)

We can remove all attributes by default, only keeping attributes that contains substring from the whitelist:  ["ink", "doc", "clippy", "allow", "deny"](Maybe you know other useful attributes?)

If the user wants to keep the attribute that is not from the whitelist, he can use the `#[ink(keep(...))]` attribute proc-macro. During expanding of that macro it will remove `ink(keep(` + `)` part and will only keep `#[...]`.

cmichi:
@xgreenx How about changing this PR so that the syntax `#[ink::contract(filter_attr = "foo, bar")]` would be supported? Filtering in that case would mean that the attributes are forwarded until explicitly specified in the filter.

xgreenx:
> How about changing this PR so that the syntax #[ink::contract(filter_attr = "foo, bar")] would be supported? Filtering in that case would mean that the attributes are forwarded until explicitly specified in the filter.

It means that we need to introduce the same syntaxis for `#[ink::trait_definition]`. Also, may exist cases where we want to have a case `#[ink(skip(...))]` only for one method in the contract -> we need support that syntaxis too. 

It is hard to imagine for me when we need to pass attributes to builders except `clippy`, `doc`, `allow` and `deny`. I think simpler will introduction of `#[ink(keep(...))]` and maybe extend in the future whitelist(it is a simple modification).

cmichi:
> It is hard to imagine for me when we need to pass attributes to builders except `clippy`, `doc`, `allow` and `deny`. 

@xgreenx Here is the list of built-in attributes: https://doc.rust-lang.org/reference/attributes.html#built-in-attributes-index, so those would need to be whitelisted as well, right?

Couldn't your `#[brush::modifiers]` attribute insert an ink! attribute `#[ink(filter_attr = "…")]`? This would then be hidden to the user, but achieve the same effect?

xgreenx:
> @xgreenx Here is the list of built-in attributes: https://doc.rust-lang.org/reference/attributes.html#built-in-attributes-index, so those would need to be whitelisted as well, right?

Not all=) We are talking about the contract's public methods so only attributes: Diagnostics, Documentation, Conditional compilation.

> Couldn't your #[brush::modifiers] attribute insert an ink! attribute #[ink(filter_attr = "…")]? This would then be hidden to the user, but achieve the same effect?

`#[brush::modifiers]` is an attribute for methods, it is expanded after `#[ink::contract]` so it can't insert any ink! related stuff before ink! expanding by itself=(

cmichi:
Ok, so having discussed this, from my pov it's fine if you implement whitelisting + users having to opt-in if they want to keep the propagation. I think this then is still a better solution than how the PR is currently approved (removing the attributes altogether).

xgreenx:
I will work on it=) Are you okay with syntaxis `#[ink(keep(...))]`?

cmichi:
> I will work on it=) Are you okay with syntaxis `#[ink(keep(...))]`?

Cool! I would put it on the top level of the contract as `#[ink::contract(keep_attr = "foo, bar")]`.

xgreenx:
> Cool! I would put it on the top level of the contract as #[ink::contract(keep_attr = "foo, bar")].

Okay, the same for `#[ink::trait_definition(keep_attr = "foo, bar")]`?

cmichi:
> > Cool! I would put it on the top level of the contract as #[ink::contract(keep_attr = "foo, bar")].
> 
> Okay, the same for `#[ink::trait_definition(keep_attr = "foo, bar")]`?

Yup, fine with me!

xgreenx:
Created a new pull request https://github.com/paritytech/ink/pull/1145

PR #1136, opened 2022-02-17
Use impl index as const generic parameter for `CallBuilder` trait impls

ascjones:
Alternative fix for #982, with the goal of being a smaller, simpler change than #983. I have copied the example from there.

Replaces the use of the associated type with a const int parameter for the `TraitCallForwarderFor` trait, which is implemented per trait for the `CallBuilder`.

Since the `TraitInfo` parameter was only used to resolve the target trait's forwarder on the call builder, this could be replaced by a const generic parameter. The value used for this is the `index` of the impl block, which is the position it appears. It is unique within the context of the contract.


xgreenx:
> The main issue I have is that this is quite a significant refactoring to solve this particular issue. To this end I explored a few alternatives to see whether it is possible to fix it with a much smaller change. The one I came up with is https://github.com/paritytech/ink/pull/1136. Please have a look and let me know what you think of that. There may be some issues with it that I have not considered.

I don't see any issues that are not covered with that change. `TraitCallForwarderFor` is a utility trait so it is okay to use index instead of the actual type. So it is a good change without potential regression=)

The idea of my change also was to reduce the redundant of the code. We have 3 places with the same code that can be covered with blanked implementation. + We don't need an introduction of the utility trait `TraitCallForwarderFor`.
Also, it was preparation for future change that we want to suggest, but we can refactor it there=)
Also, we planned to use that blanked implementation in OpenBrush to not generate the same code at the fourth time.

So both changes fix the problem and I'm okay with any of them. The first change was for the future and it is bigger=)

ascjones:
Okay thanks for the background, I wasn't aware there was another intended use for the blanket impl. 

I think if we are talking about a change which is only intended to fix that specific bug, then I prefer this one simply because it is smaller.

Then I am happy to look at the further refactoring through the lens of reducing code redundancy, and utilizing it for your future requirements.

However I am interested to see what the other guys think @HCastano @cmichi @Robbepop , should we just try and solve the immediate bug with this PR, or do the larger refactoring using the blanket impl?

ascjones:
> If someone does any changes at all to the contract code (like just adding blank lines or comments), the index and the generated code will change

I did think about only generating the index only for ink impl blocks, then having a `fn trait_index() -> Option<syn::LitInt>` on `ItemImpl` which should solve this particular issue, with an existing index only changing if the order of the impl blocks is changed or any are removed. However that would be slightly more tricky than adding the simple `.enumerate()` over all `syn::ItemImpl`s

paritytech-cicd-pr:
 These are the results when building the `examples/*` contracts from this branch with `cargo-contract 0.17.0-4195197` and comparing them to ink! `master`: 

| `accumulator`        |                  |            | 1.35 K               |                |
| `adder`              |                  |            | 2.47 K               |                |
| `contract-terminate` |                  |            | 1.23 K               | 215_704        |
| `contract-transfer`  |                  |            | 8.47 K               | 15_704         |
| `delegator`          |                  |            | 6.86 K               | 50_833         |
| `dns`                |                  |            | 9.70 K               | 47_112         |
| `erc1155`            |                  |            | 28.12 K              | 94_224         |
| `erc20`              |                  |            | 9.34 K               | 47_112         |
| `erc721`             |                  |            | 13.17 K              | 125_632        |
| `flipper`            |                  |            | 1.63 K               | 15_704         |
| `incrementer`        |                  |            | 1.56 K               | 15_704         |
| `multisig`           |                  |            | 26.25 K              | 102_922        |
| `proxy`              |                  |            | 3.36 K               | 32_147         |
| `rand-extension`     |                  |            | 4.64 K               | 15_704         |
| `subber`             |                  |            | 2.48 K               |                |
| `trait-erc20`        |                  |            | 9.64 K               | 47_112         |
| `trait-flipper`      |                  |            | 1.41 K               | 15_704         |
| `trait-incrementer`  |                  |            | 1.54 K               | 31_408         |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/179702) | Last update: Thu Feb 17 19:01:16 CET 2022

xgreenx:
I think that everything is okay. Because if the user changed the ordering of the `impls` we will also update indexes during expanding of the macro. It is only for internal usage of builders so it is fine. The problem can be if the user will try to use that trait explicitly.

To avoid that we can add `const TRAIT_ID: u32` into the definition of the trait with a default value and use that index `TraitCallForwarderFor <Trait:: TRAIT_ID>`.

xgreenx:
> Indeed that was what I came up with initially, the question being how to generate that ID and guarantee that it is globally unique? The obvious answer is to hash module_path! + trait_ident. However I decided against that to go with the locally unique id, because it gives us the guarantee of uniqueness.

It can be XoR of all selectors of the trait(we already have a selector for each method).

> So question is whether we want to support the user explicitly using the CallBuilders TraitCallForwarderFor impls?

I prefer to not expose `TraitCallForwarderFor` for the user, it is why I removed it=) But we can add a comment that it is only for internal usage only, and please don't use it.

ascjones:
> It can be XoR of all selectors of the trait(we already have a selector for each method).

I've opened another PR #1141, and we can choose now between the 3 options :see_no_evil:

ascjones:
Superseded by #1141

Issue #787, opened 2021-05-10
New off-chain engine: Implement utility methods


cmichi:
Needed for https://github.com/paritytech/ink/issues/565.

- [ ] `random`
- [ ] `weight_to_fee`
- [ ] `minimum_balance`
- [ ] `gas_left`
- [ ] `block_number`
- [ ] And copy/pasted [from here](https://github.com/paritytech/ink/pull/712#discussion_r625033109):
For the test API we should additionally add an additional `count_reads` and `count_writes` over `HashMap<Vec<u8>, usize>` that mirrors the number of read and writes to every storage database cell agnostic to an account ID. Multiple contracts may share some storage space through runtime chain extensions etc. or balance changes etc.

PR #1143, opened 2022-02-22
Implement utility methods for experimental off-chain testing env

cmichi:
Closes https://github.com/paritytech/ink/issues/787.

Well, except an implementation for `gas_left`, because I'm uncertain what a sensible logic for that would be.

Yes, the experimental off-chain env is currently limited in that it only supports `ink_env::DefaultEnvironment`, but IMHO the situation of us currently having two off-chain testing engines in every examples is worse.

So I would like to get the experimental engine into a stage where we can throw the old one out and just use this new one by default everywhere. From my pov this PR is the last prerequisite before we can do that.

### DIFF
--- a/crates/engine/src/exec_context.rs
+++ b/crates/engine/src/exec_context.rs
@@ -15,7 +15,11 @@
 use super::types::{
     AccountId,
     Balance,
+    BlockNumber,
+    BlockTimestamp,
+    Hash,
 };
+use rand::Rng;
 
 /// The context of a contract execution.
 #[cfg_attr(test, derive(Debug, PartialEq, Eq))]
@@ -36,16 +40,27 @@ pub struct ExecContext {
     pub callee: Option<AccountId>,
     /// The value transferred to the contract as part of the call.
     pub value_transferred: Balance,
+    /// The current block number.
+    pub block_number: BlockNumber,
+    /// The current block timestamp.
+    pub block_timestamp: BlockTimestamp,
+    /// The randomization entropy for a block.
+    pub entropy: Hash,
 }
 
 #[allow(clippy::new_without_default)]
 impl ExecContext {
     /// Creates a new execution context.
     pub fn new() -> Self {
+        let mut entropy: [u8; 32] = Default::default();
+        rand::thread_rng().fill(entropy.as_mut());
         Self {
             caller: None,
             callee: None,
             value_transferred: 0,
+            block_number: 0,
+            block_timestamp: 0,
+            entropy,
         }
     }
 
@@ -63,6 +78,12 @@ impl ExecContext {
         self.caller = None;
         self.callee = None;
         self.value_transferred = Default::default();
+        self.block_number = 0;
+        self.block_timestamp = 0;
+
+        let mut entropy: [u8; 32] = Default::default();
+        rand::thread_rng().fill(entropy.as_mut());
+        self.entropy = entropy;
     }
 }
 
@@ -83,6 +104,10 @@ mod tests {
         assert_eq!(exec_cont.callee(), vec![13]);
 
         exec_cont.reset();
-        assert_eq!(exec_cont, ExecContext::new());
+        exec_cont.entropy = Default::default();
+
+        let mut new_exec_cont = ExecContext::new();
+        new_exec_cont.entropy = Default::default();
+        assert_eq!(exec_cont, new_exec_cont);
     }
 }
--- a/crates/engine/src/exec_context.rs
+++ b/crates/engine/src/exec_context.rs
@@ -48,10 +48,8 @@ pub struct ExecContext {
     pub entropy: Hash,
 }
 
-#[allow(clippy::new_without_default)]
-impl ExecContext {
-    /// Creates a new execution context.
-    pub fn new() -> Self {
+impl Default for ExecContext {
+    fn default() -> Self {
         let mut entropy: [u8; 32] = Default::default();
         rand::thread_rng().fill(entropy.as_mut());
         Self {
@@ -62,6 +60,13 @@ impl ExecContext {
             block_timestamp: 0,
             entropy,
         }
+    }
+}
+
+impl ExecContext {
+    /// Creates a new execution context.
+    pub fn new() -> Self {
+        Default::default()
     }
 
     /// Returns the callee.
@@ -75,15 +80,7 @@ impl ExecContext {
 
     /// Resets the execution context
     pub fn reset(&mut self) {
-        self.caller = None;
-        self.callee = None;
-        self.value_transferred = Default::default();
-        self.block_number = 0;
-        self.block_timestamp = 0;
-
-        let mut entropy: [u8; 32] = Default::default();
-        rand::thread_rng().fill(entropy.as_mut());
-        self.entropy = entropy;
+        *self = Default::default();
     }
 }
 
--- a/crates/engine/src/ext.rs
+++ b/crates/engine/src/ext.rs
@@ -425,12 +425,6 @@ impl Engine {
     ///   It provides the same behavior in that it will likely yield the
     ///   same hash for the same subjects within the same block (or
     ///   execution context).
-    ///
-    /// - Returned hashes on the surface might appear random, however for
-    ///   testing purposes the actual implementation is quite simple and
-    ///   computes those "random" hashes by wrapping XOR of the internal
-    ///   entry hash with the eventually repeated sequence of the subject
-    ///   buffer.
     pub fn random(&self, subject: &[u8], output: &mut &mut [u8]) {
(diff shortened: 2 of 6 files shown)


paritytech-cicd-pr:

These are the results when building the `examples/*` contracts from this branch with `cargo-contract 0.17.0-ab41bd9` and comparing them to ink! `master`: 

| `accumulator`        |                  |            | 1.28 K               |                |
| `adder`              |                  |            | 2.23 K               |                |
| `contract-terminate` |                  |            | 1.23 K               | 215_704        |
| `contract-transfer`  |                  |            | 8.37 K               | 15_704         |
| `delegator`          |                  |            | 6.26 K               | 33_142         |
| `dns`                |                  |            | 9.64 K               | 47_112         |
| `erc1155`            |                  |            | 27.71 K              | 94_224         |
| `erc20`              |                  |            | 9.12 K               | 47_112         |
| `erc721`             |                  |            | 14.40 K              | 125_632        |
| `flipper`            |                  |            | 1.57 K               | 15_704         |
| `incrementer`        |                  |            | 1.49 K               | 15_704         |
| `multisig`           |                  |            | 26.38 K              | 103_006        |
| `proxy`              |                  |            | 2.98 K               | 32_082         |
| `rand-extension`     |                  |            | 4.40 K               | 15_704         |
| `subber`             |                  |            | 2.24 K               |                |
| `trait-erc20`        |                  |            | 9.39 K               | 47_112         |
| `trait-flipper`      |                  |            | 1.35 K               | 15_704         |
| `trait-incrementer`  |                  |            | 1.45 K               | 31_408         |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/180416) | Last update: Wed Feb 23 06:44:09 CET 2022

crates/engine/src/ext.rs
```diff
@@ -358,12 +404,38 @@ impl Engine {
...
+    ///   computes those "random" hashes by wrapping XOR of the internal
+    ///   entry hash with the eventually repeated sequence of the subject
+    ///   buffer.
+    pub fn random(&self, subject: &[u8], output: &mut &mut [u8]) {
+        let mut entropy_bytes = self.exec_context.entropy;
+        let len_entropy = entropy_bytes.len();
+        for (n, subject) in subject.iter().enumerate() {
+            let id = n % len_entropy;
+            entropy_bytes[id] = entropy_bytes[id] ^ subject ^ (n as u8);
+        }
```
athei, line 437: No need to do this manually. Just seed a rng:
```suggestion
        let seed = (self.exec_context.entropy, subject).encode();
        let rng = rand::rngs::StdRng::from_seed(&seed);
        let mut rng_bytes: [u8; 32] = Default::default();
        rng.fill(&rng_bytes);
```

crates/engine/src/exec_context.rs
```diff
@@ -36,16 +40,27 @@ pub struct ExecContext {
...
     pub fn new() -> Self {
+        let mut entropy: [u8; 32] = Default::default();
+        rand::thread_rng().fill(entropy.as_mut());
         Self {
             caller: None,
             callee: None,
             value_transferred: 0,
+            block_number: 0,
+            block_timestamp: 0,
+            entropy,
```
HCastano, line 63: I think we can just implement `Default` for `ExecContext` and use that instead

crates/engine/src/exec_context.rs
```diff
@@ -63,6 +78,12 @@ impl ExecContext {
         self.caller = None;
         self.callee = None;
         self.value_transferred = Default::default();
+        self.block_number = 0;
+        self.block_timestamp = 0;
+
+        let mut entropy: [u8; 32] = Default::default();
+        rand::thread_rng().fill(entropy.as_mut());
+        self.entropy = entropy;
```
HCastano, line 86: Can override with the default implementation here (if you add it)

crates/engine/src/ext.rs
```diff
@@ -114,6 +118,37 @@ pub struct Engine {
...
     pub(crate) debug_info: DebugInfo,
+    /// The chain specification.
+    pub chain_spec: ChainSpec,
+}
+
+/// The chain specification.
+pub struct ChainSpec {
+    /// The current gas price.
+    pub gas_price: Balance,
+    /// The minimum value an account of the chain must have.
```
HCastano, line 129:
```suggestion
    /// The minimum value an account of the chain must have (a.k.a the existential deposit).
```

crates/engine/src/types.rs
```diff
@@ -12,15 +12,29 @@
...
+//! Right now the `engine` crate can only be used with the `ink_env::DefaultEnvironment`.
+//!
+//! In the long-term all types in this module should be `Vec<u8>`, as to not
+//! be dependent on any specific off-chain environment type. Then the `engine`
+//! crate could be used with an arbitrary `Environment` configuration.
+//!
+//! Right now the issue is that if e.g. some test calls an `engine` API function
+//! to transfer some balance, then there has to be some `+` operation to the
+//! existing balance. And we can't just add a `Vec<u8>` onto another one, we
+//! need to know the type.
```
HCastano, line 24: This seems like an internal comment (and something that should be part of the Github issue), not something that should be in the module level comments

crates/engine/src/types.rs
```diff
@@ -12,15 +12,29 @@
...
 use derive_more::From;
 
-/// This is just a temporary solution for the MVP!
-/// As a temporary solution we choose the same type as the default
-/// `env` `Balance` type.
-///
-/// In the long-term this type should be `Vec<u8>` as well, as to not
-/// be dependent on the specific off-chain environment type, so that
-/// the `engine` crate can be used with an arbitrary `Environment` configuration.
+/// As a temporary solution we choose the same type as the `DefaultEnvironment::Hash` type.
```
HCastano, line 28:
```suggestion
/// Same type as the `DefaultEnvironment::Hash` type.
```

HCastano: And same thing for all the other types

crates/engine/src/ext.rs
```diff
@@ -358,12 +404,38 @@ impl Engine {
...
     }
 
-    pub fn weight_to_fee(&self, _gas: u64, _output: &mut &mut [u8]) {
-        unimplemented!("off-chain environment does not yet support `weight_to_fee`");
+    /// Emulates gas price calculation.
+    pub fn weight_to_fee(&self, gas: u64, output: &mut &mut [u8]) {
+        let fee = self
+            .chain_spec
+            .gas_price
+            .saturating_mul(gas.try_into().unwrap_or(u128::MAX));
```
HCastano, line 412:
```suggestion
            .saturating_mul(gas.into());
```

HCastano (commented):
Small things to clean up

crates/engine/src/types.rs
cmichi: Will create a follow-up once merged.

HCastano (approved):

crates/engine/src/ext.rs
```diff
@@ -358,12 +410,29 @@ impl Engine {
...
+    /// # Note
+    ///
+    /// - This is the off-chain environment implementation of `random`.
+    ///   It provides the same behavior in that it will likely yield the
+    ///   same hash for the same subjects within the same block (or
+    ///   execution context).
+    pub fn random(&self, subject: &[u8], output: &mut &mut [u8]) {
+        let seed = (self.exec_context.entropy, subject).encode();
+        let mut rng = rand::rngs::StdRng::from_seed(
+            seed.try_into().expect("seed must be [u8; 32]"),
```
athei, line 431: This will panic. The entropy alone is 32byte and you concat the variable length subject to it. My suggestion was wrong. Forgot that the seed is fixed length. You need to hash the `seed` in order to get it to a fixed length. Is there a test or an example that should catch that?

HCastano: I hashed the seed and added a doc test

crates/engine/src/ext.rs
```diff
@@ -358,12 +410,39 @@ impl Engine {
...
+    ///   same hash for the same subjects within the same block (or
+    ///   execution context).
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    ///    let engine = ink_engine::ext::Engine::default();
+    ///    let seed = [0u8; 32];
+    ///    let mut output = [0u8; 32];
+    ///    engine.random(&seed, &mut output.as_mut_slice());
```
athei, line 435: Let's make it consistent with the argument name.

```suggestion
    ///    let subject = [0u8; 32];
    ///    let mut output = [0u8; 32];
    ///    engine.random(&subject, &mut output.as_mut_slice());
```

athei (approved):

PR #1144, opened 2022-02-23
Replace default off-chain engine with experimental one, remove old one

cmichi:
Follow-up to https://github.com/paritytech/ink/pull/1143.

The experimental off-chain env is currently limited in that it only supports `ink_env::DefaultEnvironment`, but IMHO the situation of us currently having two off-chain testing engines in every examples is worse.


athei (approved):
I agree. I think it is fine that it only supports the default env. The testing environment isn't a 100% match with the on chain environment anyways.

HCastano:
> −4,528 

❤️

crates/env/src/backend.rs
```diff
@@ -46,7 +46,7 @@ impl ReturnFlags {
     }
 
     /// Returns the underlying `u32` representation.
-    #[cfg(not(feature = "ink-experimental-engine"))]
+    #[cfg(not(feature = "std"))]
```
HCastano, line 49: Why does this need to be behind a feature flag?

examples/erc20/lib.rs
```diff
@@ -800,9 +482,31 @@ mod erc20 {
...
-                ink_env::test::recorded_events().collect::<Vec<_>>();
-            assert_eq!(emitted_events_before.len(), emitted_events_after.len());
+            assert_eq!(
+                emitted_events_before,
+                ink_env::test::recorded_events().count()
+            )
+        }
+    }
+
+    struct PrefixedValue<'a, 'b, T> {
```
HCastano, line 492: Why's this needed?

HCastano (commented):
Two small questions. Once the CI is fixed it should be good to go

crates/env/src/backend.rs
cmichi: In `std` the function is never used.

PR #1163, opened 2022-03-03
Remove `ink-as-dependency`

HCastano:
As part of the `ink!` [RC7 release](https://github.com/paritytech/ink/blob/master/RELEASES.md#version-30-rc7) the `ink!` codegen started generating a `ContractRef`
struct alongside the `Contract` struct. Since then there has been no need for
`ink-as-dependency` because all contracts get a generated `Ref` struct which can be used 
anywhere `ink-as-dependency` would've been used.

The [release notes](https://github.com/paritytech/ink/blob/master/RELEASES.md#changed-2) even mention that the `Ref` struct was meant to be the successor to the
`ink-as-dependency` concept.

> MyContractRef is pretty much the same of what we had gotten with the old
> ink-as-dependency. It is a typed thin-wrapper around an AccountId that is mirroring the
> ink! smart contract's API and implemented traits.

This PR removes `ink-as-dependency` since it is no longer needed.


HCastano:
I'll fix the CI tomorrow 😔

xgreenx:
You can't just remove the `ink-as-dependency` feature because it is still used to prevent multiple `call`, `deploy`, and `__ink_generate_metadata`.

@Robbepop mentioned in that https://github.com/paritytech/ink/pull/665 the idea with `in-as-root` feature. All no mangle methods can be marked with that feature and `cargo-contract` can enable it during compilation.

HCastano:
> You can't just remove the `ink-as-dependency` feature because it is still used to prevent multiple `call`, `deploy`, and `__ink_generate_metadata`.
> 
> @Robbepop mentioned in that #665 the idea with `in-as-root` feature. All no mangle methods can be marked with that feature and `cargo-contract` can enable it during compilation.

Hmm yeah, I figured the `ContractRef` impl was a bit more clever, but you're right we
still need some way to prevent multiple of those `no_mangle` calls.

So the main thing I really wanted to get rid of here was the
`#[ink::contract(compile_as_dependency = ...)]` syntax from the language.

I think we can still do that without removing the `ink-as-dependency` feature flag. I
don't really see a difference between that and `ink-as-root`, the feature will still need
to be enabled/disabled _somewhere_.

xgreenx:
> I think we can still do that without removing the ink-as-dependency feature flag. I
don't really see a difference between that and ink-as-root, the feature will still need
to be enabled/disabled somewhere.

`ink-as-root` should be enabled only for the root contract, not for subcontracts, and `cargo-contract` can enable it because the developer calls `cargo contract build` for the root contract.

`ink-as-dependency` should be enabled for all subcontracts, and `cargo-contract` can't recognize what contract is sub contract=)

HCastano:
Okay so I'm gonna split this out into two PRs.

1. Remove `compile_as_dependency`
2. Replace `ink-as-dependency` with `ink-as-root`

For (1) I've opened #1168, and for (2) I've started playing around with something today, will
maybe open a PR tomorrow.

PR #1133, opened 2022-02-15
`CallBuilder` delegated calls API

VargSupercolony:
This PR adds the possibility to make delegated calls into ink!. To achieve this, we:
1. Add `seal_delegate_call` unstable interface import
2. Refactor `CallBuilder` to two types of calls: `Call` (meaning cross-contract call) and `DelegateCall` (meaning delegated calls). This is a breaking change, as now the users must specify `call_type` on the builder manually, as opposed to it being always `Call` :)
3. Add `invoke_contract_delegate` & `eval_contract_delegate` functions to backend trait & impls: we call those functions when making a delegated call
4. Rework codegen for cross-contract calls (add `set_call_type`)
5. Update docs in the ink! itself
6. Update `proxy` example, rework it to use `build_call` with `DelegateCall` call type, add upgradeable flipper to showcase simple upgradeable contract example.

Massive thanks to @yarikbratashchuk for providing initial PoCs & ideas for the implementation.

PR #1165, opened 2022-03-04
Unify `ink_env::{eval_contract, invoke_contract}`

ascjones:
WIP exploring how easy it is to unify those two functions and simplify the API. Requires explicitly setting of `()` as the return type on the call builder. 

If using the call builder API explicitly to invoke a message which does have a return value but you want to skip decoding it, as was the previous behaviour of `invoke_contract`,  the user can simply set the return type to `()`.

### DIFF
--- a/crates/env/src/call/call_builder.rs
+++ b/crates/env/src/call/call_builder.rs
@@ -84,38 +84,17 @@ where
     }
 }
 
-impl<E, Args> CallParams<E, Args, ()>
-where
-    E: Environment,
-    Args: scale::Encode,
-{
-    /// Invokes the contract with the given built-up call parameters.
-    ///
-    /// # Note
-    ///
-    /// Prefer [`invoke`](`Self::invoke`) over [`eval`](`Self::eval`) if the
-    /// called contract message does not return anything because it is more efficient.
-    pub fn invoke(&self) -> Result<(), crate::Error> {
-        crate::invoke_contract(self)
-    }
-}
-
 impl<E, Args, R> CallParams<E, Args, ReturnType<R>>
 where
     E: Environment,
     Args: scale::Encode,
     R: scale::Decode,
 {
-    /// Evaluates the contract with the given built-up call parameters.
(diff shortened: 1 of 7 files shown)


crates/env/src/call/call_builder.rs
```diff
@@ -192,7 +171,7 @@ where
 ///             .push_arg(true)
 ///             .push_arg(&[0x10; 32])
 ///     )
-///     .returns::<ReturnType<i32>>()
+///     .returns::<i32>()
```
HCastano, line 174: ❤️ that users aren't exposed to `ReturnType<T>` anymore

HCastano: In the PR description you mention that _"the user can either omit the call to `returns::<R>()`"_. Doesn't looks like this is the case, if I remove this line the example fails to compile

HCastano (approved):
Good stuff!

Heads up @xgreenx, this will cause some merge conflicts with #1133

HCastano:
CI fails due to the [`contract-introspection`](https://github.com/paritytech/ink-waterfall/blob/7247c49a3690be9ad97a5e32340b32e198d68b4e/examples/contract-introspection/lib.rs#L46) example that's part of the `ink-waterfall`. Let's get this merged and then update the waterfall

crates/env/src/call/call_builder.rs
ascjones: Oh it appears that is only the case if the arg list is not set  (which in this example it is) see https://github.com/paritytech/ink/blob/c0cccb74270f4ef9b92ea3dbde7a06b25f1bf091/crates/env/src/call/call_builder.rs#L438.

Here is an example which calls explcitily but with no args so that compiles: https://github.com/paritytech/ink/blob/378d83f3fc204e18dce6f3946e6a0f27e333d7d6/examples/proxy/lib.rs#L80

I'm not sure why it was already this way, it would make sense to either allow it to be `Unset` either way and default to `()`, or require it to be set in all cases. What do you think?

PR #1169, opened 2022-03-08
We forgot to add an ability to alter the arguments of the call

xgreenx:
The developer can't alter the default arguments of codegen.


cmichi (approved):

ascjones (commented):
Can you elaborate a bit? In the doc comments we are setting `callee` `transferred_value` and `gas_limit` directly on initialization via `call_type`. So where would we use these new `CallBuilder` methods instead?

xgreenx:
If `call_type` is initialized, the user can't alter the values inside. For example, [codegen here](https://github.com/paritytech/ink/blob/master/crates/lang/codegen/src/generator/trait_def/call_builder.rs#L391) only initializes the `callee` field(`.call_type(::ink_env::call::Call::new().callee(::ink_lang::ToAccountId::to_account_id(self)))`). And if some wants to modify `transferred_value` he can't do that.

PR #1173, opened 2022-03-10
Debug failing macOS CI

HCastano:
macOS CI on `master` is failing, trying to figure out why. Current hunch is that it's a caching issue. 

These recent failed after restoring from cache: [1](https://github.com/paritytech/ink/runs/5485085095?check_suite_focus=true), [2](https://github.com/paritytech/ink/runs/5386862451?check_suite_focus=true) 

These recent runs pass with a clean cache: [3](https://github.com/paritytech/ink/runs/5484036329?check_suite_focus=true), [4](https://github.com/paritytech/ink/runs/5425286873?check_suite_focus=true)

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -10,8 +10,8 @@ on:
       - 'README.md'
       - '.gitlab-ci.yml'
   pull_request:
-    branches:
-      - master
+      # branches:
+      #   - master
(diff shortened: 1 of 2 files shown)


cmichi:
@HCastano I think we can close this, since that CI stage seems to run fine again now.

PR #1179, opened 2022-03-13
limit codegen-units to get the maximum optimization

sabify:
Hello,

There might be a performance regression due to enabling lto without limiting the number of code generation units (codegen-units = 1).

References:
https://github.com/rust-lang/rust/issues/47745
https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,3 +23,6 @@ exclude = [
 [profile.release]
 panic = "abort"
 lto = true
+# https://github.com/rust-lang/rust/issues/47745
+# https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units
+codegen-units = 1


xgreenx:
[`codegen-units` is `1` by default](https://github.com/paritytech/cargo-contract/blob/master/src/workspace/profile.rs#L35) for release contract, so it is enabled.

ascjones:
@xgreenx is right, it is already set by `cargo-contract`

sabify:
If the `ink` is solely dependent on `cargo-contract`, how about removing the whole release profile block to avoid misconceptions?

ascjones:
Yes we should do that since the profile for the ink workspace itself is not used when building a contract.

PR #1209, opened 2022-04-05
Optimize `push_topic`: write prefix bytes directly to buffer

ascjones:
While topics/events needs a broader overhaul to reduce code size, here is a smaller incremental change which reduces e.g. erc20 from `8.5KB` to `8.2KB`.

I removes the `PrefixedValue` type and passes the prefix bytes directly to `push_topic` so they can be appended to the buffer. So it saves on bringing in some unneeded `Encode` implementations.

Draft while we wait for CI to tell us improvements.

### DIFF
--- a/crates/env/src/engine/off_chain/impls.rs
+++ b/crates/env/src/engine/off_chain/impls.rs
@@ -134,7 +134,7 @@ where
     where
         T: scale::Encode,
     {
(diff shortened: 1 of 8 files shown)


ascjones:
This doesn't work, it just makes things worse

Issue #1134, opened 2022-02-16
Macro based storage rework


athei:
Most of the existing storage types [were removed](https://github.com/paritytech/ink/pull/1111) because they are simply to inefficient in size and executed instructions. Right now we only have `Mapping`. However, this type doesn't work very well with the existing storage infrastructure and needs a band aid named `SpreadAllocate` and `initialize_contract`.

We want to pivot more in the direction of how storage works in FRAME. We should keep the overarching contract data structure but use the storage macro to generate the supporting code instead of a elaborate system of traits.

This would also allow us more flexibility in how to generate the storage keys for these data structures and more control over what code is generated. The aim is to reduce the contract sizes by removing monomorphization and copying of the root keys.

Please note that the following is only a sketch of the generated code. It is probably not how the generated code should look like. @Robbepop noted that we should not use private sub modules. Please feel free to post your suggestions and I will update the top post.

We add a new trait that will be implemented by the storage macro to supply our storage traits with a key. Our storage traits have default implementations for every function. The macro will only set the associated types.

```rust
trait HasKey {
    const KEY: [u8; 32];
}

trait StorageValue: HasKey {
    type Value: Encode + Decode;
    
    // we take self here to build in the mutability levels
    fn write(&mut self, value: &Value) {
        ink_env::set_contract_storage(Self::KEY, value.encode());
    }
    
    fn read(&self) -> Self::Value {
        ink_env::get_contract_storage(Self::KEY)
    }
}

trait StorageMap: HasKey {
    type Key: Hashable;
    type Value: Encode + Decode;
    
    fn insert(&mut self, key: Self::Key, &value: Self::Value) {
        ink_env::set_contract_storage(twoxx(Self::KEY ++ key), value.encode());
    }
    
    fn get(&self, key: Self::Key) -> Self::Value {
        ink_env::get_contract_storage(twoxx(Self::KEY ++ key));
    }
}
```

Then the storage macro creates a private module where it dumps all the generated code:

```rust
#[ink(storage)]
struct Contract {
    a: u8,
    b: Mapping<AccountId, Balance>,
}

// The macro would transfrom the struct into this:
// The underscores are just to prevent clashes with user defined modules
// Having public fields in the struct would be an error.
mod __private__ {
    pub struct AFieldStorageValue {
        /// prevent instantation outside of this generated code
        _private: ()
    }
    
    pub struct BFieldMapping {
        /// prevent instantation outside of this generated code
        _private: ()
    }
    
[...]

PR #1217, opened 2022-04-11
Storage rework

xgreenx:
Implements https://github.com/paritytech/ink/issues/1134

### Overview
It is an implementation of new storage of smart contracts. Previously each field was stored in its storage cell under its storage key. The storage key was calculated in runtime over fields iteration. In a new version, all atomic fields are stored in **one** storage cell under **one** storage key. All non-atomic fields know their storage key during compilation time.

- Removed old traits: `SpreadLayout`, `PacketLayout`, `SpreadAllocate`, `PacketAllocate`, and all related code.
- ~Removed `ink_storage::collection` and `ink_storage::lazy`. A new collections/lazy should be added in follow-up PRs~(https://github.com/paritytech/ink/pull/1271).

Introduced new traits: `StorageType`, `StorageKeyHolder`, `AtomicGuard`. All those traits are implemented for primitives and base collections.
- `StorageType` and `StorageKeyHolder` is used by a new `#[ink_lang::storage_item]` macro to automate storage key calculation during compilation.
- `AtomicGuard` prevents the usage of non-atomic structs/enums in collections or mappings. It is a guard for the user to not corrupt the storage. `AtomicGuard` can be derived for the structure or enum if all fields are atomic(it can be implemented manually. You can do that if you sure that all types are atomic).

The `scale::Encode` and `scale::Decode` is used to pull and push storage.

The `StorageKey` is `u32`. Due to some limitations of the host functions, we still use the `[u8; 32]` storage key in some places but it should be fixed with https://github.com/paritytech/substrate/issues/11029.

### Atomic vs Non-atomic
The field is atomic if it doesn't require its storage cell.

New `ink_storage::Mapping` and `ink_storage::StorageValue` allow to occupy additional storage cells. The storage key is calculated for those types during the compilation time. The rules about automatic storage key calculation you can find below. By default, those types use the `ink_storage::traits::AutoKey` policy, but anyone can specify that a storage key should be used via  `ink_storage::traits::ManualKey` like:
```rust
#[ink(storage)]
#[derive(Default)]
pub struct Erc20 {
    /// Total token supply.
    total_supply: Balance,
    /// Mapping from owner to number of owned token.
    balances: Mapping<AccountId, Balance, ManualKey<123>>,
    /// Mapping of the token amount which an account is allowed to withdraw
    /// from another account.
    allowances: Mapping<(AccountId, AccountId), Balance, ManualKey<456>>,
}
```

The specified storage key is more prior than autogenerated. Any struct or enum, that uses those types, becomes non-atomic(more non-atomic types in follow-up PRs).

### `#[ink_lang::storage_item]`

`#[ink_lang::storage_item]` should be used during the declaration of a new struct or enum if it should be part of the storage. The macro derives all required traits by default:
```rust
#[derive(
    ::ink_storage::traits::AtomicGuard,
    ::ink_storage::traits::StorageType,
    ::ink_storage::traits::StorageKeyHolder,
    ::scale::Encode,
    ::scale::Decode,
)]
#[cfg_attr(feature = "std", derive(
    ::scale_info::TypeInfo,
    ::ink_storage::traits::StorageLayout,
))]
```
But it can be disabled via the `derive` argument to implement them manually.

```rust
#[ink_lang::storage_item(derive = false)]
#[derive(Default)]
struct Contract {
    a: u16,
    b: u64,
    c: u128,
}

// Disabling of deriving allow to implement the trait manually
impl<KEY: StorageKeyHolder> AtomicGuard<true> for Contract<KEY> {}
```
If you declared and planned to use a non-atomic type twice or more, you need to declare it with a generic storage key `KEY: StorageKeyHolder`. It will propagate the storage key from parent to child to prevent the same storage key.

```rust
#[derive(Default)]
struct NonAtomic<KEY: StorageKeyHolder = AutoKey> {
    field: StorageType<u128>,
}

#[ink_lang::storage_item]
#[derive(Default)]
struct Contract {
    // Storage key - Hash(Hash("Contract::a"), Hash("NonAtomic::field"))
    a: StorageType<NonAtomic>,
    // Storage key - Hash(Hash("Contract::b"), Hash("NonAtomic::field"))
    b: StorageType<NonAtomic>,
}
// Without `NonAtomic<KEY: StorageKeyHolder = AutoKey>` the storage key in both cases is 
// Hash("NonAtomic::field") -> conflicting storage keys
```
The macro expects `StorageKeyHolder` naming. It is hardcoded and you can't use aliases.

### Storage key rules

Right now hashing is used `SHA2` because it allows to do it during compilation time.
Maybe in the future, it will be changed if `Blake2b` will support const calculation.

The rules to calculate the storage key of the field:

1. Compute the ASCII byte representation of `struct_name` and call it `S`.
1. If `variant_name` is `Some` then computes the ASCII byte representation and call it `V`.
1. Compute the ASCII byte representation of `field_name` and call it `F`.
1. Concatenate (`S` and `F`) or (`S`, `V` and `F`) using `::` as separator and call it `C`.
1. Apply the `SHA2` 256-bit hash `H` of `C`.
1. The first 4 bytes of `H` make up the storage key.

- `variant_name` is `None` for structures and unions.
- if the field is unnamed then `field_name` is `"{}"` where `{}` is a number of the field("0", "1" ...).

If the parent's storage key is passed then use the next rules: If one of the keys is zero, then return another
without hashing. If both keys are non-zero, return the hash of both keys.

If the type is a tuple, then it uses additional rules to calculate the inner storage key:

- The storage key is generated based on the tuple name and field number concatenated with `::`.
- The name of the tuple looks like: "(A)", "(A, B)" ... "(A, B, ..., J)" where `A`, `B`, ..., `J` are hardcoded.
- Fields number looks like: "0", "1", ..., "9"

The final:
For `(A)` it is `(A)::0`
For `(A, B)` it is `(A, B)::0`, `(A, B)::1`
...
For `(A, B, ..., J)` it is `(A, B, ..., J)::0`, `(A, B, ..., J)::1`, ..., `(A, B, ..., J)::9`

### Contract customizability

- Remove `ContractRootKey` trait.
- Added support of `ManualKey` for the contract's storage. Now anyone can specify the default storage key of the contract.
```rust
/// A simple ERC-20 contract.
#[ink(storage)]
#[derive(Default)]
pub struct Erc20<K: StorageKeyHolder = ManualKey<123>> {
    /// Total token supply.
    total_supply: Balance,
    /// Mapping from owner to number of owned token.
    balances: Mapping<AccountId, Balance>,
    /// Mapping of the token amount which an account is allowed to withdraw
    /// from another account.
    allowances: Mapping<(AccountId, AccountId), Balance>,
}
```
- Added a new trait `OnCallInitializer`. The contract can implement that trait to support initialization on the runtime if it is unable to pull from the storage.
It can be in several cases:
  - The contract doesn't have constructor. That initializer can be alternative for the constructor.
  - The constructor was not called due to upgrade ability, `Proxy` or `Diamond` pattern.
  - The storage was moved or corrupted.
If the trait is not implemented the behavior of the storage is the default - It should be first initialized by the constructor.

### Metadata

The metadata was reworked to be fully compatible with storage key generation rules. The storage layout now uses the `u32` layout key instead of `[u8; 32]`. Each struct or enum layout has a name and all unnamed fields have a enumerate name like "0", "1" etc. Now everyone can calculate the storage key based on the naming that was used. The metadata still is version 3, it should be updated to version 4 in https://github.com/paritytech/ink/issues/1265.

Before the generation of the metadata, it checks that the metadata is valid. Right now it can be invalid if some storage keys intersect(maybe more errors in the future).

The conflict storage key error looks like:
image

The storage layout example (it's pretty large, so it's hidden by default):

<details>
<p>

```json
{
  "storage": {
    "root": {
      "layout": {
        "struct": {
          "fields": [
            {
              "layout": {
                "leaf": {
                  "key": "0x00000000",
                  "ty": 0
                }
              },
              "name": "balance"
            },
            {
              "layout": {
                "root": {
                  "layout": {
                    "struct": {
                      "fields": [
                        {
                          "layout": {
                            "leaf": {
                              "key": "0x9cbeb3b2",
                              "ty": 0
                            }
                          },
                          "name": "a"
                        },
                        {
                          "layout": {
                            "leaf": {
                              "key": "0x9cbeb3b2",
                              "ty": 1
                            }
                          },
                          "name": "b"
                        },
                        {
                          "layout": {
                            "struct": {
                              "fields": [
                                {
                                  "layout": {
                                    "leaf": {
                                      "key": "0x9cbeb3b2",
                                      "ty": 0
                                    }
                                  },
                                  "name": "0"
                                },
                                {
                                  "layout": {
                                    "leaf": {
                                      "key": "0x9cbeb3b2",
                                      "ty": 2
                                    }
                                  },
                                  "name": "1"
                                }
                              ],
                              "name": "(A, B)"
                            }
                          },
                          "name": "c"
                        }
                      ],
                      "name": "Atomic"
                    }
                  },
                  "root_key": "0x9cbeb3b2"
                }
              },
              "name": "atomics"
            },
            {
              "layout": {
                "root": {
                  "layout": {
                    "struct": {
                      "fields": [
                        {
                          "layout": {
                            "root": {
                              "layout": {
                                "leaf": {
                                  "key": "0x70d94bd5",
                                  "ty": 0
                                }
                              },
                              "root_key": "0x70d94bd5"
                            }
                          },
                          "name": "a"
                        },
                        {
                          "layout": {
                            "root": {
                              "layout": {
                                "struct": {
                                  "fields": [
                                    {
                                      "layout": {
                                        "leaf": {
                                          "key": "0x506b3e18",
                                          "ty": 0
                                        }
                                      },
                                      "name": "0"
                                    },
                                    {
                                      "layout": {
                                        "leaf": {
                                          "key": "0x506b3e18",
                                          "ty": 2
                                        }
                                      },
                                      "name": "1"
                                    }
                                  ],
                                  "name": "(A, B)"
                                }
                              },
                              "root_key": "0x506b3e18"
                            }
                          },
                          "name": "b"
                        },
                        {
                          "layout": {
                            "root": {
                              "layout": {
                                "struct": {
                                  "fields": [
                                    {
                                      "layout": {
                                        "leaf": {
                                          "key": "0x84a13f7d",
                                          "ty": 0
                                        }
                                      },
                                      "name": "0"
                                    },
                                    {
                                      "layout": {
                                        "leaf": {
                                          "key": "0x84a13f7d",
                                          "ty": 2
                                        }
                                      },
                                      "name": "1"
                                    }
                                  ],
                                  "name": "(A, B)"
                                }
                              },
                              "root_key": "0x84a13f7d"
                            }
                          },
                          "name": "c"
                        }
                      ],
                      "name": "NonAtomic"
                    }
                  },
                  "root_key": "0x0000007c"
                }
              },
              "name": "non_atomic"
            }
          ],
          "name": "Erc20"
        }
      },
      "root_key": "0x00000000"
    }
  }
}
```
</p>
</details>

### Follow-ups
- If GATs are stabilized we can simplify the `StorageType` and use generics salt in the associated type.
- `#[ink_lang::storage_item]` may have problems with bounds for inner types. We can implement the approach from that [comment](https://github.com/paritytech/ink/pull/1217#issuecomment-1121438006).
- Implement a `Lazy` non-atomic type, that allows loading of the storage type in a lazy manner.
- Fully remove usage of `[u8; 32]`
- Update metadata to version 4 in polkadot-js


xgreenx:
The change already works for `Erc20` and `Erc1155` examples. So you can check what it looks like. If you are okay with it, we can start moving in direction of removing old traits, adding comments, tests and so on.

The size reduction:
`Erc20` from 8535 -> 7972  = -563
`Erc1155` from 17447 -> 16820 = -627

Example of storage layout of `Erc20` with custom root key:
```json
"storage": {
  "struct": {
    "fields": [
      {
        "layout": {
          "cell": {
            "key": "0x00000123",
            "ty": 0
          }
        },
        "name": "total_supply"
      },
      {
        "layout": {
          "cell": {
            "key": "0xd44dbd33",
            "ty": 1
          }
        },
        "name": "balances"
      },
      {
        "layout": {
          "cell": {
            "key": "0x429121b3",
            "ty": 9
          }
        },
        "name": "allowances"
      }
    ]
  }
},
```

The change has several differences from what we discussed. We added `AutoKey` and `ManualKey` as possible keys. If the user specifies the manual key, it overrides the auto key. 

If the user wants to use the non-atomic struct twice or more he should specify a generic with `StorageKeyHolder` instead of const with `StorageKey`. It is done to provide an ability to the user to specify the manual key for the struct. That feature is used to specify the root storage key for the contract. An example is the `Erc20` definition:

```rust
/// A simple ERC-20 contract.
#[ink(storage)]
#[derive(Default)]
pub struct Erc20<KEY: StorageKeyHolder = ManualKey<0x123>> {
    /// Total token supply.
    total_supply: Balance,
    /// Mapping from owner to number of owned token.
    balances: StorageMapping<AccountId, Balance>,
    /// Mapping of the token amount which an account is allowed to withdraw
    /// from another account.
    allowances: StorageMapping<(AccountId, AccountId), Balance>,
}
```

That feature also is useful for upgradable contracts t specify the storage key of the upgradable struct.

Also in `lib.rs` of the `Erc20` example, you can see definitions of atomic and non-atomic structures.

```rust
#[ink_lang::storage_item]
struct Atomic {
    s1: u128,
    s2: Vec<u128>,
    // Fails because `StorageType` implemented for Vec if T is not AtomicGuard<true>
    // s3: Vec<NonAtomic>,
}

#[ink_lang::storage_item]
struct NonAtomic {
    s1: StorageMapping<u128, u128>,
    s2: StorageValue<u128>,
}

#[ink_lang::storage_item]
struct Jora<KEY: StorageKeyHolder> {
    s1: StorageMapping<u128, u128>,
    s2: StorageValue<u128>,
    s3: StorageMapping<u128, Atomic>,
    s4: StorageValue<NonAtomic>,
    // Fails because: the trait `AtomicGuard<true>` is not implemented for `NonAtomic`
    // s5: StorageMapping<u128, NonAtomic>,
}
```

In the comment, you can see how it is expanded.

We need to get a feedback from you to work in direction of finilization of the change.

xgreenx:
We've also commented out on a lot of stuff in [`lazy` module](https://github.com/Supercolony-net/ink/blob/184c55000f382400f931fafe0762ba8c812393f5/crates/storage/src/lazy/mod.rs#L55). The question: Do we need to support it and update according to new storage? Or can we remove it, or leave it commented out?

crates/lang/codegen/src/generator/storage_item.rs
```diff
@@ -0,0 +1,228 @@
...
+                                    #manual_key_ident,
+                                    #manual_salt_ident
+                                >
+                            });
+                        } else {
+                            auto_key_ty_generics.push(t.ident.to_token_stream());
+                            manual_key_ty_generics.push(t.ident.to_token_stream());
+                        }
+                    }
+                    GenericParam::Lifetime(l) => {
```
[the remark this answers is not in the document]
xgreenx: It is not necessary. That part of the code doesn't touch the generic specified by the user. It searches for the "salt" generic and modifies it by passing the right storage key.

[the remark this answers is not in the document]
xgreenx: The example:

```rust
#[ink::storage_item]
struct ComplexStruct<'a, T, Salt: StorageKeyHolder, V>{ ... };

impl<'a, T, Salt: StorageKeyHolder, V, ManualSalt: StorageKeyHolder, const MANUAL_KEY: StorageKey> StorageType<Salt> for ComplexStruct<'a, T, ManualKey<MANUAL_KEY, ManualSalt>, V>
 {
     type Type = StorageMapping<K, V, ManualKey<MANUAL_KEY, ManualSalt>>;
 }

 impl<'a, T, Salt: StorageKeyHolder, V> StorageType<Salt>
     for ComplexStruct<'a, T, AutoKey, V>
 {
     type Type = ComplexStruct <'a, T, ManualKey<0, Salt>, V>;
 }
```

crates/lang/codegen/src/generator/storage_item.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        let fields = item.fields().map(|field| {
+            let key_bytes: Vec<u8> = if let Some(field_ident) = &field.ident {
+                [
+                    &ident.to_string().into_bytes()[..],
+                    &field_ident.to_string().into_bytes()[..],
+                ]
+                .concat()
+            } else {
+                ident.to_string().into_bytes()
```
ascjones, line 70: So tuple structs have colliding keys?

xgreenx: In the case of tuple structs like `struct Test1(u32, u8)` yes, only the name of the struct defines the storage key, and if someone used two structs twice(or `Mapping`, or `StorageValue`), it generates the same storage key in each field.

So `struct Test1(Mapping<u32, u8>, Mapping<u128, u64>)` will work wrongly.

We can leave it as it is and emit errors during checking of storage key collisions, or add the restriction to not allow tuple structs.

xgreenx: Another question how do we want to work with enums?
How should we calculate the storage key of each field in the variant?

It can be `NAME_OF_ENUM + NAME_OF_VARIANT + INDEX_OF_FIELD`.

ascjones:
> We've also commented out on a lot of stuff in [`lazy` module](https://github.com/Supercolony-net/ink/blob/184c55000f382400f931fafe0762ba8c812393f5/crates/storage/src/lazy/mod.rs#L55). The question: Do we need to support it and update according to new storage? Or can we remove it, or leave it commented out?

IMO we can remove these (and in turn `collections`) since they are all based on the legacy storage abstractions.

Also they are currently internal and [not exposed](https://github.com/paritytech/ink/blob/992706ee6715866d2ecbc1ffec7c3405980c58f9/crates/storage/src/lib.rs#L57) as part of the public API.

As a future task we can revisit these legacy lazy/collection types to see whether it is worth adapting them to the new storage abstractions in terms of code size/gas costs.

cmichi:
>The StorageKey is u32 but at the end, we convert it into [u8; 32] because the host function expects 32 bytes=(

Could you quickly recap why `u32` is fine?

In a next iteration we can introduce a new `seal_…` host function that takes the primitive directly.

I agree with @ascjones that we can remove `Lazy`.

From our pov it looks like the right direction, so it's fine if you go ahead with adding docs and tests.

xgreenx:
> Could you quickly recap why u32 is fine?

We [agreed on that in the issue](https://github.com/paritytech/ink/issues/1134#issuecomment-1062948188). The chance of collision is low and it is still [optimal](https://github.com/paritytech/ink/issues/1134#issuecomment-1051303774) for us. We don't need `[u8; 32]` because `contract-pallet` hashes it too.

> In a next iteration we can introduce a new seal_… host function that takes the primitive directly.

It will be resolved by https://github.com/paritytech/substrate/issues/11029. I meant that the size will be reduced more in the future, but right now it is `[u8; 32]`=)

> I agree with @ascjones that we can remove Lazy.
From our pov it looks like the right direction, so it's fine if you go ahead with adding docs and tests.

Okay, cool, thanks for review=)

athei:
> In a next iteration we can introduce a new seal_… host function that takes the primitive directly.

The next version will take a ptr/len pair. So arbitrary length. ink! will use `u32` for now. Since those keys are stored in static data we can save more space by just using the smallest possible type. We detect collisions by scanning the metadata.

xgreenx:
The current implementation of the trait that returns the type with the right storage key looks like this:
```rust
/// Returns the type that should be used for storing the value
pub trait StorageType<Salt: StorageKeyHolder> {
    /// Type with storage key inside
    type Type;
}
```

It works well if we don't want to support generics in structures that can be part of the storage. Because we can't define that kind of struct.
```rust
// Compilation error because `Salt` is not part of the struct + we will use another `ManualKey` 
struct Generic<T: StorageType<Salt>, Salt: StorageKeyHolder> {
    s1: u128,
    s2: u32,
    s3: T,
}
```

I created a new [StorageType2](https://github.com/Supercolony-net/ink/blob/7e4b56dd9ed4e3fbfcb9deacbee26ed987ae067d/crates/storage/src/traits/storage.rs#L40) trait with upcoming stabilization of [GATs](https://github.com/rust-lang/rust/pull/96709). And it helped to simplify the code with auto storage key selection, simplify codegen, and now the usage of the trait is very user-friendly. That allows us to define the structs and enums with generic now. We still have a problem with the auto-generated `AtomicGuard` trait, but it can be resolved by a manual implementation.

I played with it, and it is better than before, but still, I have issues with `#[derive(Default, Debug, ...]`. Because it requires bounds for `StorageType2::Type<Salt>: Default + Debug + ...`. And with GATs I came up with the next solution.
[Here](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=884bb41f17ae681efa16d9f54f3a200c) is an example of the code.

We can create a fully generic private structure, that structure can have a lot of bounds from the derive. After that, we create an alias to that structure with the same name and same generics. It uses already actual types there.

The user defines the next struct:
```rust
#[ink_lang::storage_item]
struct Struct<T: StprageType> {
     s1: u32,
     s2: u64,
     s3: u128,
     s4: T,
}
```

`#[ink_lang::storage_item]` generates the next code:
```rust
#[allow(non_snake_case)]
mod private_Struct {
    use super::*;
    
    #[derive(Default, Debug)]
    pub struct Struct<A: StorageType, B: StorageType, C: StorageType, D: StorageType> {
        pub s1: A::Type<ManualKey>,
        pub s2: B::Type<ManualKey>,
        pub s3: C::Type<ManualKey>,
        pub s4: D::Type<ManualKey>,
    }
    
    impl<A: StorageType, B: StorageType, C: StorageType, D: StorageType> IsAtomic for Struct<A, B, C, D>
    where
        A: IsAtomic,
        B: IsAtomic,
        C: IsAtomic,
        D: IsAtomic,
    {}
}

type Struct<T> = private_Struct::Struct<u32, u64, u128, T>;
```

For the user, it is the same type, but for us, we applied all bounds that we need. It also allows us to improve the implementation of the `AtomicGuard<true>`. Instead of evaluating it via the [`is_atomic!`](https://github.com/Supercolony-net/ink/blob/7e4b56dd9ed4e3fbfcb9deacbee26ed987ae067d/crates/storage/src/is_atomic.rs#L34) macro, we can create a generic implementation. Where if all types are `IsAtomic` then the struct is too `IsAtomic`.

If we are belive that GATs would be stabilized, I will use the implementation with GATs and remove old one.

What do you think about the idea of a private struct? 
Pros:
- Support of all derives and unlimited bounds. -> We can implement `IsAtomic` in ore natural way.
- With GATs can support generics.
Cons:
- More complex codegen.
- Maybe, not clear errors for the user with aliases.
- I didn't find other problems except for magic=) 

The idea with a private struct can be implemented without GATs, but that means that we will not support generics=) If you are okay with that idea I will implement it.

ascjones:
> It works well if we don't want to support generics in structures that can be part of the storage

Do we really need to support generics for this first iteration? The current storage struct does not allow generics so just to replicate that would be good enough, and we can add that as future enhancement?

> If we believe that GATs would be stabilized, I will use the implementation with GATs and remove the old one.

I don't know whether we can count on that, there still seems to be some contention in the stabilization PR, and it has been "coming soon" for a long time. I personally think it is a really nice feature but I'm wary of depending upon unstable features, there should be a very strong justification.

> What do you think about the idea of a private struct?

Legacy versions of `ink!` heavily depended on private sub modules in codegen but we (more specifically @Robbepop) had a major effort to remove all of them from the codegen, moving to traits + associated types for addressing concrete implementions, in order to use the Rust type system rather than the more brittle name resolution. I'm sure you are very familiar with this technique from working in the `ink! `codegen.

At the very least we should consider if it is possible to use the same technique in this situation. 

I still need to spend more time looking into your WIP here to get my head fully around it, in order to provide some more in depth feedback.

xgreenx:
> Do we really need to support generics for this first iteration? The current storage struct does not allow generics so just to replicate that would be good enough, and we can add that as a future enhancement?

It is not only about the storage of the contract it is about the `#[ink_lang::storage_item]` macro. All types used in the storage should be defined via that macro to automate key calculation. And the problem is that users can normally declare their generic types.

The workaround: implement all traits manually. So it is not critical, and we can do that in follow-up PR. If something is implemented wrongly, the user can still have bugs, but we can handle it by documentation.

> I don't know whether we can count on that, there still seems to be some contention in the stabilization PR, and it has been "coming soon" for a long time. I personally think it is a really nice feature but I'm wary of depending upon unstable features, there should be a very strong justification.

Okay, then let's integrate GATs in follow-up PR  when it will be stabilized=)

ascjones:
> The workaround: implement all traits manually. So it is not critical, and we can do that in follow-up PR. If something is implemented wrongly, the user can still have bugs, but we can handle it by documentation.

I'm okay with doing this for the first iteration. My preference is to get something as simple as possible working from this PR, then iterate upon that. Then it will be easier to review and there will be faster feedback.

xgreenx:
> I'm okay with doing this for the first iteration. My preference is to get something as simple as possible working from this PR, then iterate upon that. Then it will be easier to review and there will be faster feedback.

Yep, I got it=) I've already updated everything to use `StorageType`. All old tests are green now. Now I need to fix the latest comments and add some new tests and the change is ready for review=)

xgreenx:
@athei @ascjones @cmichi @HCastano The change is ready for the final review=)

I tried to describe all the main moments in the description.

crates/metadata/src/lib.rs
```diff
@@ -73,12 +73,14 @@ pub enum MetadataVersioned {
     /// Version 2 of the contract metadata.
     V2(MetadataVersionDeprecated),
     /// Version 3 of the contract metadata.
-    V3(InkProject),
+    V3(MetadataVersionDeprecated),
+    /// Version 4 of the contract metadata.
+    V4(InkProject),
```
cmichi, line 78: That's why the `ink-waterfall` currently fails ‒ `polkadot-js` doesn't yet support this new metadata format, uploading the `.contract` bundles from this PR there is not yet possible, the UI just displays a stack trace and `Uncaught error.` once attempted.

We'll have to create a follow-up issue for `polkadot-js` once this PR is merged.

xgreenx: The first question: Do we need a new version of metadata if only the storage layout section is affected by the change?
And the second: I can temporarily return back version 3 to get the size reduction statistic=) Is it okay?

paritytech-cicd-pr:
 These are the results when building the `examples/*` contracts from this branch with `cargo-contract 1.3.0-cef152e` and comparing them to ink! `master`: 

| `accumulator`            | +0.37 K          |            | 1.40 K               |                |
| `adder`                  | -0.20 K          |            | 1.94 K               |                |
| `contract-introspection` | +0.28 K          |            | 2.65 K               |                |
| `contract-terminate`     | +0.19 K          |            | 1.13 K               | 275_000        |
| `contract-transfer`      | +0.31 K          |            | 8.43 K               | 75_000         |
| `data-structures`        | +0.16 K          |            | 1.90 K               |                |
| `delegate-calls`         | -0.59 K          | -207       | 2.37 K               | 76_069         |
| `delegator`              | -0.38 K          | -480       | 5.99 K               | 231_801        |
| `dns`                    | -0.17 K          |            | 8.72 K               | 225_000        |
| `erc1155`                | -0.58 K          |            | 16.71 K              | 450_000        |
| `erc20`                  | -0.56 K          |            | 7.94 K               | 225_000        |
| `erc721`                 | -0.14 K          |            | 11.68 K              | 600_000        |
| `flipper`                | +0.31 K          |            | 1.62 K               | 75_000         |
| `forward-calls`          | -0.31 K          | -148       | 2.57 K               | 151_279        |
| `incrementer`            | +0.33 K          |            | 1.54 K               |                |
| `multisig`               | -1.50 K          | -1_387     | 23.62 K              | 468_769        |
| `rand-extension`         | -0.30 K          |            | 3.62 K               | 75_000         |
| `seal-code-hash`         | +0.27 K          |            | 1.84 K               |                |
| `set-code-hash`          | +0.33 K          |            | 1.90 K               | 150_000        |
| `subber`                 | -0.20 K          |            | 1.95 K               |                |
| `trait-erc20`            | -0.56 K          |            | 8.21 K               | 225_000        |
| `trait-flipper`          | +0.36 K          |            | 1.37 K               | 75_000         |
| `trait-incrementer`      | +0.37 K          |            | 1.56 K               | 150_000        |
| `updated-incrementer`    | +0.28 K          |            | 9.90 K               |                |
| `upgradeable-flipper`    | -0.03 K          |            | 1.53 K               |                |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/194412) | Last update: Wed May 18 13:47:28 CEST 2022

xgreenx:
You can see the history of size improvement via "edits" of [the comment](https://github.com/paritytech/ink/pull/1217#issuecomment-1129908772).

The initial change improved the size of complex contracts. But it makes it worse for common contracts for 300/400 bytes.

I've started objdumping, investigating, and comparing each WASM file(before the change and after). Previously most of the code was inlined inside of the `push_spread_root`. Each struct implemented `SpreadLayout` and forced the usage of its methods and the compiler decided to inline everything. But now we are using `Decode::decode` and `Encode::encode_to`, but it is from external create, and in most cases, it prefers not to inline it. It caused separate encoding/decoding functions in the bloat. A separate encoding/decoding caused not inlining of the `ext::*` `buffer::*` methods in the `on_chain` engine and again added more separate functions. So I inserted several `#[inline(always)]`(I spent some time playing with different combinations based on the WASM binary) to force the compiler to inline the simple code. You can check the history of the size and see that each optimization improves the size for all examples.

After I also found several other optimizations that help to return back the size. I didn't put them to separate PR because it all is related to `push_spread_root` and new storage.

Also, I found other potential optimizations that can be done later and that are not related to that PR:
- https://github.com/paritytech/ink/pull/1267 - Already done, it is simple.
- https://github.com/paritytech/ink/issues/1263
- https://github.com/paritytech/ink/issues/1262

xgreenx:
The last delta size report seems buggy(I guess because previous `ink-waterfall` failed and erased data regarding size). But you still can see that size is less than before=)

crates/primitives/Cargo.toml
```diff
@@ -19,6 +19,7 @@ ink_prelude = { version = "3.1.0", path = "../prelude/", default-features = fals
 scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive", "full"] }
 scale-info = { version = "2", default-features = false, features = ["derive"], optional = true }
 cfg-if = "1"
+sha2-const = { version = "0.1.2", default-features = false }
```
athei, line 22: Why does it need to be a const implementation? You hopefully never call it at runtime but only from a proc macro (or in tests). So it can just be a normal blake2 runtime implementation here.

crates/env/src/engine/on_chain/impls.rs
```diff
@@ -260,9 +287,10 @@ impl EnvBackend for EnvInstance {
     where
         R: scale::Encode,
     {
-        let mut scope = self.scoped_buffer();
-        let enc_return_value = scope.take_encoded(return_value);
-        ext::return_value(flags, enc_return_value);
+        let mut scope = EncodeScope::from(&mut self.buffer[..]);
+        return_value.encode_to(&mut scope);
+        let len = scope.len();
+        ext::return_value(flags, &self.buffer[..][..len]);
```
athei, line 293: How is this change related to this PR?

athei (commented):
I think we have no example that actually makes use of any feature but `Mapping`. Maybe we can update some of them to make use of `StorageValue` and generic structs (used multiple times).

crates/metadata/src/lib.rs
athei: I think we need a new storage version. It changes the structure of the metadata. For now it works with the old version because noone used the storage layout, yet. We will try to get all of those changes into the new version: https://github.com/paritytech/ink/issues/1265

xgreenx:
> I think we have no example that actually makes use of any feature but `Mapping`. Maybe we can update some of them to make use of `StorageValue` and generic structs (used multiple times).

Of course, we can=) It can be very actual for multisig example. But I think better to do that in follow-up PR.

> I also think that we need a better name for `StorageValue`. First, the prefix is redundant (it is in the storage module). Second, `Value` is a meaningless word in this context. We need something short that expressed that it is stored in its own storage cell or that it is loaded lazily. I think `Lazy` would actually be a good name for it.

Yea, `StorageValue` is a bad name=D But `Lazy` also is not correct. `StorageValue` uses the `get`/`set` ideology to manage data. `Lazy` means that if we use it, we will load it. Maybe `CellMapping` or `StorageCell` or `Cell`?

crates/primitives/Cargo.toml
xgreenx: The main place where we need that:
image

We need to decide how the parent's key should affect the child's key:
- We can hash it, but that means we need to do that during compilation time. I selected that way.
- We can xor it, but it doesn't work in the: Parent -> Child -> Parent. xOr = Parent ^ Child ^ Parent = Child.
- We can use overflow and simply `+` storage key. But I'm not sure is that deterministic.

crates/env/src/engine/on_chain/impls.rs
xgreenx: It is related to the PR because it removed the usage of `scale::Encode::encode_to` in the WASM binary of some examples caused by storage refactoring. It can be moved to a separate PR, of course, and it still will reduce the size.

For me, that change equally can be here, or in a separate PR=) The final decision is yours

crates/env/src/backend.rs
```diff
@@ -183,6 +186,16 @@ pub trait EnvBackend {
...
 
+    fn set_storage_value<V>(&mut self, key: &StorageKey, value: &V) -> Option<u32>
+    where
+        V: scale::Encode;
+
+    fn get_storage_value<R>(&mut self, key: &StorageKey) -> Result<Option<R>>
+    where
+        R: scale::Decode;
+
+    fn clear_storage_value(&mut self, key: &StorageKey);
```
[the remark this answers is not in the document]
xgreenx: I forgot to add comments=(I will fix that.

`Mapping` hashes the `Key` with the storage key and gets 32 bytes old storage key. We can easily convert it to a new storage key with 4 bytes and remove duplicate methods. The problem is that it will increase the chance of collision(we reduced 32 bytes to 4 bytes).

With transparent hashing of course we will have only one bunch of methods. 

I will prepare the code for transparent hashing and remove duplicates.

The signature will be next:
```rust
fn set_contract_storage<V>(&mut self, storage_key: &StorageKey, key: Option<&Key>, value: &V) -> Option<u32>
where
    K: scale::Encode;
    V: scale::Encode;
```

If the `key` is none, we store the `value` under `storage_key`. Otherwise, hash `key` and  `storage_key` like in `Mapping`. With transparent hashing we will remove hashing of keys)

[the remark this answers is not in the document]
xgreenx: Done. Now we don't have `Key` at all.

I moved the hashing logic into `set_contract_storage`, `get_contract_storage` and etc.
But it increased the size of the contracts(`Erc20`, `Erc721`, `Erc1155`) again by ~100 bytes.

But after some investigation, I found that inlining can help again=) Now the size is reduced by ~230 bytes=D

```rust
#[inline(always)]
/// Returns storage key.
fn storage_key<K>(
    &mut self,
    storage_key: &StorageKey,
    dynamic_key: Option<K>,
) -> [u8; 32]
where
    K: scale::Encode,
{
    if let Some(dynamic_key) = dynamic_key {
        self.combined_storage_key(storage_key, dynamic_key)
    } else {
        StorageKeyComposer::old_key(storage_key)
    }
}

#[inline(never)]
/// Combines the storage key with dynamic key knows at runtime.
fn combined_storage_key<K>(
    &mut self,
    storage_key: &StorageKey,
    dynamic_key: K,
) -> [u8; 32]
where
    K: scale::Encode,
{
    let mut hash = <Blake2x256 as HashOutput>::Type::default();
    self.hash_encoded::<Blake2x256, _>(&(dynamic_key, storage_key), &mut hash);
    hash
}
```

HCastano:
Hey, just starting to take a look at this. Awesome work so far! 

For the removal of the `collections` and `lazy` modules, do you mind splitting that out into its own PR?

crates/primitives/Cargo.toml
athei: Ahh okay so it is actually used in runtime code. I see.

crates/env/src/engine/on_chain/impls.rs
athei: Okay it's fine to keep it.

crates/env/src/backend.rs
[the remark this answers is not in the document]
athei: Nice. I look it much better know. The details of the actual key passed to pallet-contracts are now contained to the lower levels.

I think the issues about code size will go away once we stop messing around with `[u8; 32]` which trigger a lot of copy code to pass around. Once transparent hashing is implemented we don't hash but just concat that stuff into a `Vec<u8>` which can be moved by pointer.

athei:
I don't wanna merge this without transparent hashing as it will create just another intermediate version that is incompatible to existing storage before and after. We work on transparent hashing with highest priority.

examples/delegator/lib.rs
```diff
@@ -18,21 +14,8 @@ mod delegator {
...
-        scale::Decode,
-        SpreadLayout,
-        PackedLayout,
-    )]
-    #[cfg_attr(
-        feature = "std",
-        derive(::scale_info::TypeInfo, ::ink_storage::traits::StorageLayout)
-    )]
+    #[ink_lang::storage_item]
+    #[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
cmichi, line 18: Why did you add the `Debug` derives?

xgreenx: It was before=) 
image

crates/lang/codegen/src/generator/storage_item.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    let mut new_field = field.clone();
+    let ty = field.ty.clone();
+    let span = field.span();
+    new_field.ty = parse2(quote_spanned!(span =>
+        <#ty as ::ink_storage::traits::AutoItem<
+            ::ink_storage::traits::ManualKey<#key, #salt>,
+        >>::Type
+    ))
+    .unwrap();
```
cmichi, line 245: Could we get an error output a la 

```
.unwrap_or_else(|error| ::core::panic!("converting into storage field failed: {}", error))
```

here? Maybe with the span?

crates/lang/ir/src/ir/storage_item/config.rs
```diff
@@ -0,0 +1,84 @@
...
+use crate::{
+    ast,
+    error::ExtError as _,
+};
+use syn::spanned::Spanned;
+
+/// The ink! configuration.
+#[derive(Debug, Default, PartialEq, Eq)]
+pub struct StorageItemConfig {
+    /// `true` means automatically implement all storage related traits. It is default value.
```
cmichi, line 24:
```suggestion
    /// `true` means automatically implement all storage related traits. It is the default value.
```

crates/lang/ir/src/ir/storage_item/config.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// `false` disables implementing all storage traits. In some cases,
+    /// it can be helpful to override the default implementation of the trait.
```
cmichi, line 26:
```suggestion
    /// If set to `true`, all storage related traits are implemented automatically,
    /// this is the default value.
    /// If set to `false`, implementing all storage traits is disabled. In some cases
    /// this can be helpful to override the default implementation of the trait.
```

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,145 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
     trait_def::analyze(attr.into(), item.into()).into()
 }
 
+/// Prepares the type to be fully compatible and usable with the storage.
+/// It implements all necessary traits and calculates the storage key for types.
+/// All packed types don't have a storage key, but non-packed types
```
cmichi, line 671:
```suggestion
/// Packed types don't have a storage key, but non-packed types
```

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,145 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// (like `Mapping`, `Lazy` etc.) require calculating the storage key during compilation.
+///
+/// All structs and enums that plan to be a part of the storage better to be marked by this macro.
```
cmichi, line 674:
```suggestion
/// All structs and enums that are intended to be a part of the storage must be annotated with this macro.
```

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,145 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// - Derived manually via `#[derive(...)]`
+/// - Derived automatically via deriving of `scale::Decode` and `scale::Encode`.
+/// - Derived via this macro.
+///
+/// If the type is non-packed better to have automated storage key calculation,
+/// but you can specify it manually.
+///
+/// # Example
+///
+/// # Trait implementation
```
cmichi, line 688:
```suggestion
/// ## Trait implementation
```

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,145 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// The macro should be called before `derive` macros because it can change the type.
+///
+/// If the type is packed then the usage of the macro is optional.
+/// All required traits can be:
+/// - Derived manually via `#[derive(...)]`
+/// - Derived automatically via deriving of `scale::Decode` and `scale::Encode`.
+/// - Derived via this macro.
+///
+/// If the type is non-packed better to have automated storage key calculation,
+/// but you can specify it manually.
```
cmichi, line 684:
```suggestion
/// Consider annotating structs and enums that are intented to be a part of the storage with
/// this macro. If the type is packed then the usage of the macro is optional.
///
/// If the type is non-packed it is best to rely on automatic storage key calculation.
/// The storage key can also be specified manually with the generic `KEY` parameter though.
///
/// The macro should be called before `derive` macros because it can change the type.
///
/// All required traits can be:
/// - Derived manually via `#[derive(...)]`.
/// - Derived automatically via deriving of `scale::Decode` and `scale::Encode`.
/// - Derived via this macro.
```

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,145 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///     KeyHolder,
+///     Item,
+///     Storable,
+/// };
+///
+/// #[derive(scale::Decode, scale::Encode)]
+/// struct Packed {
+///     s1: u128,
+///     s2: Vec<u128>,
+///     // Fails because `Item` implemented only for `Vec` where T: Packed
```
cmichi, line 706:
```suggestion
///     // Fails because `Item` is only implemented for `Vec` where `T: Packed`.
```

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,145 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///
+/// #[derive(scale::Decode, scale::Encode)]
+/// #[cfg_attr(
+///     feature = "std",
+///     derive(scale_info::TypeInfo, ink_storage::traits::StorageLayout)
+/// )]
+/// struct PackedManual {
+///     s1: u32,
+///     s2: Vec<(u128, String)>,
+///     // Fails because `Item` implemented only for `Vec` where T: Packed
```
cmichi, line 718:
```suggestion
///     // Fails because `Item` is only implemented for `Vec` where `T: Packed`.
```

crates/metadata/src/layout/mod.rs
```diff
@@ -54,16 +58,14 @@ pub enum Layout<F: Form = MetaForm> {
     ///
     /// This is the only leaf node within the layout graph.
     /// All layout nodes have this node type as their leafs.
-    ///
-    /// This represents the encoding of a single cell mapped to a single key.
-    Cell(CellLayout<F>),
+    Leaf(CellLayout<F>),
+    /// The root cell defines the storage key for all sub-tree.
```
cmichi, line 62:
```suggestion
    /// The root cell defines the storage key for all sub-trees.
```

crates/metadata/src/layout/validate.rs
```diff
@@ -0,0 +1,298 @@
...
+
+/// It validates that the storage layout:
+/// - Hasn't conflicting storage keys, otherwise returns an error with a description of the conflict
+pub struct ValidateLayout {
+    first_entry: HashMap<Key, String>,
+    name_stack: Vec<String>,
+}
+
+impl ValidateLayout {
+    /// Validates the storage layout
```
cmichi, line 32:
```suggestion
    /// Validates the storage layout.
```

crates/metadata/src/layout/validate.rs
```diff
@@ -0,0 +1,298 @@
...
+    Layout,
+    MetadataError,
+    StructLayout,
+};
+use ink_prelude::collections::HashMap;
+use ink_primitives::Key;
+use scale_info::form::MetaForm;
+
+/// It validates that the storage layout:
+/// - Hasn't conflicting storage keys, otherwise returns an error with a description of the conflict
```
cmichi, line 25:
```suggestion
/// It validates that the storage layout doesn't have conflicting storage keys.
/// Otherwise an error with a description of the conflict is returned.
```

crates/primitives/src/key.rs
```diff
@@ -12,341 +12,69 @@
...
+    /// Return storage key from the string
+    pub const fn from_str(str: &str) -> Key {
+        Self::from_bytes(str.as_bytes())
     }
-}
 
-impl Display for Key {
-    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        self.write_bytes(f)
+    /// Return storage key from the bytes
```
cmichi, line 46:
```suggestion
    /// Returns the storage key from the supplied `bytes`.
```

crates/primitives/src/key.rs
```diff
@@ -12,341 +12,69 @@
...
     }
-}
 
-impl Debug for Key {
-    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        write!(f, "Key(")?;
-        self.write_bytes(f)?;
-        write!(f, ")")?;
-        Ok(())
+    /// Return storage key from the string
```
cmichi, line 41:
```suggestion
    /// Return the storage key from the supplied `str`.
```

crates/storage/codegen/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::HashSet;
+use syn::Data;
+
+/// Provides common methods for `DeriveInput`.
+///
+/// Note: It is only for internal usage in the codegen
```
cmichi, line 20:
```suggestion
/// **Note:** This is only for internal usage in the `codegen` module.
```

crates/lang/codegen/src/generator/storage_item.rs
xgreenx: We can add span, but it requires changing the return type to `Result`.

I prefer not to do that because we can't get an error here=) `ty` already was parsed early, so it is a valid type. `key` and `salt` are autogenerated, so they also are valid.

HCastano:
Alright, so imo this PR is too thorny to review in its current state. It touches too many
different (but related) parts of the code.

If we ever want to get this merged we need to make it easily reviewable. I suggest we set
up of series of stacked PRs whose individual reviews should be quick.

By stacking I mean something like:
```
master
    |- working-base
        |- feature-a [1/2]
            |- feature-b [2/2]
```

Some blog posts about stacked PRs
- https://jg.gg/2018/09/29/stacked-diffs-versus-pull-requests/
- https://timothya.com/blog/git-stack/
- https://blog.logrocket.com/using-stacked-pull-requests-in-github/

You obviously know the PR better than we do, but here's the order of the stack I would
suggest:

- Primitives
- Traits
- Derive Macros for Traits (may need to be split into PRs per macro + tests)
- IR + Generators
- `#[ink::storage_item]` Macro
- Codegen
- `ink_env` API
- Data Structures (Lazy, Mapping)
- Metadata (with associated codegen)
- Examples
- Old code removal (may also be done first, or along the way if it's not too distruptive)

If some of these PRs are 50 lines long, that's fine. If they're like 2500, please find a
way to further split them.

Thanks again for you work on this 🙇

xgreenx:
Moved to https://github.com/paritytech/ink/pull/1331

PR #1223, opened 2022-04-22
Introduce entrance crate

### LINKED ISSUE #825: Towards the ink_api entrance crate
## Description

Introduce an `ink` or `ink_api` crate that acts as a central unified re-export layer on top of all ink! related crates.

## Motivation

Currently, if you want to write an ink! smart contract your contract has to depend on several different ink! related crates such as `ink_env`, `ink_storage`, `ink_primitives`, `ink_prelude` as well as some other non-ink! crates such as `parity_scale_codec` and `scale-info` crates.
On top of that there is the need to properly forward crate features such as `std`, `ink-debug` and `ink-as-dependency` to the respective crates.
This isn't such a big deal usually because `cargo-contract` takes care of it mostly, however, it is yet another hurdle for new 
developers especially.

## Ideal Solution

The ideal is to only have to depend on a single ink related crate that manages all the details for all the contracts that depend on it.
We call this entry crate `ink_api` or sometimes even just `ink`.
This `ink_api` crate won't define anything by itself but merely acts as a layer that provides a unified way to access all the different parts of ink! via re-exports or re-definitions of functions, types, macros etc.

## Problems

So why didn't we implement this already?
The problem starts with procedural macros which are pervasive to the idea of ink!.
There currently exists no sane way to properly deal with build system related information from within a proc. macro.
For our use case it will be important to support both: using ink! crates as it happens nowadays as well as the proposed new solution via the centralized and unifying `ink` or `ink_api` crate layer.

The problem is that within a proc. macro it is hard (or impossible) to properly extract the information how exactly a project has been or will be built and therefore the proc. macro is kind of unable to properly align its generated code with respect to this. So depending on if the user actually uses this new `ink_api` crate or not the proc. macro is required to generate code a little differently and this is not possible without introducing impure hacks to take measures into the proc. macro itself.

## Outlook

The Rust team constantly improves the situation and API surrounding proc. macros. There are ideas to make inspection of build processes and other information possible to inspect from within a proc. macro in the future and we could build upon this when it is ready to be used.

Another possible way forward is to simply implement the hacks that the previous section talks about. We refer to them as "hacks" since they are not a clean solution that covers the entire space but only acts on a limited set of common patterns that users are expected to go for when designing their contract architecture. Besides that this will make ink! proc. macros impure with respect to the fact that they will have the need to inspect the file system during compilation of an ink! smart contract which is nothing you want to happen usually, also with respect to file caching and build artifacts etc.

ascjones:
Closes #825.

:warning: **Breaking Change** :warning: 

This PR:

- Instead of introducing a new top level `ink` crate, simply renames the `ink_lang` crate to `ink`.
- Reexports sub-crates from the `ink` crate.
- Changes all macro generated code to use reexported crates e.g. `::ink::env` instead of `::ink_env`.
  - Note that for now the derives in `ink_storage_derive` will produce code which assumes the `ink` crate is imported e.g. `::ink::storage`, so will not work if used as a standalone crate  (see below)

> ## Problems
>So why didn't we implement this already?
The problem starts with procedural macros which are pervasive to the idea of ink!.
There currently exists no sane way to properly deal with build system related information from within a proc. macro.
For our use case it will be important to support both: using ink! crates as it happens nowadays as well as the proposed new solution via the centralized and unifying ink or ink_api crate layer.

>The problem is that within a proc. macro it is hard (or impossible) to properly extract the information how exactly a project has been or will be built and therefore the proc. macro is kind of unable to properly align its generated code with respect to this. So depending on if the user actually uses this new ink_api crate or not the proc. macro is required to generate code a little differently and this is not possible without introducing impure hacks to take measures into the proc. macro itself.

In this case, specifically this is a problem with the storage derive macros. Where previously the derives, reexported via the `ink_storage` crate, could safely assume the presence of that crate e.g. `::ink_storage::traits::Storable`. Now it generates `::ink::storage::traits::Storable` instead, assuming the usage of the `ink` crate rather than the standalone `ink_storage` crate.

To overcome this in an opinionated way (for now), we will assume that a contract is uses the `ink` entrance crate, instead of individual crates. This means that the `ink_storage` macros will no longer work when using as a standalone crate.

I believe that this can be solved, but I will do that as a follow up so it easier to review and discuss.

~~**NOTE:** `gitlab-examples-contract-build` CI will fail because https://github.com/paritytech/cargo-contract/pull/728 is required for `cargo-contract` to work with the new entrance crate~~ I've temporarily updated the CI to use that PR branch to check whether all the examples still compile.


crates/storage/derive/src/tests/packed_layout.rs
```diff
@@ -22,20 +22,20 @@ fn unit_struct_works() {
         }
         expands to {
             const _: () = {
-                impl ::ink_storage::traits::PackedLayout for UnitStruct {
-                    fn pull_packed(&mut self, __key: &::ink_primitives::Key) {
+                impl ::ink::storage::traits::PackedLayout for UnitStruct {
```
ascjones, line 25: These tests will need to be moved to the `ink` crate if they are to work. Alternatively if these traits/derives are removed entirely by #1217 then it will no longer be an issue,

cmichi:
Could you update the `CHANGELOG.md` as well? Ideally with a note on how to migrate.

athei:
This is maybe shaving the yak but we should also re-export crates that are simply necessary for a contract to import. Otherwise it is just a chore to keep the versions in-sync. Talking about `scale` and `scale-info`.

CHANGELOG.md
```diff
@@ -6,6 +6,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+#### New `ink` crate
+The `ink_lang` crate has been replaced in [#1223](https://github.com/paritytech/ink/pull/1223) by a new top level `ink` 
+crate. All existing sub-crates are reexported and should be used via the new `ink` crate e.g. `ink::env` instead of 
+`::ink_env`. Contract authors should now import the top level `ink` crate instead of the individual crates.
+##### Migration
+- In `Cargo.toml` Replace all individual `ink_*` crate dependencies with the `ink` crate.
+- In the contract source:
+  - Remove the commonly used `use ink_lang as ink` idiom.
+  - Replace all usages of individual crates with reexports, e.g. `ink_env` -> `ink::env`
```
cmichi, line 19:
```suggestion
crate. All existing sub-crates are reexported and should be used via the new `ink` crate, so e.g. `ink::env` instead of 
`::ink_env`. Contract authors should now import the top level `ink` crate instead of the individual crates.

##### Migration
- In `Cargo.toml` Replace all individual `ink_*` crate dependencies with the `ink` crate.
- In the contract source:
  - Remove the commonly used `use ink_lang as ink` idiom.
  - Replace all usages of individual crates with reexports, e.g. `ink_env` ➜ `ink::env`.
```

cmichi (approved):
This is so nice! :rocket:

athei (commented):
I think we should rename`ink_lang_macro` to `ink_macro`.

ascjones:
> This is maybe shaving the yak but we should also re-export crates that are simply necessary for a contract to import. Otherwise it is just a chore to keep the versions in-sync. Talking about `scale` and `scale-info`.

It would be nice and that was also my original intention, however it does introduce a wrinkle: when using these reexported crates for `Encode, Decode, TypeInfo` derives, the user would be required to add the [`#[codec(crate = path::to::crate)]`](https://github.com/paritytech/parity-scale-codec/#derive-attributes) to have those derives use the reexported path.

Of course in our codegen we can apply that attribute ourselves but if the user wants to use them like `#[derive(ink::scale::Encode]` they will need to do that extra step.

So I think it is definitely worth exploring to see if we can make it work, as a follow. I would also like to see just a single crate.

athei:
I didn't think about that. This is a complication. Yeah let's put more thought into that and do it as a follow up.

athei (approved):

PR #1235, opened 2022-04-26
Fix ordering of message ids if the trait is implemented before the inherent section

xgreenx:
Fixed the ordering of ids. Before it was sorted:
inherent_ids
traits_ids

Now the ordering is not changed. It is important for dispatching logic and wildcard. Dispatching takes the global index of the method but if the order is changed, it will use the wrong id. In the case of a wildcard, it breaks it at all if the trait was implemented before the inherent section.

### DIFF
--- a/crates/lang/tests/ui/contract/pass/message-selector.rs
+++ b/crates/lang/tests/ui/contract/pass/message-selector.rs
@@ -6,20 +6,45 @@ mod contract {
     #[ink(storage)]
     pub struct Contract {}
 
+    #[ink_lang::trait_definition]
+    pub trait Messages {
+        #[ink(message)]
+        fn message_0(&self);
+
+        #[ink(message, selector = 1)]
+        fn message_1(&self);
+    }
+
+    impl Messages for Contract {
+        #[ink(message)]
+        fn message_0(&self) {}
+
+        #[ink(message, selector = 1)]
+        fn message_1(&self) {}
+    }
+
     impl Contract {
         #[ink(constructor)]
         pub fn constructor() -> Self {
             Self {}
         }
 
-        #[ink(message)]
-        pub fn message_0(&self) {}
-
-        #[ink(message, selector = 1)]
-        pub fn message_1(&self) {}
-
         #[ink(message, selector = 0xC0DE_CAFE)]
         pub fn message_2(&self) {}
+
+        #[ink(message, selector = _)]
+        pub fn message_3(&self) {}
+    }
+
+    #[ink_lang::trait_definition]
+    pub trait Messages2 {
+        #[ink(message, selector = 0x12345678)]
+        fn message_4(&self);
+    }
+
+    impl Messages2 for Contract {
+        #[ink(message, selector = 0x12345678)]
+        fn message_4(&self) {}
     }
 }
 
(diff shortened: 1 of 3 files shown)


ascjones (commented):
I see you have the example, but could you also add a `ui` test for this case too, e.g. https://github.com/paritytech/ink/blob/e8d4739649293a458c3958bec802d2d750067d98/crates/lang/tests/ui/contract/pass/message-selector.rs

ascjones (approved):
LGTM

ascjones:
At a glance looks like it is not configured to pull from forks. Maybe we need to fix that. Anyway we can go ahead and merge if we get another approval since the waterfall is not a requirement for merge. @cmichi

crates/lang/tests/ui/contract/pass/message-selector.rs
```diff
@@ -34,7 +59,7 @@ fn main() {
                 >>::IDS[0]
             },
         >>::SELECTOR,
-        [0x5A, 0x6A, 0xC1, 0x5D],
+        [0xFB, 0xAB, 0x03, 0xCE],
```
cmichi, line 62: So this is a breaking change. Will require a minor release.

examples/upgradeable-contracts/forward-calls/lib.rs
```diff
@@ -33,6 +33,30 @@ pub mod proxy {
         admin: AccountId,
     }
 
+    /// The trait provides information about `Proxy`
+    #[ink_lang::trait_definition]
+    pub trait Forwarder {
+        /// Returns `AccountId` of contract that calls are forward to
```
cmichi, line 39:
```suggestion
        /// Returns the `AccountId` of the contract that calls are forward to.
```

examples/upgradeable-contracts/forward-calls/lib.rs
```diff
@@ -33,6 +33,30 @@ pub mod proxy {
...
+        #[ink(message)]
+        fn forward_to(&self) -> AccountId;
+
+        /// Returns `AccountId` of administrator of current contract
```
cmichi, line 43:
```suggestion
        /// Returns the `AccountId` of the admin of the current contract.
```

examples/upgradeable-contracts/forward-calls/lib.rs
```diff
@@ -33,6 +33,30 @@ pub mod proxy {
         admin: AccountId,
     }
 
+    /// The trait provides information about `Proxy`
```
cmichi, line 36:
```suggestion
    /// The trait provides information about the forward pattern used in this contract.
```

cmichi, line 36: What is the reason why you added this trait btw.?

xgreenx: I added it to test the behavior of the wildcard in case of the trait is implemented before the inherent section.

But it was before I added the UI test=) So now it is not so actual but we can keep it to verify that the behavior is right during the e2e test.

examples/upgradeable-contracts/forward-calls/lib.rs
```diff
@@ -33,6 +33,30 @@ pub mod proxy {
...
+        fn forward_to(&self) -> AccountId {
+            self.forward_to
+        }
+
+        #[ink(message)]
+        fn admin(&self) -> AccountId {
+            self.admin
+        }
+    }
+
```
cmichi, line 59:
```suggestion
```

I would remove it, as it doesn't serve a function to the example. I'd like to keep the examples as minimal as possible to focus on illustrating the point (in this case how to build an upgradeable contract).

PR looks good besides that!

cmichi (approved):
Thank you!

PR #1244, opened 2022-05-03
PSP22 chain extension example

stiiifff:
Example implementation of the PSP22 Fungible Token Standard as a chain extension, supporting a multi-token system provided by the FRAME assets pallet.
It demonstrates how ink! contracts (L2) can interact with native assets (L1) from the chain runtime in a standardised way.


xgreenx:
It is bad that we did double work=( We could cooperate together to deliver but it is too late.
We did the [same in the OpenBrush](https://github.com/Supercolony-net/openbrush-contracts/pull/104).

But it is still good to have several implementations=)

We plan to create a production-ready chains extension and define its interface in the PSP. So if you have some though regarding our implementation it will be cool to hear them here or in PSP=)

I want to highlight that maybe some contract wants to transfer his tokens, so it is why we added the `Origin` enum into the chain extension where the developer can specify it(`Origin::Caller` or `Origin::Self`).

But we also realized the problem with `Origin::Caller`. Allowing anyone to transfer/approve tokens on behalf of the caller without any restriction creates a vulnerability. Any contract can be malicious and can steal all tokens. I think you need to highlight that by commenting on the implementation of `transfer` in the chain extension.

paritytech-cicd-pr:
 These are the results when building the `examples/*` contracts from this branch with `cargo-contract 1.2.0-76a6df4` and comparing them to ink! `master`: 

| `accumulator`            |                  |            | 1.04 K               |                |
| `adder`                  |                  |            | 2.14 K               |                |
| `contract-introspection` |                  |            | 2.37 K               |                |
| `contract-terminate`     |                  |            | 0.94 K               | 213_894        |
| `contract-transfer`      |                  |            | 8.13 K               | 13_894         |
| `delegate-calls`         |                  |            | 2.96 K               | 14_416         |
| `delegator`              |                  |            | 6.38 K               | 44_666         |
| `dns`                    |                  |            | 8.85 K               | 41_682         |
| `erc1155`                |                  |            | 17.35 K              | 83_364         |
| `erc20`                  |                  |            | 8.50 K               | 41_682         |
| `erc721`                 |                  |            | 11.81 K              | 111_152        |
| `flipper`                |                  |            | 1.32 K               | 13_894         |
| `forward-calls`          |                  |            | 2.89 K               | 28_475         |
| `incrementer`            |                  |            | 1.22 K               |                |
| `multisig`               |                  |            | 25.34 K              | 90_233         |
| `psp22-extension`        | +7.54 K          |            | 7.54 K               |                |
| `rand-extension`         |                  |            | 3.93 K               | 13_894         |
| `seal-code-hash`         |                  |            | 1.59 K               |                |
| `set-code-hash`          |                  |            | 1.58 K               | 27_788         |
| `subber`                 |                  |            | 2.16 K               |                |
| `trait-erc20`            |                  |            | 8.77 K               | 41_682         |
| `trait-flipper`          |                  |            | 1.01 K               | 13_894         |
| `trait-incrementer`      |                  |            | 1.19 K               | 27_788         |
| `updated-incrementer`    |                  |            | 9.70 K               |                |
| `upgradeable-flipper`    |                  |            | 1.56 K               |                |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/191641) | Last update: Tue May  3 14:34:56 CEST 2022

stiiifff:
@xgreenx Thanks for your feedback ! I was unaware of your implementation. This is just a simple, incomplete example. My motivation was to show a more advanced chain extension example than the existing `FetchRandomExtension` one, and I thought that PSP22 was a good use case. So I don't see it as double work.
Good point re. the caller !

Maar-io:
hi @stiiifff 
Do you still plan to add this as example? 
I'd like to use it for one of my workshops

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,289 @@
...
+						)
+						.encode(),
+					_ => unreachable!(),
+				};
+				trace!(
+					target: "runtime",
+					"[ChainExtension] PSP22Metadata::{:?}",
+					func_id
+				);
+				env.write(&result, false, None).map_err(|_| {
```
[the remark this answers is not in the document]
athei: `None` is correct because the `result` is of fixed size. This means the weight for writing it is contained in the overall weight for the whole function.

Wiezzel:
@HCastano I will resolve the comments.

examples/psp22-extension/README.md
```diff
@@ -0,0 +1,39 @@
...
+* Defining and calling the extension in ink!.
+* Defining the extension in Substrate.
+
+## Chain-side Integration
+
+To integrate this example into Substrate you need to do two things:
+
+* Use the code in [`psp22-extension-example.rs`](runtime/psp22-extension-example.rs)
+  as an implementation for the trait `ChainExtension` in Substrate.
+  You can just copy/paste the content of that file into e.g. your `runtime/src/lib.rs`.
```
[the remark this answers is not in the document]
Wiezzel: @HCastano So the text here should advise to create a chain extensions module and copy the code there instead of `lib.rs`?

BTW, I don't know much about chain extensions, but is there a good reason for this code to be copied rather than imported?

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,289 @@
...
+				// env.adjust_weight(charged, actual_weight)
+			},
+
+			// PSP22::decrease_allowance
+			0xfecb57d5 => {
+				trace!(
+					target: "runtime",
+					"[ChainExtension]|call|decrease_allowance"
+				);
+				return Err(DispatchError::Other("Unimplemented func_id"))
```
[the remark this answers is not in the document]
Wiezzel: Looks like a TODO for me. ;)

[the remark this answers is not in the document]
Wiezzel: After second check, turns out it currently cannot be implemented due to limitations of the `assets` pallet. There's no method to decrease allowance available.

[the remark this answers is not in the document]
xgreenx: You can use a combination of `cancel_approval` and `approve_transfer` together like we did it [here](https://github.com/Supercolony-net/openbrush-contracts/blob/f8b7762ee1faacd1d8d11c850148c9981db9c5f0/contracts/src/token/psp22_pallet/psp22_pallet.rs#L141)=)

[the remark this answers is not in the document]
Wiezzel: @xgreenx Oh, yeah, I could try that.

examples/psp22-extension/lib.rs
```diff
@@ -0,0 +1,267 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::new_without_default)]
```
HCastano, line 2: Can you address this lint?

examples/psp22-extension/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+pub enum Psp22ErrorCode {
+    TotalSupplyFailed,
+}
+
+#[derive(scale::Encode, scale::Decode)]
+#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+pub enum Psp22Error {
+    ErrorCode(Psp22ErrorCode),
+}
```
HCastano, line 90: Can we fold these into a single error enum?

examples/psp22-extension/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+    type Balance = DefaultBalance;
+    type Hash = <ink_env::DefaultEnvironment as Environment>::Hash;
+    type Timestamp = <ink_env::DefaultEnvironment as Environment>::Timestamp;
+    type BlockNumber = <ink_env::DefaultEnvironment as Environment>::BlockNumber;
+
+    type ChainExtension = crate::Psp22Extension;
+}
+
+#[ink::contract(env = crate::CustomEnvironment)]
+mod psp22_ext_test {
```
HCastano, line 135:
```suggestion
mod psp22_ext {
```

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,398 @@
...
+}
+
+#[derive(Debug, PartialEq, Encode, Decode, MaxEncodedLen)]
+struct Psp22ApproveInput<AssetId, AccountId, Balance> {
+    pub asset_id: AssetId,
+    pub spender: AccountId,
+    pub value: Balance,
+}
+
+pub struct Psp22Extension;
```
HCastano, line 80: This won't compile due to some changes with how chain extensions work. You can read more about the breaking changes here: https://github.com/paritytech/substrate/pull/11874.

There's also these two PRs which you might find useful/interesting
- https://github.com/paritytech/substrate/pull/11816
- https://github.com/paritytech/substrate/pull/12032

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,398 @@
...
+        env: Environment<E, InitState>,
+    ) -> Result<RetVal, DispatchError>
+    where
+        E: Ext<T = T>,
+        <E::T as SysConfig>::AccountId:
+            UncheckedFrom<<E::T as SysConfig>::Hash> + AsRef<[u8]>,
+    {
+        match func_id {
+            // Note: We use the PSP22 interface selectors as function IDs,
+            // there is no need but it makes sense from a convention perspective.
```
HCastano, line 371:
```suggestion
            // While we can use anything here, it makes sense from a convention perspective.
```

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,398 @@
...
+        StaticLookup,
+        Zero,
+    },
+    DispatchError,
+};
+
+#[derive(Debug, PartialEq, Encode, Decode, MaxEncodedLen)]
+struct Psp22BalanceOfInput<AssetId, AccountId> {
+    pub asset_id: AssetId,
+    pub owner: AccountId,
```
HCastano, line 48: Don't think these fields need to be `pub` (same for all the other structs)

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,398 @@
...
+                .encode()
+        }
+        // PSP22Metadata::token_decimals
+        0x7271b782 => {
+            <pallet_assets::Pallet<T> as InspectMetadata<T::AccountId>>::decimals(
+                &asset_id,
+            )
+            .encode()
+        }
+        _ => unreachable!(),
```
HCastano, line 122: This isn't actually unreachable, someone can call `metadata` without doing any checks on the incoming `func_id`. I'd change this to an error instead

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,398 @@
...
+    env: Environment<E, InitState>,
+) -> Result<(), DispatchError>
+where
+    T: pallet_assets::Config + pallet_contracts::Config,
+    <T as SysConfig>::AccountId: UncheckedFrom<<T as SysConfig>::Hash> + AsRef<[u8]>,
+    E: Ext<T = T>,
+{
+    let mut env = env.buf_in_buf_out();
+    let asset_id = env.read_as()?;
+    let result = match func_id {
```
HCastano, line 104: We can `encode()` the `result` once instead of at every match arm

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,398 @@
...
+#[derive(Debug, PartialEq, Encode, Decode, MaxEncodedLen)]
+struct Psp22ApproveInput<AssetId, AccountId, Balance> {
+    pub asset_id: AssetId,
+    pub spender: AccountId,
+    pub value: Balance,
+}
+
+pub struct Psp22Extension;
+
+fn map_err(err_msg: &'static str) -> impl FnOnce(DispatchError) -> DispatchError {
```
HCastano, line 82: I'd personally inline this, but if you do want to keep it I'd change the name so we don't end up with slightly odd `map_err(map_err("foo"))` statements

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,398 @@
...
+        sender,
+        &input.to,
+        input.value,
+        true,
+    )
+    .map_err(map_err("ChainExtension failed to call transfer"))?;
+    trace!(
+        target: "runtime",
+        "[ChainExtension]|call|transfer"
+    );
```
HCastano, line 212:
```suggestion
    );
    
```

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,398 @@
...
+        owner,
+        &input.spender,
+        allowance,
+    )
+    .map_err(map_err("ChainExtension failed to call decrease_allowance"))?;
+
+    trace!(
+        target: "runtime",
+        "[ChainExtension]|call|decrease_allowance"
+    );
```
HCastano, line 351:
```suggestion
    );
    
```

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,398 @@
...
+            0x162df8c2 | 0x6568382f | 0x4d47d921 => query::<T, E>(func_id, env)?,
+
+            // P2P22:transfer
+            0xdb20f9f5 => transfer::<T, E>(env)?,
+
+            // P2P22:transfer_from
+            0x54b3c76e => transfer_from::<T, E>(env)?,
+
+            // PSP22::approve + PSP22::increase_allowance
+            0xb20f1bbd | 0x96d6b57a => approve::<T, E>(env)?,
```
HCastano, line 386: How come you use the same function for both?

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,289 @@
...
+				})?;
+			},
+
+			// P2P22:transfer
+			0xdb20f9f5 => {
+				let mut env = env.buf_in_buf_out();
+
+				let base_weight = <T as pallet_assets::Config>::WeightInfo::transfer();
+				let overhead =
+					<T as pallet_contracts::Config>::Schedule::get().host_fn_weights.debug_message;
```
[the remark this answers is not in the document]
HCastano: Why's this a good approximation?

examples/psp22-extension/README.md
```diff
@@ -0,0 +1,50 @@
...
+See the example contract in [`lib.rs`](lib.rs).
+
+## Disclaimer
+
+:warning: This is not a feature-complete or production-ready PSP22 implementation. This
+example currently lacks proper error management, precise weight accounting, tests (these
+all  might be added at a later point).
+
+:warning: `decrease_allowance` function is currently not implemented due to limitations
+of the `assets` pallet.
```
HCastano, line 50:
```suggestion
```

examples/psp22-extension/lib.rs
```diff
@@ -0,0 +1,281 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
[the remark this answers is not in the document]
HCastano: There is still missing documentation (e.g the `Psp22Extension` struct)

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,398 @@
...
+
+            // PSP22::decrease_allowance
+            0xfecb57d5 => decrease_allowance(env)?,
+
+            _ => {
+                error!("Called an unregistered `func_id`: {:}", func_id);
+                return Err(DispatchError::Other("Unimplemented func_id"))
+            }
+        }
+        Ok(RetVal::Converging(0))
```
HCastano, line 396:
```suggestion

        Ok(RetVal::Converging(0))
```

HCastano (commented):
Thanks for updating this @Wiezzel! 

Can you fork the `substrate-contracts-node`, add this chain extension to it, and share the repo? I want to see if this code compiles. We'll need to think of a way to check with our CI in the future

examples/psp22-extension/README.md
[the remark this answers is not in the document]
HCastano: I think the original intent with the copy/paste bit was that since this was an example, you could copy it and use that as a base to work off of

examples/psp22-extension/lib.rs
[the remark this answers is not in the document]
Wiezzel: Roger that.

examples/psp22-extension/runtime/psp22-extension-example.rs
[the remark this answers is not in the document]
Wiezzel: I don't know, but @athei said so. `¯\_(ツ)_/¯`

examples/psp22-extension/lib.rs
Wiezzel: I though it cannot be done, because of `no_std`. I see that the same `allow` annotation is added in four other examples in this repo.

examples/psp22-extension/runtime/psp22-extension-example.rs
Wiezzel: Well, it's a private function of this module and it's called only once, and the call is made with a check. So I'm not sure what you have in mind when saying "someone can call `metadata` without doing any checks".

examples/psp22-extension/runtime/psp22-extension-example.rs
Wiezzel: Would `convert_err` sound ok to you?

examples/psp22-extension/runtime/psp22-extension-example.rs
Wiezzel: Good question. I'll check this with @stiiifff.

stiiifff: Because of [this](https://github.com/paritytech/substrate/blob/ca1f81cd0864ae0614aaff07fa121d2d42110586/frame/assets/src/functions.rs#L722-L726). Problem is the doc on the trait function `approve` is not clear on the behavior of subsequent calls. So it's a bit of a shortcut in the context of this sample, I think it's fine .. what you think ?

Wiezzel:
@HCastano I've made some changes according to your comments.
Here's the `substarte-contratcs-node` code with the extension installed: https://github.com/Wiezzel/substrate-contracts-node/pull/1
CI tests are failing, but it seems to me that is *not* related to any changes from this PR. I've run `cargo test -p ink_lang --test compile_tests` on master and it fails as well.

examples/psp22-extension/runtime/psp22-extension-example.rs
HCastano: This still hasn't been addressed

examples/psp22-extension/runtime/psp22-extension-example.rs
HCastano: Yeah that's fine

examples/psp22-extension/runtime/psp22-extension-example.rs
HCastano: Can you address this for the other functions too?

examples/psp22-extension/runtime/psp22-extension-example.rs
HCastano: Yeah I guess that's fine. Can you add some docs around that in source code?

HCastano (commented):
> Here's the substarte-contratcs-node code with the extension installed: https://github.com/Wiezzel/substrate-contracts-node/pull/1

Thanks for that!

> CI tests are failing, but it seems to me that is not related to any changes from this PR. I've run cargo test -p ink_lang --test compile_tests on master and it fails as well.

Can you merge `master` and see if that fixes your issues?

examples/psp22-extension/runtime/psp22-extension-example.rs
Wiezzel: Weird, I remember doing this. Maybe a commit got lost somehow. 😕

examples/psp22-extension/runtime/psp22-extension-example.rs
Wiezzel: I can do this in the `query` function, but here it won't work because of incompatible match arms (`name` and `symbol` return `Vec<u8>` while `decimals` returns `u8`).

examples/psp22-extension/runtime/psp22-extension-example.rs
Wiezzel: OK, I've thought this through and I believe we can still implement the correct behaviour by calling `cancel_approval`, as we do in `decrease_allowance` implementation. I'd like to close this PR soon, but I'll leave a 'FIXME' comment to mark thus.

examples/psp22-extension/runtime/psp22-extension-example.rs
Wiezzel: OK, I've come up with a cleaner, enum-based solution to avoid this cases completely. LMKWYT

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,414 @@
...
+    Metadata(Metadata),
+    Query(Query),
+    Transfer,
+    TransferFrom,
+    Approve,
+    IncreaseAllowance,
+    DecreaseAllowance,
+}
+
+impl TryFrom<u16> for FuncId {
```
HCastano, line 79: Yeah I like this a lot better 👍

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,414 @@
...
+            0xdb20 => Ok(Self::Transfer),
+            0x54b3 => Ok(Self::TransferFrom),
+            0xb20f => Ok(Self::Approve),
+            0x96d6 => Ok(Self::IncreaseAllowance),
+            0xfecb => Ok(Self::DecreaseAllowance),
+            _ => {
+                error!("Called an unregistered `func_id`: {:}", func_id);
+                return Err(DispatchError::Other("Unimplemented func_id"));
+            }
+        }
```
HCastano, line 101:
```suggestion
        let id = match func_id {
            // Note: We use the first two bytes of PSP22 interface selectors as function IDs,
            // While we can use anything here, it makes sense from a convention perspective.
            0x3d26 => Self::Metadata(Metadata::Name),
            0x3420 => Self::Metadata(Metadata::Symbol),
            0x7271 => Self::Metadata(Metadata::Decimals),
            0x162d => Self::Query(Query::TotalSupply),
            0x6568 => Self::Query(Query::BalanceOf),
            0x4d47 => Self::Query(Query::Allowance),
            0xdb20 => Self::Transfer,
            0x54b3 => Self::TransferFrom,
            0xb20f => Self::Approve,
            0x96d6 => Self::IncreaseAllowance,
            0xfecb => Self::DecreaseAllowance,
            _ => {
                error!("Called an unregistered `func_id`: {:}", func_id);
                return Err(DispatchError::Other("Unimplemented func_id"));
            }
        };

        Ok(id)
```

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,414 @@
...
+    Symbol,
+    Decimals,
+}
+
+#[derive(Debug)]
+enum Query {
+    TotalSupply,
+    BalanceOf,
+    Allowance,
+}
```
HCastano, line 117: Do you mind moving these above the `TryFrom` conversion and below the `FuncId` enum?

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,414 @@
...
+        trace!(
+            target: "runtime",
+            "PSP22 Transfer failed:{:?}",
+            err
+        );
+        DispatchError::Other(err_msg)
+    }
+}
+
+enum FuncId {
```
HCastano, line 69: Since this is an example I'd add some docs to explain why this is a better approach than just using the raw `func_id`s

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,414 @@
...
+            UncheckedFrom<<E::T as SysConfig>::Hash> + AsRef<[u8]>,
+    {
+        let func_id = FuncId::try_from(env.func_id())?;
+        match func_id {
+            FuncId::Metadata(func_id) => metadata::<T, E>(func_id, env)?,
+            FuncId::Query(func_id) => query::<T, E>(func_id, env)?,
+            FuncId::Transfer => transfer::<T, E>(env)?,
+            FuncId::TransferFrom => transfer_from::<T, E>(env)?,
+            // FIXME: This is a bit of a shortcut. It was made because the documentation
+            //        for Mutate::approve does not specify the result of subsequent calls.
```
HCastano, line 407: Please don't leave `FIXME`s (or `TODO`s) in the code. File an issue and we can address it in a follow up PR

.gitlab-ci.yml
```diff
@@ -128,6 +128,8 @@ examples-fmt:
         cargo +nightly fmt --verbose --manifest-path ./examples/upgradeable-contracts/${contract}/Cargo.toml -- --check;
       done
     - cargo +nightly fmt --verbose --manifest-path ./examples/upgradeable-contracts/set-code-hash/updated-incrementer/Cargo.toml -- --check
+    # This file is not a part of the cargo project, so it wouldn't be formatted the usual way
+    - rustfmt --verbose --check ./examples/psp22-extension/runtime/psp22-extension-example.rs
```
HCastano, line 132:
```suggestion
    - rustfmt +nightly --verbose --check ./examples/psp22-extension/runtime/psp22-extension-example.rs
```

HCastano (commented):
Couple of small things left but we're almost here 💪

examples/psp22-extension/runtime/psp22-extension-example.rs
Wiezzel: OK, I'll open an issue as soon as this PR is merged.

examples/psp22-extension/runtime/psp22-extension-example.rs
```diff
@@ -0,0 +1,446 @@
...
+            err
+        );
+        DispatchError::Other(err_msg)
+    }
+}
+
+// We're using enums for function IDs because contrary to raw u16 it enables
+// exhaustive matching, which results in cleaner code.
+
+enum FuncId {
```
HCastano, line 97:
```suggestion
/// We're using enums for function IDs because contrary to raw u16 it enables
/// exhaustive matching, which results in cleaner code.
enum FuncId {
```

HCastano (approved):
Thanks for the work on this!

Issue #1181, opened 2022-03-15
Create the mother of all contracts


athei:
It would be massively helpful for Ul development to create an example contract that uses as much functionality as possible. I mean that from an input/output standpoint:

* Use a lot of different complicated nested input and ouput types
* Use debug printing
* Use revert
* Use trap

This list is not exhaustive. The contract doesn't need to make sense. Just something that exercises the limits of the UI.

cc @statictype 

Would be a good issue for someone new.

cmichi:
Once we have it, we should test if the `call`, `initiate`, etc. sub-commands of `cargo-contract` behave correctly with this contract.

PR #1245, opened 2022-05-03
Create the Mother of All Contracts

agryaznov:
This PR should solve #1181. 

~~I decided to give it a narrower name, as *Mother of All Contracts* implies something really exhaustive, while this contract just solves its particular goal: to be a helpful example for UI developers demonstrating all possible inputs/outputs of an ink! contract.~~


examples/bioc/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+edition = "2021"
+publish = false
+
+[dependencies]
+ink_primitives = { version = "3.0", default-features = false }
+ink_metadata = { version = "3.0", default-features = false, features = ["derive"], optional = true }
+ink_env = { version = "3.0", default-features = false }
+ink_storage = { version = "3.0", default-features = false }
+ink_lang = { version = "3.0", default-features = false }
+ink_prelude = { version = "3.0", default-features = false }
```
athei, line 14: Can you please check the manifest of the other examples. You should use a `path` dependency here. This is probably the reason why the waterfall fails.

examples/bioc/lib.rs
```diff
@@ -0,0 +1,157 @@
+//! # Basic Input Output Contract (BIOC)
+//!
+//! This is a Basic Input Output Contract (BIOC), which is intented to
```
athei, line 3: I think we need a better name. It is definitely about input output. But the point is that it is **not** basic. I think the mother of all contracts is still a good name. While it doesn't do any **useful** stuff it should still use all of ink!s features that are somehow observable off-chain. For example, storage and events are missing from this contract.

examples/bioc/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+//! demonstrate rich i\o posibilties of ink! contracts, namely:
```
athei, line 4:
```suggestion
//! demonstrate rich I/O possibilities of ink! contracts, namely:
```

examples/bioc/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+mod bioc {
+    use ink_prelude::{
+        string::String,
+        vec::Vec,
+    };
+
+    /// No storage is needed for this contract.
+    #[ink(storage)]
+    #[derive(Default)]
+    pub struct Bioc {}
```
athei, line 27: We should have storage, too. UIs should add a feature to inspect storage, eventually. So this is also relevant to them.

examples/bioc/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+                }
+                None => Ok(()),
+            }
+        }
+
+        /// Prints the specified string into node's debug log.
+        #[ink(message)]
+        pub fn debug_log(&mut self, _str: String) {
+            ink_env::debug_println!("debug_log: {}", _str);
+        }
```
athei, line 112: The message is used. Don't prefix it with `_`. The clippy error you get was because you used a keyword `str` as an identifier.

```suggestion
        pub fn debug_log(&mut self, message: String) {
            ink_env::debug_println!("debug_log: {}", message);
        }
```

examples/bioc/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+        /// Candle auction can have no winner.
+        /// If auction is finalized, that means that the winner is determined.
+        finalized: bool,
+    }
+
+    /// Way to fail a contract execution.
+    #[derive(scale::Encode, scale::Decode, Debug, PartialEq)]
+    #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+    pub enum Failure {
+        Revert,
```
athei, line 80:
```suggestion
        Revert(String),
```

examples/bioc/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+        #[ink(message)]
+        pub fn echo_auction(&mut self, auction: Auction) -> Auction {
+            auction
+        }
+
+        /// Fails contract execution in the required way.
+        #[ink(message)]
+        pub fn revert_or_trap(&mut self, fail: Option<Failure>) -> Result<(), Failure> {
+            match fail {
+                Some(Failure::Revert) => Err(Failure::Revert),
```
athei, line 100: Let's make it equivalent with the panic.

```suggestion
                Some(Failure::Revert) => Err(Failure::Revert("Reverting on user demand!".to_string()),
```

athei (changes requested):
I still thing we should call it mother of all contracts. It should use all ink! features so that UI devs can make sure that they support them properly. Of course some features are not applicable as they are not observable for UIs. But everything that somehow interacts with them should be included in this contract.

statictype:
Could not build the contract to test it out but i can just enumerate the types of inputs we needed to handle in the UI so far. 
As for the outputs, it would be sufficient if the contract messages just return the input values.

Inputs/Outputs
1. Number: u8, i8, i32, u32, i64, u64 - these inputs validate differently in the UI so it would be useful to test them separately. could be done in unit testing if you think it's too much for the Mother
2. Boolean
3. Balance
4. AccountID
5. Vector
    - Nested Vec - `Vec<Vec<Vec<AccountID>>> `
    - No generic param Vec - not clear what the default input should be here? found this kind of input but don't remember where
    - Vec<u8> special case, polkadot.js doesn't display a normal Vector input with `+` and `-` buttons.  should accept either a hex string or a file upload. - see SBP contract `ProjectName`
6. Struct - this input basically generates an inner form - would be useful if the struct had all other types as fields
7. Enum - a nested enum
8. Option
9. Hash 

Messages 
- should have getters with and without params
- should have mutating messages with and without params
- should have multiple constructors with and without params

Events 
- the contract should contain custom events (user defined)

Errors (not sure how to test these, maybe just create some messages that will always fail?)
- dispatchable failed
- contract reverted
- debug messages

examples/mother/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+edition = "2021"
+publish = false
+
+[dependencies]
+ink_primitives = { version = "3.0.1", path = "../../crates/primitives", default-features = false }
+ink_metadata = { version = "3.0.1", path = "../../crates/metadata", default-features = false, features = ["derive"], optional = true }
+ink_env = { version = "3.0.1", path = "../../crates/env", default-features = false }
+ink_storage = { version = "3.0.1", path = "../../crates/storage", default-features = false }
+ink_lang = { version = "3.0.1", path = "../../crates/lang", default-features = false }
+ink_prelude = { version = "3.0", path = "../../crates/prelude", default-features = false }
```
athei, line 15: OR just remove the versions all together. There is no benefit of having them.

```suggestion
ink_prelude = { version = "3.0.1", path = "../../crates/prelude", default-features = false }
```

examples/mother/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+
+scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive"] }
+scale-info = { version = "2", default-features = false, features = ["derive"], optional = true }
+
+[lib]
+name = "bioc"
```
athei, line 21:
```suggestion
name = "mother"
```

examples/mother/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+                c.auction = auction;
+            })
+        }
+
+        #[ink(constructor)]
+        pub fn default() -> Self {
+            initialize_contract(|c: &mut Self| {
+                c.balances = <Mapping<AccountId, Balance>>::default();
+                c.auction = Auction::default();
+            })
```
athei, line 156:
```suggestion
            Self::new(Auction::default())
```

examples/mother/lib.rs
```diff
@@ -0,0 +1,219 @@
+//! # Mother of All Contracts
+//!
+//! This contracts is intented to
+//! demonstrate rich I/O posibilities of ink! contracts, namely:
```
athei, line 4: Is no longer only about I/O but every functionality that is important for off chain interaction.

examples/mother/lib.rs
agryaznov: Why? Shouldn't we initialize the Mapping storage field, as our documentation [states](https://paritytech.github.io/ink-docs/datastructures/mapping)?

agryaznov:
@statictype thanks for your review!
The contract builds now, please feel free to play with it. 

Regarding the input types kindly provided in your list. It seems that most of them are already there, with only these ones left to be added to the contract:

> - [ ] Nested Vec - Vec<Vec<Vec<AccountID>>>   

Currently we actually have `Vec<Option<(AccountId, Balance)>>`. Please, let me know if we need nested vectors still?

> - [ ] `Vec<u8>`
> - [ ] nested `Enum`

We have `Enum` nested into `Struct`. Would it be enough?
> - [ ] Events

Regarding Number types, I would left only `BlockNumber` (`u32`) for the contract if you really can cover others by unit tests.

examples/mother/lib.rs
agryaznov: right, gotcha

paritytech-cicd-pr:
 These are the results when building the `examples/*` contracts from this branch with `cargo-contract 1.3.0-824c2f6` and comparing them to ink! `master`: 

| `accumulator`            |                  |            | 1.04 K               |                |
| `adder`                  |                  |            | 2.13 K               |                |
| `contract-introspection` |                  |            | 2.37 K               |                |
| `contract-terminate`     |                  |            | 0.94 K               | 275_000        |
| `contract-transfer`      |                  |            | 8.13 K               | 75_000         |
| `data-structures`        |                  |            | 1.73 K               |                |
| `delegate-calls`         |                  |            | 2.96 K               | 76_276         |
| `delegator`              |                  |            | 6.38 K               | 232_281        |
| `dns`                    |                  |            | 8.84 K               | 225_000        |
| `erc1155`                |                  |            | 17.34 K              | 450_000        |
| `erc20`                  |                  |            | 8.49 K               | 225_000        |
| `erc721`                 |                  |            | 11.81 K              | 600_000        |
| `flipper`                |                  |            | 1.31 K               | 75_000         |
| `forward-calls`          |                  |            | 2.88 K               | 151_427        |
| `incrementer`            |                  |            | 1.21 K               |                |
| `mother`                 | +10.39 K         |            | 10.39 K              |                |
| `multisig`               |                  |            | 25.33 K              | 470_285        |
| `rand-extension`         |                  |            | 3.92 K               | 75_000         |
| `seal-code-hash`         |                  |            | 1.58 K               |                |
| `set-code-hash`          |                  |            | 1.57 K               | 150_000        |
| `subber`                 |                  |            | 2.15 K               |                |
| `trait-erc20`            |                  |            | 8.77 K               | 225_000        |
| `trait-flipper`          |                  |            | 1.00 K               | 75_000         |
| `trait-incrementer`      |                  |            | 1.19 K               | 150_000        |
| `updated-incrementer`    |                  |            | 9.69 K               |                |
| `upgradeable-flipper`    |                  |            | 1.55 K               |                |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/193521) | Last update: Thu May 12 20:36:24 CEST 2022

examples/mother/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+edition = "2021"
+publish = false
+
+[dependencies]
+ink_primitives = { version = "3.0.1", path = "../../crates/primitives", default-features = false }
+ink_metadata = { version = "3.0.1", path = "../../crates/metadata", default-features = false, features = ["derive"], optional = true }
+ink_env = { version = "3.0.1", path = "../../crates/env", default-features = false }
+ink_storage = { version = "3.0.1", path = "../../crates/storage", default-features = false }
+ink_lang = { version = "3.0.1", path = "../../crates/lang", default-features = false }
+ink_prelude = { version = "3.0.1", path = "../../crates/prelude", default-features = false }
```
athei, line 15: See: https://github.com/paritytech/ink/pull/1254

```suggestion
ink_primitives = { path = "../../crates/primitives", default-features = false }
ink_metadata = { path = "../../crates/metadata", default-features = false, features = ["derive"], optional = true }
ink_env = { path = "../../crates/env", default-features = false }
ink_storage = { path = "../../crates/storage", default-features = false }
ink_lang = { path = "../../crates/lang", default-features = false }
ink_prelude = { path = "../../crates/prelude", default-features = false }
```

examples/mother/lib.rs
```diff
@@ -0,0 +1,217 @@
+//! # Mother of All Contracts
+//!
+//! This contracts is intented to
+//! demonstrate rich posibilities of ink! contracts, namely:
```
athei, line 4:
```suggestion
//! This contracts is intended to make use of all features that are observable
//! by off chain tooling (for example UIs). It doesn't do anything useful beyond
//! serving off chain tooling developers with a contract to test their software against.
//! Currently, this includes the following:
```

examples/mother/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+//!
+//!   1. Use complicated nested input and ouput types.
+//!      This is done through the real use case example of data structure
+//!      needed to store a candle auction data.
+//!   2. Make contract fail with `ContractTrapped`.
+//!   3. Make contract fail with returning an Error.
+//!   4. Perform debug printing from contract into node's log.
```
athei, line 11:
```suggestion
//!   4. Perform debug printing from contract into node's log.
//!   5. Use complicated types in storage.
```

statictype:
@agryaznov still couldn't build the Mother, but i can answer your questions

> Currently we actually have `Vec<Option<(AccountId, Balance)>>`. Please, let me know if we need nested vectors still?

With this we'd like to test our recursive `Vector` React component that should support multiple levels of nesting, so we need at least `Vec<Vec<<(AccountId, Balance)>>`.


> We have Enum nested into Struct. Would it be enough?

What i meant here is nested `Enum` variants, for example: 

```
pub enum MilestoneIndex {
   M1,
   M2,
   M3,
}
pub enum ProjectStatus {
  Applied,
  PostPoned,
  Evaluation,
  Ongoing(MilestoneIndex),
  Alumni,
}

```
> Regarding Number types, I would left only BlockNumber (u32) for the contract if you really can cover others by unit tests.

👍🏻

athei:
> @agryaznov still couldn't build the Mother, but i can answer your questions

It is the linting that fails. It isn't smart enough to recognize the delegating constructor to initialize the contract.

agryaznov:
@statictype the contract is updated, looking forward your feedback!

statictype:
all good from my side, i used it with the UI and already revealed a few bugs. thanks so much!

statictype (approved):

examples/mother/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+                    panic!("Trapping on user demand!")
+                }
+                None => Ok(()),
+            }
+        }
+
+        /// Prints the specified string into node's debug log.
+        #[ink(message)]
+        pub fn debug_log(&mut self, message: String) {
+            let _ = message;
```
athei, line 221: You can merge master and remove this line now. The fix was merged.

PR #1255, opened 2022-05-16
Remove `Default` implementation for `AccountId`

### LINKED ISSUE #1110: Double check usage of `Default` `AccountId`s
Substrate had an issue recently related to having a [`Default` implementation for `AccountId`s](https://github.com/paritytech/substrate/pull/10403). We should double check our usage of `Default` and ensure that we don't have a similar issue.

HCastano:
The problem here is that in the `sr25519` and `ed25519` curves the
[zero-address public key has a known private key](https://substrate.stackexchange.com/questions/982/why-does-the-all-0-public-key-have-a-known-private-key-in-sr25519-and-ed25519).

The `Default` implementation of `AccountId` returned the zero-address. Since developers
frequently reach for `Default` it is entirely possible that they could shoot themselves
in the foot unknowingly. For example, assigning a contract owner to the zero address
and not updating it, or using it as a burn address.

There's also an argument to be made that the concept of the `Default` address doesn't
make sense anyways, and we shouldn't be giving this specific address a special treatment.

Closes #1110.


crates/storage/src/traits/impls/prims.rs
```diff
@@ -288,6 +288,53 @@ where
...
+    fn push_spread(&self, ptr: &mut KeyPtr) {
+        crate::traits::impls::forward_push_packed::<Self>(self, ptr)
+    }
+
+    fn clear_spread(&self, ptr: &mut KeyPtr) {
+        crate::traits::impls::forward_clear_packed::<Self>(self, ptr)
+    }
+}
+
+// TODO: Not sure if this can actually be implemented, since we need some sort of "sane" default
```
HCastano, line 311: Basically just blocked on this. I think we'll be able to sidestep this with the new storage re-write in #1217.

crates/env/src/call/call_builder.rs
```diff
@@ -341,7 +343,7 @@ where
 /// The default call type for cross-contract calls. Performs a cross-contract call to `callee`
 /// with gas limit `gas_limit`, transferring `transferred_value` of currency.
 pub struct Call<E: Environment> {
-    callee: E::AccountId,
+    callee: Option<E::AccountId>,
```
ascjones, line 346: I think this should remain `E::AccountId` so a `Call` can only ever be constructed with a callee. So we could pass it to the constructor e.g. `Call::new(callee)` instead of `Call::new().callee(callee)`. Or even modify the `CallBuilder` so that we don't need to construct `Call` directly.

We should prevent the user constructing invalid states using the type system where possible.

crates/env/src/engine/on_chain/impls.rs
```diff
@@ -425,7 +412,11 @@ impl TypedEnvBackend for EnvInstance {
     {
         let mut scope = self.scoped_buffer();
         let gas_limit = params.gas_limit();
-        let enc_callee = scope.take_encoded(params.callee());
+        let callee = params
+            .callee()
+            .as_ref()
+            .expect("An account ID must be set in order to call a contract.");
```
ascjones, line 418: Then we can remove this panic if we know that `callee` is always a value

ascjones (changes requested):

crates/env/src/call/call_builder.rs
ascjones: e.g. https://github.com/paritytech/ink/pull/1636

ascjones (approved):

PR #1267, opened 2022-05-19
Optimise `deny_payment`

xgreenx:
Seems we forgot to use optimization in cases if all constructors or methods don't accept payment.

PR #1233, opened 2022-04-24
Implement `ecdsa_to_eth_address()` and remove `eth_compatibility` crate

agryaznov:
- Follow up to [substrate#11087](https://github.com/paritytech/substrate/pull/11087)
- Resolves #1068 

Ported the following function from `eth_compatibility` crate to `ink_env`:
```rust
pub fn ecdsa_to_eth_address(pubkey: &[u8; 33], output: &mut [u8; 20]) -> Result<()>
``` 
`eth_compatibility` crate removed as having no reason to exist anymore

PR #1224, opened 2022-04-22
Add  `Mapping::contains(key)` and `Mapping::insert_return_size(key, val)`

agryaznov:
This adds:
``` rust
fn Mapping::contains(key) -> Option<u32>;
fn Mapping::insert_return_size(key, val) -> Option<u32>; 
```

both functions return the size of the (previously) stored value for the `key`.

Needed for [link#1](https://github.com/paritytech/link/issues/1)

PR #1284, opened 2022-06-10
Release `3.3.0` with backwards compatibility fixes

### LINKED ISSUE #1278: Restore ink! 3.x.x backward compatibility
This issue describes the reasons of [this](https://substrate.stackexchange.com/questions/2870/cargo-contract-throws-error-about-supplied-arguments-in-inkconstructor-f) StackExchange reported problem with _ink!_ `v3.0.1` 

Cargo crate dependencies version specifications [rules](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) work as follows:
`3.0.1 := >= 3.0.1 < 4.0.0`

This leads to that `3.2.0` is being the resolved version even when `=3.0.1` is specified in the contract manifest, for the following crates:
- `ink_allocator`
- `ink_engine`
- `ink_lang_codegen`
- `ink_lang_ir`
- `ink_prelude`
- `ink_storage_derive`

As a hotfix for this, I've [changed the dependencies](https://github.com/paritytech/ink/commit/0638f8208431ffb6c188170fe1132ad746dee1b1) to be specified as [tilde requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#tilde-requirements). I suggest making release `v3.0.2` out of it.

**Meanwhile, it is all really seems to me to be just about stitching to [SemVer](https://semver.org) rules**. E.g. `3.2.0` **should** have been backward compatible to `3.0.1`, if it's not then it should have been versioned `4.0.1`, instead. If these rules were followed, there would not be the need to apply tilde requirements as cargo is [SemVer compatible](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) by default. So maybe it worth re-considering _ink!_ versioning policy more rigorously.

Other questions to regard on this topic:

1.  Why should we really increment `examples/` contracts manifest versions without actually changing the contract sources itself at all?
2.  The workflow described [here](https://github.com/paritytech/ink/blob/master/RELEASES_CHECKLIST.md) does not imply fixing older releases like issuing `v3.0.2` containing hotfix for `v3.0.1`.  
    But it turns out we still should make such releases to support older versions being used by contract developers, as e.g. to fix such issues as being reported in the aforementioned SE question.

agryaznov:
*Context: user-reported issues on our SE ([first](https://substrate.stackexchange.com/questions/2721/cargo-contract-3-0-1) and [second](https://substrate.stackexchange.com/questions/2870/cargo-contract-throws-error-about-supplied-arguments-in-inkconstructor-f)) unveiled backward incompatibility introduced in the 3.1.0 release.*

The following has been done to restore backward compatibility:
- Reverted backward-incompatible piece of [#1233](https://github.com/paritytech/ink/pull/1233): removal of `eth_compatibility crate`
- Reverted backward-incompatible piece of [#1224](https://github.com/paritytech/ink/pull/1224): dependency on `[seal1] seal_set_storage`
- Reverted "Optimise deny_payment. Use eerywhere semantic of deny. ([#1267](https://github.com/paritytech/ink/pull/1267))"  
  (this one is to restore compatibiliy between minor versions of ink! crates; see @HCastano SE [answer](https://substrate.stackexchange.com/a/3000/472) in this regard)

All these reverted breaking changes are subject to be included in the upcoming MAJOR *ink! release 4*.
Meanwhile, to continue support and delivering fixes for  *ink! 3*, branch *v3.x.x* was created from *3.2.0* tag. 

---
This closes #1278.


HCastano:
Sorry, will take a look at this tomorrow. We should get this out soon

crates/env/src/api.rs
```diff
@@ -183,8 +183,22 @@ where
...
 }
 
-/// Writes the value to the contract storage under the given key and returns
-/// the size of pre-existing value at the specified key if any.
+/// The elder version of the function which is equivalent to the new one
+/// but returns nothing.
+///
+/// # Panics
+///
+/// - If the encode length of value exceeds the configured maximum value length of a storage entry.
```
HCastano, line 191:
```suggestion
/// Writes the value to the contract storage under the given key.
///
/// # Panics
///
/// - If the encode length of value exceeds the configured maximum value length of a storage entry.
///
/// # Developer Note
///
/// This is equivalent to the new [`set_contract_storage`] method, but in order to maintain old
/// behavior it returns nothing.
```

You'll probably have to fix the method link so it points to the right thing

crates/engine/src/ext.rs
```diff
@@ -227,6 +227,23 @@ impl Engine {
         });
     }
 
+    /// Silently writes the encoded value into the storage at the given key.
```
HCastano, line 230: Same thing with the `Developer Note` as I suggested elsewhere

crates/storage/src/traits/mod.rs
```diff
@@ -187,6 +187,18 @@ where
 
 /// Pushes the entity to the contract storage using packed layout.
 ///
+/// This is the silent equivalent to the newer version.
+pub fn push_packed_root_compat<T>(entity: &T, root_key: &Key)
```
HCastano, line 191: Don't we need the implementation of `Mapping::insert` to use this instead of `push_packed_root`?

RELEASES.md
```diff
@@ -1,4 +1,26 @@
 # [Unreleased]
+- Backward-incompatible piece of [#1233](https://github.com/paritytech/ink/pull/1233): removal of `eth_compatibility crate`
+- Backward-incompatible piece of [#1224](https://github.com/paritytech/ink/pull/1224): dependency on `[seal1] seal_set_storage`
+- Backward-incompatible "Optimise deny_payment. Use eerywhere semantic of deny. ([#1267](https://github.com/paritytech/ink/pull/1267))"
+  (see @HCastano SE [answer](https://substrate.stackexchange.com/a/3000/472) in this regard)
+
+# Version 3.3.0
+
+This release is to make *ink!* 3.x.x backward compatible with *substrate-contracts-node* version [0.13.0](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.13.0) again.
```
HCastano, line 9: This is just one of the reasons, we should also mention that we didn't follow SemVer

RELEASES.md
```diff
@@ -1,4 +1,26 @@
...
+- Backward-incompatible piece of [#1224](https://github.com/paritytech/ink/pull/1224): dependency on `[seal1] seal_set_storage`
+- Backward-incompatible "Optimise deny_payment. Use eerywhere semantic of deny. ([#1267](https://github.com/paritytech/ink/pull/1267))"
+  (see @HCastano SE [answer](https://substrate.stackexchange.com/a/3000/472) in this regard)
+
+# Version 3.3.0
+
+This release is to make *ink!* 3.x.x backward compatible with *substrate-contracts-node* version [0.13.0](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.13.0) again. 
+
+## Compatibility
+This version should work fine with *substrate-contracts-node* versions from [0.13.0](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.13.0) up to  [0.16.0](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.16.0)
```
HCastano, line 12:
```suggestion
This version should work fine with *substrate-contracts-node* versions from [0.13.0](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.13.0) up to [0.16.0](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.16.0)
```

RELEASES.md
```diff
@@ -1,4 +1,26 @@
 # [Unreleased]
+- Backward-incompatible piece of [#1233](https://github.com/paritytech/ink/pull/1233): removal of `eth_compatibility crate`
+- Backward-incompatible piece of [#1224](https://github.com/paritytech/ink/pull/1224): dependency on `[seal1] seal_set_storage`
+- Backward-incompatible "Optimise deny_payment. Use eerywhere semantic of deny. ([#1267](https://github.com/paritytech/ink/pull/1267))"
+  (see @HCastano SE [answer](https://substrate.stackexchange.com/a/3000/472) in this regard)
```
HCastano, line 5: Not sure if we need to mention these changes here

RELEASES.md
```diff
@@ -1,4 +1,26 @@
...
+This release is to make *ink!* 3.x.x backward compatible with *substrate-contracts-node* version [0.13.0](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.13.0) again. 
+
+## Compatibility
+This version should work fine with *substrate-contracts-node* versions from [0.13.0](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.13.0) up to  [0.16.0](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.16.0)
+
+## Changed
+*Context: user-reported issues on our SE ([first](https://substrate.stackexchange.com/questions/2721/cargo-contract-3-0-1) and [second](https://substrate.stackexchange.com/questions/2870/cargo-contract-throws-error-about-supplied-arguments-in-inkconstructor-f)) unveiled backward incompatibility introduced in 3.1.0 release.*
+
+The following has been done to restore backward compatibility:
+- Reverted backward-incompatible piece of [#1233](https://github.com/paritytech/ink/pull/1233): removal of `eth_compatibility crate`
```
HCastano, line 18:
```suggestion
- Reverted backward-incompatible piece of [#1233](https://github.com/paritytech/ink/pull/1233). This removed the `eth_compatibility crate`.
```

RELEASES.md
```diff
@@ -1,4 +1,26 @@
...
+- Reverted backward-incompatible piece of [#1224](https://github.com/paritytech/ink/pull/1224): dependency on `[seal1] seal_set_storage`
```
HCastano, line 19:
```suggestion
- Reverted backward-incompatible piece of [#1224](https://github.com/paritytech/ink/pull/1224). An old method depended on a new SEAL API (`[seal1] seal_set_storage`).
```

RELEASES.md
```diff
@@ -1,4 +1,26 @@
...
+- Reverted "Optimise deny_payment. Use eerywhere semantic of deny. ([#1267](https://github.com/paritytech/ink/pull/1267))"
```
HCastano, line 20:
```suggestion
- Reverted "Optimise deny_payment. Use everywhere semantic of deny. ([#1267](https://github.com/paritytech/ink/pull/1267))"  
```

RELEASES.md
```diff
@@ -1,4 +1,26 @@
...
+
+## Changed
+*Context: user-reported issues on our SE ([first](https://substrate.stackexchange.com/questions/2721/cargo-contract-3-0-1) and [second](https://substrate.stackexchange.com/questions/2870/cargo-contract-throws-error-about-supplied-arguments-in-inkconstructor-f)) unveiled backward incompatibility introduced in 3.1.0 release.*
+
+The following has been done to restore backward compatibility:
+- Reverted backward-incompatible piece of [#1233](https://github.com/paritytech/ink/pull/1233): removal of `eth_compatibility crate`
+- Reverted backward-incompatible piece of [#1224](https://github.com/paritytech/ink/pull/1224): dependency on `[seal1] seal_set_storage`
+- Reverted "Optimise deny_payment. Use eerywhere semantic of deny. ([#1267](https://github.com/paritytech/ink/pull/1267))"  
+  (this one is to restore compatibiliy between minor versions of ink! crates; see @HCastano SE [answer](https://substrate.stackexchange.com/a/3000/472) in this regard)
```
HCastano, line 21:
```suggestion
  This one is to restore compatibility between minor versions of ink! crates; see @HCastano's SE [answer](https://substrate.stackexchange.com/a/3000/472) in this regard.
```

HCastano (commented):
Main question I have has to do with the use of the `*_compat` methods

RELEASES.md
agryaznov: if we go with separate 3.x.x branch, I agree. This is done not to forget mentioning them in the 4.x release

RELEASES.md
cmichi, line 9:
```suggestion
This release is to make ink! 3.x.x backwards compatible with `substrate-contracts-node` version [0.13.0](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.13.0) again. 
```

RELEASES.md
HCastano: Sure, but it's a typo and we shouldn't propagate that everywhere 😄

RELEASES.md
HCastano: We're trying to alleviate two things here, incompatibility between 3.2 and earlier 3.x releases, as well as incompatibility with the 0.13 node

HCastano (changes requested):
Just realized that this won't work. 

We can't leave `set_contract_storage` as is, we need to revert it to not return anything, otherwise code like:

```rust
#[ink(message)]
pub fn foo(&self) {
    ink_env::set_contract_storage(&Default::default(), &1)
}
```

won't compile (even though it did in 3.0.1).

Instead of having the `compat` functions we need `with_size` functions (or something similarly named).

agryaznov:
> We can't leave `set_contract_storage` as is, we need to revert it to not return anything, otherwise code like:

Nice catch! Fixed for `ink_env::set_contract_storage()`.

> Instead of having the compat functions we need with_size functions (or something similarly named).

I renamed the user-facing api function. This should solve the issue I think. As a general rule, having `_compat` named old functions on the backend side should make it better to remove them all as a general cleanup before a next MAJOR release.

HCastano:
Can you remove all the `compat` functions and only use the `with_size` ones? It'll make things more consistent throughout, ensure we don't accidentally break SemVer somewhere we weren't expecting, and make removing the old functions easier when we move to 4.0

crates/env/src/backend.rs
```diff
@@ -162,9 +162,19 @@ impl CallFlags {
 
 /// Environmental contract functionality that does not require `Environment`.
 pub trait EnvBackend {
+    /// Writes the value to the contract storage under the given key.
+    ///
+    /// # Note
+    ///
+    /// This is an equivalent to the new [`set_contract_storage`][`Self::set_contract_storage`] method,
+    /// but in order to maintain legacy behavior it returns nothing.
```
HCastano, line 170: We should have this as part of the new functions, not the old one. We don't need to mention "legacy" behaviour or anything like that since it's part of the current release (and thus the current expected behaviour)

crates/storage/src/traits/impls/mod.rs
```diff
@@ -162,7 +162,7 @@ pub fn forward_push_packed<T>(entity: &T, ptr: &mut KeyPtr)
 where
     T: PackedLayout,
 {
-    push_packed_root::<T>(entity, ptr.next_for::<T>());
```
HCastano, line 165: Why are you changing this?

crates/env/src/engine/on_chain/ext.rs
```diff
@@ -375,7 +381,8 @@ mod sys {
             output_len_ptr: Ptr32Mut<u32>,
         ) -> ReturnCode;
 
-        pub fn seal_set_storage(
+        #[link_name = "seal_set_storage"]
```
HCastano, line 384: We need to do this since that's how the Contract pallet exports the `seal1` function, right? Just thinking if this can cause any sort of naming conflicts if we mix `seal0` and `seal1` methods within an ink! contract

crates/env/src/api.rs
```diff
@@ -183,18 +183,47 @@ where
...
         EnvBackend::set_contract_storage::<V>(instance, key, value)
+    });
+}
+
+/// Writes the value to the contract storage under the given key and returns
+/// the size of the pre-existing value at the specified key if any.
+///
+/// # Compatibility
+///
+/// This function requires minimum `substrate-contracts-node` version `v0.15.1`,
```
HCastano, line 214: We should link these two things as well

crates/env/src/api.rs
```diff
@@ -183,18 +183,47 @@ where
...
+/// or any node built with `substrate` version later than
+/// [#7d233c2446b5a60662400a0a4bcfb78bb3b79ff7](https://github.com/paritytech/substrate/tree/7d233c2446b5a60662400a0a4bcfb78bb3b79ff7).
```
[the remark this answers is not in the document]
agryaznov: I think using branch\tag name is better for human readability. But this actually would make us tightly coupled to the polkadot releases cadence. In this particular example, the `#7d233` commit made on May 12, while if we are to pin to the polkadot release branch instead, we have to pin to the `polkadot-v0.9.23` which appeared almost a **month** later (June 10), as it is the earliest release branch containing that commit.

crates/env/src/engine/on_chain/ext.rs
agryaznov: We import both `seal0` and `seal1` versions of the function, which have the same name in both modules, but this shouldn't be an issue in ink! once they are linked to differently named rust functions.

crates/storage/src/traits/impls/mod.rs
agryaznov: indeed

crates/env/src/engine/on_chain/ext.rs
agryaznov: double-checked, and yes, you can use both within the same ink! contract - that's the whole point ;)
```rust
        #[ink(message)]
        pub fn seal0(&self) {
            ink_env::set_contract_storage(&Default::default(), &1)
        }

        #[ink(message)]
        pub fn seal1(&self) -> Option<u32> {
            ink_env::set_contract_storage_inform(&Default::default(), &1)
        }
```

crates/storage/src/traits/mod.rs
```diff
@@ -187,21 +187,37 @@ where
...
+
+/// Pushes the entity to the contract storage using packed layout and
+/// returns the size of the pre-existing value if any.
+///
+/// The root key denotes the offset into the contract storage where the
+/// instance of type `T` is being pushed to.
+///
+/// # Note
+///
+/// This is an equivalent to the new [`push_packed_root`],
```
HCastano, line 213:
```suggestion
/// This is an equivalent to [`push_packed_root`],
```

crates/env/src/api.rs
```diff
@@ -183,18 +183,47 @@ where
...
-/// the size of pre-existing value at the specified key if any.
+/// Writes the value to the contract storage under the given key.
 ///
 /// # Panics
 ///
 /// - If the encode length of value exceeds the configured maximum value length of a storage entry.
-pub fn set_contract_storage<V>(key: &Key, value: &V) -> Option<u32>
+#[deprecated(
+    since = "3.3.0",
+    note = "`set_contract_storage_inform()` is provides more information, and will be made the standard in the future."
```
HCastano, line 193:
```suggestion
    note = "`set_contract_storage_inform()` provides more information, and will be made the standard in the future."
```

agryaznov:
also fixed spellcheck config (was done wrong in #1295) to make CI happy

paritytech-cicd-pr:

These are the results when building the `examples/*` contracts from this branch with `cargo-contract 1.4.0-25bfb96` and comparing them to ink! `master`: 

| `accumulator`            | +0.04 K          |            | 1.03 K               |                |
| `adder`                  | +0.10 K          |            | 2.13 K               |                |
| `contract-introspection` | +0.05 K          |            | 2.37 K               |                |
| `contract-terminate`     | +0.02 K          |            | 0.94 K               | 275_000        |
| `contract-transfer`      |                  |            | 8.31 K               | 75_000         |
| `data-structures`        | +0.01 K          |            | 1.73 K               |                |
| `delegate-calls`         | +0.07 K          | +22        | 2.95 K               | 76_264         |
| `delegator`              | +0.03 K          | +98        | 6.37 K               | 232_234        |
| `dns`                    | +0.08 K          |            | 8.89 K               | 225_000        |
| `erc1155`                | +0.55 K          |            | 17.29 K              | 450_000        |
| `erc20`                  | +0.07 K          |            | 8.49 K               | 225_000        |
| `erc721`                 | +0.19 K          |            | 11.81 K              | 600_000        |
| `flipper`                | +0.07 K          |            | 1.31 K               | 75_000         |
| `forward-calls`          | +0.01 K          | +5         | 2.88 K               | 151_416        |
| `incrementer`            | +0.07 K          |            | 1.21 K               |                |
| `mother`                 | +0.20 K          |            | 12.36 K              |                |
| `multisig`               | +0.01 K          | +6         | 25.29 K              | 470_246        |
| `rand-extension`         | +0.13 K          |            | 3.92 K               | 75_000         |
| `seal-code-hash`         | +0.19 K          |            | 1.58 K               |                |
| `seal-ecdsa`             | +0.09 K          |            | 1.80 K               |                |
| `set-code-hash`          | +0.07 K          |            | 1.57 K               | 150_000        |
| `subber`                 | +0.09 K          |            | 2.15 K               |                |
| `trait-erc20`            | +0.07 K          |            | 8.76 K               | 225_000        |
| `trait-flipper`          | +0.03 K          |            | 1.00 K               | 75_000         |
| `trait-incrementer`      | +0.06 K          |            | 1.19 K               | 150_000        |
| `updated-incrementer`    | +0.08 K          |            | 9.80 K               |                |
| `upgradeable-flipper`    | +0.07 K          |            | 1.55 K               |                |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/199456) | Last update: Mon Jun 20 22:13:42 CEST 2022

ascjones (approved):
LGTM

Issue #940, opened 2021-09-23
Document metadata format


cmichi:
We should document the metadata format for a stable 3.0 release. Without documentation the hurdle to build third party tools (like different UI's or CLI tools) is too high.

Ideally the `ink-docs` would contain a section on the metadata as well, maybe linking to more extensive documentation in the `ink_metadata` crate?

cmichi:
@ascjones I just merged https://github.com/paritytech/ink/pull/923, `polkadot-js/api` has been updated in https://github.com/polkadot-js/api/pull/4289. So this issue is no longer blocked.

PR #1290, opened 2022-06-13
Add JSON schema for metadata

HCastano:
This PR introduces a [JSON schema](https://json-schema.org/) which describes the contract
metadata. There are two schemas, the outer metadata (generated by `cargo-contract`) and
the versioned ink! metadata.

(This distinction only exists because `cargo-contract` is technically agnostic to the
language metadata, although in practice it's really only using the ink! metadata. In the
future we should be able to merge these into a single schema.)

The schemas provide a source of truth about what key-value pairs are supported and what
types they expect. This is useful for developers as well as external tooling. One use
case here is tooling which needs to verify that a given piece of metadata is valid.

Note that the generated process for the schemas is currently using some hacked branches
across `scale-info`, `ink`, and `cargo-contract` (the generation process is described in
the metadata crate's `README`). My goal is that in the future we'll be able to properly
upstream these changes across our repos and have an easy way to generate the schemas from
`cargo-contract`. At that point we'll be able to get rid of these hardcoded schema files.

Related issue: #940


cmichi (approved):
For some whatever reason I cannot add suggestions to the readme. I'm approving it hence now and might create a follow-up PR. 

For the `Metadata Schema Generation`, do you plan on upstreaming those changes?

HCastano:
> For the Metadata Schema Generation, do you plan on upstreaming those changes?

Yeah. I think the ideal workflow would be to have an option in `cargo-contract` to generate the schema. Will take a bit to upstream though since it cuts through a few repos

ascjones (approved):
LGTM

HCastano:
CI failures are unrelated to this PR

PR #1291, opened 2022-06-14
Add `built for ink!` badges

cmichi:


HCastano:
Where do you expect these badges to be used? In developer tooling?

PR #1311, opened 2022-07-07
Fix clippy `let_unit_value` warnings

ascjones:
Clippy was complaining about unit let bindings in the `cargo-contract` CI: https://gitlab.parity.io/parity/mirrors/cargo-contract/-/jobs/1664716

### DIFF
--- a/crates/lang/codegen/src/generator/dispatch.rs
+++ b/crates/lang/codegen/src/generator/dispatch.rs
@@ -807,7 +807,7 @@ impl Dispatch<'_> {
                 }
 
                 impl ::ink_lang::reflect::ExecuteDispatchable for __ink_MessageDecoder {
-                    #[allow(clippy::nonminimal_bool)]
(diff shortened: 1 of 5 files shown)


crates/lang/codegen/src/generator/dispatch.rs
```diff
@@ -807,7 +807,7 @@ impl Dispatch<'_> {
                 }
 
                 impl ::ink_lang::reflect::ExecuteDispatchable for __ink_MessageDecoder {
-                    #[allow(clippy::nonminimal_bool)]
+                    #[allow(clippy::nonminimal_bool, clippy::let_unit_value)]
```
ascjones, line 810: When the message returns `()` then the lint would be triggered on this line: https://github.com/paritytech/ink/blob/cc7e4773c47b07336e4aa81d7edf843cbd54b779/crates/lang/codegen/src/generator/dispatch.rs#L742

cmichi (approved):

PR #1313, opened 2022-07-11
Add `version` field to ink! metadata

### LINKED ISSUE #1289: Update `MetadataVersioned` to have `version` field
Right now versioning in the ink! metadata is done through the use of a version key which
wraps the metadata object:

```
"V3": {
    ... ink! metadata object ...
}
```

We should instead change this to be better defined by introducing a seperate `version`
field into the ink! metadata.

```
"abi": {
  "version": "3",
  "spec": { ... },
  "storage": { ... },
  "types": { ... },
}
```

This makes the meaning of the field more clear. We'll want to introduce this along side
the `V4` metadata changes.

HCastano:
Up until now the version of the ink! metadata/ABI was specified using a version key which
wrapped the whole ink! metadata struct (`MetadataVersioned(InkProject)`).

However, it is nicer to have a proper `version` field as part of the ink! metadata object
so that it can be queried in a less-ambiguous way.

Closes #1289.

### DIFF
--- a/crates/metadata/src/lib.rs
+++ b/crates/metadata/src/lib.rs
@@ -57,38 +57,32 @@ use serde::{
     Serialize,
 };
 
-/// Versioned ink! project metadata.
+/// The metadata version of the generated ink! contract.
+///
+/// The serialized metadata format (which this represents) is different from the
+/// version of this crate or the contract for Rust semantic versioning purposes.
 ///
 /// # Note
 ///
-/// Represents the version of the serialized metadata *format*, which is distinct from the version
-/// of this crate for Rust semantic versioning compatibility.
-#[derive(Debug, Serialize, Deserialize)]
-#[allow(clippy::large_enum_variant)]
-pub enum MetadataVersioned {
-    /// Version 0 placeholder. Represents the original non-versioned metadata format.
-    V0(MetadataVersionDeprecated),
-    /// Version 1 of the contract metadata.
-    V1(MetadataVersionDeprecated),
-    /// Version 2 of the contract metadata.
-    V2(MetadataVersionDeprecated),
-    /// Version 3 of the contract metadata.
-    V3(InkProject),
+/// Versions other than the `Default` are considered deprecated.
+#[derive(Default, Debug, Serialize, Deserialize)]
+pub enum MetadataVersion {
+    /// Represents the original non-versioned metadata format.
+    #[serde(rename = "0")]
+    V0,
+    #[serde(rename = "1")]
+    V1,
+    #[serde(rename = "2")]
+    V2,
+    #[default]
+    #[serde(rename = "3")]
+    V3,
 }
-
-impl From<InkProject> for MetadataVersioned {
-    fn from(ink_project: InkProject) -> Self {
-        MetadataVersioned::V3(ink_project)
-    }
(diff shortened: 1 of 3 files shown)


crates/metadata/src/lib.rs
```diff
@@ -57,38 +57,32 @@ use serde::{
...
-impl From<InkProject> for MetadataVersioned {
-    fn from(ink_project: InkProject) -> Self {
-        MetadataVersioned::V3(ink_project)
-    }
+/// Versions other than the `Default` are considered deprecated.
+#[derive(Default, Debug, Serialize, Deserialize)]
+pub enum MetadataVersion {
+    /// Represents the original non-versioned metadata format.
+    #[serde(rename = "0")]
+    V0,
```
ascjones, line 72: I wonder whether we should remove these deprecated version variants entirely. After all they will never be constructed here, and if you want to deserialize legacy versions then you just have to use an older version of the crate.

HCastano: Yeah good point, updated

crates/metadata/src/lib.rs
```diff
@@ -57,38 +57,27 @@ use serde::{
...
-impl From<InkProject> for MetadataVersioned {
-    fn from(ink_project: InkProject) -> Self {
-        MetadataVersioned::V3(ink_project)
-    }
+/// Versions other than the `Default` are considered deprecated. If you want to
+/// deserialize legacy metadata versions you will need to use an old version of
+/// this crate.
+#[derive(Default, Debug, Serialize, Deserialize)]
+pub enum MetadataVersion {
+    #[default]
```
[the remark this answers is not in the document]
ascjones: Can we just use a manual implementation then for now until 1.62 is propogated everywhere?

HCastano:
The `ink-waterfall` step fails because it's using an old `cargo-contract` `master` instead [of my branch on ](https://github.com/paritytech/cargo-contract/pull/641).

HCastano:
This builds fine with the `staging` Docker image. @cmichi should we use that instead while waiting for the `production` one to be fixed?

HCastano:
@ascjones can you take another look at this?

sergejparity (approved):

crates/metadata/src/lib.rs
```diff
@@ -57,38 +57,34 @@ use serde::{
...
-    V0(MetadataVersionDeprecated),
-    /// Version 1 of the contract metadata.
-    V1(MetadataVersionDeprecated),
-    /// Version 2 of the contract metadata.
-    V2(MetadataVersionDeprecated),
-    /// Version 3 of the contract metadata.
-    V3(InkProject),
+pub enum MetadataVersion {
+    #[serde(rename = "3")]
+    V3,
```
ascjones, line 73: Do we still need V3?

ascjones (approved):
LGTM

crates/metadata/src/lib.rs
HCastano: Guess not, removed

HCastano:
Let's wait until we sort out the release details of `cargo-contract` before merging this

HCastano:
Gonna :shipit: and get the release train started

PR #1314, opened 2022-07-11
Bump crate versions to 4.0

HCastano:
While working on https://github.com/paritytech/cargo-contract/pull/641 I realized that
ink!'s `master` branch was still based on 3.2. This makes for a strange situation in
which we're "downgrading" from ink! 3.3 to 3.2 in order to work off `master`.

I figured since the `master` branch already includes breaking changes it would be
appropriate to bump all the version numbers anyways.

### DIFF
--- a/crates/allocator/Cargo.toml
+++ b/crates/allocator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ink_allocator"
-version = "3.2.0"
+version = "4.0.0"
 authors = ["Parity Technologies <admin@parity.io>", "Robin Freyler <robin@parity.io>"]
 edition = "2021"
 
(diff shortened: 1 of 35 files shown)


HCastano:
> Are those also changes to the `examples`?
> 
> Generally we had bad experience in the past with making the examples contain changes incompatible with our published crates. People rightfully assume that they can copy/paste example code and if that code is not working out of the box it's a bad onboarding experience.

At the moment no, it's fine to copy-paste the examples and build them with 3.x dependencies. This is part of the reason why we were able to release backwards incompatible changes after 3.0.1 and not realize, haha.

However, after Green's storage changes this will not be the case.

At that point if people want working 3.x examples they will need to grab them from the 3.x branch instead of `master`.

PR #1327, opened 2022-07-20
Fixes for nightly changes 

cmichi:


athei (approved):

crates/lang/tests/ui/contract/fail/message-returns-non-codec.stderr
```diff
@@ -60,7 +60,7 @@ error[E0599]: the method `fire` exists for struct `ink_env::call::CallBuilder<De
     = note: the following trait bounds were not satisfied:
             `NonCodecType: parity_scale_codec::Decode`
 note: the following trait must be implemented
-   --> $CARGO/parity-scale-codec-3.1.5/src/codec.rs
```
HCastano, line 63: @cmichi uhh why's this been downgraded?

HCastano: Oh, nvm I see it's fixed in https://github.com/paritytech/ink/pull/1328

PR #1331, opened 2022-07-20
Storage refactoring new hope for review

xgreenx:
We decided to abandon old [PR](https://github.com/paritytech/ink/pull/1217) because it contains a lot of commits and comments and is hard to review.

The whole change was split into seven commits with their descriptions. In most cases, it is crate-based splitting. But some crates united together because they related to the same area.

You could check first the commit six with updated examples. It will help to understand how the usage changed. But after better to go in order of commits.
Commits:

1. [First commit](https://github.com/paritytech/ink/pull/1331/commits/2b26752dc38272e26e35393dc389c8f4b8fd1a6f) imports new unstable functions with transparent hashing.
Updated the API to work with the generic key `K: Encode` instead of the old `Key`.
Also, the change contains optimization to reduce the size of contracts. In most cases, it is `#[inline(always)]`; but `return_value` also got small optimization; removed usage of `extract_from_slice` where it is not needed.
1. [Second commit](https://github.com/paritytech/ink/pull/1331/commits/8dbf040b64eb18be25effea402a778811a0070d7):
    * primitives crate:
Removed old 32 bytes `Key`. Replaced it with `u32`. Added `KeyComposer`, it is a helper struct that does all manipulation with the storage key. It allows concat two storage keys into one, compute storage key for fields based on the filed, struct, enum, variants names.
Removed all tests and benches. Didn't add it for new primitives because the key is standard `u32`, and all keys are calculated during compilation.

    * storage crate:
Removed `SpreadLayout`, `PackedLayout`, `SpreadAllocate`, `PackedAllocate`, and all related helper functions.
Removed `Packed` struct cause it is the default behavior for storage right now. Added `Lazy` struct that allows `get`/`set` value from/into the storage. It is similar to `Mapping` but works with one storage key and one value.
Introduced new main traits to work with storage in `storage/src/traits/storage.rs`.
Also added a new `OnCallInitializer` trait to improve the flow with upgradable contracts and support initialization on demand by default. Added `pull_or_init` macro that inits the storage struct if it is impossible to load it. It also can be used as optimization for contracts without an explicit constructor.
Replaced implementation of old traits for main primitives with a new one. Added blanket implementation of new traits for structures that are `Packed` by default. It reduces the amount of code and adds support of generic structures but adds problems with tuples(now tuples implement new traits only if inner items are `Packed`).
Introduced `AutoKey` and `ManualKey` that allows specifying which key the user wants to use. Added support of it into all traits and structures.
Refactored `Mapping` to follow new rules.

1. [Third commit](https://github.com/paritytech/ink/pull/1331/commits/341c48b965ee4afab580a2eb74a83763201446f3):
    * metadata crate:
Updated storage layout in the metadata. Introduces the concept of roots and leafs. Root defines the storage key for all sub-tree until there is another Root. Leafs are common types that are part of the sub-tree and serialized/deserialized together into one storage key.
Replaced 32 bytes storage key with `u32`.
Added validation that all root storage keys don't overlap. Maybe better to add that error or reuse that validator on the `cargo-contract` side to show a more user-friendly error.
`RootLayout` and validator are used in codegen(next commit). The contract is wrapped into `RootLayout`, and we do validation for that tree.
Metadata now contains name for each struct/enum/variant/field. It is useful information because now the storage key is calculated based on the name of struct/enum/variant/field.

    * storage crate:
Added a new helper crate `storage/codegen`. It contains some useful functional that is used in `ink_storage_derive` and in `ink_lang_codegen` crates. It has a method that returns all types of the struct/enum/unit and a method that finds "salt" in the generics of the struct/enum. It uses the magic constant "KeyHolder"(about that you can read in the issue) to find salt, so I tried to have only one place where we are using that constant.
    * storage derive crate:
Replaced derive implementation of old trait with a new one. You can check the tests to see what the implementation looks like. `Storable` recursively calls `encode` and `decode` for all fields. `KeyHolder` return the key of the salt. `Item` uses `AutoKey` or a key specified by the user. I want to highlight that `PreferredKey` only is used with the `AutoItem` trait. If `PreferredKey` is `AutoKey`, then `AutoItem<AutoGenerated>` select auto-generated key, otherwise preferred. So `AutoItem` trait decides which key to use. It is why derive macro only set `PreferredKey`.
Updated drive for `StorageLayout`, now it uses `u32` and passes the name of the struct into metadata.

1. [Fourth commit](https://github.com/paritytech/ink/pull/1331/commits/346529c35e14649d2b6d7b52af81b5a66980a1cd) removed initialize_contract and related to initialization stuff. Now the codegen uses `pull_or_init` in the `call`. Updated `execute_constructor` to work with new storage methods.
Allowed usage of generics during the declaration of the primary contract storage. Users can specify the default storage key with `KeyHolder` via generic.
Added parser for `storage_item` macro with its config.
1. [Fifth commit](https://github.com/paritytech/ink/pull/1331/commits/b731f92269fd49ffbcf920c1810268f449c60507) removed the old codegen related to spread and packed layout. If some object implements `Decode` and `Encode`, it is `Packed`, and it uses the blanket implementation of new traits.
In dispatch, codegen started to use a new method to work with storage.
In metadata codegen added usage of new `RootLayout`. We wrap the contract into that layout because the contract has its storage key for all inner fields by default. Also added a run of validation logic during metadata generation.
Added `storage_item` macro. It transforms all types from autokey into manual key(if types support it). It calculates the storage key based on the name(it uses the `KeyComposer::compute_key` function from the primitives crate). Also, macro generates an additional `Check` structure that includes all raw fields. It helps show the user correct errors in typos, wrong types, etc.

1. [Sixth commit](https://github.com/paritytech/ink/pull/1331/commits/d6db8de2b9f54bd99af90fb09e407ecba74d4f05) updates examples to sue a new API.

1. [Seventh commit](https://github.com/paritytech/ink/pull/1331/commits/f967986d10688524570dbfff266eefa333415ecf) adds new UI tests and fixes the old one.


crates/env/src/engine/on_chain/ext.rs
```diff
@@ -587,7 +645,8 @@ macro_rules! impl_seal_wrapper_for {
                         )
                     };
                 }
-                extract_from_slice(output, output_len as usize);
+                // We don't need `extract_from_slice` here because it is stored in the own array
+                // in `get_property_little_endian` and `get_property_inplace`
```
HCastano, line 649: Sure, but we don't need this comment in here

HCastano (commented):
Yeah this is a lot easier to review!

So far I've only looked at the 1st and 6th commits (and took a peek at the expanded code), but will work through it this week

crates/storage/src/traits/storage.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/// Describes the type that should be used for storing the value and preferred storage key.
+///
+/// # Note
+///
+/// The trait is automatically implemented for [`Packed`](crate::traits::Packed) types
+/// via blank implementation.
+pub trait Item<Salt: KeyHolder> {
+    /// Type with storage key inside
```
HCastano, line 68: Nowhere in the contract with `Storable` is there a requirement that the type have a storage key, so I think this comment isn't totally accurate

crates/storage/src/traits/storage.rs
```diff
@@ -0,0 +1,97 @@
...
+    type Type: Storable;
+    /// Preferred storage key
+    type PreferredKey: KeyHolder;
```
HCastano, line 71: What do you mean by preferred? Is there a way for an `Item` to have a different storage key than the one given here?

crates/storage/src/traits/storage.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/// Automatically returns the type that should be used for storing the value.
+///
+/// Trait is used be codegen to use the right storage type.
```
HCastano, line 76: "... used _by_ codegen ..."

What does "correct" mean here?

crates/storage/src/traits/storage.rs
```diff
@@ -0,0 +1,97 @@
...
+pub trait AutoItem<Salt: KeyHolder> {
+    /// Type with storage key inside
+    type Type: Storable;
+}
+
+/// The contract can implement that trait to support initialization on the runtime
```
HCastano, line 82: "A trait to support ..."

crates/storage/src/traits/storage.rs
```diff
@@ -0,0 +1,97 @@
...
+/// if it is unable to pull from the storage.
+///
+/// It can be in several cases:
+/// - The contract doesn't have constructor. That initializer can be alternative for the constructor.
+/// - The constructor was not called due to upgrade ability, `Proxy` or `Diamond` pattern.
+/// - The storage was moved or corrupted.
+///
+/// If the trait is not implemented the behavior of the storage is default.
+/// It should be first initialized by the constructor.
+pub trait OnCallInitializer: Default {
```
HCastano, line 92: Does this need to be added as part of this PR, or can it be done in a follow up?

crates/storage/src/traits/storage.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// `Default::default` creates the instance of the contract.
+    /// After the `initialize` method is called on that instance.
+    /// The developer can do everything that he wants during initialization or do nothing.
```
HCastano, line 95: > A default instance of the contract is first created. The `initialize` method
> is then called on that instance. There are no restrictions to what a developer
> may do during the initialization phase, including doing nothing.

crates/storage/src/traits/impls/mod.rs
```diff
@@ -12,172 +12,8 @@
...
-                $crate::traits::impls::forward_allocate_packed::<Self>(ptr)
-            }
-        }
-    };
-}
-
 mod arrays;
-mod collections;
 mod prims;
 mod tuples;
```
HCastano, line 17: So these three modules just include tests and no real code related to trait implementations. I'd either move them here, or to to some tests module

crates/storage/src/lazy/mod.rs
```diff
@@ -15,7 +15,252 @@
...
+
+impl<V, KeyType> Lazy<V, KeyType>
+where
+    V: Storable,
+    KeyType: KeyHolder,
+{
+    /// Get the `value` from the contract storage.
+    ///
+    /// Panics if no `value` exists.
+    pub fn get(&self) -> V {
```
HCastano, line 142: So `Mapping::get()` returns an `Option`, maybe we should do that instead of panicking. We can think about adding a `Deref` impl or something if people want to be more 🤠 about accessing the inner contents

crates/storage/src/lazy/mapping.rs
```diff
@@ -154,7 +166,8 @@ where
     where
         Q: scale::EncodeLike<K>,
     {
-        pull_packed_root_opt(&self.storage_key(&key))
+        ink_env::get_contract_storage::<(&Key, Q), V>(&(&KeyType::KEY, key))
```
HCastano, line 169:
```suggestion
        ink_env::get_contract_storage(&(&KeyType::KEY, key))
```

crates/storage/src/lazy/mapping.rs
```diff
@@ -154,7 +166,8 @@ where
...
+            .unwrap_or_else(|error| panic!("failed to get value in mapping: {:?}", error))
```
HCastano, line 170:
```suggestion
            .unwrap_or_else(|error| panic!("Failed to get value in Mapping: {:?}", error))
```

I wonder if we can test this somehow. Maybe by clearing the underlying contract storage manually and then trying to pull. Really this shouldn't happen unless a storage migration messes up your storage layout

xgreenx (commented):
Thank you @HCastano for the review=) I hope I didn't miss any of your comments.

I applied most of your suggestions locally, updated/added comments and tests. There are several major questions regarding naming let's discuss them and after I will push all changes.

1. Hernando suggested renaming the `KeyHolder` trait into the `StorageKey` trait. During the previous review iteration, we decided to use the naming `KeyHolder` and we tried to remove all "Storage" prefixes for all entities in the `ink_storage`. The `KeyHolder` trait means that the type occupies some storage key. So `StorageKey` also fits well. 

1. The trait `Item<X: KeyHolder>` is used to manipulate the type's storage key. By default, the type will use the storage key passed as `X`. I called it `Salt` because in the implementation underhood it is used as a salt to generate the final storage key. But it is not accurate naming for public trait and may be better to name it "Key: StorageKey" or "KeyHolder: StorageKey", "ParentKey: StorageKey" or maybe another name(remember we can select a good name for bound too).

1. `Packed` vs non-`Packed`: With the introduction of blanket implementation of the `Packed` trait, it simplified the implementation of `Storable` and other traits for primitives. We can simplify the previous structure as Hernando suggested. I used the old structure because we didn't decide if we wanted to have two macros, `ink::packed` and `ink::non_packed`. With two macros, maybe we don't want to have blanket implementation. Do we want to discuss it or are we okay with the current implementation and I will simplify the structure more?

1. All functions from `ink_env` use `scale::Decode` and `scale::Encode`. `ink_env` doesn't know about `ink_storage::traits::Storable`. So [here](https://github.com/paritytech/ink/pull/1331#discussion_r932509661) I created `EncodeWrapper` and `DecodeWrapper` to transform `Storable` into `scale::Encode` or `scale::Decode`. And it is a hack=) I tried not to break the abstraction that all traits related to storage in the `ink_storage` crate. What do you think, can we move the `Storable` definition to `ink_env` and use `Storable` for `ink_env::set_contract_storage` and `get_contract_storage` instead of `scale::Decode` and `scale::Encode`? And where we should define derive macro for `Storable` in that case?

examples/upgradeable-contracts/set-code-hash/updated-incrementer/Cargo.toml
```diff
@@ -10,7 +10,7 @@ publish = false
 [dependencies]
 ink_primitives = { path = "../../../../crates/primitives", default-features = false }
 ink_metadata = { path = "../../../../crates/metadata", default-features = false, features = ["derive"], optional = true }
-ink_env = { path = "../../../../crates/env", default-features = false, features = ["ink-debug"] }
```
cmichi, line 13: I think this feature removal is unintentional.

HCastano:
>     1. Hernando suggested renaming the `KeyHolder` trait into the `StorageKey` trait. During the previous review iteration, we decided to use the naming `KeyHolder` and we tried to remove all "Storage" prefixes for all entities in the `ink_storage`. The `KeyHolder` trait means that the type occupies some storage key. So `StorageKey` also fits well.

I don't mind having the Storage prefix in this case. It makes things clearer imo.

>     2. The trait `Item<X: KeyHolder>` is used to manipulate the type's storage key. By default, the type will use the storage key passed as `X`. I called it `Salt` because in the implementation underhood it is used as a salt to generate the final storage key. But it is not accurate naming for public trait and may be better to name it "Key: StorageKey" or "KeyHolder: StorageKey", "ParentKey: StorageKey" or maybe another name(remember we can select a good name for bound too).

So a big part of the confusion for me is that the `Salt` is only used after a few layers
of trait implementations, and only really populated in one specific case.

It's not used on `Item` directly, and not even on `AutoItem` directly. Instead we use the
`Salt` when implemeting `KeyHolder` for `ManualKey`, and then `ManualKey` gets used in
the implementation of `AutoItem` (which internally uses `Item`).

I think there's room for simplification around the traits and implementations 
(Get rid of `AutoItem`? Roll the salting into `KeyHolder`? Implement `ManualKey`s later?),
but I'm a bit stuck on figuring out what the right move is here.

>     3. `Packed` vs non-`Packed`: With the introduction of blanket implementation of the `Packed` trait, it simplified the implementation of `Storable` and other traits for primitives. We can simplify the previous structure as Hernando suggested. I used the old structure because we didn't decide if we wanted to have two macros, `ink::packed` and `ink::non_packed`. With two macros, maybe we don't want to have blanket implementation. Do we want to discuss it or are we okay with the current implementation and I will simplify the structure more?

I'd stick with the current implementation. We can always have a follow up PR if we decide there
is a better approach.

>     4. All functions from `ink_env` use `scale::Decode` and `scale::Encode`. `ink_env` doesn't know about `ink_storage::traits::Storable`. So [here](https://github.com/paritytech/ink/pull/1331#discussion_r932509661) I created `EncodeWrapper` and `DecodeWrapper` to transform `Storable` into `scale::Encode` or `scale::Decode`. And it is a hack=) I tried not to break the abstraction that all traits related to storage in the `ink_storage` crate. What do you think, can we move the `Storable` definition to `ink_env` and use `Storable` for `ink_env::set_contract_storage` and `get_contract_storage` instead of `scale::Decode` and `scale::Encode`? And where we should define derive macro for `Storable` in that case?

Haven't gotten around to this yet, I'll need to think about it.

crates/storage/src/traits/storage.rs
cmichi: I think we should find a better name for `AutoItem`, something that reflects more what it's about. Maybe `StorageType`?

Regarding
>The trait is used by codegen to determine which storage key the type should have.

The word `type` here is confusing to me. We don't derive the key from the type, no? If I look at `fn convert_into_storage_field` the type is not part of the `key`.

examples/upgradeable-contracts/set-code-hash/updated-incrementer/Cargo.toml
xgreenx: I removed it to track the actual size of the contract

crates/storage/src/traits/storage.rs
xgreenx: > The word type here is confusing to me. We don't derive the key from the type, no? If I look at fn convert_into_storage_field the type is not part of the key.


The type defines which storage key we should use=)

```rust
// Here, the storage key is `123`
let storage: Contract<ManualKey<123>> = Default::default(); 
// Here, the storage key is `234`
let storage: Contract<ManualKey<234>> = Default::default();
// Here, the storage key is a hash of `234` and `123`
let storage: Contract<ManualKey<234, ManualKey<123> /* It is Salt or ParentStorageKey */ >> = Default::default();
```

> I think we should find a better name for AutoItem, something that reflects more what it's about. Maybe StorageType?

`StorageType` is the previous name of the `Item`=) And previously, it was `AutoStorageType`. I'm okay with `StorageType` but it doesn't highlight that it is related to automatically type determination. `Item` helps to determine the type too, but it also contains `PreferredKey`

xgreenx: What do you think about?:

```rust
pub trait Item<AutoGenerated: StorageKey> {
    type Type: Storable;
    type PreferredKey: StorageKey;
}

pub trait StorageType<AutoGenerated: StorageKey> {
    type Type: Storable;
}
```

or maybe simplify to `Generated`

crates/storage/src/lazy/mod.rs
```diff
@@ -18,4 +18,258 @@
...
+}
+
+impl<V, KeyType> Lazy<V, KeyType>
+where
+    V: Storable + Default,
+    KeyType: StorageKey,
+{
+    /// Reads the `value` from the contract storage.
+    ///
+    /// Returns `Default::default()` if no `value` exists.
```
HCastano, line 150:
```suggestion
    /// Returns the default value for the storage type if no `value` exists.
```

crates/storage/src/lazy/mod.rs
```diff
@@ -18,4 +18,258 @@
...
+
+impl<V, KeyType> Lazy<V, KeyType>
+where
+    V: Storable,
+    KeyType: StorageKey,
+{
+    /// Writes the given `value` to the contract storage.
+    pub fn set(&mut self, value: &V) {
+        push_storage(&KeyType::KEY, value);
+    }
```
HCastano, line 167: Should move this to the same impl block as `get`

crates/storage/src/lazy/mod.rs
```diff
@@ -15,7 +15,252 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::traits::ManualKey;
+
+    #[test]
+    fn set_and_get_work() {
+        ink_env::test::run_test::<ink_env::DefaultEnvironment, _>(|_| {
+            let mut storage: Lazy<u8, ManualKey<123>> = Lazy::new();
```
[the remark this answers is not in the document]
HCastano: I still don't think my first question was answered. Pretty sure we can just have 

```suggestion
            let mut storage: Lazy<u8> = Lazy::new();
```

And these tests would all be valid. You're not testing any behaviour related to `ManualKey` here

crates/storage/src/lazy/mod.rs
```diff
@@ -18,4 +18,258 @@
...
+            storage.set(&2);
+            assert_eq!(storage.get(), Some(2));
+
+            Ok(())
+        })
+        .unwrap()
+    }
+
+    #[test]
+    fn set_and_get_work_for_two_lazy_with_same_manual_key() {
```
HCastano, line 241: We should make a note somewhere, maybe as part of the doc comments of `Lazy` as to why this is powerful but also dangerous.

I can picture cases where developers copy paste some fields in a struct and end up with two fields that modify the same area in storage inadvertently

crates/storage/src/lazy/mod.rs
```diff
@@ -18,4 +18,258 @@
...
+
+            Ok(())
+        })
+        .unwrap()
+    }
+
+    #[test]
+    fn gets_or_default_if_no_key_set() {
+        ink_env::test::run_test::<ink_env::DefaultEnvironment, _>(|_| {
+            let storage: Lazy<u8, ManualKey<123>> = Lazy::new();
```
HCastano, line 257:
```suggestion
            let storage: Lazy<u8> = Lazy::new();
```

crates/storage/src/lazy/mod.rs
```diff
@@ -18,4 +18,258 @@
...
+
+            Ok(())
+        })
+        .unwrap()
+    }
+
+    #[test]
+    fn gets_fails_if_no_key_set() {
+        ink_env::test::run_test::<ink_env::DefaultEnvironment, _>(|_| {
+            let storage: Lazy<u8, ManualKey<123>> = Lazy::new();
```
HCastano, line 268:
```suggestion
            let storage: Lazy<u8> = Lazy::new();
```

crates/storage/src/lazy/mod.rs
```diff
@@ -18,4 +18,258 @@
...
+            assert_eq!(storage.get_or_default(), 0);
+
+            Ok(())
+        })
+        .unwrap()
+    }
+
+    #[test]
+    fn gets_fails_if_no_key_set() {
```
HCastano, line 266:
```suggestion
    fn gets_returns_none_if_no_value_was_set() {
```

crates/storage/src/lazy/mod.rs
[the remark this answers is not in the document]
HCastano: Still want this added :)

crates/storage/src/lazy/mod.rs
```diff
@@ -18,4 +18,258 @@
...
+///
+/// impl MyContract {
+///     #[ink(constructor)]
+///     pub fn new() -> Self {
+///         let mut instance = Self::default();
+///         let caller = Self::env().caller();
+///         instance.owner.set(&caller);
+///         instance.balance.set(&123456);
+///         instance
+///     }
```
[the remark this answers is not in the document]
xgreenx: It shows how to use the `Lazy` in your contract and how you can set the value.

crates/storage/src/lazy/mod.rs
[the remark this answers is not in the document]
xgreenx: Oh, yeas. We can do it right now=) I didn't allow `Lazy<T, AutoKey>` in the first implementation(the idea was to force you to use the `ink::storage_item` that will replace `AutoKey` with `ManualKey`).

crates/storage/src/traits/pull_or_init.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::traits::{
+    pull_storage,
+    storage::OnCallInitializer,
+};
+use ink_primitives::{
+    traits::Storable,
+    Key,
+};
+
+pub struct PullOrInit<T: Storable> {
```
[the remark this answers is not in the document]
xgreenx: Doc is hidden so it doesn't make sense)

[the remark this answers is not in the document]
HCastano: If you're looking at it in your IDE it could be useful ;)

crates/primitives/derive/src/tests/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+mod storable;
+
+#[macro_export]
+macro_rules! test_derive {
```
HCastano, line 18: 👍

crates/primitives/derive/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+
+extern crate proc_macro;
+
+mod storable;
+
+#[cfg(test)]
+mod tests;
+
+use self::storable::storable_derive;
+synstructure::decl_derive!(
```
HCastano, line 28: So I know we moved the `Storable` trait to primitives, but I'd avoid moving the proc macro here if possible

crates/primitives/derive/src/storable.rs
```diff
@@ -0,0 +1,140 @@
...
+    let variant: &synstructure::VariantInfo = &s.variants()[0];
+    let decode_body = variant.construct(|field, _index| {
+        let ty = &field.ty;
+        let span = ty.span();
+        quote_spanned!(span =>
+            <#ty as ::ink_primitives::traits::Storable>::decode(__input)?
+        )
+    });
+    let encode_body = variant.each(|binding| {
+        let span = binding.ast().ty.span();
```
HCastano, line 34: nit: Can we get the type name in here so it's something like: `__balances_binding` instead?

crates/primitives/derive/src/storable.rs
```diff
@@ -0,0 +1,140 @@
...
+                let ty = &field.ty;
+                let span = ty.span();
+                quote_spanned!(span =>
+                    <#ty as ::ink_primitives::traits::Storable>::decode(__input)?
+                )
+            })
+        })
+        .enumerate()
+        .fold(quote! {}, |acc, (index, variant)| {
+            let index = index as u8;
```
HCastano, line 77: Enums don't have a limit on the number of variants, so we can actually end up producing invalid code like this:

```rust
0u8 => Foo::A0,
1u8 => Foo::A1,
... => ...,
254u8 => Foo::A254,
255u8 => Foo::A255,
0u8 => Foo::A256,
1u8 => Foo::A257,
```

I'd stick with `usize` here.

HCastano (changes requested):
Issue with `Storable` derive macro and enums needs to be fixed

crates/storage/codegen/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+use syn::Data;
+
+/// Provides common methods for `DeriveInput`.
+///
+/// **Note:** This is only for internal usage in the `codegen` module.
+pub trait DeriveUtils {
+    /// Finds the salt of the structure, enum or union.
+    /// The salt is any generic that has bound `StorageKey`.
+    /// In most cases it is parent storage key or auto-generated storage key.
+    fn find_salt(&self) -> Option<syn::TypeParam>;
```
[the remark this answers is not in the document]
xgreenx: After renaming in the trait, I added the comment here that it is "parent storage key or auto-generated storage key". Because in most cases, it is used as salt to generate the final storage key=)

crates/storage/codegen/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+impl DeriveUtils for syn::DeriveInput {
+    fn find_salt(&self) -> Option<syn::TypeParam> {
+        self.generics.params.iter().find_map(|param| {
+            if let syn::GenericParam::Type(type_param) = param {
+                if type_param.bounds.len() == 1 {
+                    let bound = type_param.bounds.first().unwrap();
+                    if let syn::TypeParamBound::Trait(trait_bound) = bound {
+                        let segments = &trait_bound.path.segments;
+                        if !segments.is_empty()
+                            && segments.last().unwrap().ident == "StorageKey"
```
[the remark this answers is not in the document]
xgreenx: We already checked that it is not empty above `!segments.is_empty()`=)

crates/storage/codegen/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Return all types of the input.
+    fn all_types(&self) -> Vec<syn::Type>;
+}
+
+impl DeriveUtils for syn::DeriveInput {
+    fn find_salt(&self) -> Option<syn::TypeParam> {
+        self.generics.params.iter().find_map(|param| {
+            if let syn::GenericParam::Type(type_param) = param {
+                if type_param.bounds.len() == 1 {
+                    let bound = type_param.bounds.first().unwrap();
```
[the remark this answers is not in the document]
xgreenx: We already checked it)

crates/storage/derive/src/lib.rs
```diff
@@ -15,163 +15,94 @@
...
+    ///     StorageKey,
+    ///     AutoItem,
+    ///     AutoKey,
+    ///     ManualKey,
+    /// };
+    /// use ink_primitives::traits::Storable;
     ///
-    /// # ink_env::test::run_test::<ink_env::DefaultEnvironment, _>(|_| {
-    /// #[derive(SpreadLayout)]
+    /// #[derive(Default, Item, Storable)]
```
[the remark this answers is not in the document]
xgreenx: It is because of the blanket implementation of `Storable`. Rust tries to describe what should be implemented to use blanket implementation=) We can do nothing with it

crates/storage/derive/src/lib.rs
```diff
@@ -15,163 +15,94 @@
...
+    /// // (ManualKey<321> | ManualKey<123>) -> ManualKey<321>
+    /// assert_eq!(321, <NamedFieldsStorage<ManualKey<321>> as AutoItem<ManualKey<123>>>::Type::KEY);
     /// ```
-    spread_layout_derive
+    item_derive
 );
 synstructure::decl_derive!(
-    [PackedLayout] =>
-    /// Derives `ink_storage`'s `PackedLayout` trait for the given `struct` or `enum`.
+    [StorageKey] =>
```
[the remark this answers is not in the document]
xgreenx: The user can manually implement each trait or some traits. It is why `storage_item(derive = false)` exists. The user wants to do that for the case of types with generics.

crates/storage/derive/src/lib.rs
```diff
@@ -15,163 +15,94 @@
...
-    ///     pull_packed_root
+    ///     AutoItem,
+    ///     StorageKey,
+    ///     ManualKey,
+    ///     AutoKey,
     /// };
     ///
-    /// # ink_env::test::run_test::<ink_env::DefaultEnvironment, _>(|_| {
-    /// #[derive(Encode, Decode, SpreadLayout, PackedLayout)]
+    /// #[derive(StorageKey)]
```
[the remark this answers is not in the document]
xgreenx: Cause of blanket implementation=)

crates/storage/derive/src/item.rs
```diff
@@ -0,0 +1,87 @@
...
+    ToTokens,
+};
+use syn::{
+    parse2,
+    GenericParam,
+};
+
+fn item_inner(s: synstructure::Structure) -> TokenStream2 {
+    let ident = s.ast().ident.clone();
+    let salt_ident = format_ident!("__ink_generic_salt");
```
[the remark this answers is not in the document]
xgreenx: It is a salt to generate a combined key, so here is the actual name=)

crates/primitives/derive/src/storable.rs
xgreenx: Didn't get you=)

crates/primitives/derive/src/storable.rs
xgreenx: `scale::Encode` supports only `255` variants by default=) 
изображение

paritytech-cicd-pr:
 These are the results when building the `examples/*` contracts from this branch with `cargo-contract 1.5.0-ebd388e` and comparing them to ink! `master`: 

|                       | Δ Optimized Size | Δ Used Gas | Total Optimized Size | Total Used Gas |
|:-----------------------|------------------:|------------:|----------------------:|----------------:|
| `accumulator`         | +0.12 K          |            | 1.12 K               |                |
| `adder`               | -0.34 K          |            | 1.69 K               |                |
| `contract-terminate`  | +0.29 K          |            | 1.21 K               | 275_000        |
| `contract-transfer`   | +0.11 K          |            | 8.47 K               | 75_000         |
| `delegate-calls`      | -0.76 K          | -255       | 2.12 K               | 75_987         |
| `delegator`           | -0.60 K          | -669       | 5.74 K               | 231_469        |
| `dns`                 | -1.01 K          |            | 7.80 K               | 225_000        |
| `erc1155`             | -1.38 K          |            | 15.77 K              | 450_000        |
| `erc20`               | -1.29 K          |            | 7.13 K               | 225_000        |
| `erc721`              | -0.88 K          |            | 10.74 K              | 600_000        |
| `flipper`             | +0.11 K          |            | 1.34 K               | 75_000         |
| `forward-calls`       | -0.49 K          | -201       | 2.38 K               | 151_211        |
| `incrementer`         | +0.09 K          |            | 1.23 K               |                |
| `mother`              | -3.18 K          |            | 9.06 K               |                |
| `multisig`            | -2.52 K          | -2_060     | 23.27 K              | 468_423        |
| `payment-channel`     | -2.12 K          |            | 5.82 K               |                |
| `rand-extension`      | -1.02 K          |            | 2.77 K               | 75_000         |
| `set-code-hash`       | +0.02 K          |            | 1.51 K               |                |
| `subber`              | -0.35 K          |            | 1.71 K               |                |
| `trait-erc20`         | -1.23 K          |            | 7.46 K               | 225_000        |
| `trait-flipper`       | +0.16 K          |            | 1.13 K               | 75_000         |
| `trait-incrementer`   | +0.15 K          |            | 1.27 K               | 150_000        |
| `updated-incrementer` | -8.27 K          |            | 1.51 K               |                |
| `upgradeable-flipper` | -0.19 K          |            | 1.29 K               |                |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/208693) | Last update: Wed Aug 17 15:10:27 CEST 2022

crates/storage/codegen/src/lib.rs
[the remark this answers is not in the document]
HCastano: Exactly, so let's make that fact explicit with the `expect`

crates/storage/codegen/src/lib.rs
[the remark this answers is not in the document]
HCastano: I know, but it's good practice to have that instead of doing a raw unwrap

crates/primitives/derive/src/storable.rs
HCastano: I made a [Gist](https://gist.github.com/HCastano/3dea267d2c59d5909abca08104db3bfa) with the issue.

Seeing things like [this](https://gist.github.com/HCastano/3dea267d2c59d5909abca08104db3bfa#file-flipper-expanded-rs-L2732) in the expanded code worries me. I haven't been able to figure out how to exploit it, but we definitely should not allow it

xgreenx: [Fixed](https://github.com/paritytech/ink/pull/1331/commits/5ce970d41be41f52dabc8629252cf21cee8a027f)

crates/engine/src/ext.rs
```diff
@@ -269,14 +269,16 @@ impl Engine {
     }
 
     /// Removes the storage entries at the given key.
```
ascjones, line 271:
```suggestion
    /// Removes the storage entries at the given key.
    /// Returns the size of the previously stored value at the key if any.
```

crates/env/src/api.rs
```diff
@@ -183,49 +183,56 @@ where
...
+{
     <EnvInstance as OnInstance>::on_instance(|instance| {
-        EnvBackend::contract_storage_contains(instance, key)
+        EnvBackend::contains_contract_storage::<K>(instance, key)
     })
 }
 
-/// Clears the contract's storage key entry.
-pub fn clear_contract_storage(key: &Key) {
+/// Clears the contract's storage entry under the given storage key.
```
ascjones, line 229:
```suggestion
/// Clears the contract's storage entry under the given storage key.
///
/// If a value was stored under the specified storage key, the size of the value is returned.
```

crates/env/src/backend.rs
```diff
@@ -162,26 +162,33 @@ impl CallFlags {
...
-    /// Returns the size of a value stored under the specified key is returned if any.
-    fn contract_storage_contains(&mut self, key: &Key) -> Option<u32>;
+    /// Returns the size of a value stored under the given storage key is returned if any.
+    fn contains_contract_storage<K>(&mut self, key: &K) -> Option<u32>
+    where
+        K: scale::Encode;
 
-    /// Clears the contract's storage key entry.
-    fn clear_contract_storage(&mut self, key: &Key);
+    /// Clears the contract's storage key entry under the given storage key.
```
ascjones, line 188:
```suggestion
    /// Clears the contract's storage key entry under the given storage key.
    ///
    /// Returns the size of the previously stored value at the specified key if any.
```

ascjones (commented):
Update `clear_storage` docs

crates/primitives/derive/src/storable.rs
HCastano: When reading the generated code this is `__binding0`, `__binding1`, etc. So just for the dev's sake I figured it would be nicer to make it more readable. Again, it's a nitpick so you don't have to change anything if you don't see value in it

crates/storage/derive/src/lib.rs
```diff
@@ -15,200 +15,115 @@
...
+    /// assert_eq!(<NamedFieldsManualKey<ManualKey<123>> as StorageKey>::KEY, 123);
     /// ```
-    spread_allocate_derive
+    key_holder_derive
 );
-
 synstructure::decl_derive!(
     [StorageLayout] =>
-    /// Derives `ink_storage`'s `StorageLayout` trait for the given `struct` or `enum`.
+    /// Derives `ink_storage`'s [`StorageKey`](ink_storage::traits::StorageLayout) trait for the
```
cmichi, line 102:
```suggestion
    /// Derives `ink_storage`'s [`StorageLayout`](ink_storage::traits::StorageLayout) trait for the
```

crates/storage/derive/src/lib.rs
```diff
@@ -15,200 +15,115 @@
...
+    /// given `struct` or `enum`.
```
cmichi, line 103:
```suggestion
    /// Derives `ink_storage`'s [`StorageLayout`](ink_storage::traits::StorageLayout) trait
    /// for the given `struct` or `enum`.
```

crates/primitives/derive/src/storable.rs
xgreenx: I got your idea. I tried to do that, but it means removing the usage of `synstructure::Structure` and implementing everything by myself. I prefer to avoid it=D

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,149 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
     trait_def::analyze(attr.into(), item.into()).into()
 }
 
+/// Prepares the type to be fully compatible and usable with the storage.
+/// It implements all necessary traits and calculates the storage key for types.
+/// Packed types don't have a storage key, but non-packed types
```
HCastano, line 671: Can you link to the `Packed` trait docs?

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,149 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// (like `Mapping`, `Lazy` etc.) require calculating the storage key during compilation.
+///
+/// Consider annotating structs and enums that are intented to be a part of
+/// the storage with this macro. If the type is packed then the usage of the
+/// macro is optional.
+///
+/// If the type is non-packed it is best to rely on automatic storage key
+/// calculation. The storage key can also be specified manually with the
+/// generic `KEY` parameter though.
```
HCastano, line 680: Is this relevant to the usage of `storage_item`?

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,149 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// All required traits can be:
+/// - Derived manually via `#[derive(...)]`.
+/// - Derived automatically via deriving of `scale::Decode` and `scale::Encode`.
+/// - Derived via this macro.
+///
+/// # Example
+///
+/// ## Trait implementation
+///
+/// ```
```
HCastano, line 693: I'm having a hard time following all these docs. There's a lot of variance with the examples and there's no real indication as to when each case would apply.

Plus, only a couple show how to actually use `storage_item` (which is what the main focus of these docs should be).

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,149 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///     s6: PackedGeneric<Packed>,
+///     s7: NonPackedGeneric<Packed>,
+///     // Fails because: the trait `ink_storage::traits::Packed` is not implemented for `NonPacked`
+///     // s8: Mapping<u128, NonPacked>,
+/// }
+/// ```
+///
+/// ## Header Arguments
+///
+/// The `#[ink::storage_item]` macro can be provided with some additional comma-separated
```
HCastano, line 781:
```suggestion
/// The `#[ink::storage_item]` macro can be provided with an additional comma-separated
```

crates/lang/macro/src/lib.rs
```diff
@@ -665,6 +666,149 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// header arguments:
```
HCastano, line 782:
```suggestion
/// header argument:
```

crates/lang/macro/src/lib.rs
xgreenx: I can't because it is dev dependency

xgreenx:
Yeah, I also don't like it, but it seems it is a problem for all chain extensions=( Right now, we use the same version as you, but maybe you want to change it in the future

athei (commented):
Looked through the examples. Played around with it. Checked the docs and the naming. This implements the original design as we discussed it.

The metadata makes **some** sense to me. Have you checked whether it is enough to completely enumerate and decode storage?

crates/metadata/src/layout/mod.rs
```diff
@@ -54,16 +58,14 @@ pub enum Layout<F: Form = MetaForm> {
     ///
     /// This is the only leaf node within the layout graph.
     /// All layout nodes have this node type as their leafs.
-    ///
-    /// This represents the encoding of a single cell mapped to a single key.
-    Cell(CellLayout<F>),
+    Leaf(CellLayout<F>),
```
ascjones, line 61: Can we rename `CellLayout` to `LeafLayout` then?

ascjones (commented):
Review of layout metadata

crates/storage/src/pull_or_init.rs
```diff
@@ -0,0 +1,124 @@
...
+pub struct PullOrInit<T: Storable> {
+    marker: core::marker::PhantomData<fn() -> T>,
+}
+
+impl<T: OnCallInitializer + Storable> PullOrInit<T> {
+    #[allow(dead_code)]
+    pub fn pull_or_init(key: &Key) -> T {
+        let maybe_instance = ink_env::get_contract_storage::<Key, T>(key);
+        match maybe_instance {
+            Ok(None) | Err(_) => {
```
ascjones, line 43: Possible attack vector: this could be triggered by a call to an unrestricted `message`? I'm not sure what the attack would be but it means that any user could have control about when this is called if it has not already been initialized.

So not as safe as a constructor.

ascjones: Maybe could add a check to ensure only the owner can cause this to be invoked

crates/primitives/src/key.rs
```diff
@@ -39,314 +24,135 @@ use scale_info::{
...
-            let offset = i * len_chunk;
-            write!(
-                f,
-                "_{:02X}{:02X}{:02X}{:02X}",
-                bytes[offset],
-                bytes[offset + 1],
-                bytes[offset + 2],
-                bytes[offset + 3]
-            )?;
+pub type Key = u32;
```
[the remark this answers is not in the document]
ascjones: Agree that it should be its own type. Looks like it does work with the `const fn` https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5942cc2c00828d6050f9abccf90d2160

crates/metadata/src/layout/mod.rs
```diff
@@ -400,17 +442,21 @@ impl IntoPortable for ArrayLayout {
...
     /// The fields of the struct layout.
     fields: Vec<FieldLayout<F>>,
 }
 
 impl StructLayout {
     /// Creates a new struct layout.
-    pub fn new<F>(fields: F) -> Self
+    pub fn new<N, F>(name: N, fields: F) -> Self
     where
+        N: Into<&'static str>,
```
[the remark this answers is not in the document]
xgreenx: So, do I need to implement it here, or it will be done in https://github.com/paritytech/ink/pull/1357?=)

crates/storage/src/pull_or_init.rs
xgreenx: How do we know who is the owner of the contract?=)

ascjones: Is it a possible attack vector though?

crates/storage/src/traits/mod.rs
```diff
@@ -37,188 +36,52 @@ pub use self::layout::{
...
-use ink_primitives::Key;
+use ink_primitives::{
+    traits::Storable,
+    Key,
+};
 pub use ink_storage_derive::{
-    PackedLayout,
-    SpreadAllocate,
-    SpreadLayout,
+    StorableHint,
```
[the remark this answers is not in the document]
xgreenx: But in that case, we also need to move traits to that level. The idea is that you need to import only `Storage Layout`, and you have a trait and derive in one import

crates/metadata/src/layout/mod.rs
```diff
@@ -72,17 +74,17 @@ pub enum Layout<F: Form = MetaForm> {
...
+    key: Key,
 }
 
 impl serde::Serialize for LayoutKey {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: serde::Serializer,
     {
-        serde_hex::serialize(&self.key, serializer)
+        serde_hex::serialize(&self.key.to_be_bytes(), serializer)
```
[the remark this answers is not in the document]
xgreenx: The default serializer does it in LE. I want to keep the storage key and the selector the same in BE.

crates/storage/src/pull_or_init.rs
xgreenx: It can be called on any message(read and write via transaction) one time per contract(while storage is empty). If the attacker will somehow corrupt the storage, then yes, he again can call the initialization.

examples/erc20/lib.rs
```diff
@@ -60,23 +57,16 @@ mod erc20 {
...
-        }
-
-        /// Default initializes the ERC-20 contract with the specified initial supply.
-        fn new_init(&mut self, initial_supply: Balance) {
+            let mut instance = Erc20::default();
             let caller = Self::env().caller();
-            self.balances.insert(&caller, &initial_supply);
-            self.total_supply = initial_supply;
+            instance.balances.insert(&caller, &initial_supply);
+            instance.total_supply = initial_supply;
```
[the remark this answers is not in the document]
xgreenx: It is not possible=( 
изображение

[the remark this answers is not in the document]
xgreenx: But it is possible with `let mut balances: Mapping<AccountId, Balance, _> = Default::default();`

xgreenx:
We can't use a separate struct for `Key` because, in that case, we can't use `const KEY: Key`.
изображение

It is what makes the types unique. The type declaration contains the storage key as part of the type `Mapping<_, _, ManualKey<123>>`.

crates/metadata/src/layout/mod.rs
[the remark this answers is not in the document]
ascjones: Can do that in the other one, I suppose it is harmless

examples/erc20/lib.rs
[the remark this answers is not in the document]
ascjones: I've tested it and you can do `let mut balances = Mapping::default()` or even `Mapping::new()` which is a bit less ugly. It must be able to infer the type params from the `balances.insert`

crates/storage/src/traits/mod.rs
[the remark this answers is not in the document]
ascjones: Okay so why not have the traits at the top level too? Isn't that normal to have the main traits exported at the crate level?

crates/storage/src/pull_or_init.rs
ascjones: > one time per contract(while storage is empty)

Also possibly after a contract is upgraded and the storage layout has changed?

Maybe it's not that bad, it just feels like a potential footgun to have a hook which allows any user to run the migration code, even unintentionally.

crates/storage/src/traits/mod.rs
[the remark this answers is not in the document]
xgreenx: I used the same structure as in the previous version of the ink!=)
I can move it, but it is a decision on the ink! team=) @HCastano @cmichi

examples/erc20/lib.rs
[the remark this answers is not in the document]
xgreenx: `balances.insert` is a generic function, it is why the compiler can't get the type of the `K` in the `Mapping`=)
изображение

crates/storage/src/pull_or_init.rs
xgreenx: Yes, also after the upgrade=) The user who starts using it should know how to use it. We plan to highlight in the OpenBrush documentation about upgradable contracts

crates/metadata/src/layout/mod.rs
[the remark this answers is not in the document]
ascjones: But just using `u32` would not encode it so the byte order is not relevant? But if we want it as bytes then okay it should be the same as the selector.

crates/storage/src/pull_or_init.rs
ascjones: Okay, I'd be curious to know what the other guys think about this @cmichi @HCastano @athei. I'm not an expert in upgradable contract patterns.

> The user who starts using it should know how to use it.

Yes of course, which is why we should think very carefully when adding such features.

examples/erc20/lib.rs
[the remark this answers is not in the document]
ascjones: Wait but the compiler *is* able to infer the `K` and `V` from the usage. e.g. the following works:
```
            let mut balances = Mapping::default(); // or Mapping::new()
            let caller = Self::env().caller();
            balances.insert(&caller, &total_supply);
```
I think from your original error it couldn't infer the type of `balances` itself just from the usage. Providing the explicit `Mapping` hint appears to be enough.

crates/storage/src/traits/mod.rs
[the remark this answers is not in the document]
HCastano: @ascjones Yeah we said we'd leave them here for now, but we can always discuss a move in a follow up PR

crates/storage/src/pull_or_init.rs
HCastano: @xgreenx we should pull this out of this PR and discuss it elsewhere. It deserves a bit more thinking. We can get rid of the proxy examples for now

crates/lang/codegen/src/generator/storage_item.rs
```diff
@@ -0,0 +1,249 @@
...
+                quote_spanned!(span =>
+                    #field_name: #ty
+                )
+            });
+        let generics = self.item.generics();
+        let salt = self.item.salt();
+
+        quote! {
+            const _: () = {
+                struct Check #generics {
```
HCastano, line 204: Why's this `Check` struct required?

xgreenx: To return a user-friendly error if the user uses the wrong types. Without it, we will return that some invalid types are not implemented `StorageHint`.

crates/storage/src/pull_or_init.rs
ascjones: Indeed after reading up properly about the diamond patterns etc. it will be really useful to have a feature like this. Agree it is worth discussing as a feature on its own.

ascjones (approved):
LGTM!

HCastano (approved):
Green, thanks so much for this 💚! It's definitely been quite an effort on both sides, and
it's about time we get this into the hands of users. There's inevitably gonna be
follow up PRs as we discover bugs, edge cases, and quirks. Hopefully we catch
all the major ones before this hits a mainnet deployment 😅

xermicus (approved):

cmichi:
@xgreenx As Hernando wrote above: we want to give you a tip for the huge amount of work you've put into this PR (and the previous one from which this one was created + the smaller ones which were split off). In total you worked on this project for over six months, starting with the discussions in https://github.com/paritytech/ink/issues/1134, specifying it out. And then over four months ago you created the first PR and met with a number of people from the core team in person two times to discuss it. 

Thank you so much! We're very happy to have you be part of our ecosystem!

```
```
or
```
```

xgreenx:
> It's definitely been quite an effort on both sides, and
> it's about time we get this into the hands of users.

Yea, we finally did it!=)

> There's inevitably gonna be follow up PRs as we discover bugs, 
> edge cases, and quirks. Hopefully we catch all the major ones 
> before this hits a mainnet deployment 😅

I hope the number of problems will be as few as possible=) But definitely, we will have follow-up PRs to improve the feature, documentation, etc.


> Thank you so much! We're very happy to have you be part of our ecosystem!

💚💚💚

PR #1336, opened 2022-07-21
Fix overflow when creating `Layout` in bump allocator tests

### LINKED ISSUE #1323: [ci] Failing fuzz tests on master (18 Jul 2022)
The CI job [1682117](https://gitlab.parity.io/parity/mirrors/ink/-/jobs/1682117) just failed.

The offending commit is [Bump crate versions to 4.0 (#1314)](https://github.com/paritytech/ink/commit/8ba26d89409ca5330c507bffa3999d77fac8d82f).

HCastano:
Looks like I didn't read the `Layout::from_size_align` docs carefully enough and I
should've been checking for `isize` overflows instead of `usize` overflows.

Instead of hand checking this I instead decided to use `from_size_align` directly to
determine if something is valid/invalid.

I've been running the fuzz tests on my machine for a few hours and nothing's failed yet.

Closes #1323.

~~Draft since I need to double check I'm not accidentally changing the intention of some
heap overflow tests.~~

### DIFF
--- a/crates/allocator/src/bump.rs
+++ b/crates/allocator/src/bump.rs
@@ -289,16 +289,14 @@ mod fuzz_tests {
 
     #[quickcheck]
     fn should_allocate_arbitrary_sized_bytes(n: usize) -> TestResult {
-        // If `n` is going to overflow we don't want to check it here (we'll check the overflow
-        // case in another test)
-        if n.checked_add(PAGE_SIZE - 1).is_none() {
-            return TestResult::discard()
-        }
-
         let mut inner = InnerAlloc::new();
 
-        let layout =
-            Layout::from_size_align(n, size_of::<usize>()).expect(FROM_SIZE_ALIGN_EXPECT);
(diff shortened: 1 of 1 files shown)


crates/allocator/src/bump.rs
```diff
@@ -321,28 +314,6 @@ mod fuzz_tests {
         TestResult::passed()
     }
 
-    #[quickcheck]
-    fn should_not_allocate_arbitrary_bytes_if_they_overflow(n: usize) -> TestResult {
```
cmichi, line 325: Could you shortly explain why you think the test is no longer necessary?

HCastano: I explain in the commit which removed it (https://github.com/paritytech/ink/pull/1336/commits/fe85b7e93afc55c91d1488e58ec61ca52b7e4e1c) 😉

cmichi (approved):

PR #1339, opened 2022-07-25
Fix CI for nightly-2022-07-23

HCastano:


HCastano:
The PR CI uses the `production` Docker image, which is outdated in terms of its `nightly` version. However, the `master` CI is using `staging`, which is what I'm targeting with this fix. If we do mange to fix the `master` CI we'll be able to produce a new `production` image and fix our PRs.

Merging even though the pipelines are red.

PR #1357, opened 2022-08-17
Enable `ink_metadata` to be used for metadata generation outside of `ink!`

xermicus:
In order to fix [this issue](https://github.com/paritytech/cargo-contract/issues/659) together with [this change](https://github.com/paritytech/scale-info/pull/164/), we want to enable metadata construction at runtime. In Solang, we'd like to rely on the official `ink!` crates instead of maintaining our own solution. Additionally, it will be useful for anyone attempting to produce metadata for substrate contracts outside `cargo-contract` (e.g. when other languages want to add a Substrate target).

So far, this PR:
- Adds new constructors where needed
- Opens up some builder to be generic over `Form`
- Makes `String` the default String type

Ideally this should not break any existing API.


crates/metadata/src/specs.rs
```diff
@@ -171,7 +171,7 @@ impl<S> ContractSpecBuilder<S> {
     /// Sets the documentation of the contract specification.
     pub fn docs<D>(self, docs: D) -> Self
     where
-        D: IntoIterator<Item = &'static str>,
+        D: IntoIterator<Item = String>,
```
ascjones, line 174: Could add an extra generic param `S: ToString` and then `Item = S` to make it a non breaking api change?

xermicus: Good Idea I'll try that

crates/metadata/src/tests.rs
```diff
@@ -63,13 +63,11 @@ fn spec_contract_json() {
                         vec!["i32"].into_iter().map(AsRef::as_ref),
                     ))
                     .done()])
-                .docs(Vec::new())
```
xermicus, line 66: Because it requires a type hint now, this might technically still be an API change introducing regressions somewhere...

crates/metadata/src/specs.rs
```diff
@@ -397,19 +428,19 @@ pub struct MessageSpec<F: Form = MetaForm> {
     ///
     /// In case of trait provided messages and constructors the prefix
     /// by convention in ink! is the label of the trait.
-    label: F::String,
+    pub label: F::String,
```
ascjones, line 431: Don't forget to remove all these `pub` modifiers to require use of builders

xermicus: almost forgot :monkey: thanks!

xermicus: `#[must_use]` is set for the `new<T>(ty: T)` function, however there is no other means to construct it anyways? There is no builder for `ReturnTypeSpec` (probably because there is only one fuild anyways). Should I implement one or can we rather just ditch the must_use?

crates/metadata/src/utils.rs
```diff
@@ -58,11 +58,26 @@ where
...
 }
 
-/// Strips a single whitespace at the start and removes trailing spaces
-pub fn trim_extra_whitespace(item: &str) -> &str {
-    if let Some(stripped) = item.strip_prefix(' ') {
-        stripped.trim_end()
-    } else {
-        item.trim_end()
+pub trait DocString {
+    fn trim_extra_whitespace(item: Self) -> Self;
```
SkymanOne, line 62: Nit: Add docs with examples

SkymanOne (approved):
LGTM

crates/metadata/src/layout/mod.rs
```diff
@@ -241,6 +247,10 @@ where
     pub fn ty(&self) -> &F::Type {
         &self.ty
     }
+
+    pub fn new_from_ty(key: LayoutKey, ty: <F as Form>::Type) -> Self {
```
ascjones, line 251: I think this constructor should become `new` because it allows setting all fields. The existing `new<T>` can be renamed to `from_key<T>` or similar to be consistent.

crates/metadata/src/specs.rs
```diff
@@ -942,6 +1024,21 @@ impl<F> EventParamSpec<F>
 where
     F: Form,
 {
+    /// Creates a new event paramater specification builder from a custom type.
+    pub fn new_custom(label: F::String, ty: TypeSpec<F>) -> EventParamSpecBuilder<F> {
```
ascjones, line 1028: `EventParamSpecBuilder` already has `of_type` to mutate the type, so you don't need this.

crates/metadata/src/specs.rs
```diff
@@ -1029,31 +1133,33 @@ impl IntoPortable for ReturnTypeSpec {
...
         }
     }
-}
 
-impl<F> ReturnTypeSpec<F>
-where
-    F: Form,
-{
+    /// Create a new return type specification from a custom type.
+    pub fn new_custom(ty: Option<TypeSpec<F>>) -> Self {
```
ascjones, line 1159: Do we need this, can just use the `new` constructor above?

crates/metadata/src/specs.rs
```diff
@@ -1079,15 +1185,15 @@ impl IntoPortable for MessageParamSpec {
...
+            label: self.label.to_string(),
             ty: self.ty.into_portable(registry),
         }
     }
 }
 
 impl MessageParamSpec {
     /// Constructs a new message parameter specification via builder.
-    pub fn new(label: &'static str) -> MessageParamSpecBuilder {
+    pub fn new(label: <MetaForm as Form>::String) -> MessageParamSpecBuilder {
```
ascjones, line 1196: Make this generic over `Form` and remove the other constructor

crates/metadata/src/specs.rs
```diff
@@ -1102,6 +1208,15 @@ impl<F> MessageParamSpec<F>
 where
     F: Form,
 {
+    /// Constructs a new message parameter specification via builder.
+    pub fn new_custom(label: F::String, ty: TypeSpec<F>) -> Self {
```
ascjones, line 1212: You can then remove this and use `MessageParamSpecBuilder::of_type`

crates/metadata/src/specs.rs
```diff
@@ -366,21 +393,25 @@ impl<S, P> ConstructorSpecBuilder<S, P> {
     /// Sets the documentation of the message specification.
     pub fn docs<D>(self, docs: D) -> Self
     where
-        D: IntoIterator<Item = &'static str>,
+        D: IntoIterator<Item = <F as Form>::String>,
+        F::String: DocString,
```
ascjones, line 397:
```suggestion
        F::String: AsRef<[u8]>,
```
This might work to avoid requiring the `DocString` trait and the new impls.

crates/metadata/src/specs.rs
```diff
@@ -366,21 +393,25 @@ impl<S, P> ConstructorSpecBuilder<S, P> {
...
     {
         let mut this = self;
         debug_assert!(this.spec.docs.is_empty());
         this.spec.docs = docs
             .into_iter()
-            .map(trim_extra_whitespace)
+            .map(DocString::trim_extra_whitespace)
```
ascjones, line 403:
```suggestion
            .map(|s| trim_extra_whitespace(s.as_ref()))
```
Then you *should* be able to do something like this.

crates/metadata/src/specs.rs
```diff
@@ -888,6 +966,10 @@ where
     pub fn display_name(&self) -> &DisplayName<F> {
         &self.display_name
     }
+
+    pub fn new_from_ty(ty: <F as Form>::Type, display_name: DisplayName<F>) -> Self {
```
ascjones, line 970: Again I would make this the `new` constructor, and change the existing `new<T>` to e.g. `of_type<T>` or maybe `with_type<T>` as other constructors are called on this.

crates/metadata/src/utils.rs
```diff
@@ -58,11 +58,26 @@ where
     deserializer.deserialize_str(Visitor)
 }
 
-/// Strips a single whitespace at the start and removes trailing spaces
-pub fn trim_extra_whitespace(item: &str) -> &str {
```
ascjones, line 62: This might be able to be restored and remove the duplicate impls, see usage site above with `AsRef<[u8]>` bound

ascjones (changes requested):
Constructors need to be tidied up a bit. 

Since this will part of `ink 4.0`, we are not constrained here by having to make the changes non breaking as we were with `scale-info` so a bit more flexibility.

crates/metadata/src/specs.rs
xermicus: Neat idea!

crates/metadata/src/specs.rs
xermicus: The reason for it was that `Type::new()` is currently only implemented for `MetaForm`, but this should be doable

ascjones: Yes we should open up any idiomatic `new` constructors to be generic

crates/metadata/src/utils.rs
xermicus: Got rid of it completely :grin:

crates/metadata/src/utils.rs
xermicus: Not even needed, was able to nail it using just str refs :ok_hand:

xermicus:
@ascjones was able to tidy this up quite a bit. I think I'm done implementing all your code review suggestions, thanks! :heart: Only downside is I had to break the API (just slightly) so I'm gonna wait for the CI, but I think you can glance over the code already if you feel like it

ascjones:
> Only downside is I had to break the API (just slightly)

Breaking the API here is totally fine since this will be part of a major release `4.0` anyway.

ascjones (approved):
LGTM just needs CI fixing

ascjones:
> LGTM just needs CI fixing

Oh it's a `subxt` thing, I will look at that now

xermicus:
Yeah right. I mean I don't know whether API breaking is the right term, I just thought by making some constructors more generic it might then require an additional type hint here or there. But the CI failures seem unrelated to the change

xermicus:
> > LGTM just needs CI fixing
> 
> Oh it's a `subxt` thing, I will look at that now

Do you want me to wait with merging?

Issue #1178, opened 2022-03-11
mulitsig example contract: can not call `add_owner`, `remove_owner`, `replace_owner`


achimcc:
Whenever I try to invoke any of the functions `add_owner`, `remove_owner`, `replace_owner` of the `multisig` ink! contract from the examples folder, I receive the following error:

> ERROR: Module error: Contracts: ContractTrapped
> 
> Contract trapped during execution.

I did test it with apps, contracts ui and cargo-contract with:

cargo-contract 0.17.0-c6f9de0-x86_64-linux-gnu
ink! 3.0.0-rc9
substrate-contracts-node 0.8.0-382b446-x86_64-linux-gnu

Also not successfull with canvas on rococco.

I also checked that they are currently not tested by ink! waterfall, I'd suggest to add tests for them.

agryaznov:
Well, as a multisig's own essense dictates, this method should be invoked not by a single user account but rather by the multisig contract itself!
For that, one of the contract owners should use `submit_transaction()` for submitting the tx which should call `add_owner()`, then it should get enough confirmations from the multisig wallet owners by `confirm_transaction()`, and only after that in could get invoked by `invoke_transaction()`.

Wow! It may seem a lil bit tricky, but feel free to look at the contract [docs](https://github.com/paritytech/ink/blob/master/examples/multisig/lib.rs#L299) which hepls to get into it by the example.  

---

Anyway, this contract had a problem indeed, but in a different aspect: it should had have calling flags being setup to allow re-entrancy for that tricky methods which call the contract's own message. Otherwise, it fails (fairly enough) with a `contracts:ReentranceDenied`. This is to be fixed by the upcoming PR.

PR #1360, opened 2022-08-23
Allow reentrancy for multisig example contract

agryaznov:
`examples/multisig`: Messages `invoke_transaction()` and `eval_transaction()` while making the call should have the `allow_reentry` call flag being set to true, otherwise the tx invocation fails with `contract:ReentranceDenied`.

_(this has been found while re-producing #1178 )_

### DIFF
--- a/examples/multisig/lib.rs
+++ b/examples/multisig/lib.rs
@@ -64,10 +64,13 @@ use ink_lang as ink;
 
 #[ink::contract]
 mod multisig {
-    use ink_env::call::{
-        build_call,
-        Call,
-        ExecutionInput,
+    use ink_env::{
+        call::{
+            build_call,
+            Call,
+            ExecutionInput,
+        },
+        CallFlags,
     };
     use ink_prelude::vec::Vec;
     use ink_storage::{
@@ -546,6 +549,7 @@ mod multisig {
                         .gas_limit(t.gas_limit)
                         .transferred_value(t.transferred_value),
                 )
+                .call_flags(CallFlags::default().set_allow_reentry(true))
                 .exec_input(
                     ExecutionInput::new(t.selector.into()).push_arg(CallInput(&t.input)),
                 )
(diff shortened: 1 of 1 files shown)


examples/multisig/lib.rs
```diff
@@ -546,6 +549,7 @@ mod multisig {
                         .gas_limit(t.gas_limit)
                         .transferred_value(t.transferred_value),
                 )
+                .call_flags(CallFlags::default().set_allow_reentry(true))
```
athei, line 552: I think we should store this in the `Transaction` whether reentrancy should be enabled.

agryaznov: hm. Indeed, as the transaction can have any other .callee aside from the mutlisig itself, this makes sense

examples/multisig/lib.rs
```diff
@@ -135,6 +138,8 @@ mod multisig {
         pub transferred_value: Balance,
         /// Gas limit for the execution of the call.
         pub gas_limit: u64,
+        /// The call flag allowing reentry
```
athei, line 141:
```suggestion
        /// If set to true the transaction will be allowed to re-enter the multisig contract.
```

agryaznov: well, not necessarily multisig contract - it will apply to the contract we call which is specified by `t.callee`

examples/multisig/lib.rs
```diff
@@ -135,6 +138,8 @@ mod multisig {
...
+        pub allow_reentry: bool,
```
xgreenx, line 142: Better to pass the `CallFlag` itself to Allow more control:)
Also, we can add a note that allowing reentrancy may cause vulnerabilities

examples/multisig/lib.rs
athei: But multisig is always the caller and hence the one who blocks reenter.

examples/multisig/lib.rs
athei: Good idea.

agryaznov: Well, this would require implementing `parity_scale_codec::WrapperTypeEncode` for `ink_env::CallFlags` which is out of the scope of this contract. While this PR is just a bugfix of the contract.

examples/multisig/lib.rs
athei: Yes. Let's not shake the yak here if you don't want to.

athei (approved):

HCastano:
@agryaznov can you merge master? Should be good to go after that

Issue #1346, opened 2022-08-02
smart contract's balance doesn't change in tests


vni:
Hi!

I'm playing with ink! smart contracts and in tests it seems like there is a bug (or my misunderstanding)  when the user transfers tokens to contract, the balance of the contract doesn't change.

Here is the link to simple deposit contract code (lib.rs, without Cargo.toml):
https://gist.github.com/vni/fb897f32244e8ff6931b286bb926a191

All the ink_* dependencies are of version = "3"
(ink_primitives, ink_metadata, ink_env, ink_storage, ink_lang)

Contract was created with `cargo +nightly contract new simple_deposit`

Here is my question on substrate.stackexchange.com:
https://substrate.stackexchange.com/questions/3951/smart-contracts-balance-doesnt-change-in-tests

Contract was built with +nightly.
`cargo-contract 1.4.0-unknown-x86_64-unknown-linux-gnu`

**Describe the bug**
In tests I check the contract balance before and after the user deposits funds to the contract. I expect that the balance_before and balance_after are different, but they are the same. Contract balance doesn't change.

**Expected behavior**
I expect that in tests when I do some kind of fund transfer the balance of the contract will be increased and I can check it in tests.
```        
        #[ink::test]
        fn deposit() {
            let mut simple_deposit = SimpleDeposit::new(15);
            let balance_before = simple_deposit.env().balance();

            set_caller::<Environment>(bob_account());
            set_value_transferred::<ink_env::DefaultEnvironment>(15);
            let res = simple_deposit.deposit();
            assert!(res.is_ok());

            let balance_after = simple_deposit.env().balance();

            assert_ne!(balance_before, balance_after);
        }
```

Actually, this may be a bug or I may not understand something. So, I'm here for help. : )
Thanks in advance!!

agryaznov:
Currently, the off-chain Engine does not provide means for automatically moving funds along while making contract (payable) message invocation. So for now such a transfers can only be done in tests by explicitly accompanying such calls with `ink_env::test::set_account_balance()`. 
  
  This is not the best possible developer experience, and we could think about providing more handy way to do this in ink!.

  My suggestions are:

  1. Add a new fn `Engine::transfer_in(account_id, value)` which transfers amount from caller to callee.
     This is just a companion to [`Engine::transfer()`](https://github.com/paritytech/ink/blob/master/crates/engine/src/ext.rs#L182) (which transfers funds out of the contract).
[...]

PR #1379, opened 2022-09-01
Add `ink_env::pay_with_call!` helper macro for off-chain emulation of sending payments with contract msg calls

agryaznov:
Solves #1346


examples/contract-transfer/lib.rs
```diff
@@ -65,7 +65,9 @@ pub mod give_me {
     #[cfg(test)]
     mod tests {
         use super::*;
+        use ink_env::test::transfer_in;
```
HCastano, line 68: I don't like that we have to import this and not even use it (since it's used from within the macro). Can we instead import this in the macro?

crates/env/src/lib.rs
```diff
@@ -163,3 +163,16 @@ cfg_if::cfg_if! {
...
+        #[macro_export]
+        /// Prepend contract message call with value transfer. Used for tests in off-chain environment.
+        macro_rules! pay_with_call {
+            ($con:ident . $msg:ident ( $($params:ty)? ) , $amt:expr) => {{
+                transfer_in::<Environment>($amt);
+                $con.$msg($($params:ty)?)
+            }}
+        }
+    }
+}
```
HCastano, line 178: Can we replace this with a function that accepts a closure instead?

Something like:

```rust
pub fn pay_with_call_fn<F, O, T>(f: F, amount: T::Balance) -> O
where
    F: FnOnce() -> O,
    T: Environment<Balance = u128>,
{
    transfer_in::<T>(amount);
    f()
}
```

This way we can make it more easily discoverable with the rest of the testing functions

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -209,6 +209,42 @@ where
     })
 }
 
+/// Transfers value from the caller account to the contract.
```
HCastano, line 212: Some doc comments related on how to set the caller and callee accounts in a test would be useful

examples/contract-transfer/lib.rs
```diff
@@ -104,17 +106,27 @@ pub mod give_me {
             // given
             let accounts = default_accounts();
             let give_me = create_contract(100);
+            let contract_account = give_me.env().account_id();
 
             // when
-            // Push the new execution context which sets Eve as caller and
-            // the `mock_transferred_value` as the value which the contract
-            // will see as transferred to it.
+            // Push the new execution context which sets initial balnaces and
```
HCastano, line 112:
```suggestion
            // Push the new execution context which sets initial balances and
```

crates/env/src/lib.rs
agryaznov: I was thinking to impl it this way at first. But how to make it in the way so that it can work with messages returning something? I want to cover all possible message signatures, the ones taking parameters and the ones returning any value. I'm not sure if I get how your example solves this.

crates/env/src/engine/off_chain/test_api.rs
agryaznov: Well, the functions that make this are documented. I mean, if this action (manually setting caller and callee) did not require special docs beforehand (like e.g. [here](https://github.com/paritytech/ink/pull/1379/files/c49d0ffa99091e9350273f0c433fd75e6f0a74bf#diff-d388925fdaa4ef71cf03648e80d5e087cbb248271c13a396dd7d09516e59c93eR202)), why should it require it here. For end-user facing case the example is specially updated in this PR.

examples/contract-transfer/lib.rs
agryaznov: Makes sense, yeah

crates/env/src/engine/off_chain/test_api.rs
HCastano: I'm not sure what you're linking to. 

Either way, we should always be looking to improve our documentation if possible. If it's easy and reasonable to add example usage, we should do it

crates/env/src/lib.rs
```diff
@@ -163,3 +163,17 @@ cfg_if::cfg_if! {
...
+cfg_if::cfg_if! {
+    if #[cfg(feature = "std")] {
+        #[macro_export]
+        /// Prepend contract message call with value transfer. Used for tests in off-chain environment.
+        macro_rules! pay_with_call {
+            ($con:ident . $msg:ident ( $($params:ty)? ) , $amt:expr) => {{
+                use ink_env::test::transfer_in;
+                transfer_in::<Environment>($amt);
+                $con.$msg($($params:ty)?)
+            }}
```
HCastano, line 176:
```suggestion
        macro_rules! pay_with_call {
            ($contract:ident . $message:ident ( $($params:ty)? ) , $amount:expr) => {{
                ::ink_env::test::transfer_in::<::ink_env::Environment>($amount);
                $contract.$message($($params:ty)?)
            }}
        }
```

We should be more explicit with the paths here.

Also, can we move this into the `ink_env::test` module instead of having it be at top level? That way we can have it be closer to the test functions

examples/contract-transfer/lib.rs
```diff
@@ -66,6 +66,7 @@ pub mod give_me {
     mod tests {
         use super::*;
         use ink_lang as ink;
+        use ink_lang::codegen::Env;
```
HCastano, line 69: nit: I'd import this in `transfer_works` since it's only used here

crates/env/src/lib.rs
HCastano: Hmm alright fair enough, let's stick with the macro

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -355,3 +393,12 @@ pub fn assert_contract_termination<T, F>(
     assert_eq!(value_transferred, expected_value_transferred_to_beneficiary);
     assert_eq!(beneficiary, expected_beneficiary);
 }
+
+#[macro_export]
+/// Prepend contract message call with value transfer. Used for tests in off-chain environment.
```
HCastano, line 398:
```suggestion
/// Prepend contract message call with value transfer. Used for tests in off-chain environment.
#[macro_export]
```

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -355,3 +393,12 @@ pub fn assert_contract_termination<T, F>(
...
+macro_rules! pay_with_call {
+            ($contract:ident . $message:ident ( $($params:ty)? ) , $amount:expr) => {{
```
HCastano, line 400: This is all indented weird

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -355,3 +393,12 @@ pub fn assert_contract_termination<T, F>(
...
+                ::ink_env::test::transfer_in::<Environment>($amount);
```
HCastano, line 401:
```suggestion
                ::ink_env::test::transfer_in::<$crate::Environment>($amount);
```

examples/contract-transfer/lib.rs
```diff
@@ -101,20 +101,31 @@ pub mod give_me {
...
+            set_balance(contract_account, 0);
             set_sender(accounts.eve);
-            ink_env::test::set_value_transferred::<ink_env::DefaultEnvironment>(10);
 
             // then
-            // there must be no panic
-            give_me.was_it_ten();
+            // we use helper macro to emulate method invocation coming with payment,
+            // and there must be no panic
+            ink_env::pay_with_call!(give_me.was_it_ten(), 10);
```
HCastano, line 120: Ah looks like we can't refer to the macro outside of the root...for discovery it is nice that we have it alongside the other test functions though, but users may find it confusing. Let's leave it and see if we get any SE questions wondering about this

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -200,6 +200,7 @@ where
 }
 
 /// Sets the value transferred from the caller to the callee as part of the call.
+/// Please note that the acting accounts should be set with [`set_caller()`] and [`set_callee()`] beforehand.
```
HCastano, line 203:
```suggestion
///
/// Please note that the acting accounts should be set with [`set_caller()`] and [`set_callee()`] beforehand.
```

crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -209,6 +210,43 @@ where
     })
 }
 
+/// Transfers value from the caller account to the contract.
+/// Please note that the acting accounts should be set with [`set_caller()`] and [`set_callee()`] beforehand.
```
HCastano, line 214:
```suggestion
///
/// Please note that the acting accounts should be set with [`set_caller()`] and [`set_callee()`] beforehand.
```

HCastano (approved):
Some small things to fix, but looks good!

crates/env/src/engine/off_chain/test_api.rs
agryaznov: nope, this won't compile

we can leave it just `Environment` here because this type [is always declared](https://github.com/paritytech/ink/blob/ag-call-payable-offchain/crates/lang/codegen/src/generator/env.rs#L35) in the generated contract code.

armyhaylenko:
hey @agryaznov, we tested out the macro and it does not work.
since the macro is defined as `($contract:ident . $message:ident ( $($params:ty)? ) , $amount:expr)`, you can't really pass multiple params there, like `ink_env::pay_with_call!(contract.foobar(arg1, arg2, arg3.quox), 1000)`. it'll lead to compile errors. and why is `params` a type?

PR #1384, opened 2022-09-05
Document storage compatibility

### LINKED ISSUE #1247: Document what "storage compatibility" for `set_code_hash`
Follow-up to https://substrate.stackexchange.com/questions/2352/old-and-new-storage-compatibility-inkenvset-code-hash.

We should document what "storage compatibility" actually means. As I wrote in the answer for the StackExchange question it's the same requirements as in the OpenZeppelin docs.

This explanation of "storage compatibility" should also be linked on https://github.com/paritytech/ink/tree/master/examples/upgradeable-contracts.

SkymanOne:
Closes #1247

### DIFF
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -621,6 +621,57 @@ where
 /// # Errors
 ///
 /// `ReturnCode::CodeNotFound` in case the supplied `code_hash` cannot be found on-chain.
+///
+/// # Storage Compatibility
+/// When the smart contract code is modified, it is important to observe an additional restriction
+/// that is imposed on this procedure: you cannot change the order in which the contract state variables
+/// are declared, nor their type.
+///
+/// If the storage of your contract looks like this:
+/// ```rust
+/// #[ink(storage)]
+/// pub struct YourContract {
+///     x: u32,
+///     y: bool,
+/// }
+/// ```
+/// The procedures listed below will make it invalid:
+///
+/// Changing the order of variables:
+/// ```rust
+/// #[ink(storage)]
+/// pub struct YourContract {
+///     y: bool,
+///     x: u32,
+/// }
+/// ```
+/// Removing existing variable:
+/// ```rust
+/// #[ink(storage)]
+/// pub struct YourContract {
+///     x: u32,
+/// }
+/// ```
+/// Changing type of a variable:
+/// ```rust
+/// #[ink(storage)]
+/// pub struct YourContract {
+///     x: u64,
+///     y: bool,
+/// }
+/// ```
+/// Introducing a new variable before any of the existing ones
+/// ```rust
+/// #[ink(storage)]
+/// pub struct YourContract {
+///     z: Balance,
+///     x: u32,
+///     y: bool,
+/// }
+/// ```
+///
+/// Please refer to the [OpenZeppelin](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#modifying-your-contracts)
+/// for more details and examples.
 pub fn set_code_hash(code_hash: &[u8; 32]) -> Result<()> {
     <EnvInstance as OnInstance>::on_instance(|instance| instance.set_code_hash(code_hash))
 }
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -623,9 +623,13 @@ where
 /// `ReturnCode::CodeNotFound` in case the supplied `code_hash` cannot be found on-chain.
 ///
 /// # Storage Compatibility
-/// When the smart contract code is modified, it is important to observe an additional restriction
-/// that is imposed on this procedure: you cannot change the order in which the contract state variables
+/// When the smart contract code is modified, it is important to observe an additional virtual restriction
+/// that is imposed on this procedure: you should not change the order in which the contract state variables
 /// are declared, nor their type.
+///
+/// Violating the restriction will not prevent a successful compilation,
+/// but will result in the mix-up of values or failure to read the storage correctly.
+/// This can result in severe errors in the application utilising the contract.
 ///
 /// If the storage of your contract looks like this:
 /// ```rust
--- a/crates/env/src/api.rs
+++ b/crates/env/src/api.rs
@@ -637,7 +637,6 @@ where
 /// If the storage of your contract looks like this:
 ///
 /// ```ignore
-/// use ink_lang as ink;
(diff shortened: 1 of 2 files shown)


crates/env/src/api.rs
```diff
@@ -621,6 +621,57 @@ where
 /// # Errors
 ///
 /// `ReturnCode::CodeNotFound` in case the supplied `code_hash` cannot be found on-chain.
+///
+/// # Storage Compatibility
+/// When the smart contract code is modified, it is important to observe an additional restriction
+/// that is imposed on this procedure: you cannot change the order in which the contract state variables
+/// are declared, nor their type.
```
ascjones, line 628: Maybe it should be made absolutely clear that it is not a restriction per se, there is nothing to prevent you doing it. It is rather the consequences of having new storage data types which are incompatible with the existing stored data. In the best case it will result in the new contract simply failing to load the data from storage, in the worst case it could read the existing data into the new storage struct incorrectly, leading to unwanted side effects.

cmichi: Nit: Could you add an empty line after the `#`, so that we have it consistent everywhere? Also something like 80 CPL would be great for consistency.

SkymanOne: > Nit: Could you add an empty line after the `#`, so that we have it consistent everywhere? Also something like 80 CPL would be great for consistency.

Addressed this in the last commit

examples/upgradeable-contracts/README.md
```diff
@@ -27,3 +27,10 @@ more information on proxy patterns.
   This effectively replaces the code which is executed for the contract address.
 * The other contract (`updated-incrementer`) needs to be deployed on-chain.
 * State is stored in the storage of the originally instantiated contract (`incrementer`).
+
+## Storage Compatibility
+
+When working on the contract upgradeability, it is important to observe additional rules that are imposed on
+the modifications of storage:
+
+Please refer to the section of [Storage Compatibility](https://paritytech.github.io/ink/ink_env/fn.set_code_hash.html) in the ink! documentation.
```
cmichi, line 36:
```suggestion
Please refer to the section of [Storage Compatibility](https://paritytech.github.io/ink/ink_env/fn.set_code_hash.html) in the ink! crate documentation.
```

crates/env/src/api.rs
```diff
@@ -621,6 +621,66 @@ where
...
+/// it is important to observe an additional virtual restriction
+/// that is imposed on this procedure:
+/// you should not change the order in which the contract state variables
+/// are declared, nor their type.
+///
+/// Violating the restriction will not prevent a successful compilation,
+/// but will result in the mix-up of values or failure to read the storage correctly.
+/// This can result in severe errors in the application utilizing the contract.
+///
+/// If the storage of your contract looks like this:
```
cmichi, line 637:
```suggestion
/// If the storage of your contract looks like this:
///
```

crates/env/src/api.rs
```diff
@@ -621,6 +621,66 @@ where
...
+/// ```ignore
+/// use ink_lang as ink;
+/// #[ink(storage)]
+/// pub struct YourContract {
+///     x: u32,
+///     y: bool,
+/// }
+/// ```
```
cmichi, line 645:
```suggestion
/// ```
///
```

crates/env/src/api.rs
```diff
@@ -621,6 +621,66 @@ where
...
+/// The procedures listed below will make it invalid:
+///
+/// Changing the order of variables:
```
cmichi, line 648:
```suggestion
/// Changing the order of variables:
///
```

crates/env/src/api.rs
```diff
@@ -621,6 +621,66 @@ where
...
+/// ```ignore
+/// #[ink(storage)]
+/// pub struct YourContract {
+///     y: bool,
+///     x: u32,
+/// }
+/// ```
+/// Removing existing variable:
```
cmichi, line 656:
```suggestion
///
/// Removing existing variable:
///
```

crates/env/src/api.rs
```diff
@@ -621,6 +621,66 @@ where
...
+/// ```ignore
+/// #[ink(storage)]
+/// pub struct YourContract {
+///     x: u32,
+/// }
+/// ```
+/// Changing type of a variable:
```
cmichi, line 663:
```suggestion
///
/// Changing type of a variable:
///
```

crates/env/src/api.rs
```diff
@@ -621,6 +621,66 @@ where
...
+/// ```ignore
+/// #[ink(storage)]
+/// pub struct YourContract {
+///     x: u64,
+///     y: bool,
+/// }
+/// ```
+/// Introducing a new variable before any of the existing ones
```
cmichi, line 671:
```suggestion
///
/// Introducing a new variable before any of the existing ones
///
```

crates/env/src/api.rs
```diff
@@ -638,39 +646,40 @@ where
...
 ///     y: bool,
 /// }
 /// ```
 /// Introducing a new variable before any of the existing ones
-/// ```rust
+/// ```ignore
 /// #[ink(storage)]
 /// pub struct YourContract {
-///     z: Balance,
+///     z: Vec[u32],
```
ascjones, line 675:
```suggestion
///     z: Vec<u32>,
```

crates/env/src/api.rs
```diff
@@ -621,6 +621,75 @@ where
...
+///
+/// ```ignore
+/// #[ink(storage)]
+/// pub struct YourContract {
+///     x: u64,
+///     y: bool,
+/// }
+/// ```
+///
+/// Introducing a new variable before any of the existing ones
```
cmichi, line 679:
```suggestion
/// Introducing a new variable before any of the existing ones:
```

crates/env/src/api.rs
```diff
@@ -621,6 +621,75 @@ where
...
+/// are declared, nor their type.
+///
+/// Violating the restriction will not prevent a successful compilation,
+/// but will result in the mix-up of values or failure to read the storage correctly.
+/// This can result in severe errors in the application utilizing the contract.
+///
+/// If the storage of your contract looks like this:
+///
+/// ```ignore
+/// use ink_lang as ink;
```
cmichi, line 640:
```suggestion
```

For consistency.

cmichi (approved):

cmichi:
Thank you!

PR #1385, opened 2022-09-05
remove single whitespace prefix from comments and docs

SkymanOne:
Closes #613 

Only a single white space is removed which comes under the assumption that the user uses the standard cargo formatting style.

# Example of trimming
```json
...
"docs": [
  "# Flips the current value of the Flipper's boolean.",
  "Let's try to use some code",
  "```rust",
  "fn main() {",
  "    \"Hello, World!\"",
  "}",
  "```"
],
...
```

### DIFF
--- a/crates/metadata/src/specs.rs
+++ b/crates/metadata/src/specs.rs
@@ -14,7 +14,7 @@
 
 #![allow(clippy::new_ret_no_self)]
 
-use crate::serde_hex;
+use crate::{serde_hex, utils::trim_extra_whitespace};
 #[cfg(not(feature = "std"))]
 use alloc::{
     format,
@@ -367,7 +367,7 @@ impl<S, P> ConstructorSpecBuilder<S, P> {
     {
         let mut this = self;
         debug_assert!(this.spec.docs.is_empty());
-        this.spec.docs = docs.into_iter().map(str::trim).collect::<Vec<_>>();
+        this.spec.docs = docs.into_iter().map(trim_extra_whitespace).collect::<Vec<_>>();
         this
     }
 }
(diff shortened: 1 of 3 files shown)


crates/metadata/src/utils.rs
```diff
@@ -56,3 +57,11 @@ where
 
     deserializer.deserialize_str(Visitor)
 }
+
+pub fn trim_extra_whitespace(item: &'static str) -> &'static str {
```
ascjones, line 61:
```suggestion
pub fn trim_extra_whitespace(item: &str) -> &str {
```
Lifetimes can be elided

crates/metadata/src/specs.rs
```diff
@@ -586,7 +592,10 @@ impl<S, M, P, R> MessageSpecBuilder<S, M, P, R> {
     {
         let mut this = self;
         debug_assert!(this.spec.docs.is_empty());
-        this.spec.docs = docs.into_iter().collect::<Vec<_>>();
+        this.spec.docs = docs
+            .into_iter()
+            .map(trim_extra_whitespace)
```
ascjones, line 597: Good spot, could you update the `[trim_docs` test](https://github.com/paritytech/ink/blob/47b1fa3291dd7551002e780ace31df4d80e8539d/crates/metadata/src/tests.rs#L180) (or add a new one) to cover this, and also to test the updated trimming logic?

crates/metadata/src/utils.rs
```diff
@@ -56,3 +57,11 @@ where
 
     deserializer.deserialize_str(Visitor)
 }
+
+pub fn trim_extra_whitespace(item: &str) -> &str {
```
ascjones, line 61: Please also add a short comment to this public function

ascjones (approved):

PR #1389, opened 2022-09-08
Combine storage traits and derives into single crate

ascjones:
Follow up to #1331. Reviewers don't be alarmed this is just moving a bunch of stuff around.

This PR:

- Creates new `ink_storage_traits` crate
- Moves `Storable` definition and derive there (together with other traits)
- Moves some types to primitives to remove dependency on `ink_env`

It is to make life easier for storage derive macro crate identifiers in #1223.

- Currently `ink_primitives` contains the `Storable` trait definition and derive.
  - This is not a natural place for it, just required to avoid circular depenency (see below) 
- `ink_storage` -> `ink_env`, so `ink_env` cannot depend on `ink_storage` (circular dependency)


xgreenx (approved):
It looks clearly now=)
I like that `Storable` in the right place right now. And I definitely like definition of the `AccountId`, `Hash` in the `ink_ptimitives`=)

cmichi (approved):

PR #1390, opened 2022-09-12
Add IDE files to gitignore

DoubleOTheven:
* Add `.vscode` and `*.swp` (Vim) files to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,10 @@
 /examples/**/target/
 /design/
 
+# IDEs
+.vscode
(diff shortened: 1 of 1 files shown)


Robbepop (changes requested):
Those gitignores do not belong to the root gitignore file but to the local one.

ascjones:
Yes indeed, you can do something like ` echo .vscode > .git/info/exclude`, or just edit the ` .git/info/exclude` file directly.

Issue #1234, opened 2022-04-25
Add `[ink::e2e-test]` with a new E2E testing crate à la `ink_waterfall`


cmichi:
Persisting the results from our discussion on Saturday:

- We want to remove the existing off-chain testing environment (`engine`) and replace it with an E2E testing framework.
- This E2E testing framework will be a new crate based on `ink_waterfall`. So we want to make the waterfall a product as part of ink!.
- We will retain the `[ink::test]` syntax, but it will then instead spawn a `substrate-contracts-node` (if one is not already spawned) and execute real transactions via `cargo-contract instantiate/call/…`. 

The reason behind this is that we found that we need a proper E2E testing framework to keep up with competing products. In the past users found it very irritating that they can't test e.g. the call builder in our off-chain testing environment. As we move more towards interoperability (XCM & Co.) the need for E2E testing will need to be prioritized. By making the waterfall part of ink! in this way we can in the future also use this setup to test contracts that interact with parachains.

It will also remove the need of duplicating the API and logic of `pallet-contracts` in ink!.

ascjones:
I would add that there are potentially benefits to removing some abstractions required for the offchain env in the `ink!` codebase. It would certainly remove a lot of code, and allow more flexibility for more radical reductions in code size.

h4x3rotab:
E2E testing framework is good. However there are many things to consider. For example:

1. Performance. Usually running the unit tests takes less than one second. If we keep only e2e tests, afaik, we need at least minutes to bring up the node, submit the transactions, and setup the browser. It slows down the development. One solution I can come up with is to build a customized node with instant seal engine (just like [Ganache](https://trufflesuite.com/ganache/)), and run polkadot.js based test scripts without involving the browser. Although I haven't tried RedSpot, I guess it go with this way. Generally I would be happy if the test can be done in 30s.

2. Customization. A lot of parachains will build their own `ChainExtension`. Any E2E test framework should allow the the developer to use their own configuration in pallet-contracts. Taking Phala Fat Contract as a special example, it has a heavily tweaked runtime environment with its own RPC API outside the substrate node. It would be nice if the E2E test framework should be flexible enough to support our use case.

[...]

PR #1391, opened 2022-09-13
CI: Use stable for `clippy` and `cargo test`

cmichi:
This is needed for https://github.com/paritytech/ink/issues/1234.

The reason is that going forward `cargo-contract` will require stable and the E2E tests use it for building the contract. If the CI runs nightly clippy on our example contracts and later builds them in stable it runs into:

```
error[E0514]: found crate `core` compiled by an incompatible version of rustc
```

I'm aware of the discussion that happened in https://github.com/paritytech/scripts/pull/453, namely @HCastano was bringing this up:

>I've thought about this a bit and I'm not sure we want to do this, since contributors to ink! will have to use `RUSTC_BOOTSTRAP=1` locally to check that Clippy passes. I'd prefer if they were told to use nightly instead.

I agree, it's a bit unfortunate, but working around the above compiler error would require significant work to our CI setup and how it uses `CARGO_TARGET_DIR`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,6 +97,11 @@ You can have a look at the [PR#1284](https://github.com/paritytech/ink/pull/1284
 
 Our [continuous integration (CI)](https://github.com/paritytech/ink/blob/master/.gitlab-ci.yml) will check for the following properties of all changes.
 
+*Note:* We use Rust `stable` together with `export RUSTC_BOOTSTRAP=1` in our CI.
+So you have to export this environment variable locally too!
+Setting it will enable nightly features in stable Rust. So it's definitely a hack,
+but we decided on it since using nightly in the CI came with a lot of maintenance burden.
+
 1. Is `rustfmt` happy with it ?
     - `cargo fmt --all`
 1. Is `clippy` happy with it?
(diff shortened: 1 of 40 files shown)


CONTRIBUTING.md
```diff
@@ -36,35 +36,46 @@ contributing in the first place anyways, right?)
...
-   This might seem pedantic but we believe that in essence this is going to improve overall comment and documentation quality.
+   This might seem pedantic, but we believe that in essence this is going to improve overall comment and documentation quality.
 1. If possible try to sign your commits, e.g. using GPG keys. For more information about this go [here](https://help.github.com/en/articles/signing-commits).
 
+Verify the following locally, otherwise the CI will fail:
+
+1. Is `rustfmt` happy with it ?
+    - `cargo +nightly fmt --all`
+1. Is `clippy` happy with it?
+    - `cargo clippy --all-targets --all-features`
```
ascjones, line 67:
```suggestion
    - `cargo clippy --all-targets --all-features` -- -D warnings;
```

CONTRIBUTING.md
```diff
@@ -36,35 +36,46 @@ contributing in the first place anyways, right?)
...
+1. Does the code still compile?
+    - `cargo check --all-features`
+1. Do all the examples still compile?
+    - `cargo contract check --manifest-path ./examples/.../Cargo.toml`
+1. Is the `wasm-32` target still compiling?
```
ascjones, line 72:
```suggestion
1. Is the `wasm32` target still compiling?
```

ascjones (approved):
Awesome this means having to update the UI tests less often

PR #1393, opened 2022-09-13
Use XXH32 instead of sha256 for const hashing

xermicus:
Would it be feasible to use ~~[FNV](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function)~~ `xxhash` instead of sha256 for Key hashing, since we use only 32bits of the sha hash anyways? With ~~FNV~~ `xxhash` being a much simpler algorithm, it would allow us to use `ink_metadata` with stable rust without relying on `RUSTC_BOOTSTRAP`. Also what do you think @xgreenx ?


xgreenx:
I thought about this solution(in the first implementation I used a custom hash function). The problem is that the chance of getting collision is higher, and it will make the implementation of the same storage key calculation for other languages harder because they need find implementation of this not popular function=) The same for documentation, we need to describe the usage of an unusual hashing algorithm.

But if other members of the ink! team like this idea, of course, go with it=)

xermicus:
Thanks for your feedback! At first I shared your concerns about collisions, but as I thought a bit about it I came to a different conclusion. My guess was that reducing sha2 to 32 bits only has significant impact on its collision rate as well. And FNV1A is actually not that bad: According to [this stack overflow answer](https://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed/145633#145633), there are only ~4 collisions per ~200K hashes of UUIDS.

This left me wondering, how bad is FNV1A compared to 32bit SHA256? I quickly hacked together [my own very small and totally unscientific experiment](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=0b76f70bf0be3fec23a33536887727a1). If you run this multiple times, you'll observe that the SHA2 option, if at all, does only insignificantly better than FNV1a. In fact, sometimes there are even more collisions for SHA2:

```
Testing for 200000 UUIDs with 32 runs...
FNV1A collisions: 137 (avg. 4.28125 per 200000 UUIDs)
SHA2_32 collisions: 155 (avg. 4.84375 per 200000 UUIDs)
```

Based of that it seems like we don't really gain much in terms of collision resistance by using 32bit SHA2 instead of FNV1A. This only applies to hashing UUIDs though (behavior in our use case might be different!).

> the usage of an unusual hashing algorithm

I don't think FNV is "unusual". To my knowledge it is a solid (in terms of speed vs collision resistance) and often used non cryptographic hash.

> But if other members of the ink! team like this idea, of course, go with it=)

I'd be very happy to hear some more opinions on this one as well :blush:

ascjones (commented):
I am in favour of this because it is a huge headache to require people who depend on `ink` std libs to run `nightly` or `RUSTC_BOOTSTRAP=1`.

However I am not an expert on the suitability or not of hashing algorithms for unique key generation, but I found this instructive https://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed.

So I would like somebody else here with more knowledge to approve this.

crates/primitives/Cargo.toml
```diff
@@ -19,7 +19,7 @@ derive_more = { version = "0.99", default-features = false, features = ["from",
 ink_prelude = { version = "4.0.0-alpha.1", path = "../prelude/", default-features = false }
 scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive", "full"] }
 scale-info = { version = "2", default-features = false, features = ["derive"], optional = true }
-sha2-const = { version = "0.1.2", default-features = false }
+const-fnv1a-hash = "1.1.0"
```
athei, line 22: Might it even possible to just put it under `build-dependencies` as it is only used in a `const` context?

xermicus: hmm, what happens if I want to use the metadata crate at runtime to create the ABI file?

xermicus:
As discussed internall, switched to xxhash32, since it is used already in substrate for non cryptographic hashes.

I used 0 as seed value, it is a valid seed and seems to be the common choice (`xxh32sum` utility and LZ4 compression use 0 as seed for example).

crates/primitives/Cargo.toml
athei: But it is still a `const` function. So it can only be evaluated during compilation. It won't work on values only known at runtime. How are dynamic keys hashes by the way (map keys)? They can't be hashes by this crate obviously. Is there still some SHA in there somewhere?

xermicus: Sure, but if I wanted to use `ink_primitives::KeyComposer::from_str(...)` for example in another project. How would this still be possible if we move all of it into a build script?

athei: Point is you can't even do that with the current code as `from_str` is a `const` function. If you want to run that function on a runtime value you would need to add a non-const function and this would be the time when you move the dependency from `build-dependencies` to `dependencies`.

xermicus: I'm not sure whether I understand correctly. `const` function can be run at runtime too?

athei: No. The function will always be executed during compilation of the program.

athei (approved):

crates/primitives/src/key.rs
```diff
@@ -69,8 +65,7 @@ impl KeyComposer {
     /// 1. If `variant_name` is not empty then computes the ASCII byte representation and call it `V`.
     /// 1. Compute the ASCII byte representation of `field_name` and call it `F`.
     /// 1. Concatenate (`S` and `F`) or (`S`, `V` and `F`) using `::` as separator and call it `C`.
-    /// 1. Apply the `SHA2` 256-bit hash `H` of `C`.
-    /// 1. The first 4 bytes of `H` make up the storage key.
+    /// 1. The `FNV1A` 32-bit hash of `C` is the storage key.
```
HCastano, line 68: @xermicus this should be `XXHASH` instead, eh?

xermicus: thanks! :see_no_evil:

PR #1395, opened 2022-09-14
Add E2E testing framework MVP

cmichi:
MVP for https://github.com/paritytech/ink/issues/1234.

I've created a fresh PR to make it easier for reviews.

Contains only E2E tests for our `contract-transfer` example so far, no cross-contract tests yet. Once we've nailed down the API I'll add E2E tests for the other examples in follow-up PRs. The best way to look at this PR is, I think, to look at the `#[ink::e2e_test]` tests for the `contract-transfer` example. 

I've annotated a couple of `TODO(#xxx)` comments ‒ I'll create the issues + add the number before merging, I'm just waiting for approvals beforehand.


crates/env/src/engine/e2e/xts.rs
```diff
@@ -0,0 +1,360 @@
...
+            gas_limit: DRY_RUN_GAS_LIMIT.into(),
+            storage_deposit_limit: storage_deposit_limit.map(|l| {
+                l.try_into()
+                    .expect("unable to convert `storage_deposit_limit`")
+            }),
+            input_data: Bytes(data),
+        };
+        let params = rpc_params![call_request];
+        self.ws_client
+            .request("contracts_call", params)
```
ascjones, line 293: The dry run should use the `state_call` rpc, since I believe these RPCs will be deprecated (cc @athei). See https://github.com/paritytech/cargo-contract/pull/701

crates/env/src/engine/e2e/xts.rs
```diff
@@ -0,0 +1,360 @@
...
+                log_info(&format!("signed and submitted call with extrinsic hash {:?}", tx_progress.extrinsic_hash()));
+                tx_progress
+            })
+            .unwrap_or_else(|err| {
+                panic!(
+                    "error on call `sign_and_submit_then_watch_default`: {:?}",
+                    err
+                );
+            })
+            // TODO(#xxx) It should be configurable to use `.wait_for_finalized_success` instead.
```
ascjones, line 342: Since this is for testing I don't think waiting for finalization is necessary?

crates/env/Cargo.toml
```diff
@@ -50,6 +50,27 @@ secp256k1 = { version = "0.24", features = ["recovery", "global-context"], optio
...
+
+# Substrate
+sp-rpc = "6.0.0"
+sp-core = "6.0.0"
+sp-keyring = "6.0.0"
+sp-runtime = "6.0.0"
+
+# TODO(#xxx) `smart-bench_macro` needs to be forked.
+smart-bench-macro = { git = "https://github.com/paritytech/smart-bench", branch = "cmichi-ink-e2e-test-mvp", package = "smart-bench-macro"}
+
```
ascjones, line 73: Do we really want to add all these dependencies to the `ink_env` crate? 

This might even merit a standalone so that if you don't want to use it then it won't bring in all these dependencies.

crates/env/src/engine/e2e/xts.rs
cmichi: Hmm, so I agree, but I'm not sure that in the future if we merge XCM into ink! and have the e2e for it, that cross-chain testing will work without waiting for finalization. But I guess we can figure it out then!

crates/env/src/engine/e2e/xts.rs
cmichi: More involved than I thought (multiple repos), I've added a ToDo comment, will create a follow-up issue.

crates/env/Cargo.toml
cmichi: A standalone crate will unfortunately have to depend on `ink_env`, due to the types (e.g. the `Environment` trait). This would introduce a cyclic dependency.

cmichi:
CI currently fails because https://github.com/paritytech/cargo-contract/pull/728 is not yet merged.

crates/env/src/engine/e2e/xts.rs
athei: Why is it involved? You add new code. You can use `state_call` as well here.

cmichi: Missing `Encode` derive in `smart-bench` and `submit`.

athei: On which types? AFAIK there is no support for state call in subxt. You need to define them in this crate anyways (as you have done for the RPC arguments).

cmichi: Migrated it to `state_call` now.

athei: Thanks for not adding new users to the deprecated API.

crates/env/src/engine/e2e/xts.rs
```diff
@@ -0,0 +1,343 @@
...
+    },
+    OnlineClient,
+};
+
+/// The gas limit for contract instantiate and call dry runs.
+const DRY_RUN_GAS_LIMIT: u64 = 500_000_000_000;
+
+// TODO(#xxx) Should be fetched automatically.
+#[subxt::subxt(runtime_metadata_path = "metadata/contracts-node.scale")]
+pub(super) mod api {}
```
athei, line 52: I don't see that being used anywhere.

crates/env/src/engine/e2e/xts.rs
```diff
@@ -0,0 +1,343 @@
...
+pub struct InstantiateWithCode<B> {
+    #[codec(compact)]
+    value: B,
+    #[codec(compact)]
+    gas_limit: crate::types::Gas,
+    storage_deposit_limit: Option<B>,
+    code: Vec<u8>,
+    data: Vec<u8>,
+    salt: Vec<u8>,
+}
```
athei, line 65: Shouldn't you use the generated data from the subxt macro here?

athei (commented):
You can't really collect the output of a contract call with this, right? Because you only dry-run for gas_limit but do not allow to explicitly dry-run for calling non mutable massages.

crates/env/src/engine/e2e/xts.rs
cmichi: This is the recommended way of initializing `subxt` in a way that it generates the runtime types from the `*.scale` file: https://docs.rs/subxt/latest/subxt/#generating-runtime-types.

crates/env/src/engine/e2e/xts.rs
cmichi: By generated data you mean. that the `InstantiateWithCode` struct should be available in some generated form from `subxt`?

crates/env/src/engine/e2e/xts.rs
athei: Yes but you are not using it to generate the transaction. Instead you are doing this manually.

crates/env/src/engine/e2e/xts.rs
athei: Yes. This is the whole point of reading in the metadata. Use `api::tx().contracts().call()` to create your `StaticTxPayload`.

crates/env/src/engine/e2e/xts.rs
athei: Yes let's not worsen the status quo for some future stuff.

crates/env/src/engine/e2e/xts.rs
ascjones: Yes this would be the ideal way to do it. However I believe you would need to access the non-public `call_data` field for this, since [`StaticTxPayload::encode_call_data`](https://github.com/paritytech/subxt/blob/5ff849318b75ba32ca905ec9d34cc1d16a858b31/subxt/src/tx/tx_payload.rs#L81) prepends the pallet and call indices.

I believe there is a `subxt` release coming shortly, so you should get this in quickly.

crates/env/src/engine/e2e/xts.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        Self {
+            client,
+            ws_client,
+            _phantom: Default::default(),
+        }
+    }
+
+    /// Dry runs the instantiation of the given `code`.
+    pub async fn instantiate_with_code_dry_run(
```
SkymanOne, line 160: This looks to me very similar to the same function definition in `cargo-contract`. Wouldn't it be good to import this method from `cargo-contract` for the sake of single point of truth and DRY?

crates/env/src/engine/e2e/xts.rs
```diff
@@ -0,0 +1,343 @@
...
+            });
+        scale::Decode::decode(&mut bytes.as_ref()).expect("decoding failed")
+    }
+
+    /// Submits an extrinsic to instantiate a contract with the given code.
+    ///
+    /// Returns when the transaction is included in a block. The return value
+    /// contains all events that are associated with this transaction.
+    #[allow(clippy::too_many_arguments)]
+    pub async fn instantiate_with_code(
```
SkymanOne, line 196: Same here as above

crates/env/src/engine/e2e/xts.rs
```diff
@@ -0,0 +1,343 @@
...
+            .unwrap_or_else(|err| {
+                panic!(
+                    "error on call `fetch_events`: {:?}",
+                    err
+                );
+            })
+    }
+
+    /// Dry runs a call of the contract at `contract` with the given parameters.
+    pub async fn call_dry_run(
```
SkymanOne, line 256: Same

crates/env/src/engine/e2e/xts.rs
```diff
@@ -0,0 +1,343 @@
...
+                panic!("error on ws request `contracts_call`: {:?}", err);
+            });
+        scale::Decode::decode(&mut bytes.as_ref()).expect("decoding failed")
+    }
+
+    /// Submits an extrinsic to call a contract with the given parameters.
+    ///
+    /// Returns when the transaction is included in a block. The return value
+    /// contains all events that are associated with this transaction.
+    pub async fn call(
```
SkymanOne, line 287: Same as above

SkymanOne (commented):
These are my thoughts. I understand that we don't want to loosely couple `cargo-contract` and `ink`, but I personally see some code duplication as an issue because we'll have to change the code in 2 different places if specs are changed.

crates/env/src/engine/e2e/xts.rs
cmichi: After implementing it I ran into a problem that makes it currently not possible to use the from-metadata-generated `InstantiateWithCode` & co structs. The reason is that those structs have a well-defined type, e.g. `type Balance = u128`. 

My E2E methods on the other hand are generalized, they just require `value: <E as Environment>::Balance`. In order to make my testing methods work with the generated structs I would have to require `value: u128` hardcoded. I don't want to do this though, as it would limit the E2E testing framework to support only chains with `value: u128`. 

I've discussed it with Andrew and we agree that the best option is to leave as is.

I still need to generate the types via `#[subxt::subxt(runtime_metadata_path = "metadata/contracts-node.scale")]` though, as they are required to determine:
```rust
.as_event::<xts::api::system::events::ExtrinsicFailed>()
```

crates/env/src/engine/e2e/xts.rs
cmichi: Thanks for flagging it. My view is that in most cases DRY makes total sense, but as you already noted we would couple `cargo-contract` and ink! quite closely together by introducing a dependency here. I would rather have clear boundaries between the two for now.

cmichi:
> You can't really collect the output of a contract call with this, right? Because you only dry-run for gas_limit but do not allow to explicitly dry-run for calling non mutable massages.

@athei Can't fully follow you here, can you elaborate?

cmichi:
> > > You can't really collect the output of a contract call with this, right? Because you only dry-run for gas_limit but do not allow to explicitly dry-run for calling non mutable massages.
> > 
> > 
> > @athei Can't fully follow you here, can you elaborate?
> 
> It looks to me that you are not returning the result (as in the output buffer) to the user of the testing framework.

The return value of invoking `call()` on the E2E testing framework is a `Result` with an `Ok` of:

```rust
/// Result of a contract call.
pub struct CallResult<C: subxt::Config, E: Environment> {
    /// The result of the dry run, contains debug messages
    /// if there were any.
    pub dry_run: ContractExecResult<E::Balance>,
    /// Events that happened with the contract instantiation.
    pub events: TxEvents<C>,
}
```

The `dry_run` here is all that the `state_call` RPC for `ContractsApi_call` returns. This is how `CallResult.dry_run` looks:

```
ContractResult {
        gas_consumed: 5501174466,
        gas_required: 74999922688,
        storage_deposit: StorageDeposit::Charge(0),
        debug_message: [],
        result: Ok(
                ExecReturnValue {
                        flags: (empty),
                        data: Bytes([0, 57, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
                }
        )
}
```

This should cover anything, right?

athei (approved):

athei:
> So all of this is returned to the user as part of executing `.call()`. This should cover anything, right?

yes

crates/env/Cargo.toml
```diff
@@ -50,6 +50,26 @@ secp256k1 = { version = "0.24", features = ["recovery", "global-context"], optio
...
+tokio = { version = "1.18.2", features = ["rt-multi-thread"] }
+log = "0.4"
+env_logger = "0.8"
+subxt = "0.24.0"
+
+# Substrate
+sp-rpc = "6.0.0"
+sp-core = "6.0.0"
+sp-keyring = "6.0.0"
+sp-runtime = "6.0.0"
```
ascjones, line 68: I think all of these new e2e dependencies should be `optional = true`, since the `e2e` code is conditional for `std`.

Arguably we may even want an additional feature to prevent them being compiled for the metadata generation which is `std`, there are some hefty dependencies here.

crates/env/Cargo.toml
```diff
@@ -50,6 +50,26 @@ secp256k1 = { version = "0.24", features = ["recovery", "global-context"], optio
...
+
+# TODO(#xxx) `smart-bench_macro` needs to be forked.
+smart-bench-macro = { git = "https://github.com/paritytech/smart-bench", branch = "cmichi-ink-e2e-test-mvp", package = "smart-bench-macro"}
+
```
ascjones, line 72: I think all of these new e2e dependencies should be `optional = true`, since the `e2e` code is conditional for `std`.

Arguably we may even want an additional feature to prevent them being compiled for the metadata generation which is `std`, there are some hefty dependencies here.

PR #1399, opened 2022-09-16
overflowCheck

andropixels:
while adding a by:i32  to the value:i32  it will be good to add them securely with checked_add() method !

### DIFF
--- a/examples/incrementer/lib.rs
+++ b/examples/incrementer/lib.rs
@@ -20,7 +20,7 @@ mod incrementer {
 
         #[ink(message)]
         pub fn inc(&mut self, by: i32) {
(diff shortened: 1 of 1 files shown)


shawntabrizi:
@andropixels you should read the information here: https://ink.substrate.io/faq/#overflow-safety

> Being written in Rust, ink! can provide compile-time overflow/underflow safety. Using a Rust compiler configuration, you can specify whether you want to support overflowing math, or if you want contract execution to panic when overflows occur.

So, AFAIK, all math in ink! is already "safe", and does not need such explicit checks.

andropixels:
Thanks for the information  ser !!

PR #1400, opened 2022-09-16
Move storage derives to `ink` crate

ascjones:
As discussed in #1223, one problem is that the storage traits derive macros  (`Storable`, `StorableHint`, `StorageKey`, `StorageLayout`) generate code which assumes the top level `ink` crate is in scope e.g. `::ink::storage::traits::StorageKey`. Therefore if a user were to import the standalone `ink_storage` crate and use those derives, it would not work.

This PR solves that by moving the storage derives into the `ink` crate itself, ensuring it is only possible to use those derives together with the `ink` crate.


crates/ink/ir/src/ir/utils.rs
```diff
@@ -144,3 +144,24 @@ where
         name
     ))
 }
+
+/// Finds the salt of a struct, enum or union.
+/// The salt is any generic that has bound `StorageKey`.
+/// In most cases it is parent storage key or auto-generated storage key.
```
cmichi, line 150:
```suggestion
/// In most cases it is the parent storage key or the auto-generated storage key.
```

cmichi (approved):

xiyu1984:
Hey @ascjones ,
Where is the `::ink_storage::traits::push_spread_root::<Self>(self, &root_key);` in 4.0.0-alpha.3? 
I have tried `::ink::storage::traits::push_spread_root::<Self>(self, &root_key);` but still not found.

image

Are there any suggestions? Thanks~

ascjones:
Hi @xiyu1984, those methods and traits have been removed as part of the storage rework: see the release notes: https://github.com/paritytech/ink/blob/master/CHANGELOG.md#storage-rework

xiyu1984:
> Hi @xiyu1984, those methods and traits have been removed as part of the storage rework: see the release notes: https://github.com/paritytech/ink/blob/master/CHANGELOG.md#storage-rework

Thanks a lot. My use case is that I need a re-entry call to a smart contract. That is,  Contract A calls contract B, and then contract B calls back to contract A to update a state(a member of contract A). In the last version of ink, I use `push_spread_root` and `pull_spread_root` to make sure the state in contract A is updated. The related code can be seen [here](https://github.com/dantenetwork/protocol-stack-for-ink/blob/fddb9cbd5e634aa1964fe9bc1231a9ca3f172b32/contracts/callee/lib.rs#L155). Are there any suggestions of how to implement it currently?

xgreenx:
@xiyu1984 I can't compile it because I don't have a `payload` crate, but the fix should look like this https://github.com/dantenetwork/protocol-stack-for-ink/pull/10

xiyu1984:
> @xiyu1984 I can't compile it because I don't have a `payload` crate, but the fix should look like this [dantenetwork/protocol-stack-for-ink#10](https://github.com/dantenetwork/protocol-stack-for-ink/pull/10)

Yes, it works~ Thank you very much~ I remember that I solved the re-entry problem last time with the help of you. So, thanks again!

PR #1401, opened 2022-09-16
Allow `pay_with_call` to take multiple arguments

HCastano:
We were using the wrong metavariable when accepting function arguments to the
`pay_with_call` macro, as reported [here](https://github.com/paritytech/ink/pull/1379#issuecomment-1248008357).


HCastano:
Looks like Codecov is broken on `master` too 😒

crates/ink/tests/ui/pay_with_call/pass/multiple_args.rs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        #[ink(message)]
+        pub fn message0(&self) {}
+
+        #[ink(message)]
+        pub fn message1(&self, _arg1: u8) {}
+
+        #[ink(message)]
+        pub fn message2(&self, _arg1: u8, _arg2: u8) {}
```
agryaznov, line 19:
```suggestion
        pub fn message2(&self, _arg1: u8, _arg2: (u8, String)) {}
```

crates/ink/tests/ui/pay_with_call/pass/multiple_args.rs
```diff
@@ -0,0 +1,29 @@
...
+
+        fn check_compiles(&self) {
+            ink::env::pay_with_call!(self.message0(), 0);
+            ink::env::pay_with_call!(self.message1(0), 0);
+            ink::env::pay_with_call!(self.message2(0, 0), 0);
```
agryaznov, line 24:
```suggestion
            ink::env::pay_with_call!(self.message2(0, (255, "something")), 0);
```

just a suggestion to make 2nd param of some structural type

agryaznov, line 24:
```suggestion
            ink::env::pay_with_call!(self.message2(4*64, 0), 0);
```
just to make sure it really works with expr; fun fact: it will compile regardless the value of the expr is out of bounds

agryaznov (approved):

crates/ink/tests/ui/pay_with_call/pass/multiple_args.rs
HCastano: I don't get what you mean here, can you clarify?

agryaznov: nevermind, just wanted to pass it a lil more interesting expr than just `0`; `4*64` evaluates to `256` which is out of `u8` bounds; still this would compile, while passing just `256` - not

PR #1402, opened 2022-09-16
Automated money market example

DoubleOTheven:
Here the bug I found related to `Mapping`. It appears that the `K` in `Mapping<K, V>` does not get scoped to the variable and is used as the storage key instead. What happens is that any variable using mapping will have its data overwritten if you use the same key - `self.env().caller()`, for example.

`examples/amm/lib.rs:299` has a test for the `faucet()` function:

## Contract Function

```rs
        #[ink(message)]
        pub fn faucet(&mut self, amount_token1: Balance, amount_token2: Balance) {
            let token1 = self.token1_balance.get(&caller).unwrap_or(0);
            let token2 = self.token2_balance.get(&caller).unwrap_or(0);

            self.token1_balance
                .insert(&caller, &(token1 + amount_token1));
            self.token2_balance
                .insert(&caller, &(token2 + amount_token2));
       }
```

## Test

```rs
        #[ink::test]
        fn faucet_works() {
            let mut contract = Amm::new(0);
            contract.faucet(100, 200);

            assert_eq!(contract.balances(), (100, 200, 0));
        }
```

Test fails with the error:

```sh
test amm::tests::faucet_works ... FAILED

failures:

---- amm::tests::faucet_works stdout ----
thread 'amm::tests::faucet_works' panicked at 'assertion failed: `(left == right)`
  left: `(200, 200, 200)`,
 right: `(100, 200, 0)`', lib.rs:303:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    amm::tests::faucet_works
```


123oli:
amm has been implemented in ink! by [figment.io](https://learn.figment.io/tutorials/build-polkadot-amm-using-ink)

DoubleOTheven:
Thanks, I am aware of that. I'm just sharing this PR with the team for a potential issue

examples/amm/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+        pub fn new(_fee: Balance) -> Self {
+            ink_lang::utils::initialize_contract(Self::new_init);
+
+            Self {
+                fee: if _fee >= 1000 { 0 } else { _fee },
+                ..Default::default()
+            }
+        }
+
+        fn new_init(&mut self) {}
```
HCastano, line 74: The reason your test fails is because of the incorrect usage of `initialize_contract`. Here you are returning a manually constructed instance of the contract, `Amm`, which initializes everything to its default value. This means all three `Mapping`'s in the `Amm` struct will have a default storage key of `0`.

```
[lib.rs:302] &contract = Amm {
    total_shares: 0,
    total_token1: 0,
    total_token2: 0,
    token1_balance: Mapping {
        offset_key: Key(0x_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000),
    },
    token2_balance: Mapping {
        offset_key: Key(0x_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000),
    },
    lp_shares: Mapping {
        offset_key: Key(0x_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000),
    },
    fee: 0,
}
```

What needs to be done instead is to use the instance of the contract returned by `initialize_contract`. You can see an example of how to do this correctly [here](https://ink.substrate.io/datastructures/mapping/).

 ```suggestion
        pub fn new(_fee: Balance) -> Self {
            ink_lang::utils::initialize_contract(|contract: &mut Self| {
                contract.fee = if _fee >= 1000 { 0 } else { _fee };
            })
        }
```

One more note: the examples are still using `alpha.1` - we should probably release `alpha.2` which removes the need for `SpreadAllocate` and `initialize_mapping` which would make this issue irrelevant

File: examples/amm/Cargo.toml [hunk 2]
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "amm"
+version = "0.1.0"
+authors = ["[your_name] <[your_email]>"]
```
HCastano, line 4: Can you copy the fields from the other examples here?

examples/amm/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+edition = "2021"
+
+[dependencies]
+ink_primitives = { version = "4.0.0-alpha.1", default-features = false }
+ink_metadata = { version = "4.0.0-alpha.1", default-features = false, features = ["derive"], optional = true }
+ink_env = { version = "4.0.0-alpha.1", default-features = false }
+ink_storage = { version = "4.0.0-alpha.1", default-features = false }
+ink_lang = { version = "4.0.0-alpha.1", default-features = false }
```
HCastano, line 12: We should be using `master` here, not releases

examples/amm/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+crate-type = [
+	"cdylib",
+]
+
+# This is necessary to get things to compile. Without it the compiler throws an error. 
+# Setting overflow checks to "false" for an AMM is typically not a good idea...
+# "This typically goes back to a known bug in the Rust compiler:
+# https://github.com/rust-lang/rust/issues/78744"
+[profile.release.package.compiler_builtins]
+overflow-checks = false
```
HCastano, line 29: I thought we fixed this ages ago, are you sure it's still needed?

examples/amm/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+        lp_shares: Mapping<AccountId, Balance>,
+        fee: Balance,
+    }
+
+    #[ink(impl)]
+    impl Amm {
+        fn has_valid_amount(
+            &self,
+            balance: &Mapping<AccountId, Balance>,
+            _qty: Balance,
```
HCastano, line 41: We don't use the same underscore convention as in Solidity - the Rust compiler will make it clear if you're using the variable in the wrong context. Rust also uses [underscores in a different context](https://stackoverflow.com/questions/48361537/why-do-underscore-prefixed-variables-exist#48361729), which makes it confusing for not only the compiler but also other Rust developers

examples/amm/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+        total_token2: Balance,
+        token1_balance: Mapping<AccountId, Balance>,
+        token2_balance: Mapping<AccountId, Balance>,
+        lp_shares: Mapping<AccountId, Balance>,
+        fee: Balance,
+    }
+
+    #[ink(impl)]
+    impl Amm {
+        fn has_valid_amount(
```
HCastano, line 38: Can you move all these internal helpers after the `message` implementations?

examples/amm/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+
+            self.token1_balance
+                .insert(&caller, &(token1 + amount_token1));
+            self.token2_balance
+                .insert(&caller, &(token2 + amount_token2));
+        }
+
+        /// Returns the balances of the user
+        #[ink(message)]
+        pub fn balances(&self) -> (Balance, Balance, Balance) {
```
HCastano, line 90: It's unclear what the return types are here. Can you do one/all of the following:
- Pack them into a struct with named fields
- Add typedefs (e.g `type Token1 = Balance`)
- Add doc comments to the function

examples/amm/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+            let caller = self.env().caller();
+            let t1 = self.token1_balance.get(&caller).unwrap_or(0);
+            let t2 = self.token2_balance.get(&caller).unwrap_or(0);
+            let my_shares = self.lp_shares.get(&caller).unwrap_or(0);
+            (t1, t2, my_shares)
+        }
+
+        /// Returns the amount of tokens locked in the pool,total shares issued & trading fee param
+        #[ink(message)]
+        pub fn pool_details(&self) -> (Balance, Balance, Balance, Balance) {
```
HCastano, line 100: Same thing here

examples/amm/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+            &mut self,
+            amount_token1: Balance,
+            amount_token2: Balance,
+        ) -> Result<Balance, Error> {
+            self.has_valid_amount(&self.token1_balance, amount_token1)
+                .unwrap();
+            self.has_valid_amount(&self.token2_balance, amount_token2)
+                .unwrap();
+
+            let share;
```
HCastano, line 120: Can you initialize this before usage?

examples/amm/lib.rs
```diff
@@ -0,0 +1,373 @@
...
+        #[ink(message)]
+        pub fn provide_liquidity(
+            &mut self,
+            amount_token1: Balance,
+            amount_token2: Balance,
+        ) -> Result<Balance, Error> {
+            self.has_valid_amount(&self.token1_balance, amount_token1)
+                .unwrap();
+            self.has_valid_amount(&self.token2_balance, amount_token2)
+                .unwrap();
```
HCastano, line 118: Make sure to handle these errors correctly instead of unwrapping. Ideally a `Result` is returned from the function, this lets callers potentially handle any error.

No `unwrap`s should be left in the example by the end

examples/amm/lib.rs
```diff
@@ -0,0 +1,373 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
```
HCastano, line 2: Can you add some documentation explaining the example? E.g a high level overview of what an AMM is. These examples are used as learning resources and should have good documentation throughout

HCastano (commented):
Quick review since you mentioned it's still a WIP

HCastano:
Quick note on the `dont-merge` tag. We typically use it when something has been approved and _could_ presumably be merged but shouldn't, or in CI related issues (e.g we just want to trigger a CI run)

HCastano:
> amm has been implemented in ink! by [figment.io](https://learn.figment.io/tutorials/build-polkadot-amm-using-ink)

@123oli it's fine for there to be multiple implementations of the same thing, especially as we explore this ecosystem and design space

examples/amm/Cargo.toml
DoubleOTheven: I get this error without it:

```sh
ERROR: Validation of the Wasm failed.


ERROR: An unexpected panic function import was found in the contract Wasm.
This typically goes back to a known bug in the Rust compiler:
https://github.com/rust-lang/rust/issues/78744

As a workaround try to insert `overflow-checks = false` into your `Cargo.toml`.
This will disable safe math operations, but unfortunately we are currently not
aware of a better workaround until the bug in the compiler is fixed.


ERROR: An unexpected import function was found in the contract Wasm: _ZN4core9panicking5panic17hca097ff786771598E.
The only allowed import functions are those starting with one of the following prefixes:
seal, memory
```

jolow99:
@123oli and the figment example isn't updated to be compatible with latest version of Ink with only the mapping data structures

jolow99:
@HCastano is there an estimated release date for alpha.2?

HCastano:
@jolow99 you can find the latest ink! release [here](https://github.com/paritytech/ink/releases/tag/v4.0.0-alpha.3)

HCastano:
@DoubleOTheven can you avoid squashing and force-pushing commits to a PR branch? As a reviewer it makes it hard to track what has or hasn't been fixed. It's much nicer to be able to go through commits. We end up squashing the entire PR when we merge it to `master` anyways, so it doesn't matter if the commit history is overly verbose

examples/amm/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+//! or as a Liquidity Provide (LP), //! who is financially incentivised to deposit equal amounts of token1 and token2. Each time
+//! a token is swapped a small percentage is given to all the LPs, and each share is proportional to the amount of liquidity
+//! that a LP provides. Pretty simple, right? You can learn more about AMMs [here](https://www.coindesk.com/learn/2021/08/20/what-is-an-automated-market-maker/).
+//!
+//! This code example was inspired by [this blog](https://learn.figment.io/tutorials/build-polkadot-amm-using-ink).
+//!
+//! ## Warning
+//!
+//! This contract is an *example*. It is neither audited nor endorsed for production use.
+//! Do **not** rely on it to keep any
```
HCastano, line 14: This is cut off

examples/amm/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+//! ## Tokens and AMMs
+//!
+//! If you are familiar with Solidity you may have worked with AMM contracts that integrate multiple ERC20 contracts for
+//! token pairs. This is a composable contract model. In this code example, however, we demonstrate a more simple version where all tokens
+//! are maintained by this contract (one contract instead of many composed contracts).
+//!
+//! ## The Faucet
+//!
+//! There is a `faucet()` function built in to this contract to fund your wallet balance. You can use this to //! test dive the features
+//! in a staging environment, and in tests.
```
HCastano, line 25: Can you also add a section about the workflow and usage? E.g, first you must add liquidity to a pool, then you can swap, et.c

examples/amm/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+//!
+//! ## Error Handling
+//!
+//! Any function that modifies the state returns a `Result` type and does not changes the state
+//! if the `Error` occurs.
+//! The errors are defined as an `enum` type. Any other error or invariant violation
+//! triggers a panic and therefore rolls back the transaction.
```
HCastano, line 32: We don't need this

examples/amm/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+    pub struct Amm {
+        total_lp_tokens: LPTokens,
+        total_token1: Token1,
+        total_token2: Token2,
+        token1_balances: Mapping<AccountId, Token1>,
+        token2_balances: Mapping<AccountId, Token2>,
+        lp_token_balances: Mapping<AccountId, LPTokens>,
+        fee: Balance,
+    }
+
```
HCastano, line 70: You're gonna need to add some events as well, otherwise the UI development will end up being hard

examples/amm/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+            self.total_lp_tokens += lp_tokens_for_deposit;
+
+            let lp_token_balance = self.lp_token_balances.get(&caller).unwrap_or(0);
+            self.lp_token_balances
+                .insert(&caller, &(lp_token_balance + lp_tokens_for_deposit));
+
+            Ok(lp_tokens_for_deposit)
+        }
+
+        /// Returns the amount of token1 a caller can receive in exchange for the given amount of token2
```
HCastano, line 199: Can you make sure the comments between the two estimate functions match?

examples/amm/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+            if lp_tokens > self.total_lp_tokens {
+                return Err(Error::InvalidLPAmount);
+            }
+
+            let amount_token1 = lp_tokens * self.total_token1 / self.total_lp_tokens;
+            let amount_token2 = lp_tokens * self.total_token2 / self.total_lp_tokens;
+            Ok((amount_token1, amount_token2))
+        }
+
+        /// This is used for LP to redeem the liquidity plus interest eared in exchange for LP tokens
```
HCastano, line 238:
```suggestion
        /// This is used for LP to redeem the liquidity plus interest earned in exchange for LP tokens
```

examples/amm/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+    #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+    pub struct PoolDetails {
+        total_token1: Token1,
+        total_token2: Token2,
+        total_lp_tokens: LPTokens,
+        fee: Balance,
+    }
+
+    impl PoolDetails {
+        pub fn default() -> Self {
```
HCastano, line 100: Can we not derive `Default` on `PoolDetails`?

examples/amm/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+/// This is a multiplier used to calculate the total LP Tokens a LP receives for providing liquidity to the pool
+const PRECISION: u128 = 1_000_000;
+
+#[ink::contract]
+mod amm {
+    use ink_storage::{traits::SpreadAllocate, Mapping};
+
+    #[derive(Debug, PartialEq, Eq, scale::Encode, scale::Decode)]
+    #[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+    pub enum Error {
```
HCastano, line 48: We can create a typedef for our `Result` in order to make handling errors more ergonomic

```
```

examples/amm/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+//! The errors are defined as an `enum` type. Any other error or invariant violation
+//! triggers a panic and therefore rolls back the transaction.
+//!
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use ink_lang as ink;
+
+/// This is a multiplier used to calculate the total LP Tokens a LP receives for providing liquidity to the pool
+const PRECISION: u128 = 1_000_000;
```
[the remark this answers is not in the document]
DoubleOTheven: It is arbitrary. I can have the caller inject that

examples/amm/lib.rs
DoubleOTheven: Nice. I love that

examples/amm/lib.rs
DoubleOTheven: Yes we can :)

examples/amm/lib.rs
```diff
@@ -0,0 +1,570 @@
...
+        /// The first deposit is crucial to get right because it sets up a mathematical relationship between the two
+        /// tokens assuming that they are of equal value.
+        #[ink(message)]
+        pub fn provide_liquidity(
+            &mut self,
+            token1_amount: Token1,
+            token2_amount: Token2,
+        ) -> Result<LPTokens, Error> {
+            self.has_balance(&self.token1_balances, token1_amount)?;
+            self.has_balance(&self.token2_balances, token2_amount)?;
```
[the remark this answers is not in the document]
DoubleOTheven: There is a check on line 245 that does exactly this.

examples/amm/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "amm"
+version = "4.0.0-alpha.1"
```
HCastano, line 3: We're on `alpha.3` now, so you'll probably need to merge `master`

examples/amm/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+crate-type = [
+	"cdylib",
+]
+
+# This is necessary to get things to compile. Without it the compiler throws an error.
+# Setting overflow checks to "false" for an AMM is typically not a good idea...
+# "This typically goes back to a known bug in the Rust compiler:
+# https://github.com/rust-lang/rust/issues/78744"
+[profile.release.package.compiler_builtins]
+overflow-checks = false
```
HCastano, line 26: You should be able to get rid of this now too, I imagine it part of your problems stemmed from being on the `3.x` series of releases

examples/amm/lib.rs
```diff
@@ -0,0 +1,725 @@
+//! ## Automated Money Market (AMM)
+//!
+//! An AMM is a decentralized marketplace to swap tokens. Usually swap fees are
+//! significantly cheaper than centralized exchanges, and this system allows anyone to
+//! participate as either the Swapper (a user interested in swapping one token for
+//! another) or as a Liquidity Provide (LP), //! who is financially incentivised to
```
HCastano, line 6:
```suggestion
//! another) or as a Liquidity Provide (LP), who is financially incentivized to
```

examples/amm/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+//! If you are familiar with Solidity you may have worked with AMM contracts that
+//! integrate multiple ERC20 contracts for token pairs. This is a composable contract
+//! model. In this code example, however, we demonstrate a more simple version where all
+//! tokens are maintained by this contract (one contract instead of many composed
+//! contracts).
+//!
+//! ## The Faucet
+//!
+//! There is a `faucet()` function built in to this contract to fund your wallet balance.
+//! You can use this to //! test dive the features in a staging environment, and in tests.
```
HCastano, line 30:
```suggestion
//! You can use this to test dive the features in a staging environment, and in tests.
```

examples/amm/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+    pub struct Amm {
+        total_lp_tokens: LPTokens,
+        total_token1: Token1,
+        total_token2: Token2,
+        token1_balances: Mapping<AccountId, Token1>,
+        token2_balances: Mapping<AccountId, Token2>,
+        lp_token_balances: Mapping<AccountId, LPTokens>,
+        fee: Balance,
+        precision: u128,
+    }
```
HCastano, line 66: In Rust the doc comment convention is different than this. Here are a few resources about it:
- https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#making-useful-documentation-comments
- https://doc.rust-lang.org/stable/rust-by-example/meta/doc.html
- https://rust-lang.github.io/api-guidelines/documentation.html

```suggestion
    /// An automated market maker  (AMM) contract.
    #[derive(Default)]
    #[ink(storage)]
    pub struct Amm {
        /// The aggregate total of all liquidity provider tokens created or burned
        total_lp_tokens: LPTokens,
        /// ... 
        total_token1: Token1,
        /// ...
        total_token2: Token2,
        token1_balances: Mapping<AccountId, Token1>,
        token2_balances: Mapping<AccountId, Token2>,
        lp_token_balances: Mapping<AccountId, LPTokens>,
        fee: Balance,
        precision: u128,
    }
```

examples/amm/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+        InvalidLPAmount,
+        NonEquivalentAmount,
+        SlippageExceeded,
+        ThresholdNotReached,
+        ZeroAmount,
+        ZeroLiquidity,
+    }
+
+    /// We create an ergonomic helper so we can do things like Result<Balance> instead of
+    /// Result<Balance, Error>
```
HCastano, line 96:
```suggestion
    /// The AMM contract's `Result` type.
    /// 
    /// We create this ergonomic helper so we can use `Result<Balance>` instead of
    /// `Result<Balance, Error>`.
```

examples/amm/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+    pub type Result<T> = core::result::Result<T, Error>;
+
+    /// Struct to hold Account balances
+    #[derive(Default, Copy, PartialEq, Eq, Debug, Clone, Decode, Encode)]
```
HCastano, line 100: We should be using the [`#[ink(storage_item)`](https://paritytech.github.io/ink/ink/attr.storage_item.html) macro anything that goes into storage instead of manually adding all these derives

examples/amm/lib.rs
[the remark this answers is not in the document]
HCastano: It's fine for it to be arbitrary, but we should at least mention that in the documentation along with any possible effects of increasing/decreasing the value. 

Since it's an example there's no need to make it configurable

examples/amm/lib.rs
```diff
@@ -184,14 +200,24 @@ mod amm {
         /// struct members the right way.
         #[ink(constructor)]
         pub fn new(fee: Balance, precision: u128) -> Self {
-            ink_lang::utils::initialize_contract(|contract: &mut Self| {
-                contract.fee = if fee >= 1000 { 0 } else { fee };
-                contract.precision = if fee < 1_000_000 {
+            Self {
+                fee: if fee >= 1000 { 0 } else { fee },
+                precision: if fee < 1_000_000 {
```
HCastano, line 205: As I mentioned, it's fine to hardcode the precision

examples/amm/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+        #[ink(topic)]
+        swap_amount: Token2,
+        #[ink(topic)]
+        swap_type: SwapType,
+    }
+
+    impl Amm {
+        // ==============================================================================
+        // 1. Constructor
+        // ==============================================================================
```
HCastano, line 191: Sorry, I didn't mean that we should label the sections - just having the messages in that order is enough

```suggestion
```

examples/amm/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+        /// The constructor takes two arguments: fee, and precision. fee it used to
```
HCastano, line 192:
```suggestion
        /// fee it used to
```

We can see this from the function signature 😉

HCastano, line 192:
```suggestion
        /// The constructor takes two arguments: fee, and precision. The `fee` is used to
```

examples/amm/lib.rs
```diff
@@ -0,0 +1,725 @@
...
+        /// calculate the swap fee to be given to liquidity providers (LPs). Precision is used
+        /// as a multiplier for more granular fractions of shares for LPs.
+        ///
+        /// NOTE:
+        /// We init this struct using a function. Simply calling
+        /// ink_lang::utils::initialize_contract(1000) will not create the correct storage
+        /// keys "under the hood" for tests. The function format is required to initialize
+        /// struct members the right way.
```
HCastano, line 200: This is outdated now

HCastano:
Gonna close this, free free to re-open if you start working on it again.

PR #1412, opened 2022-09-22
Update release checklist

ascjones:
First, a post-mortem of how I messed up the `4.0.0-alpha.2` release:

1. Following the existing release checklist, I bumped the versions to `alpha.2`
2. For the value of `[level]` running `cargo release`, I used the value `alpha`, following the docs of `cargo release --help` which suggests to use the pre release label. Our checklist was not clear as to what this value should be.
3. The effect of this was that it caused the `cargo release` tool to bump all the versions of the crates *again* to `alpha.3`, but crucially not updating the inter-workspace crate dependency versions. 
4. Had I checked the output of the dry run correctly I would have seen these proposed version bumps, but there was a lot of output so all I was looking for was the "Success" at the bottom as documented and didn't scroll up to see those bumps.
5. Then I ran the release for real, it performed the bumps to `alpha.3`, successfully published the "leaf" crates but then the dependent crates failed to publish because they still had `alpha.2` versions which were never published. Note that this is a case where the dry run succeeded but the publish itself failed.
6. To fix this I did an immediate new `alpha.3` release with the remaining crates.

This PR updates the release process to try and avoid this mistake, instructing the user to provide the **exact** version to be deployed. Note that it would be possible to use the publish process to bump inter workspace dependencies with `--dependent-version upgrade`, though this does not work as thoroughly as the existing step to simply find/replace versions.

Another important change is that instead of first merging to `master` and then releasing, we should first release from the PR branch and then merge. This is because it is possible (as above) for the dry run to succeed but the actual publish to fail and require changes. This requires "locking" the master branch so no commits happen while the release is happening, which I think is feasible in this project with only a handful of devs and not that often commits. However dependabot might be an issue... This is just to ensure that the code bundled to `crates.io` is the same as the code at the git release tag, so if there is a race condition then a special cherry-picked branch could be created and the release tag made from there.

### DIFF
--- a/RELEASES_CHECKLIST.md
+++ b/RELEASES_CHECKLIST.md
@@ -35,6 +35,10 @@ We'll be using [`cargo-release`](https://github.com/crate-ci/cargo-release) to r
 steps though, and we hope to make this more streamlined in the future.
 
 1. Create a new feature branch off `master`.
+1. Bump the version in all TOML files to the new version.
+    ```
+    find . -type f -name *.toml -exec sed -i -e 's/$OLD_VERSION/$NEW_VERSION/g' {} \;
+    ```
 1. Make sure you've moved the changes in the `CHANGELOG.md` from `[Unreleased]` into a new
    section for the release.
 1. Check that all notable PRs since the last release are now in the new release section,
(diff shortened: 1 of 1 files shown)


RELEASES_CHECKLIST.md
```diff
@@ -35,6 +35,8 @@ We'll be using [`cargo-release`](https://github.com/crate-ci/cargo-release) to r
 steps though, and we hope to make this more streamlined in the future.
 
 1. Create a new feature branch off `master`.
+2. Copy the release notes that appear in the [`CHANGELOG.md`](https://github.com/paritytech/ink/blob/master/CHANGELOG.md)
```
HCastano, line 38:
```suggestion
1. Copy the release notes that appear in the [`CHANGELOG.md`](https://github.com/paritytech/ink/blob/master/CHANGELOG.md)
```

RELEASES_CHECKLIST.md
```diff
@@ -49,19 +51,36 @@ steps though, and we hope to make this more streamlined in the future.
...
+        this will automatically bump and commit the changes to the `Cargo.toml` manifests. We have already done that in
+        an earlier step so it is not necessary.
     - We don't want `cargo-release` to create any releases or push any code, we'll do
        that manually once we've actually published to `crates.io`.
-1. If there are no errors, merge the release PR into `master`.
-1. Publish with `export PUBLISH_GRACE_SLEEP=5 && cargo release [level] -v --no-tag --no-push --execute`
+1. Check the output of the dry run:
+   - Does not show any automatic bumping of crate versions.
+   - Runs without error.
+1. Following a successful dry run, we can now publish to crates.io. This will be done from the release branch, since it
```
HCastano, line 65: Can you move the second and third sentences here down to another line? The first line should be an easy to digest "action item", and then we add more details after

RELEASES_CHECKLIST.md
```diff
@@ -35,6 +35,8 @@ We'll be using [`cargo-release`](https://github.com/crate-ci/cargo-release) to r
 steps though, and we hope to make this more streamlined in the future.
 
 1. Create a new feature branch off `master`.
+2. Copy the release notes that appear in the [`CHANGELOG.md`](https://github.com/paritytech/ink/blob/master/CHANGELOG.md)
+   into the PR description. This will cause the individual PRs to be linked to the release in which they are included.
```
HCastano, line 39: Same thing about action items here

HCastano (commented):
Can you also wrap everything to 90ish characters?

HCastano (approved):

Issue #1414, opened 2022-09-26
Consider add dlmalloc-rs as allocator or opt-out the BumpAllocator


kvinwang:
Since `wee_alloc` has been removed in https://github.com/paritytech/ink/pull/1403, BumpAllocator is the only choice now.
However, BumpAllocator is not suitable for our off-chain execution env, because it doesn't free any memory at all.

There are two options to resolve it in my mind:
- Support [dlmalloc-rs](https://github.com/alexcrichton/dlmalloc-rs), which is the allocator for wasm in std, as an opt-in replacement to the bump allocator.
- Add a feature `no_allocator` to crate `ink_allocator` to allow downstream to opt-out of the bump allocator.

HCastano:
@kvinwang can you give us a couple more details as to why a non-freeing allocator is a problem for your off-chain env?

kvinwang:
> @kvinwang can you give us a couple more details as to why a non-freeing allocator is a problem for your off-chain env?

In our case, we run the contract with off-chain workers inside TEE.
When a user calls a contract via RPC, it is not only for metering the gas but also for business serving. It can do much more heavy computing than the on-chain env. For example, we can do HTTP requests, parse the response, GET/PUT S3 storage, and make oracles in the contract. Those abilities may cause contracts to allocate and free heap memory frequently. If the allocator doesn't free memory at all, it would run out of memory soon.

tolak:
> @kvinwang can you give us a couple more details as to why a non-freeing allocator is a problem for your off-chain env?

And I found a new reason for this kind of requirement, see [comment in PR](https://github.com/paritytech/ink/pull/1415#issuecomment-1281854741)

kvinwang:
Another [issue](https://github.com/paritytech/ink/issues/1535) with the bump allocator. Should we consider this feature request?

HCastano:
@kvinwang yeah we should consider this.

We initially removed `wee_alloc` due to size constrains, but for off-chain execution that's a bit less of a concern. We can still bring it back and keep it behind a feature flag.

Can I ask you to compare the contract sizes of contracts which use a) our bump allocator b) `wee_alloc` and c) `dlmalloc-rs`?

kvinwang:
> Can I ask you to compare the contract sizes of contracts which use a) our bump allocator b) `wee_alloc` and c) `dlmalloc-rs`?

Sorry for the latency to reply due to my Covid-19.
I just compiled 3 versions with different allocators, and got result below:
```bash
[...]

PR #1415, opened 2022-09-26
Add feature no-allocator to ink_allocator

kvinwang:
Fixes https://github.com/paritytech/ink/issues/1414


tolak:
I was trying to use some types from substrate FRAME, which will lead to a `global_allocator` conflict with the one used ink, that makes me think about should we need an approach to enable/disable the `global_allocator` like this.

Issue #1417, opened 2022-09-27
New nightly Rust breaks the building of all contracts with ink! 


xgreenx:
The error came up with the latest nightly build of the rust. I guess after stabilization of generics in the associated types: https://github.com/rust-lang/rust/pull/96709

ink! generates the code like this to get the selectors of the methods inside the `match` statement:

<img width="833" alt="image" src="https://user-images.githubusercontent.com/18346821/192554566-ff5fef1e-c491-496f-b9e9-3bd7c42dc37e.png">

And it causes the compilation error. It can be fixed for a new version of ink! by moving this code from the `match` statement for each selector into a separate constant, and using those constants in the `match`. But all old versions of ink! should be compiled with old Rust(Or we need to patch each release with a minor fix).

PR #1418, opened 2022-09-27
Moved constants from `match` to a separate constants

xgreenx:
It fixes https://github.com/paritytech/ink/issues/1417. It can be used to patch all old releases of ink!.


ascjones (approved):

athei:
We need to backport to the v3.x.x and make a release.

PR #1424, opened 2022-09-28
E2E: Add issue numbers + make E2E dependencies optional

cmichi:
Follow-up to https://github.com/paritytech/ink/pull/1395.


ascjones (approved):
Thank you :pray:. The next improvement would be to make the contract metadata `std` build not build these dependencies. Haven't tested it yet but I assume this would take considerably longer now on the first build.

PR #1429, opened 2022-09-29
Extract e2e testing framework to its own crate

### LINKED ISSUE #1428: E2E: Include e2e deps only for `cargo test`
Follow-up to https://github.com/paritytech/ink/pull/1424.

At the moment we have a ton of heavy-weight `std` dependencies needed for `#[ink::e2e_test]` in [`crates/env/Cargo.toml`](https://github.com/paritytech/ink/blob/master/crates/env/Cargo.toml) (there's a comment indicating which ones). Those deps are only needed when executing `cargo test`. Unfortunately there is no way in `cargo` to conditionally include dependencies for `cfg(test)`. 

The problem is that `cargo-contract` for metadata generation builds the contract in `std` as well ([here is the manifest file for the metadata generation](https://github.com/paritytech/cargo-contract/blob/0e4a286e5de5d9e1685d52453a2750317569a002/crates/cargo-contract/templates/tools/generate-metadata/_Cargo.toml))

We need to find a way to include those E2E deps only for `cargo test`, but not for the metadata generation of `cargo-contract`. Right now the build time of `cargo contract build` (and also just `cargo build`) has exploded, because all those unnecessary E2E deps are build as well.

ascjones:
A WIP experiment for now just to see whether it is possible. If it is it will resolve #1428.

Pulls all of the e2e testing framework and the macro into their own crates.

- Keeps all e2e related code and dependencies together
- Only builds e2e dependencies for testing, prevents pulling in a big dependency tree for ink crates


cmichi:
Thanks a lot for this! I don't see how it will resolve https://github.com/paritytech/ink/issues/1428, but we should merge it nonetheless, as it's more structured this way.

I don't think it will resolve our issue because of the following:
* E.g. `flipper` includes the `ink` dependency.
* You execute `cargo test` on `flipper`.
* This will result in the direct `dev-dependencies` define in `flipper/Cargo.toml` being built.
* However: The `ink` crate will be built without the information that `flipper` is built in a developer/testing environment. So you have no way to know in the `ink` crate if it is being built for `dev-dependencies` or regular `dependencies`.
* This is unfortunately documented behavior ([here](https://doc.rust-lang.org/cargo/reference/specifyingdependencies.html#development-dependencies)): 
>Dev-dependencies are not used when compiling a package for building, but are used for compiling tests, examples, and benchmark. These dependencies are not propagated to other packages which depend on this package.

I think what we have to do is require every contract to have a `Cargo.toml` like this:

```
[dependencies]
ink = { version = "4.0", default-features = false }

[dev-dependencies]
ink = { version = "4.0", default-features = false, features = ["as-dev-dependency"] }
```

The we can have the `crates/ink/Cargo.toml` contain:
```
[features]
as-dev-dependency = [
   "ink_e2e"
]
```

ascjones:
> I don't see how it will resolve https://github.com/paritytech/ink/issues/1428

The idea is that `ink_e2e` is a standalone crate (not reexported by the main `ink` crate).
```
[dev-dependencies]
ink_e2e = { path = "../../crates/e2e" }
```

ascjones:
Waiting for https://github.com/paritytech/cargo-contract/pull/763 to be merged and released, then update `ink!` CI in this PR

examples/contract-transfer/lib.rs
```diff
@@ -183,18 +183,16 @@ pub mod give_me {
     #[cfg(test)]
     mod e2e_tests {
         use super::*;
-        use ink::env::e2e::*;
```
ascjones, line 186: This was previously required to bring the reexported `tokio` crate into scope for the `tokio::test` macro. In order to make this properly hygienic, I have replaced the usage of the `tokio::test` macro with explicit async execution (copied from `tokio::test`) https://github.com/paritytech/ink/pull/1429/commits/7e63d1ced263664f9a527a71d36c4d2e4bd9cd73

cmichi (approved):
Thank you, Andrew!

PR #1432, opened 2022-10-11
Implement cross-contract testing in E2E framework

cmichi:
Needed for https://github.com/paritytech/ink/issues/1234.

Can be reviewed commit-by-commit, though the PR is also not that big in total.


crates/e2e/macro/src/codegen.rs
```diff
@@ -190,3 +163,49 @@ impl InkE2ETest {
...
+fn build_contract(manifest_path: &str) -> String {
+    use std::process::{
+        Command,
+        Stdio,
+    };
+    let output = Command::new("cargo")
+        .args([
+            "+stable",
+            "contract",
+            "build",
```
ascjones, line 178: Since we control `cargo-contract` it might be really nice to have a lib to allow invocation of contract builds...would certainly simplify the code here.

Made a quick start on it here https://github.com/paritytech/cargo-contract/pull/787. 

If you agree it is a good idea I can complete that.

athei (approved):
Can we de-dup the metadata? Maybe just a soft link.

ascjones (approved):

PR #1442, opened 2022-10-19
Remove random function

athei:
We need to remove away from on-chain randomness. We cannot rely on hosting chains to provide unpredictable randomness. Like on ethereum that should be provided by oracles or maybe a chain extension of a chain has a way to provide randomness. It should not be part of the core API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Remove random function from ink!.
+
 ## Version 4.0.0-alpha.3
 
 ### Breaking Changes
--- a/crates/engine/Cargo.toml
+++ b/crates/engine/Cargo.toml
@@ -23,8 +23,6 @@ sha2 = { version = "0.10" }
 sha3 = { version = "0.10" }
 blake2 = { version = "0.10" }
 
-rand = { version = "0.8" }
-
 # ECDSA for the off-chain environment.
 secp256k1 = { version = "0.24", features = ["recovery", "global-context"], optional = true }
 
--- a/crates/engine/src/exec_context.rs
+++ b/crates/engine/src/exec_context.rs
@@ -17,9 +17,7 @@ use super::types::{
     Balance,
     BlockNumber,
     BlockTimestamp,
-    Hash,
 };
-use rand::Rng;
 
 /// The context of a contract execution.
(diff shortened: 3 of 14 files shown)


ascjones (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Remove random function from ink!.
```
cmichi, line 9:
```suggestion
### Removed
- Remove random function ‒ [#1442](https://github.com/paritytech/ink/pull/1442)
```

cmichi: Could you please add some more information in a section `### How to replace random` here?

My concern is that we're just removing it without providing guidance on how to substitute this functionality. 

Is there no pallet out there which we could create a chain extension with?

athei: On ethereum contracts need to look for randomness on their own anyways. They rely on oracle contracts AFAIK. If we had some reliable way to provide randomness in the general case we wouldn't remove it. I don't think there is anything at the moment.

@burdges  has some ideas but nothing readily available. So we can't really give any advice beyond "good luck bro".

forgetso: This pseudo random function was useful enough for our contract but now its removed. Any guidance on how to recreate this functionality would be appreciated.

forgetso: We need a random seed per block that is _good enough_ to randomly read a value from a vector within our contract. The operation is not high value and the seed calculated by the validators was fulfilling this purpose very well. Are the validators no longer committing this random seed to each block?

Could the current block hash be made available in `ink!` as a substitute?

athei: > We need a random seed per block that is _good enough_ to randomly _read_ a value from a vector within our contract. There is no transaction involved. The operation is not high value and the seed calculated by the validators was fulfilling this purpose very well. Are the validators no longer committing this random seed to each block?
> 
> Could the current block hash be made available in `ink!` as a substitute?

We would need **way** for information to give some advice and to determine if using predictable randomness is actually safe.  I recommend opening an issue on [stack exchange](https://substrate.stackexchange.com).

forgetso: I've provided more info on [stack exchange](https://substrate.stackexchange.com/questions/6745/how-can-i-generate-a-random-seed-now-that-self-env-random-is-removed-in-ink).

PR #1446, opened 2022-10-21
Add `Result<Self, Error>` return type for constructors

### LINKED ISSUE #988: Support Result<Self, E> returning ink! constructors
Currently it is not possible to return anything but `Self` in ink! constructors.
This is an artificial limitation created to make things a bit easier, however, technically as well as from a design perspective nothing speaks against introducing fallible ink! constructors.
They will behave in the same way to fallible ink! messages in that they will revert state in case `Result::Err` is returned.

Mainly `ink_lang_ir` as well as `ink_lang_codegen` crates are required to be modified in order to implement this feature.

After this change ink! constructors can either return `Self` or `Result<Self, E>` where `E` is some error type:

```rust
use ink_lang as ink;

#[ink::contract]
mod contract {
    #[ink(storage)]
    pub struct Contract {
        value: Balance,
    }

    enum Error {
        TooMuchInitValue,
    }

    impl Contract {
        #[ink(constructor)]
        fn new() -> Self {
            self.value = 0;
        }

        #[ink(constructor)]
        fn with_balance(balance: Balance) -> Result<Self, Error> {
            if balance >= 1000 {
                return Err(Error::TooMuchInitValue)
            }
            Self { value: balance }
        }
    }
}
```

SkymanOne:
This PR closes #988.

The PR also includes an example `return_err` to illustrate the usage:

# Addition
If we define storage and error similar to
```rust
/* Snippet */
#[ink(storage)]
pub struct ReturnErr {
    count: i32,
}

#[derive(Debug, PartialEq, Eq, scale::Encode, scale::Decode)]
#[cfg_attr(feature = "std", derive(::scale_info::TypeInfo))]
pub enum Error {
    Foo,
}
/* Snippet */
```
We can now create a constructor that can *safely* fail without the need to `panic!`
```rust
/* Snippet */
#[ink(constructor, payable)]
// can also return core::result::Result<T, Error> for explicitness
pub fn new2(fail: bool) -> Result<T, Error> {
    if fail {
        Err(Error::Foo)
    } else {
        Ok(Default::default())
    }
}
    /* Snippet */
```

Alternatively, you can define a type alias for convenience:
```rust
 /* Snippet */
#[ink(constructor, payable)]
pub fn new2(fail: bool) -> Result<Self> {
    if fail {
        Err(Error::Foo)
    } else {
        Ok(Default::default())
    }
}
/* Snippet */
```


examples/return_err/lib.rs
```diff
@@ -0,0 +1,74 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[ink::contract]
+mod return_err {
+
+    #[ink(storage)]
+    pub struct ReturnErr {
+        instantiated: bool,
```
athei, line 8: This field does not make sense. Because when you return `Result::Err(_)` the codegen should emit a call to `seal_return` with the revert flag set. Exactly the same behavior as for a message. So an `Err` implies `instantiated: false`.

examples/return_err/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    impl Default for ReturnErr {
+        fn default() -> Self {
+            Self { instantiated: true }
+        }
+    }
+
+    /// Type alias for the contract's result type.
+    pub type Result<T> = core::result::Result<T, Error>;
```
athei, line 25: I don't like this alias. It would make sense if you used this error in more than one occasion. But the error isn't an enum which suggests it is not to be used for more than one function. Yet, you are using the highly contended `Result` identifier for it.

athei (commented):
> We can not create a constructor that can safely fail without the need to panic!

I don't parse this sentence at all. Can you elaborate?


Also, why are we reliant on type aliases here while we aren't for messages? For messages we can just write any `Result<T, E>` as return type. For constructors you would just stricten that to `Result<Self, E>` and nothing more.

crates/ink/ir/src/ir/item_impl/constructor.rs
```diff
@@ -92,7 +92,34 @@ impl Constructor {
...
+            if let syn::PathArguments::AngleBracketed(angle_args) =
+                &path.segments.last().unwrap().arguments
+            {
+                if let Some(syn::GenericArgument::Type(ty)) = angle_args.args.first() {
+                    return Self::type_is_self_val(ty)
+                }
+            }
+        }
+        false
+    }
```
Robbepop, line 120: This looks as if you are checking the return type of an ink! constructor by checking if the identifier of the returned value type is equal to `"Result"`. Therefore you use syntactical checks instead of making use of the powerful Rust type system.

This violates the rules of proc. macro hygiene and is probably also the reason why this implementation depends on the `type Result` alias that @athei has already mentioned.

Let me be clear about this: ink! or any other Rust proc. macro that wants to be taken seriously should never make use of syntactical analysis for thing like this and instead should _always_ make use of the powerful Rust type system whenever it is possible and it is very well possible here and we discussed about this some time ago in chat.

If you need more elaboration please feel free to ask me and I will happily provide you with answers.

examples/return_err/lib.rs
SkymanOne: We can actually use `Result<Self, ()>` on its own without aliases. If we want to define a type alias for convenience, we need to explicitly provide a full path. I should have made it clearer.

examples/return_err/lib.rs
SkymanOne: Well, this has nothing to do with `Err` and others. We enforce to have messages in the contract, so I just came up with this storage as demo

SkymanOne:
> > We can not create a constructor that can safely fail without the need to panic!
> 
> I don't parse this sentence at all. Can you elaborate?
> 
> Also, why are we reliant on type aliases here while we aren't for messages? For messages we can just write any `Result<T, E>` as return type. For constructors you would just stricten that to `Result<Self, E>` and nothing more.

We aren't relying on type aliases at all. They are just used for convenience because I've seen them used in other examples like like DNS or ERC721. We can use `Result<Self, ()>` on its own

> > We can not create a constructor that can safely fail without the need to panic! 

This is a typo, should be: 

> We can **now** create a constructor that can safely fail without the need to panic!

crates/ink/ir/src/ir/item_impl/constructor.rs
SkymanOne: While I agree with the philosophy, it is a little bit tricky in this case to do:
* We have `is_result_type!` macro that check `Result` based on type. However, it sits in another crate, and I will need to import the whole crate for a single macro.
* Even though this might not be a problem. The real problem lies in `Self`. We still need to extract `Ok` variant and **syntacticly** check for `Self`. Otherwise, if we want to utilise the type system in this case too, we need to know the storage type and then compare it to the `Ok` variant return type. However, my understanding is that `ir` crate sits in isolation from the global `ink` crate and context and makes no assumptions what happens "*beyond the syntax*", it solely works with token stream. 
* Hence, that led me to thinking that we only want to ensure that `Result` keyword with `Self` are present in the token stream. The real type checks happen later down the pipeline when we generate dispatchable messages and metadata

athei:
> This is a typo, should be:
> 
> > We can **now** create a constructor that can safely fail without the need to panic!

Ahh lol yeah this really fried my brain.

> We aren't relying on type aliases at all. They are just used for convenience because I've seen them used in other examples like like DNS or ERC721. We can use Result<Self, ()> on its own

Good. I just misread the UI test I think.

examples/return_err/lib.rs
Robbepop: And does this work?

```rust
type Foo = Result<MyContract, MyError>;

#[ink(constructor, payable)]
pub fn another_new(fail: bool) -> Foo { ... }
```
or this?
```rust
pub trait Foo {
    type Bar;
}
impl Foo for MyContract {
    type Bar = Result<Self, MyError>;
}
impl MyContract {
    #[ink(constructor, payable)]
    pub fn another_new(fail: bool) -> <MyContract as Foo>::Bar { ... }
}
```
or does this actually not think that the following code represents a fallible ink! constructor?
```rust
type Result = MyContract;

#[ink(constructor, payable)]
pub fn another_new(fail: bool) -> Result { ... }
```
Rust proc. macro hygiene requires proc. macros to work in any semantic contexts regardless of the surrounding context. With a syntactical analysis you can never achieve this and generated code is extremely brittle and easily malicious.

SkymanOne: Nope the following examples above would not works because the only syntax we accept is `Result<Self, ...>`. Where the `Result` is coming from is not a concern at this stage because we are working with tokens. I can use `is_result_type!()` to also ensure that the `Result` is of the tight type.

SkymanOne: no, message return types are only checked against *not being `Self`*

athei: I mean with regard to `Result`. They also need to know whether you return a result to make sure that in case of an `Err` the proper flags are passed to `seal_return`.  Are they able to detect a `Result` behind an alias like asked for by @Robbepop ?

crates/ink/ir/src/ir/item_impl/constructor.rs
Robbepop: Syntactical checks are red flags for proc. macros generally. So yes.

examples/return_err/lib.rs
SkymanOne: I have not touched `Message`s  but to answer question - yes, they do support type aliasing and the examples above would works for messages

athei: Then this leads to the question: Why handle it differently here? The code that handles this seem to exist. Why not reuse it? It seems to be very similar.

SkymanOne: Because Message's have more permissive syntax checks? Because they can handle more return types by their nature (i.e. anything, but `Self`).

I also checked how messages handle `-> Result<Self, ()>` and while it doesn't compile (as expected). It doesn't give any meaningful error either. I understand that you all included towards type checking, but it doesn't really give us the same guarantees on the code correctness on the syntax level and affect the developer experience IMHO.

crates/ink/ir/src/ir/item_impl/constructor.rs
ascjones: One possible avenue of investigation to replace the syntactica checkl is the use of [`DispatchableConstructorInfo::Output`](https://github.com/paritytech/ink/blob/109536d43abab7d6eed5000c4b208b74c49aa106/crates/ink/src/reflect/dispatch.rs#L335) to constrain the return type to either be `Self::Storage` or `Result<Self::Storage>`. Might require an additional trait with auto impl for the `Result`, or maybe there is something already there we can reuse.

SkymanOne: > One possible avenue of investigation to replace the syntactica checkl is the use of [`DispatchableConstructorInfo::Output`](https://github.com/paritytech/ink/blob/109536d43abab7d6eed5000c4b208b74c49aa106/crates/ink/src/reflect/dispatch.rs#L335) to constrain the return type to either be `Self::Storage` or `Result<Self::Storage>`. Might require an additional trait with auto impl for the `Result`, or maybe there is something already there we can reuse.

That might be a good option. I was thinking of using an enum with to variants. `Raw(Storage)` and `Result(Storage)`

crates/ink/src/codegen/dispatch/execution.rs
```diff
@@ -146,37 +133,38 @@ pub trait ConstructorReturnType<C>: private::Sealed {
...
     }
 
     #[inline]
-    fn return_value(&self) -> &Self::ReturnValue {
-        &self.0
+    fn error_value(&self) -> &Self::Error {
+        self.0
+            .as_ref()
+            .err()
+            .expect("value returned only when error occurs. qed")
```
Robbepop, line 167: What exactly are we using this for? Could its uses be replaced by `as_result` instead? I don't like that it panics. When constructors were still infallible this API made sense but now I feel like it doesn't add too much value.

crates/ink/tests/ui/contract/fail/constructor-return-result-invalid.stderr
```diff
@@ -0,0 +1,14 @@
...
+   |         ^^^ the trait `codegen::dispatch::execution::ConstructorReturnType<Contract>` is not implemented for `codegen::dispatch::execution::private::Seal<Result<u8, contract::Error>>`
+   |
+   = help: the following other types implement trait `codegen::dispatch::execution::ConstructorReturnType<C>`:
+             codegen::dispatch::execution::private::Seal<C>
+             codegen::dispatch::execution::private::Seal<Result<C, E>>
+note: required by a bound in `execute_constructor`
+  --> src/codegen/dispatch/execution.rs
+   |
+   |     private::Seal<R>: ConstructorReturnType<Contract>,
+   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `execute_constructor`
```
Robbepop, line 14: Ideally those error messages are more user friendly and readable. For example `ConstructorReturnType` could be renamed to `ConstructorReturnsSelfOrResultOfSelf` or something like that.
Also `codegen::dispatch::execution::private::Seal<C>` could rename `C` to `Contract` or even try renaming it to `Self`? Same applies for the line below. I am not sure these are improvements but I am sure we can do better than the current state of error message reports.

Robbepop (commented):
Looks a lot better now!

crates/ink/ir/src/ir/item_impl/constructor.rs
```diff
@@ -421,31 +416,6 @@ mod tests {
...
-            syn::parse_quote! {
-                #[ink(constructor)]
-                pub fn my_constructor() -> Result<Self, ()> {}
-            },
-        ];
-        for item_method in item_methods {
-            assert_try_from_fails(item_method, "ink! constructors must return Self")
-        }
-    }
-
```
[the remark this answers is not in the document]
SkymanOne: This is already tested in UI tests

crates/ink/src/codegen/dispatch/execution.rs
SkymanOne: We only return a *value* back to the caller when the constructor fails. In case of constructors, the return value is encoded error, that's why we explicitly unwrap the error. Hence, it makes more sense to rename the function to `error_value()` because it errors the values back to the caller.

This was a [suggestion from Andrew](https://github.com/paritytech/ink/pull/1446#discussion_r1004316076)

crates/ink/tests/ui/contract/fail/constructor-return-result-invalid.stderr
SkymanOne: I personally see no benefit of putting longer name of a trait. The error message is still confusing and unclear.

My proposal is to add a lint for constructor. It will be a follow-up issue

crates/ink/src/codegen/dispatch/execution.rs
```diff
@@ -78,10 +75,10 @@ where
             //
             // We need to revert the state of the transaction.
             ink_env::return_value::<
-                <private::Seal<R> as ConstructorReturnType<Contract>>::ReturnValue,
+                <private::Seal<R> as ConstructorReturnType<Contract>>::Error,
             >(
                 ReturnFlags::default().set_reverted(true),
-                result.return_value(),
+                result.error_value(),
```
Robbepop, line 81: instead of `error_value` you could (and should) be using the `Err(error_value)` that you just matched against.

crates/ink/src/codegen/dispatch/execution.rs
Robbepop: See https://github.com/paritytech/ink/pull/1446#discussion_r1006621274.

crates/ink/tests/ui/contract/fail/constructor-return-result-invalid.stderr
Robbepop: I am okay with a proper lint, too, if it significantly improves error messages for smart contracts. I'd hope though that lints are kind of non-optional for ink! since what is the point then?

SkymanOne: Lints are options with https://github.com/paritytech/cargo-contract/pull/799

ascjones: We have made them optional for now since there currently are none, to avoid having to install `cargo dylint`. This can be switched back to make them compulsory again once we have lints again.

crates/ink/src/codegen/dispatch/execution.rs
ascjones: Yes it would be better just to match against the error value

crates/ink/tests/ui/contract/fail/constructor-return-result-invalid.stderr
Robbepop: Okay understood. Originally I had in mind that those ink! specific lints are mandator and viewed similar to Rust and its borrow checker. This allows to implement additional safety guards into ink! and also rely on them across ink! smart contract dependencies in the same way that Rust crates can depend on the safety guarantees that safe Rust brings with it.

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -256,16 +256,19 @@ impl Dispatch<'_> {
                 let payable = constructor.is_payable();
                 let selector_id = constructor.composed_selector().into_be_u32().hex_padded_suffixed();
                 let selector_bytes = constructor.composed_selector().hex_lits();
+                let output_tuple_type = constructor.output().map(quote::ToTokens::to_token_stream)
+                    .unwrap_or_else(|| quote! { () });
```
Robbepop, line 260:
```suggestion
                    .unwrap_or_else(proc_macro2::TokenStream::new);
```

Robbepop: have not checked if this is valid syntactically but looks cleaner to me

crates/ink/src/codegen/dispatch/execution.rs
```diff
@@ -119,65 +113,34 @@ pub trait ConstructorReturnType<C>: private::Sealed {
     ///
     /// # Note
     ///
-    /// For infallible constructors this is `core::convert::Infallible`.
+    /// For infallible constructors this is `()`.
```
Robbepop, line 116: I kinda dislike that this is now `()` for infallible errors since that means to me that we technically cannot have `()` as error type for ink! constructors. So
```rust
#[ink(constructor)]
fn foo() -> Result<Self, ()>
```
is not accepted? I would not argue that I'd like having `()` type as error type in particular but as proc. macro developers we should be generous with what input we accept. Therefore I strongly suggest having a special type (again) that indicates infallible ink! constructor "error" types defined in the `ink_lang` crate alongside other similar marker types.

SkymanOne: The code snippet you provided is valid and will result in successful build of a contract

SkymanOne: But I have returned `Infallible` back

SkymanOne: `Infallible` doesn't implement `Encode` so it can't be used as an error type

Robbepop (approved):
LGTM

crates/ink/tests/ui/contract/fail/constructor-return-result-invalid.stderr
ascjones: Absolutely. Once we have these checks then we can make the linting mandatory again.

examples/return_err/lib.rs
```diff
@@ -0,0 +1,75 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
ascjones, line 1: I think this belongs in the `mother` example, I don't think we need an example for each feature, this is covered by the `pass` UI tests.

ascjones (approved):
LGTM!

Issue #1207, opened 2022-04-01
Change return type in metadata to allow common errors


athei:
Currently, the metadata defines the return type of a message (or constructor) as exactly the type that is returned by the message itself.

This doesn't allow us to return any error that isn't defined by the contract itself. For example, the dispatch of a contract call to the individual messages is not handled by user code but by code generated by ink! macros. The way how the metadata is defined won't allow ink! to return an error in case a message isn't found (as the return type is defined by user code). The only thing ink! can emit in this case is a trap which prevents the caller from learning about a missing message.

What we should to is define the return type of messages (and constructors) in terms of an enum that allows us to define common errors which are defined within the ABI rather than within user contract code. As long as we only extend the enum and make it clear that one should [not match it exhaustively](https://rust-lang.github.io/rfcs/2008-non-exhaustive.html) it will be forward compatible:

```rust
enum Error<T> {
    Custom(T),
    UnknownMessage,
}
```

athei:
After thinking about it again a better approach would be to define the return type of a contract message/constructor like this:
```rust
type ContractResult<ContractReturn, LangErr> = Result<ContractReturn, LangErr>;
```

`ContractReturn` itself would be a `Result` for fallible contracts in case of of ink!. I think this approach is more flexible because it does allow for contracts to return any type and not only a result. It also doesn't impose a fixed list of lang errors to all languages. Any contract language can emit whatever they want for `LangErr`.

Additionally, this might be easier from a codegen standpoint than unifying everything into a single error enum.

PR #1450, opened 2022-10-25
Add support for language level errors (`LangError`)

HCastano:
Implements the Message related changes of #1207.

The idea here is that every message now returns a `Result<Message::Output, LangError>`,
where a `LangError` is a type of error which doesn't originate from the contract itself,
nor from the underlying execution environment (so the Contracts pallet in this case).

An example of where this would arise is if a caller tries to use a non-existent message
selector for a contract. Previously, the contract would trap and not allow the caller to
do any sort of error handling if it encountered a non-existent selector. 

This breaks the ABI in two ways: first, all contract messages now have a `Result` return
type, and second a new field, `lang_err`, will be introduced as part of the contract
spec. The second change allows other languages, such as Solang, to use an equivalent
`LangError`.

<details><summary>If you're curious, click here for a snippet of the new metadata for the Flipper contract.</summary>
<p>

```json
"messages": [
  {
    "args": [],
    "docs": [
      " Flips the current value of the Flipper's boolean."
    ],
    "label": "flip",
    "mutates": true,
    "payable": false,
    "returnType": {
      "displayName": [
        "ink",
        "MessageResult"
      ],
      "type": 1
    },
    "selector": "0x633aa551"
  }],
"lang_error": {
  "displayName": [
    "ink",
    "LangError"
  ],
  "type": 3
},
{
  "id": 3,
  "type": {
    "def": {
      "variant": {
        "variants": [
          {
            "index": 1,
            "name": "CouldNotReadInput"
          }
        ]
      }
    },
    "path": [
      "ink_primitives",
      "LangError"
    ]
  }
}

```

</p>
</details>

TODO:
- [x] Add new `LangError` type to metadata
- [x] Add E2E test for demonstrating how to handle `LangError`s
- [x] Make CI green

TODOs in Follow-Ups:
- [ ] Implement changes for constructors
- [ ] Implement changes for traits
- [ ] Refactor dispatch logic to bubble up specific dispatch error
    - Will happen at the expense of always having an extra error variant in `LangError`
 - [ ] Make the `CallBuilder` API not require an explicit `Result` 

cc @xgreenx since you were the original poster for the issue


crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -435,16 +436,38 @@ impl Dispatch<'_> {
...
+                let dispatchable = match ::ink::env::decode_input::<
+                        <#storage_ident as ::ink::reflect::ContractMessageDecoder>::Type>() {
+                        Ok(decoded_dispatchable) => {
+                            ::ink::env::debug_println!("Result from `decode_input` {:?}", &decoded_dispatchable);
+                            decoded_dispatchable
+                        },
+                        Err(_decoding_error) => {
+                        ::ink::env::debug_println!("Result from `decode_input` {:?}", &_decoding_error);
+
+                        // TODO: Here will pick out the `Dispatch` error we're interested in telling the
```
ascjones, line 448: I think the only error to raise here is the `CouldNotReadInput`, the rest are already handled by `execute_dispatchable` :point_down:

ascjones (commented):
This approach looks good, I think a unification of the `LangError` and `DispathError` is a good idea if you didn't already have that planned.

crates/env/src/engine/on_chain/impls.rs
```diff
@@ -425,9 +425,23 @@ impl TypedEnvBackend for EnvInstance {
...
+                }
+            }
+            Err(ext::Error::CalleeReverted) => {
+                // let decoded = scale::Decode::decode(&mut &output[..])?;
+                // Ok(decoded)
+                if let Ok(decoded) = scale::Decode::decode(&mut &output[..]) {
+                    Ok(decoded)
+                } else {
+                    ::core::panic!("failed to decode reverted message output");
+                }
```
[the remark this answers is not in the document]
HCastano: Yep, I wanted to tell what path was being taken depending on the return types

crates/ink/src/reflect/dispatch.rs
```diff
@@ -515,7 +515,7 @@ pub trait ExecuteDispatchable {
 }
 
 /// An error that can occur during dispatch of ink! dispatchables.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, ::scale::Encode, ::scale::Decode)]
 pub enum DispatchError {
```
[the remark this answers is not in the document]
HCastano: No, for now they're gonna have to be separate. Not every `DispatchError` maps to a `LangError`. Similarly, we may want to add `LangErrors` that are errors which don't necessarily happen during dispatch (I can't think of any just yet though)

crates/metadata/src/lib.rs
```diff
@@ -28,6 +28,8 @@ pub mod layout;
 mod specs;
 mod utils;
 
+pub use ink_primitives::LangError;
```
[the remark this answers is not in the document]
HCastano: > So this reexport is for the situation where a client has a dep on ink_metadata (and not ink which already exports it via ink::LangError) and just wants to easily grab the LangError type?

Yep.

I wasn't able to have the type directly in the `ink_metadata` crate because we only compile it for `std`, and we need to use this type as part of the codegen (which is `no_std`).

deploy-contract.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -ex
+
+F="${1:-pass}"
+
+cargo contract build --manifest-path examples/flipper/Cargo.toml --skip-linting
```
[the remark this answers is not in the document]
HCastano: Yeah, I haven't gotten around to updating my local version yet, haha

examples/cross_chain_test/lib.rs
```diff
@@ -0,0 +1,37 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[ink::contract]
+mod cross_chain_test {
```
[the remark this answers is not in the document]
HCastano: Ah sorry, I should've mentioned this in the PR description - this example is only being used for development purposes

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -435,16 +436,38 @@ impl Dispatch<'_> {
...
+                        // This is okay since we're going to only be encoding the `Err` variant
+                        // into the output buffer anyways.
+                        ::ink::env::return_value::<::core::result::Result<(), ::ink::LangError>>(
+                            ::ink::env::ReturnFlags::default().set_reverted(true), &error
+                        );
+                    }
+                };
+
+                <<#storage_ident as ::ink::reflect::ContractMessageDecoder>::Type
+                    as ::ink::reflect::ExecuteDispatchable>::execute_dispatchable(dispatchable)
```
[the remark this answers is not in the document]
HCastano: So right now the only error that we get from `execute_dispatchable` is a `DispatchError::PaidUnpayableMessage`, which we don't handle with `LangError`. We may look into adding this in the future though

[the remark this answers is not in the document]
ascjones: What about `InvalidParameters` and `UnknownSelector` e.g https://github.com/paritytech/ink/blob/a4d653b8482f6bc26c038177895a8ed119eff39f/crates/ink/codegen/src/generator/dispatch.rs#L735

crates/ink/codegen/src/generator/dispatch.rs
HCastano: Not quite. Within the call stack of `decode_input` we actually raise a bunch of `DispatchError`s, but they get [swallowed by `__ink_MessageDispatch`'s `scale::Decode` implementation](https://github.com/paritytech/ink/blob/beccff6409100e30a37144d4acfff2ac83b7bf29/crates/ink/codegen/src/generator/dispatch.rs#L813). We then map this to the more generic `CouldNotReadInput` error

xermicus:
@HCastano Can we use `TypeSpec` in the contract spec instead, so it will work for Meta and Portable form? I was so free to push my proposed changes directly to your branch (also merged master), please have a look. If you are not comfortable with this just nuke my commits

xermicus:
> Note you can also open a PR against a PR so the author has the choice to accept or reject the changes.

:100: next time I try to be less annoying sorry

crates/metadata/src/specs.rs
```diff
@@ -81,8 +63,8 @@ impl IntoPortable for ContractSpec {
                 .into_iter()
                 .map(|event| event.into_portable(registry))
                 .collect::<Vec<_>>(),
-            docs: registry.map_into_portable(self.docs),
-            lang_error: registry.register_type(&self.lang_error),
+            docs: self.docs.into_iter().map(|s| s.into()).collect(),
```
HCastano, line 66: @xermicus was this change required?

HCastano:
@xermicus So the metadata now looks like

```json
"lang_error": {
  "displayName": [
    "ink",
    "LangError"
  ],
  "type": 3
},
{
  "id": 3,
  "type": {
    "def": {
      "variant": {
        "variants": [
          {
            "index": 1,
            "name": "CouldNotReadInput"
          }
        ]
      }
    },
    "path": [
      "ink_primitives",
      "LangError"
    ]
  }
}
```

Does this look better to you? Also, not sure what the advantage of going through the spec builder is though. Really we're going to end up hardcoding this error into the metadata anyways

crates/ink/ir/src/ir/item_impl/message.rs
```diff
@@ -291,6 +291,25 @@ impl Message {
...
+        }
+        .map(quote::ToTokens::to_token_stream)
+        .unwrap_or_else(|| quote::quote! { () });
+
+        let wrapped_return = quote::quote! {
+            ::core::result::Result<#return_type, ::ink::LangError>
+        };
+
+        let wrapped_return = syn::parse2::<syn::Type>(wrapped_return)
+            .expect("TODO: Failed to parse into Type");
```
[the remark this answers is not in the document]
HCastano: Nice, TIL

crates/ink/codegen/src/generator/dispatch.rs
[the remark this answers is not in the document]
HCastano: Those are part of the `DecodeDispatch` implementation, not `execute_dispatchable`. 

https://github.com/paritytech/ink/blob/a4d653b8482f6bc26c038177895a8ed119eff39f/crates/ink/codegen/src/generator/dispatch.rs#L827

crates/ink/codegen/src/generator/item_impls.rs
```diff
@@ -255,15 +255,20 @@ impl ItemImpls<'_> {
...
+            .wrapped_output()
+            .map(|ty| quote::ToTokens::to_token_stream(&ty))
+            .expect("This should always be Some atm");
         let statements = message.statements();
         quote_spanned!(span =>
             #( #attrs )*
-            #vis fn #ident(#receiver #( , #inputs )* ) #output_arrow #output {
-                #( #statements )*
+            #vis fn #ident(#receiver #( , #inputs )* ) -> #output {
+                let #mut_token  message_body = || { #( #statements )* };
```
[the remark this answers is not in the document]
HCastano: I've added this and it compiles fine

```rust
#[ink(message)]
pub fn res(&self) -> Result<bool, ()> {
    let h = Self::helper()?;
    Ok(h == 0)
}
                                        
fn helper() -> Result<u8, ()> {
    Ok(1)
}
```

Did you have something else in mind for a potential failure case?

crates/metadata/src/specs.rs
xermicus: I don't think so, this looks more like an oversight on my side, sorry.

xermicus:
@HCastano 

> Does this look better to you?

Yes, this LGTM!

> Also, not sure what the advantage of going through the spec builder is though. Really we're going to end up hardcoding this error into the metadata anyways

The way it was hardcoded breaks it for runtime metadata generation. E.g. in solang, I still need to be able to set this, even though it is kind of "hardcoded". Additionally, the dispatcher in solang could decide to return something else instead.

HCastano:
I made a change to generate two sets of functions, the "original" unchecked function and the new "checked" functions (which are the ones with the `Result<T, LangError>` return types. 

This means that from a source code point of view we don't really have any changes. Old code, such as tests for the `examples`, doesn't need to handle the added `Result` type.

From the ABI's point of view though, the only message that exist are the checked variants, in which the `Result` should be handled.

@ascjones @athei wdyt?

crates/primitives/src/lib.rs
```diff
@@ -41,3 +41,14 @@ pub use self::{
...
+/// An error that comes from the smart contracting language, but not necessarily the smart contract
+/// itself nor the underlying execution environment.
+#[non_exhaustive]
+#[repr(u32)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, ::scale::Encode, ::scale::Decode)]
+#[cfg_attr(feature = "std", derive(::scale_info::TypeInfo))]
+pub enum LangError {
+    /// Failed to read execution input for the dispatchable.
+    CouldNotReadInput = 1u32,
+}
```
athei, line 54: Aren't we missing `UnknownSelector` here? The one that green initially requested this feature for?

athei:
> I made a change to generate two sets of functions, the "original" unchecked function and the new "checked" functions (which are the ones with the `Result<T, LangError>` return types.
> 
> This means that from a source code point of view we don't really have any changes. Old code, such as tests for the `examples`, doesn't need to handle the added `Result` type.
> 
> From the ABI's point of view though, the only message that exist are the checked variants, in which the `Result` should be handled.
> 
> @ascjones @athei wdyt?

Sounds good to me.

crates/primitives/src/lib.rs
HCastano: Kinda. I mentioned this in one of the `TODO` points:

> Refactor dispatch logic to bubble up specific dispatch error (we can maybe do this
later, at the expense of always having an extra error variant in LangError)

So tl;dr, we can't easily pull out `UnknownSelector` atm.

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -295,6 +295,7 @@ impl Dispatch<'_> {
             .map(|message| {
                 let message_span = message.span();
                 let message_ident = message.ident();
+                let checked_message_ident = format_ident!("{}_checked", message_ident);
```
ascjones, line 298: Could lift this to a method on `message`, can use in the other place as well.

crates/ink/codegen/src/generator/item_impls.rs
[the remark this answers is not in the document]
ascjones: No, I obviously underestimated the type inference

crates/ink/codegen/src/generator/item_impls.rs
```diff
@@ -273,8 +273,7 @@ impl ItemImpls<'_> {
 
             #( #attrs )*
             #vis fn #checked_ident(#receiver #( , #wrapped_inputs )* ) -> #wrapped_output {
-                let #mut_token message_body = || { #( #statements )* };
-                ::core::result::Result::Ok(message_body())
+                ::core::result::Result::Ok(self.#ident( #( #input_bindings , )* ))
```
ascjones, line 276: Shouldn't it be the other way around: the unchecked calls the checked and just panics if an error is returned?

HCastano: It could be, but the thing is that the error case will never be hit. The only way to get to the error case is to go through the message dispatch mechanism - that's why we always wrap the body with `Ok`.

That said, if you have strong feelings about this I can change the unchecked implementation to always `expect()` the checked call.

crates/primitives/src/lib.rs
athei: > at the expense of always having an extra error variant in LangError

Which extra error are you talking about specifically?

But in general that is fine. We that we can add new errors to `LangError` without breaking ABI as no one is allowed to exhaustively match on it.

Another thing: Didn't we agree to move this enum to the metadata crate so it is unified across all languages? Or is this TODO.

HCastano: > Which extra error are you talking about specifically?

We'd have `LangError::CouldNotReadInput` as a potential error that doesn't get used if we start exposing more specific errors.

> Another thing: Didn't we agree to move this enum to the metadata crate so it is unified across all languages? Or is this TODO.

We re-export it in the metadata crate. See this [conversation](https://github.com/paritytech/ink/pull/1450/files#r1012870921).

crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -256,7 +264,7 @@ impl Metadata<'_> {
                 let mutates = message.receiver().is_ref_mut();
                 let ident = message.ident();
                 let args = message.inputs().map(Self::generate_dispatch_argument);
-                let ret_ty = Self::generate_return_type(message.output());
+                let ret_ty = Self::generate_return_type(Some(&message.wrapped_output()));
```
xermicus, line 267: What if the metadata crate would transparently "wrap" the return type spec into a `Result<message::Output, ::ink::LangError> for you? On a first glance this _should_ be possible in `ContractSpecBuilder::done()`, since the language error is available in the contract spec.

It would make it a slightly more pleasant API (especially for other languages where there is no `Result` type available) and it would ensure that all message return type specs are wrapped correctly in the metadata.

xermicus (changes requested):

examples/cross_chain_test/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+                    ExecutionInput,
+                    Selector,
+                },
+                DefaultEnvironment,
+            };
+
+            let result = build_call::<DefaultEnvironment>()
+                .call_type(Call::new().callee(address))
+                .exec_input(ExecutionInput::new(Selector::new(selector)))
+                .returns::<Result<(), ::ink::LangError>>()
```
ascjones, line 31: If messages are always expected to return `Result<T, LangError>`, this should be made implicit in the call builder.

crates/primitives/src/lib.rs
ascjones: Ah ha, now I understand why this is tricky because `UnknownSelector` is raised in `decode_dispatch` and then just swallowed into a ::scale::Error`. So a bigger refactoring to access that error required.

```
impl ::scale::Decode for __ink_ConstructorDecoder {
            fn decode<I>(input: &mut I) -> ::core::result::Result<Self, ::scale::Error>
            where
                I: ::scale::Input,
            {
                <Self as ::ink::reflect::DecodeDispatch>::decode_dispatch(input)
                    .map_err(::core::convert::Into::into)
            }
        }
```

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -772,12 +788,10 @@ impl Dispatch<'_> {
...
 
-                    if ::core::any::TypeId::of::<#message_output>() != ::core::any::TypeId::of::<()>() {
-                        // In case the return type is `()` we do not return a value.
-                        ::ink::env::return_value::<#message_output>(
-                            ::ink::env::ReturnFlags::default(), &result
-                        )
-                    }
+                    ::ink::env::return_value::<#message_output>(
+                        ::ink::env::ReturnFlags::default(), &result
+                    )
```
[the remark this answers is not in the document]
ascjones: And in fact it looks like this means that if a message returns `Result::Err` then that will now be the inner value so it would be `Ok(Err(err))`, which means that the contract will no longer be reverted and will be written to storage!

crates/ink/codegen/src/generator/item_impls.rs
ascjones: Sorry I was confusing this with the implementation in `ContractRef`. 

So now my question is: why generate this inherent impk in the first place? This is just generating the inherent impls so I don't think this will ever have enough information to generate a `LangError`.

We should lift this `Ok` as high as possible in the dispatch logic so it is obvious that at the moment that it always returns `Ok`. Because at the moment this has caused potential bug I mentioned in `dispatch` whereby the message returning `Err` gets wrapped in an `Ok` and the contract is not reverted.

ascjones (commented):
We need a test case to check that if a fallible message returns `Err` then the contract is still reverted, it looks like it might not be.

If I understand correctly, the general idea of this PR is to get the ABI change in at the top level (messages always returning `Result<T, LangError>`) in ASAP, then fleshing out the propogation of more errors as a follow up.

I think this is a good approach in order to get a `4.0-beta` out soon: the ABI will be set. However we should make it more explicit maybe by wrapping `result` in `Ok` here: https://github.com/paritytech/ink/blob/e2d81944bbfc524d788bd5e31d133a0b8063e3d6/crates/ink/codegen/src/generator/dispatch.rs#L779

crates/ink/codegen/src/generator/metadata.rs
xermicus: So I think the `ReturnTypeSpec` from the metadata crates deserves a refactor. I started with it here. However I'd wait for #1491 first.

Also since this is just some metadata API change I think it should not hold up our beta release.

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -316,7 +314,7 @@ impl Dispatch<'_> {
 
                         const CALLABLE: fn(&mut Self::Storage, Self::Input) -> Self::Output =
```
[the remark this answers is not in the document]
HCastano: This is not relevant anymore since we've removed the `Result` wrapping on the inherent methods

examples/cross_chain_test/lib.rs
HCastano: Yeah I was thinking of doing that in a follow up PR since it's a source-code level breaking change, not an ABI one

crates/ink/codegen/src/generator/item_impls.rs
HCastano: I didn't want to "lie" about the inherent return type and what was reflected in the ABI, which is why I wrapped the function return types and use that in the metadata generation.

After talking with Alex I realized that they had different consumers, so we could "lie" - but I didn't think of reverting those changes.

Either way, I've accepted your suggestion to move this further up the call stack, it makes sense

examples/lang_err_integration_tests/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "lang_err_integration_tests"
```
ascjones, line 2: Other examples are using `-` as a separator so we should be consistent

examples/lang_err_integration_tests/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+version = "4.0.0-alpha.3"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2021"
+publish = false
+
+[dependencies]
+ink = { path = "../../crates/ink", default-features = false }
+
+scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive"] }
+scale-info = { version = "2", default-features = false, features = ["derive"], optional = true }
```
ascjones, line 12:
```suggestion
scale-info = { version = "2.3", default-features = false, features = ["derive"], optional = true }
```

examples/lang_err_integration_tests/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+
+            let flipper_ink_acc_id = E2EAccountId(flipper_acc_id.clone().into());
+            let invalid_selector = [0x00, 0x00, 0x00, 0x00];
+            let call_result = client
+                .call(
+                    &mut ink_e2e::charlie(),
+                    contract_acc_id.clone(),
+                    lang_err_integration_tests::messages::call(
+                        flipper_ink_acc_id,
+                        invalid_selector,
```
ascjones, line 133: This tests the failure case but we don't test the success case? We could modify the `e2e_can_flip_correctly` test to flip via `lang_err_integration_tests::messages::call`

ascjones (approved):
LGTM.

If the intent is to just get in the ABI changes ASAP so we can do the beta release, I would say the minimum follow up we need is `Implement changes for constructors`.

I would also like to see a :heavy_check_mark:  from @athei that 

  1. This ABI (looking at the `return_value` invocations) is as planned, because it will now be set in stone.
  2. Agrees that this "ABI first, fill in the actual error details later" is the best approach given the focus on getting a `beta` out ASAP.

examples/lang_err_integration_tests/Cargo.toml
HCastano: But `cargo-contract` told me otherwise 😆 

```
❯ cargo contract new beep-boop
Contract names can only contain alphanumeric characters and underscores
```

xermicus (approved):

examples/lang_err_integration_tests/lib.rs
HCastano: I figured there wasn't really a need for it since we check for it here:

https://github.com/paritytech/ink/blob/476ed6515aeffc04f92d4dbd73aa1fec1dcf61b2/examples/lang_err_integration_tests/lib.rs#L203-L206

It's only the failure case that needs to be tested using the `CallBuilder` since we can't craft an invalid message using the E2E tests.

examples/lang-err-integration-tests/contract-ref/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        #[ink(message)]
+        pub fn flip(&mut self) {
+            self.flipper.flip();
+        }
+
+        #[ink(message)]
+        pub fn flip_check(&mut self) {
+            self.flipper
+                .flip_checked()
```
ascjones, line 36: Would testing `err_flip_checked` be useful do you think? As in it will be a `Ok(Err(()))`

ascjones (commented):
:heart: the comprehensive e2e tests

ascjones (approved):

athei (approved):
I think adding an error for `NotPayable` could be quite easy? Can be a follow up but calling `return_value` instead of using `?` seems easy enough.

Apart from that it looks good to me. I compiled some simple contracts and the metadata looks sensible.

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -435,16 +435,34 @@ impl Dispatch<'_> {
...
+                        // At this point we're unable to set the `Ok` variant to be the any "real"
+                        // message output since we were unable to figure out what the caller wanted
+                        // to dispatch in the first place, so we set it to `()`.
+                        //
+                        // This is okay since we're going to only be encoding the `Err` variant
+                        // into the output buffer anyways.
+                        ::ink::env::return_value::<::core::result::Result<(), ::ink::LangError>>(
+                            ::ink::env::ReturnFlags::default().set_reverted(true),
+                            &error,
+                        );
```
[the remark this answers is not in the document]
HCastano: Yeah I knew I'd miss something with this, haha

HCastano:
> I think adding an error for NotPayable could be quite easy? Can be a follow up but calling return_value instead of using ? seems easy enough.

Yeah now that this is in place it should be almost trivial to add that. Figured it would be better left for a follow-up though

xgreenx:
> I made a change to generate two sets of functions, the "original" unchecked function and the new "checked" functions (which are the ones with the Result<T, LangError> return types.

I think “unchecked” fits more for methods that return `MessageResult`, because we didn’t check the result it is why you can handle it=)

athei:
Most idiomatic would be using `try_` as a prefix. This is what the standard library will use to allow panic free allocation soon. For example: `Vec::try_push`.

PR #1452, opened 2022-10-27
patch scale-info dep

xermicus:


ascjones:
Done in https://github.com/paritytech/ink/pull/1357

PR #1456, opened 2022-10-30
replaces .expect() with .unwrap_or_else()

gangov:
this PR merges the changes where we put .unwrap_or_else() instead of .expect() for more detail.

### DIFF
--- a/crates/e2e/src/xts.rs
+++ b/crates/e2e/src/xts.rs
@@ -211,8 +211,9 @@ where
             .unwrap_or_else(|err| {
                 panic!("error on ws request `contracts_instantiate`: {:?}", err);
             });
-        scale::Decode::decode(&mut bytes.as_ref())
-            .unwrap_or_else(|err| panic!("decoding failed: {}", err))
+        scale::Decode::decode(&mut bytes.as_ref()).unwrap_or_else(|err| {
+            panic!("decoding ContractInstantiateResult failed: {}", err)
+        })
     }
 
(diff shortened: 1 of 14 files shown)


gangov:
some extra commits because of the fmt command. I'm not sure if this needs an update, but in the README.md file we have the following as an example: 

`Is rustfmt happy with it?

    cargo +nightly fmt --all`

while the actual thing executed in GitHub was `cargo +nightly fmt --verbose --all -- --check`

both results were giving different outputs and on top of that, the `fmt` was complaining about missing whitespace that IntelliJ was automatically omitting.

crates/e2e/src/xts.rs
```diff
@@ -211,7 +211,8 @@ where
             .unwrap_or_else(|err| {
                 panic!("error on ws request `contracts_instantiate`: {:?}", err);
             });
-        scale::Decode::decode(&mut bytes.as_ref()).expect("decoding failed")
+        scale::Decode::decode(&mut bytes.as_ref())
+            .unwrap_or_else(|err| panic!("decoding failed: {}", err))
```
ascjones, line 215: Would be useful to know what failed to decode e.g. `"decoding ContractInstantiateResult failed`. Same for other `"decoding failed"` messages

ascjones (approved):

crates/e2e/src/xts.rs
gangov: sure thing, fixed in the follow up commit

PR #1458, opened 2022-10-30
Conditional Codegen for Messages, Constructors and Events

### LINKED ISSUE #1231: Messages and contstructor cannot be conditionally compiled
Minimal reproducer: https://ink-playground.substrate.io/?id=d666be89f36785e487036bf43d7b175b

It is the same for constructors. I would expect for the message to be just not included in the contract. Instead I get this error:

```
error[E0599]: the function or associated item `call` exists for struct `lol::Lol`, but its trait bounds were not satisfied
--> lib.rs:18:16
|
8 | pub struct Lol {}
| --------------
| |
| function or associated item `call` not found for this
| doesn't satisfy `lol::Lol: std::ops::Fn<_>`
...
18 | pub fn call(&self) {}
| ^^^^ function or associated item cannot be called on `lol::Lol` due to unsatisfied trait bounds
|
= note: the following trait bounds were not satisfied:
`lol::Lol: std::ops::Fn<_>`
which is required by `&lol::Lol: std::ops::Fn<_>`
note: the following trait must be implemented
= help: items from traits can only be used if the trait is implemented and in scope
= note: the following trait defines an item `call`, perhaps you need to implement it:
candidate #1: `std::ops::Fn`
```

SkymanOne:
This PR closes #1231 

It allows constructors, messages and events to be conditionally compiled (i.e. have `#[cfg(...)]` attribute)

# Example

```rust
/* --Snippet -- */
#[ink(message)]
#[cfg(feature = "foo")]
pub fn another_get(&self) -> String {
    self.value.clone()
}
/* --Snippet -- */

/* --Snippet -- */
#[ink(constructor)]
#[cfg(feature = "another_foo")]
pub fn another_new(val: String) -> Self {
    Self { value: val }
}
/* --Snippet -- */

/* --Snippet -- */
#[ink(event)]
#[cfg(feature = "foo")]
pub struct FeaturedAuctionEchoed {
    auction: Auction,
}
/* --Snippet -- */
```

### DIFF
--- a/crates/ink/codegen/src/generator/metadata.rs
+++ b/crates/ink/codegen/src/generator/metadata.rs
@@ -323,22 +323,26 @@ impl Metadata<'_> {
 
     /// Generates ink! metadata for all user provided ink! event definitions.
     fn generate_events(&self) -> impl Iterator<Item = TokenStream2> + '_ {
-        self.contract.module().events().map(|event| {
-            let span = event.span();
-            let ident = event.ident();
-            let docs = event.attrs().iter().filter_map(|attr| attr.extract_docs());
-            let args = Self::generate_event_args(event);
-            quote_spanned!(span =>
-                ::ink::metadata::EventSpec::new(::core::stringify!(#ident))
-                    .args([
-                        #( #args ),*
-                    ])
-                    .docs([
-                        #( #docs ),*
-                    ])
-                    .done()
-            )
-        })
+        self.contract
+            .module()
+            .events()
+            .filter(|event| !is_conditionally_excluded(event.attrs()))
+            .map(|event| {
+                let span = event.span();
+                let ident = event.ident();
+                let docs = event.attrs().iter().filter_map(|attr| attr.extract_docs());
+                let args = Self::generate_event_args(event);
+                quote_spanned!(span =>
+                    ::ink::metadata::EventSpec::new(::core::stringify!(#ident))
+                        .args([
+                            #( #args ),*
+                        ])
+                        .docs([
+                            #( #docs ),*
+                        ])
+                        .done()
+                )
+            })
     }
 
     /// Generate ink! metadata for a single argument of an ink! event definition.
--- a/crates/ink/codegen/src/generator/metadata.rs
+++ b/crates/ink/codegen/src/generator/metadata.rs
@@ -27,8 +27,6 @@ use quote::{
 };
 use syn::spanned::Spanned as _;
 
-use super::is_conditionally_excluded;
-
 /// Generates code to generate the metadata of the contract.
 #[derive(From)]
 pub struct Metadata<'a> {
@@ -116,7 +114,6 @@ impl Metadata<'_> {
             .module()
             .impls()
             .flat_map(|item_impl| item_impl.iter_constructors())
-            .filter(|constructor| !is_conditionally_excluded(constructor.attrs()))
             .map(|constructor| Self::generate_constructor(constructor))
     }
 
@@ -210,7 +207,6 @@ impl Metadata<'_> {
             .impls()
             .filter(|item_impl| item_impl.trait_path().is_none())
             .flat_map(|item_impl| item_impl.iter_messages())
-            .filter(|message| !is_conditionally_excluded(message.attrs()))
             .map(|message| {
                 let span = message.span();
                 let docs = message
@@ -261,7 +257,6 @@ impl Metadata<'_> {
                     })
             })
             .flatten()
-            .filter(|(_, message)| !is_conditionally_excluded(message.attrs()))
             .map(|((trait_ident, trait_path), message)| {
                 let message_span = message.span();
                 let message_ident = message.ident();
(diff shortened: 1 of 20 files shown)


crates/ink/codegen/src/generator/mod.rs
```diff
@@ -70,3 +73,32 @@ pub use self::{
...
+/// #[ink(message)]
+/// #[cfg(feature = "foo")]
+/// #[cfg(feature = "baz")]
+/// pub fn get(&self) -> String {
+///     self.value.clone()
+/// }
+/// ```
+/// The function would would iterate over the `cfg` attributes
+/// and return `true` if any of the feature flags are set
+pub fn is_conditionally_excluded(attrs: &[Attribute]) -> bool {
```
ascjones, line 92: I would change this to the positive e.g. `any_conditional_cfg` since the most common usage is negated e.g. `!is_conditionally_excluded`. Easier to read and understand.

crates/ink/codegen/src/generator/mod.rs
```diff
@@ -70,3 +73,32 @@ pub use self::{
...
+    for attr_tokens in attrs
+        .iter()
+        .filter(|a| a.path.is_ident("cfg"))
+        .map(|a| &a.tokens)
+    {
+        let _s = attr_tokens.to_token_stream();
+        if !cfg!(_s) {
```
ascjones, line 99: Not 100% sure of the logic here I will check back tomorrow

crates/ink/codegen/src/generator/mod.rs
ascjones: Okay maybe a better name is something like `all_cfg_predicates_true`, since as I understand it:
```
#[cfg(feature = "foo")]
#[cfg(feature = "baz")]
```
is the equivalent of:
```
#[cfg(all(feature = "foo", feature = "bar"))]
```

crates/ink/codegen/src/generator/mod.rs
```diff
@@ -70,3 +73,32 @@ pub use self::{
     storage_item::StorageItem,
     trait_def::TraitDefinition,
 };
+
+/// Check if the the conditional compilation expression
+/// for any `cfg` attributes is satisfied.
```
ascjones, line 78: As I read the implementation, it is actually "check if any cfg predicate is not satisfied"

crates/ink/codegen/src/generator/mod.rs
SkymanOne: I extract tokens from the `cfg` attribute and pass them into `cfg!` macro which allows to filter our code spans

crates/ink/codegen/src/generator/mod.rs
```diff
@@ -70,3 +73,32 @@ pub use self::{
...
+/// ```ignore
+/// #[ink(message)]
+/// #[cfg(feature = "foo")]
+/// #[cfg(feature = "baz")]
+/// pub fn get(&self) -> String {
+///     self.value.clone()
+/// }
+/// ```
+/// The function would would iterate over the `cfg` attributes
+/// and return `true` if any of the feature flags are set
```
ascjones, line 91: .."and return `true` if any of the feature flags are **not** set.

crates/ink/codegen/src/generator/mod.rs
ascjones: Ok I think the logic is correct but it is at odds with the comments

crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -114,6 +116,7 @@ impl Metadata<'_> {
             .module()
             .impls()
             .flat_map(|item_impl| item_impl.iter_constructors())
+            .filter(|constructor| !is_conditionally_excluded(constructor.attrs()))
```
ascjones, line 119: I reckon if you encapsulate this filter inside `iter_constructors` and `iter_messages` you could get rid of a lot of these `filter` expressions

SkymanOne: I was thinking about that, but I don't think it is a good idea because there might be a case when we want to get all messages no matter whether they have `cfg` attributes or not

ascjones: I believe for all current usages that is not the case though.

If we are sticking to conditional compilation semantics then if `cfg` predicates not there it would be as if the code is not there at all.

athei (approved):

athei:
Just make cure it builds with the feature enabled `catgo check --features foo`.

crates/ink/ir/src/ir/item_impl/mod.rs
```diff
@@ -373,3 +379,35 @@ impl ItemImpl {
...
+/// #[ink(message)]
+/// #[cfg(feature = "foo")]
+/// #[cfg(feature = "baz")]
+/// pub fn get(&self) -> String {
+///     self.value.clone()
+/// }
+/// ```
+/// The function would would iterate over the `cfg` attributes
+/// and return `true` if any of the feature flags are set
+pub fn is_code_span_enabled(attrs: &[Attribute]) -> bool {
```
ascjones, line 398: A more natural place for this would be `ir/utils.rs`

crates/ink/ir/src/ir/item_impl/mod.rs
```diff
@@ -373,3 +379,35 @@ impl ItemImpl {
...
+        .iter()
+        .filter(|a| a.path.is_ident("cfg"))
+        .map(|a| &a.tokens);
+    if cfg_attrs.clone().count() == 0 {
+        return true
+    }
+    for attr_tokens in cfg_attrs {
+        let _s = attr_tokens.to_token_stream();
+        if cfg!(_s) {
+            return true
```
ascjones, line 409: I believe this logic is wrong in the case of multiple `cfg` attributes e.g.
```
fn main() {
    foo_bar();
}

#[cfg(feature = "foo")]
#[cfg(feature = "bar")]
fn foo_bar() {}
```
`cargo build --features foo` -> :red_circle: 
`cargo build --features bar` -> :red_circle: 
`cargo build --features foo,bar` -> :green_circle: 

So having separate cfg attrs seems to be the equivalent of `#[cfg(all(feature = "foo", feature = "bar"))]` 

The implementation here is instead `#[cfg(any(feature = "foo", feature = "bar"))]`.

ascjones: The logic would be easier to read when expressed e.g. in the form:

```rust
pub fn is_code_span_enabled(attrs: &[Attribute]) -> bool {
    attrs
        .iter()
        .filter(|a| a.path.is_ident("cfg"))
        .all(|a| {
            let predicate = &a.tokens.to_token_stream();
            cfg!(predicate)
        })
}
```

SkymanOne:
> Just make cure it builds with the feature enabled `cargo check --features foo`.

Checked. It works

crates/ink/ir/src/ir/item_impl/mod.rs
```diff
@@ -373,3 +379,35 @@ impl ItemImpl {
...
+    let cfg_attrs = attrs
+        .iter()
+        .filter(|a| a.path.is_ident("cfg"))
+        .map(|a| &a.tokens);
+    if cfg_attrs.clone().count() == 0 {
+        return true
+    }
+    for attr_tokens in cfg_attrs {
+        let _s = attr_tokens.to_token_stream();
+        if cfg!(_s) {
```
ascjones, line 408: Does this even work with the token stream? `cfg!` is a compile time macro.

I've tried it and it doesn't appear to work. Even with the literal `cfg!(feature = "foo")` it doesn't work for some reason. `cfg!(unix)` literal does work.

I think we need to rethink the approach here, maybe see whether we can carry those attributes through to the generated code (I know it is everywhere)

ascjones (commented):
I believe this does not work

PR #1460, opened 2022-11-01
Generate metadata for constructor return type

### LINKED ISSUE #1457: Reflect fallible constructors in metadata
This is a follow up  for https://github.com/paritytech/ink/pull/1446 which added support for fallible constructors with with a custom error type: `Result<Self, T>`. Please note that constructors are still allowed to return only `Self` in case of them being infallible. 

## Current behaviour

The data returned by a constructor is not mentioned in the metadata. All constructors return nothing according to metadata. This was fine before #1446 because they never wrote anything into the output buffer. However, now the `Err` will be written to the output buffer but only when an error is returned (nothing is written on `Ok`).

## Changes Needed

- The fact that constructors write data into the output buffer needs to be mentioned in metadata. It is necessary for callers so that they can decode the output buffer. Any constructor returning `Result<Self, T>` should have the following return type in metadata `Result<(), T>`.  `Self` is replaced by `()` because ink! constructors do not return any data from successful constructors. 
- Encode `Ok(())` into the output buffer in case of a successful construction. This is redundant information but it is necessary to make the constructor behave as advertised in the metadata. This will allow us to emit data from a constructor in the future while being backwards compatible.

Infallible constructors (` -> Self`) will continue do not return anything. So they don't have a return type in metadata and never copy anything to the output buffer.

cc @SkymanOne @xermicus @HCastano

SkymanOne:
This is a follow up for #1446  and closes #1457

### DIFF
--- a/crates/ink/codegen/src/generator/metadata.rs
+++ b/crates/ink/codegen/src/generator/metadata.rs
@@ -167,7 +167,7 @@ impl Metadata<'_> {
     /// Generates the ink! metadata for the given type.
     fn generate_type_spec(ty: &syn::Type) -> TokenStream2 {
         fn without_display_name(ty: &syn::Type) -> TokenStream2 {
-            quote! { ::ink::metadata::TypeSpec::new::<#ty>() }
+            quote! { ::ink::metadata::TypeSpec::of_type::<#ty>() }
         }
         if let syn::Type::Path(type_path) = ty {
             if type_path.qself.is_some() {
--- a/crates/ink/codegen/src/generator/metadata.rs
+++ b/crates/ink/codegen/src/generator/metadata.rs
@@ -193,7 +193,7 @@ impl Metadata<'_> {
         }
     }
 
-    /// Generates the ink! metadata for the given type.
+    /// Generates the ink! metadata specs for the given type of a constructor.
     fn generate_constructor_type_spec(ty: &syn::Type) -> TokenStream2 {
         fn without_display_name(ty: TokenStream2) -> TokenStream2 {
             quote! {
@@ -355,6 +355,9 @@ impl Metadata<'_> {
     }
 
     /// Generates ink! metadata for the given return type of a constructor.
+    /// If constructor result type is not `Result`,
+    /// the metadata will not display any metadata for return type.
+    /// Otherwise, the return type is `Result<(), E>`.
     fn generate_constructor_return_type(ret_ty: Option<&syn::Type>) -> TokenStream2 {
         /// Returns `true` if the given type is `Self`.
         fn type_is_self_val(ty: &syn::Type) -> bool {
@@ -387,6 +390,8 @@ impl Metadata<'_> {
         }
     }
 
+    /// Helper function which replace all occurrences of `Self`
+    /// with `()`.
     fn replace_self_with_unit(ty: &syn::Type) -> TokenStream2 {
         if ty.to_token_stream().to_string().contains("< Self") {
             let s = ty.to_token_stream().to_string().replace("< Self", "< ()");
--- a/crates/ink/codegen/src/generator/metadata.rs
+++ b/crates/ink/codegen/src/generator/metadata.rs
@@ -40,11 +40,14 @@ impl GenerateCode for Metadata<'_> {
     fn generate_code(&self) -> TokenStream2 {
         let contract = self.generate_contract();
         let layout = self.generate_layout();
+        let type_transform = self.generate_storage_type_transform();
 
         quote! {
             #[cfg(feature = "std")]
             #[cfg(not(feature = "ink-as-dependency"))]
             const _: () = {
+                #type_transform
+
                 #[no_mangle]
                 pub fn __ink_generate_metadata() -> ::ink::metadata::InkProject  {
                     let layout = #layout;
@@ -105,6 +108,14 @@ impl Metadata<'_> {
                     #( #docs ),*
                 ])
                 .done()
+        }
+    }
+
+    /// Generates a default implementation of `TransformType` for the storage type.
+    fn generate_storage_type_transform(&self) -> TokenStream2 {
+        let storage = self.contract.module().storage().ident();
+        quote! {
+            impl ::ink::metadata::TransformType for #storage { }
         }
     }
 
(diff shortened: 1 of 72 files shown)


crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -316,6 +354,48 @@ impl Metadata<'_> {
...
+                quote! {
+                    if !::ink::is_result_type!(#type_token) {
+                        ::ink::metadata::ReturnTypeSpec::new(::core::option::Option::None)
+                    } else {
+                       ::ink::metadata::ReturnTypeSpec::new(#type_spec)
+                    }
+                }
+            }
+        }
+    }
```
Robbepop, line 388: This again does syntactical checks instead of using the Rust type system properly and will fail for various syntactic contexts. To do this properly you'd need to take a similar approach as in `generate_trait_messages` with Rust type system reflection such as
```rust
                let is_payable = quote! {{
                    <<::ink::reflect::TraitDefinitionRegistry<<#storage_ident as ::ink::reflect::ContractEnv>::Env>
                        as #trait_path>::__ink_TraitInfo
                        as ::ink::reflect::TraitMessageInfo<#local_id>>::PAYABLE
                }};
```
We do have the information for ink! messages and constructors as well. They are identified using their selectors which are always known at proc. macro compile time. Therefore you can access them in this scope just like in the linked example.

SkymanOne: The metadata codegen takes place after the WASM blob compilation, hence at this stage all types have already been evaluated by compiler and we can work with syntax.

In fact, I *generally* do not do syntactical checks. I only check if the token is `Self` here because metadata generation takes place outside the storage definition, hence `Self` can not be evaluated there. Same logic applies for `Result<Self, ...>`.

However, if the return type is the explicit type of a contract or `Result` with explicit contract type for `Ok` variant, then I do the type evaluation, hence why we have `TransformResult` factory.
I can be sure that the `Self` is valid in this context is because the WASM build is already done by now. 

I am not how message selectors are relevant here.

crates/metadata/src/specs.rs
```diff
@@ -994,6 +1028,34 @@ where
...
+}
+
+impl<O, E> TransformResult<core::result::Result<O, E>>
+where
+    E: TypeInfo + 'static,
+{
+    /// Produces [`TypeSpec`] for a `Result<(), E>`.
+    pub fn new_type_spec<S>(segments_opt: Option<S>) -> TypeSpec
+    where
+        S: IntoIterator<Item = &'static str>,
```
xermicus, line 1049: My gut feeling is this should rather be for `<F as Form>::String` or it will break runtime metadata construction

xermicus (changes requested):

crates/metadata/src/specs.rs
xermicus: Nevermind. After giving this some more thought I think external consumers (e.g. other languages relying on runtime metadata construction) don't need this because they can produce the correct type first hand anyways.

SkymanOne:
After few tweaks and additional tests here is the summary of what has been achieved:

The metadata is successfully generated for these edges cases:
- Return type is `Self` then no return type is indicated in metadata
- Return type is storage type then no return type is indicated in metadata too
- Return type is alias to the storage type, same behaviour as above
- Return type is `Result`, then metadata indicates that `Ok` is `()` and `Err` is the type specified by developer
- Return type is alias to `Result` - same behaviour as above
- Return type is alias to `Result` and `Ok` variant is a type alias to the contract type - same behaviour as above

crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -316,6 +358,55 @@ impl Metadata<'_> {
...
+                }
+            }
+            Some(ty) => {
+                if type_is_self_val(ty) {
+                    return quote! { ::ink::metadata::ReturnTypeSpec::new(::core::option::Option::None)}
+                }
+                let segments = Self::generate_constructor_type_segments(ty);
+                let type_token = Self::replace_self_with_unit(ty);
+                quote! {
+                    if !::ink::is_result_type!(#type_token) {
```
xermicus, line 387: If the return type can only be `Self` or `Result<Self, E>` and you are sure that this is already enforced by some earlier compilation stage, isn't this is a redundant check that can be removed?

See: #1473

crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -316,6 +358,55 @@ impl Metadata<'_> {
         }
     }
 
+    /// Generates ink! metadata for the given return type of a constructor.
+    /// If constructor result type is not `Result`,
+    /// the metadata will not display any metadata for return type.
+    /// Otherwise, the return type is `Result<(), E>`.
+    fn generate_constructor_return_type(ret_ty: Option<&syn::Type>) -> TokenStream2 {
+        /// Returns `true` if the given type is `Self`.
+        fn type_is_self_val(ty: &syn::Type) -> bool {
```
xermicus, line 367: Please move this directly into the match statement, I find it confusing. See https://github.com/paritytech/ink/pull/1473

crates/metadata/src/specs.rs
```diff
@@ -994,6 +1028,46 @@ where
...
 }
 
+/// This is wrapper for [`TypeSpec`] which acts as a factory.
+/// The whole purpose of the factory is to replace the type of
+/// `Ok` variant of `Result` with `()` because constructors do not return
+/// any data upon successful instantiation.
+///
+/// # Important Note
+/// Only use this factory with constructors!
+pub trait TransformType {
```
xermicus, line 1038: I find the name of the Trait (and the trait method) confusing. Proposal in #1473

crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -107,6 +111,14 @@ impl Metadata<'_> {
         }
     }
 
+    /// Generates a default implementation of `TransformType` for the storage type.
+    fn generate_storage_type_transform(&self) -> TokenStream2 {
+        let storage = self.contract.module().storage().ident();
+        quote! {
+            impl ::ink::metadata::TransformType for #storage { }
```
xermicus, line 118:
```suggestion
            impl ::ink::metadata::TransformType for #storage {}
```

xermicus (changes requested):
LGTM! 

> The metadata is successfully generated for these edges cases:
> ...

Is this explicitly tested for somewhere? In the metadata I see only one case (`<Result<u8, ()>` being tested for. Might be worthwhile to add some contract testcases.

Also don't forget about documentation (if not already done by the original PR).

xermicus:
Made some more minor changes: #1475 

Also this PR fixes something, but the CI was passing prior/without the fix. Doesn't this mean that we should add the missing test case here as well?

crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -316,6 +358,45 @@ impl Metadata<'_> {
         }
     }
 
+    /// Generates ink! metadata for the given return type of a constructor.
+    /// If constructor result type is not `Result`,
+    /// the metadata will not display any metadata for return type.
```
xermicus, line 355:
```suggestion
    /// the metadata will not display any type spec for the return type.
```

crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -442,4 +514,29 @@ mod tests {
             ],
         )
     }
+
+    #[test]
+    fn constructor_return_type_works() {
+        let expected_no_ret_type_spec = ":: ink :: metadata :: ReturnTypeSpec :: new (:: core :: option :: Option :: None)";
```
ascjones, line 520: It's not a great way to test it using strings like this. It's brittle and dependent upon the codegen impl. 

If we can come up with a good solution with types + reflection this will be easier to test.

crates/ink/codegen/src/generator/metadata.rs
ascjones: I think we may be able to refactor this to make it use the Rust type system, I'm looking into this now.

ascjones (commented):
The resulting metadata looks to be correct, but the main problem is that at the moment the actually returned value will not be an encoded instance of the type in the metadata.

Also I think we can utilize the type system here as @Robbepop suggests, I am looking into this now.

crates/ink/codegen/src/generator/metadata.rs
ascjones: I've done an experiment with this in https://github.com/paritytech/ink/pull/1491, take a look and see what you think.

PR #1461, opened 2022-11-03
Add `Mapping::take()`

SkymanOne:
This PR adds a new function `Mapping::take()` which retrieves and removed a value under the given key from the storage.

### DIFF
--- a/crates/engine/src/ext.rs
+++ b/crates/engine/src/ext.rs
@@ -253,6 +253,22 @@ impl Engine {
         }
     }
 
+    /// Removes the storage entries at the given key.
+    /// Returns previously stored value at the key if any.
+    pub fn take_storage(&mut self, key: &[u8], output: &mut &mut [u8]) -> Result {
+        let callee = self.get_callee();
+        let account_id = AccountId::from_bytes(&callee[..]);
+
+        self.debug_info.inc_writes(account_id);
+        match self.database.remove_contract_storage(&callee, key) {
+            Some(val) => {
+                set_output(output, &val);
+                Ok(())
+            }
+            None => Err(Error::KeyNotFound),
+        }
+    }
+
     /// Returns the size of the value stored in the contract storage at the key if any.
     pub fn contains_storage(&mut self, key: &[u8]) -> Option<u32> {
         let callee = self.get_callee();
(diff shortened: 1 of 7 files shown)


agryaznov (approved):

crates/engine/src/ext.rs
```diff
@@ -253,6 +253,22 @@ impl Engine {
         }
     }
 
+    /// Removes the storage entries at the given key.
+    /// Returns previously stored value at the key if any.
```
HCastano, line 257:
```suggestion
    /// Removes the storage entry at `key`, returning the previous value `key` from storage.
```

HCastano (commented):
Some small things, but generally looks good.

Also, do you know what the version of the contracts pallet introduced `seal_take`? We'll need to know so we can mention it in the release notes (oh btw, can you add an entry to the `CHANGELOG` for this)

SkymanOne:
> Some small things, but generally looks good.
> 
> Also, do you know what the version of the contracts pallet introduced `seal_take`? We'll need to know so we can mention it in the release notes (oh btw, can you add an entry to the `CHANGELOG` for this)

Yes.
https://github.com/paritytech/substrate/blob/master/frame/contracts/CHANGELOG.md#v300-2021-02-25

crates/env/src/api.rs
```diff
@@ -214,6 +214,21 @@ where
     })
 }
 
+/// Removes the `value` at `key`,
+/// returning the previous `value` at `key` from storage
+/// # Errors
```
HCastano, line 219:
```suggestion
/// Removes the `value` at `key`, returning the previous `value` at `key` from storage.
///
/// # Errors
```

crates/env/src/backend.rs
```diff
@@ -180,6 +180,17 @@ pub trait EnvBackend {
         K: scale::Encode,
         R: Storable;
 
+    /// Removes the `value` at `key`,
+    /// returning the previous `value` at `key` from storage
+    /// if any
+    /// # Errors
```
HCastano, line 186:
```suggestion
    /// Removes the `value` at `key`, returning the previous `value` at `key` from storage if
    /// any.
    ///
    /// # Errors
```

HCastano (approved):
Small nits with the wrapping, but otherwise looks good

ascjones (approved):

PR #1464, opened 2022-11-03
Add tests for metadata construction at runtime

xermicus:
Amendment for #1357

### DIFF
--- a/crates/metadata/src/layout/tests.rs
+++ b/crates/metadata/src/layout/tests.rs
@@ -14,6 +14,7 @@
 
 use super::*;
 use ink_primitives::Key;
+use scale_info::Path;
 
 #[test]
 fn layout_key_works() {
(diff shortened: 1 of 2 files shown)


HCastano (approved):

crates/metadata/src/layout/tests.rs
```diff
@@ -318,3 +319,35 @@ fn unbounded_layout_works() {
...
+                "name": "Storage",
+                "fields": [
+                    {
+                    "name": "Field",
+                    "layout": {
+                      "leaf": {
+                        "key": "0x00000000",
+                        "ty": 123
+                      }
+                    }
```
ascjones, line 346: Needs one more indent I think

ascjones (approved):

crates/metadata/src/layout/tests.rs
xermicus: good catch

PR #1473, opened 2022-11-05
test changes

xermicus:


crates/metadata/src/specs.rs
```diff
@@ -1035,12 +1035,12 @@ where
 ///
 /// # Important Note
 /// Only use this factory with constructors!
-pub trait TransformType {
+pub trait ConstructorReturnType {
```
SkymanOne, line 1038: The name is misleading IMO. It doesn't reflect the functionality of the trait. The whole idea behind this trait is to *transform*/*convert* the return type into either `()` or `Result<(), E>` . this doesn't necessary mean it can only be applied to constructors. Additionally we already have `ConstructorReturnType<C>` in [`ink::codegen::dispatch`](https://github.com/paritytech/ink/blob/6b24de9a55474fa47737663e08ef10661d073a9a/crates/ink/src/codegen/dispatch/execution.rs#L108) module. This can also arise confusion in future

crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -363,34 +363,24 @@ impl Metadata<'_> {
...
-        }) if path.is_ident("Self"))
-        }
-
         match ret_ty {
             None => {
                 quote! {
                     ::ink::metadata::ReturnTypeSpec::new(::core::option::Option::None)
                 }
             }
+            Some(syn::Type::Path(syn::TypePath { qself: None, path }))
```
SkymanOne, line 372: Definitely cleaner approach!

SkymanOne (changes requested):
Thanks for help! I like cleaner approach to the syntactical check, but renaming the trait may not be a good idea

crates/metadata/src/specs.rs
xermicus: > Additionally we already have ConstructorReturnType<C>

Ooh that's my bad. I wasn't aware of this. So I do agree we should change it to something not already present.

> The whole idea behind this trait is to transform/convert the return type into either () or Result<(), E>

I can't see the "transforming" part for the trait name, neither the "new" part in the spec. Implementing this trait for some type `T` indicates what the return type _spec_ of_ this `T` will be inside the metadata, given `T` is used as the result of a constructor.

> this doesn't necessary mean it can only be applied to constructors.

I can't think of another use case here. Besides that, how would this interface exactly work? It only allows to implement one "transformation" per particular type. How would this be useful e.g. for a type that is returned by a constructor and is also used for something else which needs a conflicting "transformation"? I don't think we should prematurely solve this problem right now.

xermicus: Inspired by `ConstructorReturnType` I renamed the trait to `ConstructorReturnSpec`. Lines [like this](https://github.com/paritytech/ink/blob/c377c8bee0fef250bf18b67f95234de819ce3cd5/crates/metadata/src/tests.rs#L61) look much cleaner to me now, what do you think?

SkymanOne:
LGTM!

PR #1475, opened 2022-11-07
remove unecessary helper fn

xermicus:


crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -40,13 +40,13 @@ impl GenerateCode for Metadata<'_> {
     fn generate_code(&self) -> TokenStream2 {
         let contract = self.generate_contract();
         let layout = self.generate_layout();
-        let type_transform = self.generate_storage_type_transform();
+        let storage = self.contract.module().storage().ident();
```
SkymanOne, line 43:
```suggestion
        let storage_type_transform = self.contract.module().storage().ident();
```

SkymanOne (changes requested):
I think

SkymanOne (approved):
LGTM!

PR #1476, opened 2022-11-07
Fix clippy checks :t-rex: 

ascjones:


HCastano (approved):
How come this wasn't caught as part of the CI in #1461?

ascjones:
> How come this wasn't caught as part of the CI in #1461?

I assume because the CI passed prior to `1.65` making it through

PR #1482, opened 2022-11-08
Remove `seal_` prefixes from FFI

agryaznov:
Follow-up to [substrate#11444](https://github.com/paritytech/substrate/issues/11444)

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -143,7 +143,7 @@ like the following excerpt:
 ```rust
 #[link(wasm_import_module = "seal0")]
 extern "C" {
-    pub fn seal_get_storage(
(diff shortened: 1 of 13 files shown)


athei:
From the CI:
```
ERROR: An unexpected import function was found in the contract Wasm: get_storage
```

I think we need a fix for `cargo contract` which expects that imports start with `seal_`.

ascjones:
> From the CI:
> 
> ```
> ERROR: An unexpected import function was found in the contract Wasm: get_storage
> ```
> 
> I think we need a fix for `cargo contract` which expects that imports start with `seal_`.

https://github.com/paritytech/cargo-contract/pull/816

ascjones (approved):

PR #1490, opened 2022-11-10
Add more detailed error and fail during initialisation

### LINKED ISSUE #1470: Check if `#[ink_e2e::test(ws_url = "…")]` is reachable
The E2E tests run against a node behind `ws_url`. If you take a look at the implementation you'll see that we default to `ws://localhost:9944`. Right now there's just a strange "connection error" when the developer forgets to run a local node.

## ToDo

(1) Please check that a connection to this `ws_url` can be created before executing any tests. The best place to add this check is [here](https://github.com/paritytech/ink/blob/da094c6dfa529c7888364e67655f5d888bbbfb6a/crates/e2e/macro/src/codegen.rs):
```rust
::ink_e2e::INIT.call_once(|| {
    ::ink_e2e::env_logger::init();
    // this closure is executed just once, before running any tests.
});
```

(2) If the connection cannot be established please panic with a meaningful error, i.e. sth like "No node is running behind …".

(3) Add a test that checks for this behavior.

SkymanOne:
This PR closes #1470.

If there is no node running behind the url, the following error will be thrown:
```
failures:

---- give_me::e2e_tests::e2e_contract_must_transfer_value_to_sender stdout ----
thread 'give_me::e2e_tests::e2e_contract_must_transfer_value_to_sender' panicked at '

Error establishing connection to a node at ws://0.0.0.0:9944
Make sure your node is running.', /Users/skyman/Projects/ink/crates/e2e/src/client.rs:286:21

---- give_me::e2e_tests::e2e_sending_value_to_give_me_must_fail stdout ----
thread 'give_me::e2e_tests::e2e_sending_value_to_give_me_must_fail' panicked at 'Once instance has previously been poisoned', lib.rs:188:9
```

And no tests will be run.

### DIFF
--- a/crates/e2e/src/tests.rs
+++ b/crates/e2e/src/tests.rs
@@ -1,15 +1,10 @@
-#[test]
+#[tokio::test]
 #[should_panic(
     expected = "Error establishing connection to a node at ws://0.0.0.0:9944. Make sure you run a node behind the given url!"
 )]
-fn fail_initialization_with_no_node() {
-    let client_init =
-        crate::Client::<crate::PolkadotConfig, ink_env::DefaultEnvironment>::new(
-            "ws://0.0.0.0:9944",
-        );
-    tokio::runtime::Builder::new_current_thread()
-        .enable_all()
-        .build()
-        .unwrap_or_else(|err| panic!("Failed building the Runtime during test: {}", err))
-        .block_on(client_init);
+async fn fail_initialization_with_no_node() {
+    let _ = crate::Client::<crate::PolkadotConfig, ink_env::DefaultEnvironment>::new(
+        "ws://0.0.0.0:9944",
+    )
+    .await;
 }
--- a/crates/e2e/src/tests.rs
+++ b/crates/e2e/src/tests.rs
@@ -1,0 +1,15 @@
+#[test]
+#[should_panic(
+    expected = "\n\nError establishing connection to a node at ws://0.0.0.0:9944\nMake sure your node is running."
+)]
+fn fail_initialization_with_no_node() {
+    let client_init =
+        crate::Client::<crate::PolkadotConfig, ink_env::DefaultEnvironment>::new(
+            "ws://0.0.0.0:9944",
+        );
+    tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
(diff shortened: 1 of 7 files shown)


crates/e2e/src/tests.rs
```diff
@@ -0,0 +1,15 @@
+#[test]
```
ascjones, line 1: What we really want to do is fully test the `#[ink_e2e::test(ws_url = "…")]` macro here

SkymanOne: It is not possible to use E2E macros outside the contract's project:
```
ERROR: Cannot infer the root project id
[2022-11-10T20:19:57Z ERROR ink_e2e_macro::codegen] `cargo-contract` stderr: 
 --> crates/e2e/tests/sample.rs:5:5
  |
5 |     #[ink_e2e::test(ws_url = "ws://0.0.0.0:9944")]
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: contract build for Cargo.toml failed

error: could not compile `ink_e2e` due to previous error
```

crates/e2e/src/tests.rs
```diff
@@ -0,0 +1,15 @@
...
+#[should_panic(
+    expected = "Error establishing connection to a node at ws://0.0.0.0:9944. Make sure you run a node behind the given url!"
+)]
+fn fail_initialization_with_no_node() {
+    let client_init =
+        crate::Client::<crate::PolkadotConfig, ink_env::DefaultEnvironment>::new(
+            "ws://0.0.0.0:9944",
+        );
+    tokio::runtime::Builder::new_current_thread()
```
ascjones, line 10: You could avoid this boilderplate by using[ `#[tokio::test]`](https://docs.rs/tokio/0.2.3/tokio/attr.test.html). Needs the `macro` feature of tokio.

ascjones (approved):
LGTM

crates/e2e/src/tests.rs
ascjones, line 1:
```suggestion
#[tokio::test]
```

ascjones: Need to make it an `async fn ` too

Issue #1483, opened 2022-11-08
 Fix and refactor fallible constructors


ascjones:
Following up on #1460, we should:

1. Test whether this actually works either via a client (need to update e.g. `cargo-contract` to handle constructor return values), or a cross contract instantiation. See https://github.com/paritytech/ink/pull/1460#discussion_r1016886652. If it doesn't work we need to fix it.
2. Refactor the metadata generation to see whether we can use fhe type system to generate the return type metadata. See https://github.com/paritytech/ink/pull/1460#discussion_r1010656515

PR #1174, opened 2022-03-10
Extracted push and pull of the contract in the dispatching logic

xgreenx:
Extracted push and pull of the contract in the dispatching logic to reduce the size of the contracts.

It is part of the https://github.com/paritytech/ink/pull/1017. It is easy to review and it can be part of a stable release=)

PR #1491, opened 2022-11-10
Rework fallible constructors

ascjones:
Follow up to #1460 and https://github.com/paritytech/ink/pull/1446. Closes https://github.com/paritytech/ink/issues/1483.

This PR:

- Builds constructor return type metadata using types only.
- Combine those types with in the dispatch logic, 
- Refactor the dispatch logic closer to how messages do it from https://github.com/paritytech/ink/pull/1174 (may even reduce contract size).

Test with: `cargo test --manifest-path ./examples/lang-err-integration-tests/constructors-return-value/Cargo.toml --features e2e-tests`


crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -555,23 +561,52 @@ impl Dispatch<'_> {
...
+                                &contract,
+                            );
+                            // On success we return the `Ok(())` value for callers.
+                            ::ink::env::return_value::<::core::result::Result<&(), ()>>(
+                                ::ink::env::ReturnFlags::default(), &::core::result::Result::Ok(&())
+                            )
+                        },
+                        ::core::result::Result::Err(err) => {
+                           ::ink::env::return_value::<::core::result::Result<(), & #constructor_value :: Error>>(
+                                ::ink::env::ReturnFlags::default(), &::core::result::Result::Err(err)
```
SkymanOne, line 606: Shouldn't we revert the transaction here **and** return `Error`?

ascjones: Indeed it absolutely should!

crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -131,11 +128,32 @@ impl Metadata<'_> {
...
+                Some(
+                    if #constructor_info ::IS_RESULT {
+                        ::ink::metadata::TypeSpec::of_type::<
+                            ::core::result::Result<
+                                (),
+                                #constructor_info ::Error
+                            >
+                        >()
+                    } else {
+                        ::ink::metadata::TypeSpec::of_type::< #constructor_info ::Output>()
```
SkymanOne, line 152: I think this is misleading to add storage type to metadata as a return type. If the constructor is not `Result` nothing is returned

SkymanOne:
I am also getting clippy warning for constructors attribute:
```
this expression creates a reference which is immediately dereferenced by the compiler
```
Currently debugging this

crates/ink/codegen/src/generator/metadata.rs
ascjones: You are right, will fix.

ascjones: Since the ABI as it stands in this PoC PR always returns a result regardless, I have updaetd the metadata to reflect that. How we want it to be, I'll leave it up to you.

examples/lang-err-integration-tests/constructors-return-value/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "constructors_return_value"
```
ascjones, line 2: I've added this "example" into your integration tests folder @HCastano, because already all the CI scripts and waterfall tests deal with this special case folder. Perhaps we could introduce an `integration-tests` folder and have feature specific integration tests in there.

SkymanOne (approved):
LGTM!

examples/lang-err-integration-tests/constructors-return-value/Cargo.toml
HCastano: > Perhaps we could introduce an integration-tests folder and have feature specific integration tests in there.

Yep, I was thinking the same thing

HCastano (approved):
Good stuff!

PR #1492, opened 2022-11-10
Add E2E tests for `Mapping` functions

### LINKED ISSUE #1472: Add E2E test for `take_storage`
Please add an `#[ink_e2e::test]` that uses `Mapping::take_storage`.

Afterwards tell @agryaznov, so that he stabilize it in `pallet-contracts`.

### LINKED ISSUE #1488: `Mapping::insert()` fails in E2E tests
# Summary
When trying to use a message contains `Mapping::insert()` in E2E test, the test errors at dry-run stage for this specific function.

# Reproducer
1. Checkout [gn/mapping-take-tests](https://github.com/paritytech/ink/tree/gn/mapping-take-tests)
2. Navigate to [examples/mappings](https://github.com/paritytech/ink/tree/gn/mapping-take-tests/examples/mappings) folder
3. Run [`substrate-contracts-node V0.21.0`](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.21.0) - `substrate-contracts-node --dev -linfo,runtime::contracts=debug`
4. Run `RUST_LOG=debug cargo test`
5. You should get the following error message
```
***
2022-11-09T16:02:34Z INFO  ink_e2e] [test: insert_and_get_works] call: [E8, 8D, 16, 46, 01, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00]
[2022-11-09T16:02:34Z INFO  ink_e2e] [test: insert_and_get_works] call dry run: Err(Module(ModuleError { index: 7, error: 25, message: None }))
[2022-11-09T16:02:34Z INFO  ink_e2e] [test: insert_and_get_works] call dry run debug message: 
test mappings::e2e_tests::insert_and_get_works ... FAILED

failures:

---- mappings::e2e_tests::insert_and_get_works stdout ----
thread 'mappings::e2e_tests::insert_and_get_works' panicked at 'call failed: CallDryRun', lib.rs:117:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    mappings::e2e_tests::insert_and_get_works

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.23s

error: test failed, to rerun pass '--lib'
***
```

Module error that the message refers to is [`StorageDepositLimitExhausted`](https://github.com/paritytech/substrate/blob/a6da808575fe403ab2bd7f6cd009896cdc6fd71a/frame/contracts/src/lib.rs#L824)

# Edge cases
* Commenting out the code block containing `Mapping::take()` does not cause the problem, so it is definitely to do something with the function.
* Specifying `storage_limit` has no effect
* Executing normal & dry-run call via `cargo-contract` and PolkadotJS app *do not* result in error
* The implementations of dry-run call both in `cargo-contract` and `E2E` crate seem to be the same.

SkymanOne:
This PR fixes #1488 and closes #1472 by adding E2E tests for all `Mapping` functions and correctly submitting the origin of the dry-run call during E2E test


examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+            let caller = Self::env().caller();
+            self.balances.get(caller)
+        }
+
+        /// Demonstrates the usage of `Mappings::insert()`.
+        ///
+        /// Assigns the value to a given account.
+        /// Returns the size of the pre-existing balance at the specified key if any.
+        ///
+        /// Returns `None` if the account was non-existent.
```
HCastano, line 43:
```suggestion
        /// Assigns the value to a given account.
        ///
        /// Returns the size of the pre-existing balance at the specified key if any. Returns
        /// `None` if the account was not previously in the `Mapping`.
```

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+        #[ink(message)]
+        pub fn insert_balance(&mut self, value: Balance) -> Option<u32> {
+            let caller = Self::env().caller();
+            self.balances.insert(caller, &value)
+        }
+
+        /// Demonstrates the usage of `Mappings::size()`.
+        ///
+        /// Returns the size of the pre-existing value at the specified key if any.
```
HCastano, line 52: Similar suggested change here

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+            let caller = Self::env().caller();
+            self.balances.remove(caller);
+        }
+
+        /// Demonstrates the usage of `Mappings::take()`.
+        ///
+        /// Returns the balance of a given account
+        /// removing it from storage.
+        ///
+        /// Returns `0` if the account is non-existent.
```
HCastano, line 84:
```suggestion
        /// Returns the balance of a given account removing it from storage.
        ///
        /// Returns `None` if the account is not in the `Mapping`.

```

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+        #[ink(message)]
+        pub fn take_balance(&mut self) -> Option<Balance> {
+            let caller = Self::env().caller();
+            self.balances.take(caller)
+        }
+    }
+
+    #[cfg(test)]
```
HCastano, line 92:
```suggestion
        #[cfg(all(test, feature = "e2e-tests"))]
```

We now have these feature flagged since we don't want people to need a node running all the time just to run tests (see `lang-err-integration-tests`). Can you also add the feature to the contract's `Cargo.toml`?

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+                .expect("instantiate failed")
+                .account_id;
+
+            // when
+            let insert = mappings::messages::insert_balance(1_000);
+            let size = client
+                .call(&mut ink_e2e::alice(), contract_id.clone(), insert, 0, None)
+                .await
+                .expect("call failed")
+                .value;
```
HCastano, line 114: So most of these calls are gonna need an extra `unwrap` due to the recent `LangError` changes that were introduced

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+
+            let insert = mappings::messages::insert_balance(2_000);
+            let _ = client
+                .call(&mut ink_e2e::bob(), contract_id.clone(), insert, 0, None)
+                .await
+                .expect("call failed");
+
+            // then
+            let contains = mappings::messages::contains_balance();
+            let is_there = client
```
HCastano, line 150: I'd add a check to ensure that this was `not_there` before the call to `insert`

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+            // given
+            let constructor = mappings::constructors::new();
+            let contract_id = client
+                .instantiate(&mut ink_e2e::charlie(), constructor, 0, None)
+                .await
+                .expect("instantiate failed")
+                .account_id;
+
+            // when
+            let first_insert = mappings::messages::insert_balance(1_000_000);
```
HCastano, line 172: Bit of a nitpick, but I'd keep all the `insert` amounts consistent between tests unless there's a reason for changing them. 

I have to do a double check each test to reason about why this is `X` instead of `Y`

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+                    insert,
+                    0,
+                    None,
+                )
+                .await
+                .expect("call failed")
+                .value;
+
+            // then
+            assert!(size.is_some());
```
HCastano, line 198: You should also assert that you have the correct value here

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+
+            // then
+            let contains = mappings::messages::contains_balance();
+            let is_not_there = client
+                .call(&mut ink_e2e::eve(), contract_id.clone(), contains, 0, None)
+                .await
+                .expect("call failed")
+                .value;
+
+            assert_eq!(balance, Some(4_000));
```
HCastano, line 275: I'd move this check to right after the `take` call

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+            let take = mappings::messages::take_balance();
+            let balance = client
+                .call(&mut ink_e2e::eve(), contract_id.clone(), take, 0, None)
+                .await
+                .expect("call failed")
+                .value;
+
+            // then
+            let contains = mappings::messages::contains_balance();
+            let is_not_there = client
```
HCastano, line 269:
```suggestion
            let is_there = client
```

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+                .call(&mut ink_e2e::eve(), contract_id.clone(), contains, 0, None)
+                .await
+                .expect("call failed")
+                .value;
+
+            assert_eq!(balance, Some(4_000));
+            assert!(!is_not_there);
```
HCastano, line 276:
```suggestion
            assert!(!is_there);
```

HCastano (commented):
Nothing major. You're gonna need to do some fixes for the `LangError` changes

examples/mappings/lib.rs
SkymanOne: Yes, so we don't get race condition in transaction pool

examples/mappings/lib.rs
```diff
@@ -0,0 +1,281 @@
...
+    #[cfg(test)]
+    mod e2e_tests {
+        type E2EResult<T> = std::result::Result<T, Box<dyn std::error::Error>>;
+
+        #[ink_e2e::test]
+        async fn insert_and_get_works(
+            mut client: ink_e2e::Client<C, E>,
+        ) -> E2EResult<()> {
+            // given
+            let constructor = mappings::constructors::new();
```
ascjones, line 101: This PR will need updating now since https://github.com/paritytech/ink/pull/1518

ascjones (approved):
LGTM, but I would wait until `substrate-contracts-node` is updated and the tests actually pass until merging. Otherwise our master CI will be red.

PR #1495, opened 2022-11-11
Fix fallible message revert, make `Ok` explicit

ascjones:
Suggestions for #1450.

:warning: :beer: :monkey:  Friday night untested suggestions :monkey: :beer: :warning: 

- Fix possible issue with fallible message returning `Err` not reverting
- Move up message dispatch always `Ok` directly to where `return_value` is invoked for clarity.
- Removes inherent `_checked` impl since it is now not used and has not use from user code (always returns `Ok`)


SkymanOne:
I don't see any changed touching the constructors. Can you please provide more details on the issue with constructors?

HCastano:
> I don't see any changed touching the constructors. Can you please provide more details on the issue with constructors?

It was a typo

PR #1496, opened 2022-11-15
Fix E2E tests by updating `Weight` type

ascjones:
- Updates to use new `sp_weights::Weight` type for `gas_limit`.
- Depends on new versions of [`sp-core`](https://github.com/paritytech/substrate/pull/12599) etc and [`subxt`](https://github.com/paritytech/subxt/pull/709)

Since the e2e test crate is already using `git` dependencies and is not currently released to `crates.io`, I am happy to convert the substrate crate dependencies to git. Once all dependencies are released to crates.io we can use those, and then this crate can be released.


athei:
Don't the call sides need to be fixed now that the tests are required to pass `Weight` instead of `u64`?

ascjones:
> Don't the call sides need to be fixed now that the tests are required to pass `Weight` instead of `u64`?

The changes here already change the call argument to `Weight`?

athei:
Was thinking about the actual tests in examples which need to pass `Weight`.

ascjones:
Appears that the dry run to fetch the estimated `Weight` happens transparently https://github.com/paritytech/ink/blob/66c05c9b0a9139990fc2ac8e81090f668ea9e053/crates/e2e/src/client.rs#L574

athei (approved):

HCastano (approved):
Nice!

I tested this with the [v0.22.1](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.22.1) release of the contracts node and it worked 👍

PR #1500, opened 2022-11-16
Polished refactor for constructor return type

SkymanOne:
This is my iteration over #1491
- Fixes unnecessary reference issue
- Adds docs
- Adds test assertion for `Error` variant
- Indicated the right return type in metadata for `Self` and `Result<Self>`


crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -338,6 +326,34 @@ impl Metadata<'_> {
...
+        quote_spanned!(span=>
+            ::ink::metadata::ReturnTypeSpec::new(
+                if #constructor_info ::IS_RESULT {
+                    Some(::ink::metadata::TypeSpec::with_name_str::<
+                        ::core::result::Result<
+                            (),
+                            #constructor_info ::Error
+                        >
+                    >(
+                        "core::result::Result"
```
ascjones, line 347: I'm not sure of the value of adding this to the display name...really we would want the generic parameters e.g. `Result<(), MyError>`. I know that is slightly tricky but not impossible. The other way would be to just have no display name, anyway the UI should detect whether we are dealing with a result using the type definition.

Anyway, leave it and I can always modify that in my PR

ascjones (approved):
Haven't decided about the display name but it should be harmless

PR #1504, opened 2022-11-16
Return `LangError`s from constructors

HCastano:
Follow-up to #1450, and closes the constructor part of #1207.

As in #1450, constructors now return a `Result<Constructor::Output, LangError>`
which can be handled by callers.

<details><summary>If you're curious, click here for a snippet of the new metadata.</summary>
<p>

```json
    "constructors": [
  {
    "args": [
      {
        "label": "init_value",
        "type": {
          "displayName": [
            "bool"
          ],
          "type": 0
        }
      }
    ],
    "docs": [
      "Infallible constructor"
    ],
    "label": "new",
    "payable": false,
    "returnType": {
      "displayName": [
        "ink_primitives",
        "ConstructorResult"
      ],
      "type": 1
    },
    "selector": "0x9bae9d5e"
  },
                                        
{
  "id": 1,
  "type": {
    "def": {
      "variant": {
        "variants": [
          {
            "fields": [
              {
                "type": 2
              }
            ],
            "index": 0,
            "name": "Ok"
          },
          {
            "fields": [
              {
                "type": 3
              }
            ],
            "index": 1,
            "name": "Err"
          }
        ]
      }
    },
    "params": [
      {
        "name": "T",
        "type": 2
      },
      {
        "name": "E",
        "type": 3
      }
    ],
    "path": [
      "Result"
    ]
  }
},
{
  "id": 2,
  "type": {
    "def": {
      "tuple": []
    }
  }
},
{
  "id": 3,
  "type": {
    "def": {
      "variant": {
        "variants": [
          {
            "index": 1,
            "name": "CouldNotReadInput"
          }
        ]
      }
    },
    "path": [
      "ink_primitives",
      "LangError"
    ]
  }
}
```
</p></details>

TODOs (Edit: These are gonna be done in #1512)
- [ ] ~~Add test showing how to handle `LangError`~~ 
- [ ] ~~ContractRef implementations~~


crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -343,20 +343,28 @@ impl Metadata<'_> {
...
+                            #constructor_info::Error
+                        >>
                     >(
-                        "core::result::Result"
+                        "ink_primitives::ConstructorResult"
                     )
                 )
                 } else {
-                    ::core::option::Option::None
+                    ::core::option::Option::Some(::ink::metadata::TypeSpec::with_name_str::<
```
ascjones, line 360: `ReturnTypeSpec` should now always expect a `TypeSpec` instead of an `Option<TypeSpec>`. There is one other usage to take care of for trait output types.

crates/ink/ir/src/ir/item_impl/constructor.rs
```diff
@@ -227,6 +227,21 @@ impl Constructor {
...
+    pub fn wrapped_output(&self) -> syn::Type {
+        let return_type = self
+            .output()
+            .map(quote::ToTokens::to_token_stream)
+            .unwrap_or_else(|| quote::quote! { () });
+
+        syn::parse_quote! {
+            ::ink::ConstructorResult<#return_type>
+        }
+    }
```
ascjones, line 244: Looks to be unused

examples/lang-err-integration-tests/constructors-return-value/lib.rs
```diff
@@ -183,21 +192,21 @@ pub mod constructors_return_value {
...
 
             assert!(
-                decoded_result.is_err(),
-                "Fallible constructor should have failed"
+                decoded_result.is_ok(),
+                "Constructor dispatch should have succeeded"
             );
 
-            let result = client
-                .instantiate(&mut ink_e2e::charlie(), constructor, 0, None)
```
ascjones, line 197: We still want to actually attempt to run the constructor to make sure it fails

ascjones (commented):
Looks good, we may want to resolve the type of the `ReturnValueSpec` to be non optional before merging.

ascjones:
Another thing to note is that the contract sizes have gone up very slightly again here: so something to investigate for future betas. e.g. `erc20` has gone from `7.6` to `7.7`

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -607,28 +624,32 @@ impl Dispatch<'_> {
...
                             ::ink::env::set_contract_storage::<::ink::primitives::Key, #storage_ident>(
                                 &<#storage_ident as ::ink::storage::traits::StorageKey>::KEY,
                                 contract,
                             );
+
                             // only fallible constructors return success `Ok` back to the caller.
-                            if #constructor_value :: IS_RESULT {
-                                ::ink::env::return_value::<::core::result::Result<&(), ()>>(
+                            if #constructor_value::IS_RESULT {
+                                ::ink::env::return_value::<::ink::ConstructorResult<::core::result::Result<(), ()>>>(
```
athei, line 636: Doesn't really matter because how SCALE works but I think this helps with readability (at least I stumbled over it).

```suggestion
                                ::ink::env::return_value::<::ink::ConstructorResult<::core::result::Result<(), &#constructor_value::Error>>>(
```

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -607,28 +624,32 @@ impl Dispatch<'_> {
...
                                     ::ink::env::ReturnFlags::new_with_reverted(false),
-                                    &::core::result::Result::Ok(&())
+                                    &::core::result::Result::Ok(::core::result::Result::Ok(())),
```
athei, line 638:
```suggestion
                                    &::ink::ConstructorResult::Ok(::core::result::Result::Ok(())),
```

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -607,28 +624,32 @@ impl Dispatch<'_> {
...
                                 )
                             } else {
-                                ::core::result::Result::Ok(())
+                                ::ink::env::return_value::<::ink::ConstructorResult<()>>(
+                                    ::ink::env::ReturnFlags::new_with_reverted(false),
+                                    &::core::result::Result::Ok(()),
```
athei, line 643:
```suggestion
                                    &::ink::ConstructorResult::Ok(()),
```

athei (approved):
Metadata and `return_value` invocations look good.

athei (changes requested):
I think we forgot the reversion

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -607,28 +624,32 @@ impl Dispatch<'_> {
...
-                        ::core::result::Result::Err(err) => {
-                           ::ink::env::return_value::<::core::result::Result<(), & #constructor_value :: Error>>(
-                                ::ink::env::ReturnFlags::new_with_reverted(true),
-                                &::core::result::Result::Err(err)
-                            )
                         }
+                        ::core::result::Result::Err(err) => ::ink::env::return_value::<
+                            ::ink::ConstructorResult<::core::result::Result<(), &#constructor_value::Error>>,
+                        >(
+                            ::ink::env::ReturnFlags::new_with_reverted(false),
```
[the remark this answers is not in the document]
ascjones: The test should be fixed to check for that see https://github.com/paritytech/ink/pull/1504#discussion_r1026257768

athei:
I saw a change to unify those cases and get away with a single `return_value`. Revert if you don't like.

athei:
Okay then let's be nice to the compiler and also unify it for the `call` case.

crates/ink/codegen/src/generator/dispatch.rs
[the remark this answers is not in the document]
HCastano: Copy-pasta strikes again 🙈

HCastano:
The change related to handling `LangError`s from constructors ended up being a bit more
annoying that I expected. I've opened up a separate PR for those (#1512).

The current behaviour of this PR is as follows:
- Instantiations through normal dispatch:
    - These are fine, we never raise `LangError`s
- `CreateBuilder` or `ContractRef` instantiations:
    - These work, but we are not able to handle the `LangError`
    - Instead, we get a `ContractReverted` error from the Contracts pallet (so kinda like
      the old behaviour)

Since the most important thing here are the metadata changes, we _could_ in principle
merge this. The changes from #1512 would "just" be code level changes.

athei:
> Instantiations through normal dispatch:
> 
>     * These are fine, we never raise `LangError`s

I don't get it. When I use the call builder I define myself what I expect to receive. And I would need to define that I get `Result<(), LangError>` or the decoding fails because that is what the callee copies into the output buffer for a fallible constructor.

> `CreateBuilder` or `ContractRef` instantiations:
> 
>     * These work, but we are not able to handle the `LangError`
> 
>     * Instead, we get a `ContractReverted` error from the Contracts pallet (so kinda like
>       the old behaviour)

Shouldn't it be `ContractTrapped` because you basically `unwrap` the outer `Result`?  `ContractReverted` that you called `seal_return` which doesn't make sense given that you would need to choose an output buffer to return that conforms with the callers return type.

HCastano:
> I don't get it. When I use the call builder I define myself what I expect to receive. And I would need to define that I get `Result<(), LangError>` or the decoding fails because that is what the callee copies into the output buffer for a fallible constructor.
> 

What we want to talk about here isn't the `CallBuilder`, but the _`CreateBuilder`_. At
the moment the `CreateBuilder` only supports a return type of an `AccountId`, which is
comes from `seal_instantiate`. In #1512 I try and add support for reading the `LangError`
from the output buffer of `seal_instatiate`.

When I talk about "normal dispatch" I'm referring to calls that go through something like
`cargo-contract` or the E2E tests . By default neither is able to use an invalid selector
(I guess you could hack the metadata or fiddle at the call site to do this) so these
should not return a `LangError`.

> Shouldn't it be `ContractTrapped` because you basically `unwrap` the outer `Result`? `ContractReverted` implies that you called `seal_return` which doesn't make sense given that you would need to choose an output buffer to return that conforms with the callers return type.

If I log the return of the `CreateBuilder::instantiate()` call I get `CalleeReverted`. We're failing
to decode the selector and so we hit this:

https://github.com/paritytech/ink/blob/b63ef910b6c02a563cd70b081adbe90ba24ffb12/crates/ink/codegen/src/generator/dispatch.rs#L438-L441

which calls `seal_return`.

athei:
> What we want to talk about here isn't the `CallBuilder`, but the _`CreateBuilder`_. At
> the moment the `CreateBuilder` only supports a return type of an `AccountId`, which is
> comes from `seal_instantiate`. In #1512 I try and add support for reading the `LangError`
> from the output buffer of `seal_instatiate`.

Okay I got it. Yes this is a bit more of a high level stuff then. Yes we can leave it that way  for the beta.

> When I talk about "normal dispatch" I'm referring to calls that go through something like
> cargo-contract or the E2E tests . By default neither is able to use an invalid selector
> (I guess you could hack the metadata or fiddle at the call site to do this) so these
> should not return a LangError.

Yeah those off-chain callers are also not that interesting for ABI. As a matter of fact none of the UIs even displays data coming back from constructors (cause we didn't return anything up until now). We are good here.


> If I log the return of the `CreateBuilder::instantiate()` call I get `CalleeReverted`. We're failing to decode the selector and so we hit this:
> 
> https://github.com/paritytech/ink/blob/b63ef910b6c02a563cd70b081adbe90ba24ffb12/crates/ink/codegen/src/generator/dispatch.rs#L438-L441

This is the **callee** side. But I think this is about the **caller**. How does the caller handle getting a **Err(LangError)**?

ascjones (commented):
LGTM. Please check my remaining unresolved comments

ascjones (approved):
I'm fine to delay the `CreateBuilder` changes in #1512 to a subsequent `beta` release, based on the philosophy of stablizing the ABI but allowing breaking code changes.

crates/ink/codegen/src/generator/metadata.rs
HCastano: I don't know enough about all the metadata types to be able to remove this and still get things compiling, so in the interest of time I'm gonna leave it a a follow-up

crates/ink/codegen/src/generator/dispatch.rs
```diff
@@ -420,16 +420,33 @@ impl Dispatch<'_> {
...
+                        // into the output buffer anyways.
+                        ::ink::env::return_value::<::ink::ConstructorResult<()>>(
+                            ::ink::env::ReturnFlags::new_with_reverted(true),
+                            &error,
+                        );
+                    }
+                };
+
+                <<#storage_ident as ::ink::reflect::ContractConstructorDecoder>::Type
+                    as ::ink::reflect::ExecuteDispatchable>::execute_dispatchable(dispatchable)
```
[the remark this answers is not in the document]
HCastano: Yeah that seems fair enough. I'll leave it for a follow-up since we'll want to change it for messages too

HCastano:
> This is the callee side. But I thought we are discussing the caller here. How does the caller handle getting a Err(LangError)?

Sorry, maybe I'm missing something - but right now they don't. There are two scenarios:
1. Calling an invalid selector with `cargo-contract`: `ContractReverted`, nothing to do
2. Calling an invalid selector via `CreateBuilder`: `ContractReverted`, but can continue its own execution

athei:
> 2\. selector

Checked the code. Looks alright. We don't even decode the return buffer for instantiate. So we are save here. So from my side we can merge this and release the first beta. #1512 is not needed for the beta.

crates/ink/codegen/src/generator/metadata.rs
HCastano: Opened #1515

examples/lang-err-integration-tests/call-builder/lib.rs
```diff
@@ -55,14 +55,42 @@ mod call_builder {
...
+            .gas_limit(0)
+            .endowment(0)
+            .exec_input(ExecutionInput::new(Selector::new(selector)).push_arg(init_value))
+            .salt_bytes(&[0xDE, 0xAD, 0xBE, 0xEF])
+            .params()
+            .instantiate();
+
+            // NOTE: Right now we can't handle any `LangError` from `instantiate`, we can only tell
+            // that our contract reverted (i.e we see error from the Contracts pallet).
+            result.ok().map(|id| ink::ToAccountId::to_account_id(&id))
```
deuszx, line 82: How do you recommend handling instantiating other contracts inside a constructor?

PR #1505, opened 2022-11-17
Add doc tests for `Mapping`

doutv:
I use the example code here: https://use.ink/4.0.0-alpha.1/datastructures/mapping, plus some additional comments.

### DIFF
--- a/crates/storage/src/lazy/mapping.rs
+++ b/crates/storage/src/lazy/mapping.rs
@@ -38,7 +38,7 @@ use scale::{
 /// A mapping of key-value pairs directly into contract storage.
 ///
 /// # Examples
-/// ```
+/// ```rust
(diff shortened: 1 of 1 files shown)


doutv:
@ascjones Could you help me review this PR? Thanks for your time!

HCastano (changes requested):
Hey thanks for your PR. Unfortunately you've used the docs for the ink! 3.x series of releases, while we're working towards 4.0. This means your example doesn't compile (this is why the CI doesn't pass).

In the future make sure all the CI steps pass, and if you have any questions around how to do that let us know.

In terms of how to move forward with this, we can expand the existing doc test to make it the main example by adding functions which set and get values from the `Mapping`. We can then have a small follow-up section which shows the `ManualKey` stuff. How does that sound?

Issue #1510, opened 2022-11-18
Migrate `ink-waterfall` tests to `e2e` tests


ascjones:
Currently the ink-waterfall does not work, see https://github.com/paritytech/ink-waterfall/issues/41 and https://github.com/polkadot-js/apps/issues/8364. 

We need to have continued confidence that all our examples are working.

Let's migrate all the waterfall tests to be E2E tests in this repo. This is just a list of all files from https://github.com/paritytech/ink-waterfall/tree/819b12969a068de10b7984e5ce9957f79a0a53a7/src/tests.

Let's share the load and migrate them one-by one:

- [ ] set_code_hash.rs
- [x] contract_terminate.rs https://github.com/paritytech/ink/pull/1511
- [ ] data_structures.rs
- [ ] trait_incrementer.rs
- [ ] trait_erc20.rs
- [ ] erc721.rs
- [ ] contract_transfer.rs
- [ ] seal_code_hash.rs**
- [ ] seal_ecdsa.rs
- [ ] contract_introspection.rs
- [ ] incrementer.rs**
- [ ] erc1155.rs
- [ ] multisig.rs
- [ ] dns.rs
- [ ] trait_flipper.rs
- [ ] rand_extension.rs
- [x] flipper.rs https://github.com/paritytech/ink/pull/1573
- [ ] delegator.rs
- [ ] forward_calls.rs
- [x] erc20.rs https://github.com/paritytech/ink/pull/1513

PR #1511, opened 2022-11-18
Migrate `contract-terminate` waterfall test to e2e

ascjones:
Part of #1510.

Migrates the `contract-terminate` test from `ink-waterfall` https://github.com/paritytech/ink-waterfall/blob/819b12969a068de10b7984e5ce9957f79a0a53a7/src/tests/contract_terminate.rs#L33.

Change the behaviour of the `ink_e2e::Client::call` method so that it does not panic if it fails to decode the dry run result. This is because `contract_terminate` never returns so no data is written to the output buffer, even though the call itself succeeds.


crates/e2e/src/client.rs
```diff
@@ -573,6 +573,50 @@ where
...
+            events: call_result.events,
+        })
+    }
+
+    /// Executes a `call` for the contract at `account_id`.
+    ///
+    /// Returns when the transaction is included in a block. The return value
+    /// contains all events that are associated with this transaction.
+    ///
+    /// The raw bytes of the
```
HCastano, line 611: Missing rest of the

examples/contract-terminate/lib.rs
```diff
@@ -52,4 +52,55 @@ pub mod just_terminates {
...
+
+            assert!(
+                call_res.value.is_empty(),
+                "Terminated contract never returns"
+            );
+
+            let contains_event = |pallet_name: &str, variant_name: &str| {
+                call_res.events.iter().any(|evt| {
+                    let evt = evt.unwrap();
+                    // println!("{} {}", evt.pallet_name(), evt.variant_name());
```
HCastano, line 93:
```suggestion
```

examples/contract-terminate/lib.rs
```diff
@@ -52,4 +52,55 @@ pub mod just_terminates {
...
+                )
+                .await
+                .expect("terminate_me messages failed");
+
+            assert!(
+                call_res.value.is_empty(),
+                "Terminated contract never returns"
+            );
+
+            let contains_event = |pallet_name: &str, variant_name: &str| {
```
HCastano, line 90: This would be useful to have as part of the `ink_e2e` crate

examples/contract-terminate/lib.rs
```diff
@@ -52,4 +52,55 @@ pub mod just_terminates {
...
+            let contract_acc_id = client
+                .instantiate(&mut ink_e2e::alice(), constructor, 0, None)
+                .await
+                .expect("instantiate failed")
+                .account_id;
+
+            // when
+            let terminate_me = contract_terminate::messages::terminate_me();
+            let call_res = client
+                .call_raw(
```
HCastano, line 75: Why does this need to be a raw call as opposed to a regular call?

examples/contract-terminate/lib.rs
```diff
@@ -52,4 +52,55 @@ pub mod just_terminates {
...
+                .await
+                .expect("terminate_me messages failed");
+
+            assert!(
+                call_res.value.is_empty(),
+                "Terminated contract never returns"
+            );
+
+            let contains_event = |pallet_name: &str, variant_name: &str| {
+                call_res.events.iter().any(|evt| {
```
HCastano, line 91: nit:

```suggestion
                call_res.events.iter().any(|event| {
```

crates/e2e/src/client.rs
ascjones: Removed this fn now in favour of making `call` succeed even if a decoding error

examples/contract-terminate/lib.rs
ascjones: Removed `call_raw` now. But the reason is that `terminate` never returns: https://github.com/paritytech/ink/blob/e883ce5088553c93b49493e43185ce05485399d3/crates/env/src/api.rs#L345

So the return buffer is empty even though the call "succeeds" we can't decode into `Result<(), LangError>`.

Previously `call` would have panicked attempting to do the decode but it returns now.

HCastano (approved):

Issue #1498, opened 2022-11-15
E2E test "panicked at 'error on call `sign_and_submit_then_watch_default`"


HCastano:
If you try and instantiate or call a contract message from different E2E test using the
same `signer` you end up running into the following error:

```
❯ cargo +nightly test --manifest-path ./examples/flipper/Cargo.toml

running 4 tests
test flipper::tests::it_works ... ok
test flipper::tests::default_works ... ok
test flipper::e2e_tests::e2e_first_instantiation ... FAILED
test flipper::e2e_tests::e2e_second_instantiation ... ok

failures:

---- flipper::e2e_tests::e2e_first_instantiation stdout ----
thread 'flipper::e2e_tests::e2e_first_instantiation' panicked at 'error on call `sign_and_submit_then_watch_default`: Rpc(ClientError(Call(Custom(ErrorObject { code: ServerError(1014), message: "Priority is too low: (318 vs 318)", data: Some(RawValue("The transaction has too low priority to replace another transaction already in the pool.")) }))))', /Users/hcastano/Workspace/ParityTech/ink/crates/e2e/src/xts.rs:259:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    flipper::e2e_tests::e2e_first_instantiation

test result: FAILED. 3 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.17s

error: test failed, to rerun pass `--lib`
```

However, if you change the `signer` in one of the tests then everything passes.

I'm not sure if the error lies in the E2E tests, or in `subxt`, but either way I imagine
there needs to be some better nonce management here.

This can be reproduced with `ink@7789caf`, `cargo-contract@3e24722`, and
`substrate-contracts-node@0.21.0` using the following tests with `Flipper`:

```rust
#[cfg(test)]
mod e2e_tests {
    type E2EResult<T> = std::result::Result<T, Box<dyn std::error::Error>>;

    #[ink_e2e::test]
    async fn e2e_first_instantiation(
        mut client: ink_e2e::Client<C, E>,
    ) -> E2EResult<()> {
        let constructor = flipper::constructors::default();
        client
            .instantiate(&mut ink_e2e::alice(), constructor, 0, None)
            .await
            .expect("Instantiate `flipper` failed");

        Ok(())
    }

    #[ink_e2e::test]
    async fn e2e_second_instantiation(
        mut client: ink_e2e::Client<C, E>,
[...]

PR #1512, opened 2022-11-19
Handle `LangError` from instantiate

HCastano:
This is a follow up to #1504, in which `LangError`s were returned from constructors. In
that PR we had no way to read the output buffer from `seal_instantiate` and bubble any 
information back up to the contract caller.

This PR adds support for reading this output buffer for both infallible and fallible
constructors.

Because of type checking problems I've had to add a new set of APIs for handling fallible
constructors. I'm still looking for a way to unify them, as mentioned in [this comment](https://github.com/paritytech/ink/pull/1512/files#r1041508207).
If you have any ideas for improvements here I'd be happy to hear them.

I've changed the `create_builder` API slightly to be more in-line with the `call_builder`
one (i.e removing the generic in favour of a `returns` method). 

For infallible constructors, the main changes are:
- The `CreateBuilder` now returns a `ConstructorResult` (introduced in #1504)
    - This can optionally be handled using the new `try_instantiate` method

For fallible constructors:
- We've added `intantiate_fallible` and `try_instantiate_fallible` `CreateBuilder` methods
    - Like above, `ConstructorError`s are only bubbled up in the `try_*` variant
    - Edit: This API is clunky, but Andrew's proposing a nicer solution in
      https://github.com/paritytech/ink/pull/1591
- We've added a new environment API, `ink_env::instantiate_fallible_contract`

I've also updated the codegen so that `ContractRef`s can be used to instantiate contracts,
as an alternative to the "raw" `CreateBuilder` API.

Notes for reviewers:
- Keep in mind that there are two ways that the output buffer of `seal_instantiate` can
  be written to:
    - By the ink! codegen
    - Manually by the callee contract using `ink_env::return_value`
        - This second one is the one we need to worry about since people can (on purpose
          or by accident) encode erroneous values in there
- Related to the above, make sure to take a close look at the different output buffer
  decoding scenarios
- Take a look at the E2E tests to see if they make sense
- Right now the E2E test fail in the CI, not because they're actually broken but because
  of #1498
  (run them locally if you don't believe me 😆)


crates/env/src/call/create_builder.rs
```diff
@@ -121,12 +121,31 @@ where
     /// Instantiates the contract and returns its account ID back to the caller.
     #[inline]
     pub fn instantiate(&self) -> Result<R, crate::Error> {
-        crate::instantiate_contract(self).map(FromAccountId::from_account_id)
+        crate::instantiate_contract(self)
+            .map(|inner| {
+                inner.unwrap_or_else(|error| {
+                    panic!("Received a `LangError` while instatiating: {:?}", error)
```
ascjones, line 127:
```suggestion
                    panic!("Received a `LangError` while instantiating: {:?}", error)
```
Worth mentioning too in the comment that this method will panic on `LangError`, and the alternative of using `try_instantiate`

crates/env/src/call/create_builder.rs
```diff
@@ -339,54 +365,97 @@ where
     }
 }
 
-impl<E, GasLimit, Args, Salt, R>
+impl<E, CodeHash, GasLimit, Endowment, Args, Salt>
+    CreateBuilder<E, CodeHash, GasLimit, Endowment, Args, Salt, Unset<ReturnType<()>>>
+where
+    E: Environment,
+{
+    /// Sets the type of the returned value upon the execution of the call.
```
ascjones, line 373:
```suggestion
    /// Sets the type of the returned value upon the execution of the constructor.
```

crates/env/src/call/create_builder.rs
```diff
@@ -339,54 +365,97 @@ where
...
+    ///
+    /// Constructors are not able to return arbitrary values. Instead a successful call to a
+    /// constructor returns the address at which the contract was instantiated.
+    ///
+    /// Therefore this must always be a reference (i.e `ContractRef`) to the contract you're trying
+    /// to instantiate.
+    #[inline]
+    pub fn returns<R>(
+        self,
+    ) -> CreateBuilder<E, CodeHash, GasLimit, Endowment, Args, Salt, Set<ReturnType<R>>>
```
ascjones, line 385: Should we add the `R: FromAccountId<E>` constraint here too? I see we have it below on the final impl but it might give a more helpful compiler error if we do it here?

crates/env/src/call/create_builder.rs
```diff
@@ -339,54 +365,97 @@ where
...
     GasLimit: Unwrap<Output = u64>,
     Args: scale::Encode,
     Salt: AsRef<[u8]>,
-    R: FromAccountId<E>,
+    RetType: FromAccountId<E>,
 {
     /// Instantiates the contract using the given instantiation parameters.
     #[inline]
-    pub fn instantiate(self) -> Result<R, Error> {
+    pub fn instantiate(self) -> Result<RetType, Error> {
```
ascjones, line 446: Possibly worth copying my suggested comment update of `CreateParams::instantiate` which mentions that this panics in case of a `LangError`

crates/env/src/engine/on_chain/impls.rs
```diff
@@ -499,21 +499,30 @@ impl TypedEnvBackend for EnvInstance {
...
+
+        match instantiate_result {
+            Ok(()) => {
+                let account_id = scale::Decode::decode(&mut &out_address[..])?;
+                Ok(Ok(account_id))
+            }
+            Err(ext::Error::CalleeReverted) => {
+                // We don't wrap manually with an extra `Err` like we do in the `Ok` case since the
+                // buffer already comes back in the form of `Err(LangError)`
+                let out = scale::Decode::decode(&mut &out_return_value[..])?;
```
ascjones, line 521: Couldn't there exist a contract that would revert the contract but return `Ok(())` e.g.
```
::ink::env::return_value::<::ink::ConstructorResult<()>>(
    ::ink::env::ReturnFlags::new_with_reverted(true),
    Ok(()), 
);
```
This could mislead the caller into handling the `Ok` case?

examples/lang-err-integration-tests/call-builder/lib.rs
```diff
@@ -205,14 +207,7 @@ mod call_builder {
                     None,
                 )
                 .await
-                .expect("Client failed to call `call_builder::call_instantiate`.")
-                .value
-                .expect("Dispatching `call_builder::call_instantiate` failed.");
-
-            assert!(
-                call_result.is_some(),
```
ascjones, line 213: Should this check now be retained but changed to `call_result.value.is_none()` now that it returns `None` on success with no `LangError`

examples/lang-err-integration-tests/call-builder/lib.rs
```diff
@@ -254,14 +249,7 @@ mod call_builder {
                     None,
                 )
                 .await
-                .expect("Client failed to call `call_builder::call_instantiate`.")
-                .value
-                .expect("Dispatching `call_builder::call_instantiate` failed.");
-
-            assert!(
-                call_result.is_none(),
```
ascjones, line 262: Ditto here, check for `Some(LangError)`

ascjones (commented):
Just the question mark about the handling of the decoding of the `LangError`, other than that it looks good.

examples/lang-err-integration-tests/call-builder/lib.rs
HCastano: Ah shit 🙈

crates/env/src/engine/on_chain/impls.rs
HCastano: Hmm yeah this is a good observation. I've added a test that shows that you can indeed revert from a constructor and continue execution by returning `Ok(AccountId)`. I've also added a quick patch which asserts that the content in the buffer is an error (the test passes without that patch).

I'm not entirely sure what the implications of either case are though...
- In the `Revert + Ok` case, misleading the caller seems like the main reason to do this, but maybe there are legit reasons to do encode an `Ok` back? I can't think of any though.
- In the `Revert + Err` case, the callee can encode the "wrong" `LangError` into the output buffer, potentially misleading the caller too, and there's nothing we can do to help there.

Either way it'll be up to the contract developer to make sure they understand that there are risks to calling other contracts.

HCastano: @athei would like to hear your opinion here

crates/env/src/engine/on_chain/impls.rs
```diff
@@ -517,8 +517,12 @@ impl TypedEnvBackend for EnvInstance {
...
             Err(ext::Error::CalleeReverted) => {
                 // We don't wrap manually with an extra `Err` like we do in the `Ok` case since the
-                // buffer already comes back in the form of `Err(LangError)`
-                let out = scale::Decode::decode(&mut &out_return_value[..])?;
+                // buffer already comes back in the form of `Err(LangError)` (assuming it's encoded
+                // by the ink! codegen and not the contract).
+                let out = ink_primitives::ConstructorResult::<E::AccountId>::decode(
+                    &mut &out_return_value[..],
+                )?;
+                assert!(out.is_err(), "The callee reverted, but did not encode an error in the output buffer.");
```
ascjones, line 525: I wonder if this would itself be considered a type of `LangError`. e.g. `LangError::InvalidConstructorRevertedReturnValue` (obvs less verbose)

crates/env/src/engine/on_chain/impls.rs
ascjones: The assumption is that the user is not calling `return_value` directly, and if they are they are using it in an ABI compatible way which is either `Revert + 'Err(LangError)` or `Revert + Ok(Err(ContracErr))`.

So perhaps we could return an `Err(LangError)` here if this ABI is not satisfied, e.g. as mentioned in https://github.com/paritytech/ink/pull/1512#discussion_r1036968362

crates/env/src/engine/on_chain/impls.rs
ascjones: In fact this looks like this `assert` would fail if a fallible constructor returns an error: https://github.com/paritytech/ink/blob/507d01dcb2848c3deb16691bac9ec0c15bcda900/crates/ink/codegen/src/generator/dispatch.rs#L635-L643

Because it will revert, but the outer `Ok` is hardcoded.

crates/env/src/engine/on_chain/impls.rs
ascjones: Okay maybe what we want is something along the lines of
```
let out = ink_primitives::ConstructorResult::<()>::decode(
    &mut &out_return_value[..],
)?;
match out {
    // This branch when a fallible constructor returns `Err`
    Ok(()) => Err(ext::Error::CalleeReverted),
    /// This branch handles an actual `LangErr`
    Err(lang_err) => Ok(Err(lang_err))
}
```

crates/env/src/engine/on_chain/impls.rs
HCastano: Thanks for bringing up the case for fallible constructors. 

Even without the `assert!` this would've failed to decode. This is because with fallible
constructors we need to decode into a `ConstructorResult<Result<(), E>>`,
instead of a `ConstructorResult<()>` or `ConstructorResult<AccountId>`.

I need to think a bit more about what changes will be required.

crates/env/src/engine/on_chain/impls.rs
```diff
@@ -499,21 +499,117 @@ impl TypedEnvBackend for EnvInstance {
...
+                    .expect("Failed to decode into `ConstructorResult<()>`");
+
+                match out {
+                    ink_primitives::ConstructorResult::Ok(()) => {
+                        let out = <ink_primitives::ConstructorResult<
+                            core::result::Result<(), ContractError>,
+                        > as scale::Decode>::decode(
+                            &mut &out_return_value[..]
+                        )
+                        .expect("Failed to decode into a `ContractError`");
```
HCastano, line 589: @ascjones the fact that both these decode calls work is sketching me out. From what I remember `Ok(())` would be encoded as `0`, and `Ok(Ok())` as `0, 0`, right?

ascjones: SCALE parses the bytes as it sees them so for `0, 0` the `Result<()>` parser just reads the first byte and returns successfully. 

However if the encoded bytes are `0`, then the nested `Result<Result<()>>` will fail because the it will try to decode the inner `Result` and there will be no more data.

`0, 0` -> `Result<()>` :heavy_check_mark:  `Result<Result<()>>` :heavy_check_mark: 
`0`     -> `Result<()> ` :heavy_check_mark: `Result<Result<()>>` :x:

crates/env/src/call/create_builder.rs
```diff
@@ -339,54 +395,140 @@ where
...
+    RetType: FromAccountId<E>,
+    ContractError: scale::Decode,
+{
+    /// Instantiates the contract using the given instantiation parameters.
+    ///
+    /// # Note
+    ///
+    /// On failure this returns an [`ink_primitives::LangError`] which can be handled by the caller.
+    #[inline]
+    pub fn try_instantiate_fallible(
```
HCastano, line 524: Ideally we don't have this as part of the `CreateBuilder`, and instead we allow the implementation to dispatch the correct method based off the concrete types used (so whatever is set with `returns::<T>()`. 

I haven't been able to figure out a way to do this yet, @ascjones you got any ideas?

HCastano:
@ascjones still a bit of a WIP, but can you take another look for me pls 🙏

crates/env/src/engine/on_chain/impls.rs
```diff
@@ -573,38 +572,39 @@ impl TypedEnvBackend for EnvInstance {
...
-                        }
+                let decoding_result = <ink_primitives::ConstructorResult<
+                    core::result::Result<(), ContractError>,
+                > as scale::Decode>::decode(
+                    &mut &out_return_value[..]
+                );
+
+                match decoding_result {
+                    Ok(constructor_result) => {
+                        let contract_result = constructor_result.expect(
```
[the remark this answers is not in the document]
HCastano: Yep, that's the expected behaviour right now. See [this test](https://github.com/paritytech/ink/pull/1512/files#diff-08296d3fe4e7f90d1194dfe93440e596383ccdca622436ea75a420141de36334R443).

The panicking isn't necessarily ideal, but it is an "invalid" scenario

crates/env/src/engine/on_chain/impls.rs
```diff
@@ -573,38 +572,39 @@ impl TypedEnvBackend for EnvInstance {
...
+                        let contract_error = contract_result.expect_err(
+                            "Since the contract reverted, we only expect an `Error` from the constructor. \
+                             Otherwise we would be in the `AccountId` branch.");
+
+                        Ok(ink_primitives::ConstructorResult::Ok(Err(contract_error)))
                     }
-                    ink_primitives::ConstructorResult::Err(lang_err) => {
-                        Ok(ink_primitives::ConstructorResult::Err(lang_err))
+                    Err(_) => {
+                        // If we hit this branch it likely means dispatch failed, but we need to
```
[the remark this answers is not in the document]
HCastano: No, because in the branch above we would try and decode into a `Result<Result<()>>`, which as you mentioned earlier we'd fail to decode if we only had a `Result<()>` encoded into the buffer. In the piece of code you linked to we're encoding a `Result<(), LangError>` into the buffer.

This is also why we do another decode operation in this branch.

crates/env/src/engine/on_chain/impls.rs
```diff
@@ -573,38 +572,39 @@ impl TypedEnvBackend for EnvInstance {
...
-                            _ => unreachable!("We already decoded into a `ConstructorResult::Ok` earlier."),
-                        }
+                let decoding_result = <ink_primitives::ConstructorResult<
+                    core::result::Result<(), ContractError>,
+                > as scale::Decode>::decode(
+                    &mut &out_return_value[..]
+                );
+
+                match decoding_result {
+                    Ok(constructor_result) => {
```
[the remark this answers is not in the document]
HCastano: Probably, but at first I wanted to be more verbose with the `match` statements so I could spot the different codepaths we go down

crates/env/src/engine/on_chain/impls.rs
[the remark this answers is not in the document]
ascjones: In that case it is always the `Err` variant though, so it doesn't matter what type is specified as the `Ok` variant in the `return_value` generic parameters. It could be `()`, `Result<()>`, `Result<Result<Result<()>>`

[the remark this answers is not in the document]
ascjones: See https://github.com/paritytech/ink/pull/1532/files#diff-b240442c842f1df1ff5fdebd6d455289e5cfbbca404b8925e01ce4446db30a8eR128 for how I understand this test case. Correct me if I am still wrong about this.

crates/env/src/engine/mod.rs
```diff
@@ -37,3 +41,109 @@ cfg_if! {
...
+{
+    let out: ConstructorResult<Result<(), ContractError>> =
+        scale::Decode::decode(out_return_value)?;
+
+    match out {
+        ConstructorResult::Ok(ContractResult::Ok(())) => {
+            // Since the contract reverted we don't expect an `Ok` return value from the
+            // constructor, otherwise we'd be in the `AccountId` decoding branch.
+            Err(crate::Error::Decode(
+                "TODO: probably shouldn't be a `Decode` error".into(),
```
HCastano, line 66: @ascjones I'm tempted to just panic here like we do in the infallible constructor case. It doesn't seem appropriate to let the caller handle this since it's either bad encoding by a) the codegen or b) the callee

examples/lang-err-integration-tests/call-builder/lib.rs
```diff
@@ -201,35 +244,249 @@ mod call_builder {
...
+            let call_result = client
+                .call(&mut ink_e2e::bob(), call, 0, None)
+                .await
+                .expect(
+                    "Client failed to call `call_builder::call_instantiate_fallible`.",
+                )
+                .value
+                .expect("Dispatching `call_builder::call_instantiate_fallible` failed.");
+
+            assert!(
```
HCastano, line 483: I wanna point out that this is technically incorrect, since the dispatch to `try_revert_new` did succeed, and then callee manually put this in the buffer.

The caller would need to make sure that they're not calling contracts which encode invalid values to the error output buffer.

crates/env/src/engine/mod.rs
ascjones: Panicking would be okay, the alternative would be to raise a `LangError`. It means we capture the reason why this failed, and is in keeping with the semantics of the top level `try_instantiate` method,

The question is whether it fits in the category of `LangError`s that we had in mind. I'm not 100% sure on this but there is at least a case to be made for it - it is a possible scenario which we can handle gracefully in this branch.

ascjones:
Ping me when you're back and we can catch up on this and get it merged

ascjones:
See https://github.com/paritytech/ink-workshop/pull/31

crates/env/src/call/create_builder.rs
ascjones: I might have an idea for how to handle this, currently verifying whether it will work or not

HCastano:
> See [paritytech/ink-workshop#31](https://github.com/paritytech/ink-workshop/pull/31)
> 
> _Edit_, sorry I realize this is for a `call` and not an `instantiate`. Still, it highlights a potential footgun if we are requiring the user to explicitly specify the `Result<_, LangError>`

Yeah, this is something I wanted to address with https://github.com/paritytech/ink/pull/1525

crates/env/src/call/create_builder.rs
ascjones: *WIP* here: https://github.com/paritytech/ink/pull/1591

ascjones: Ok the PR for my idea basically works now: https://github.com/paritytech/ink/pull/1591. So please have a look and let me know what you think.

ascjones (approved):
LGTM!

The dual method API with `fallible` alternatives is a bit clunky, however #1591 provides a solution to that.

Issue #1474, opened 2022-11-05
CI: Migrate `cargo test` to `cargo-nextest`


cmichi:
## Motivation

[`cargo-nextest`](https://nexte.st) can be used instead of `cargo test`. It runs the tests as well, but is more performant. We should replace our usage of `cargo test` with it.

## ToDo

- [ ] Install `cargo-nextest` in the Docker image that use in our CI. The image is called `ink-ci-linux`, you can find the Dockerfile [here](https://github.com/paritytech/scripts/blob/master/dockerfiles/ink-ci-linux/Dockerfile). Just add a `cargo install` there.
- [ ] Once merged you need to rebuild the Docker image. This is done every night. You can also trigger it by hand [here](https://gitlab.parity.io/parity/mirrors/scripts/-/pipeline_schedules), by click the play button: 
<img width="1054" alt="Bildschirmfoto 2022-11-05 um 21 49 57" src="https://user-images.githubusercontent.com/241530/200140789-2a25abc0-37c8-4d79-8f36-3787315ffcb0.png">. If you don't see a play button, ask in the "ink! <> CI/CD" channel. If the last build failed (as in this picture), ask the current 🦖 for help.
- [ ] Once the Docker build succeeded you can create an ink! PR that replaces `cargo test` in our `.gitlab-ci.yml` with `cargo-nextest` invocations.
- [ ] Check if our GitHub actions under `.github/` also need to be migrated.

## Follow-up

In case this ticket worked out good and we don't see issues: please create a follow-up issue for `cargo-contract`.

cmichi:
I'm closing this PR and the issue as the results are inconclusive, we might revisit after https://github.com/paritytech/ink/issues/1520.

PR #1514, opened 2022-11-22
Replacing `cargo test` with `cargo nextest`

lean-apple:
3rd step in the migration to cargo nextest, as presented here https://github.com/paritytech/ink/issues/1474

One step of the CI can't be replaced at the moment with the flag --doc as it is explained here, 
https://nexte.st/

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -238,7 +238,7 @@ dylint:
     - cargo clippy --verbose -- -D warnings;
     # Needed until https://github.com/mozilla/sccache/issues/1000 is fixed.
     - unset RUSTC_WRAPPER
-    - cargo test --verbose --all-features
+    - cargo nextest run --verbose --all-features
 
 #### stage:                        workspace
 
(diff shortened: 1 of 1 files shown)


HCastano:
Do you have any benchmarks between this and the previous setup?

lean-apple:
Except the official ones, https://nexte.st/book/benchmarks.html, if we compare the times for the different steps in the CI, using cargo nextest, it is faster with most of the time but not always. Also the difference is not huge. We should probably repeat a certain number of times to get more proper figures/data (or at least schedule a job once a day during one-two weeks).

HCastano:
> Except the official ones, https://nexte.st/book/benchmarks.html, if we compare the times for the different steps in the CI, using cargo nextest, it is faster with most of the time but not always. Also the difference is not huge. We should probably repeat a certain number of times to get more proper figures/data (or at least schedule a job once a day during one-two weeks).

Yes, I'd like to see some stats using our actual workflows. We may want to run the `nextest` variants alongside our normal workflow, but have them be optional steps, so we can more easily compare them

HCastano:
@lean-apple are there any updates here?

lean-apple:
Not really concluant updates, neither the times we got with the scheduled job here, https://gitlab.parity.io/parity/mirrors/ink/-/pipelines, neither my local tests with a common script on some steps using nextest, show a stable decrease of CI times,
With @cmichi we suspect that some tests aren't run in parallel, maybe a way to improve and maximise the use of nextest stayed here https://github.com/paritytech/ink/issues/1520

HCastano:
Okay I'm gonna close this then. Can you make a brief write-up of your findings in #1474 in case we want to revisit this in the future?

PR #1518, opened 2022-11-25
E2E: remove `smart-bench-macro`, use existing codegen

### LINKED ISSUE #1421: E2E testing: Fork `smart-bench-macro`
Follow-up to https://github.com/paritytech/ink/pull/1395.

The `smart-bench-macro` from https://github.com/paritytech/smart-bench needs to be forked and the dependency in `crates/env/Cargo.toml` replaced with a local crate.

I've got a branch `cmichi-ink-e2e-test-mvp` in that repo that already contains necessary changes.

The fork has to be cleaned up for the stuff that is actually needed in this repo.

ascjones:
Closes #1421. Will allow `ink_e2e` crate to be released.

- Remove dependency on unreleased `smart-bench-macro` crate
- Instead use existing codegen which is used for building messages from `ContractRef`

Note that we can improve this API, especially for building messages, but my goal here is primarily to remove the dependency and make no changes to the existing codegen, which would make this PR a bit more complicated to implement and review.

### Constructors
Nice and simple and clean, can just use the inherent method of the same name on `ContractRef`
```rust
let constructor = GiveMeRef::new();`
```

### Messages
A bit more complicated, I've tried to make it as simple as possible without making any modifications to the existing codegen
```rust
let transfer = ink_e2e::build_message::<GiveMeRef>(contract_acc_id)
    .call(|contract| contract.give_me(120));
```
This can/should be improved in the future. See https://github.com/paritytech/ink/pull/1518#discussion_r1035070784


crates/e2e/src/builders.rs
```diff
@@ -0,0 +1,176 @@
...
+    Unset<<E as Environment>::Balance>,
+    Set<ExecutionInput<Args>>,
+    Unset<ink_env::call::state::Salt>,
+    R,
+>;
+
+/// Shim onto the `CreateBuilder` to allow access to the `ExecutionInput` arguments.
+pub struct ConstructorBuilder<E: Environment, Args: Encode, R> {
+    inner: CreateBuilderPartial<E, Args, R>,
+}
```
HCastano, line 48: I don't really understand the need for this. You're exposing the `exec_input` getter anyways, and you should be able to use the `CreateBuilder` directly

crates/e2e/src/builders.rs
```diff
@@ -0,0 +1,176 @@
...
+            .code_hash(ink_primitives::Clear::clear())
+            .salt_bytes(Vec::new())
+            .params()
+            .exec_input()
+            .encode()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Message<E: Environment, RetType> {
```
HCastano, line 73: Would be nice to have some docs here

crates/e2e/src/builders.rs
```diff
@@ -0,0 +1,176 @@
...
+        &self.exec_input
+    }
+}
+
+/// Convenience method for building messages for the default environment
+pub fn build_message<Ref>(
+    account_id: <ink_env::DefaultEnvironment as Environment>::AccountId,
+) -> MessageBuilder<ink_env::DefaultEnvironment, Ref>
+where
+    Ref: TraitCallBuilder + FromAccountId<ink_env::DefaultEnvironment>,
```
HCastano, line 126: Seems strange to me that everything else is generic over the environment but here we enforce the `DefaultEnvironment`.

I guess most users would end up using this as `build_message<DefaultEnvironment, ContractRef>` anyways, so maybe it's okay

crates/e2e/src/builders.rs
```diff
@@ -0,0 +1,176 @@
...
+            .exec_input()
+            .encode()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Message<E: Environment, RetType> {
+    account_id: E::AccountId,
+    exec_input: Vec<u8>,
+    marker: std::marker::PhantomData<RetType>,
```
HCastano, line 76: nit

```suggestion
    _return_type: std::marker::PhantomData<RetType>,
```

crates/e2e/src/builders.rs
```diff
@@ -0,0 +1,176 @@
...
+    account_id: E::AccountId,
+    contract_ref: Ref,
+}
+
+impl<E, Ref> MessageBuilder<E, Ref>
+where
+    E: Environment,
+    Ref: TraitCallBuilder + FromAccountId<E>,
+{
+    pub fn from_account_id(account_id: E::AccountId) -> Self {
```
HCastano, line 142: Methods should be documented

crates/e2e/src/builders.rs
```diff
@@ -0,0 +1,176 @@
...
+            account_id,
+            contract_ref,
+        }
+    }
+
+    pub fn account_id(&self) -> &E::AccountId {
+        &self.account_id
+    }
+
+    pub fn call<F, Args, RetType>(mut self, mut message: F) -> Message<E, RetType>
```
HCastano, line 154: Would be nice to even have a doc comment showing the expected usage pattern

HCastano (commented):
I've got a couple of questions, but this generally looks good!

The expanded codegen looks _way_ cleaner, and being able to use the `Environment` types is great 😄

crates/e2e/src/builders.rs
ascjones: Originally I wanted to add the ability to *optionally* set the other properties e.g. `endowment`, `salt` etc but decided it was beyond the scope of this PR. However I kept the builder wrapper, which as you say is no longer necessary. 

I have changed it to a plain function and we can always add the builder back in the future if necessary.

crates/e2e/src/builders.rs
ascjones: Indeed it is a convenience method so the user does not have to specify that generic parameter, which at the moment is hardcoded in the codegen https://github.com/paritytech/ink/blob/e1f10786be6399cafd68899a82635e6968d8075b/crates/e2e/macro/src/codegen.rs#L151.

Trying to make the creating of call messages as ergonomic as possible. Not there yet but this is a start.

crates/e2e/src/builders.rs
HCastano: @ascjones FYI, I made this compile

crates/env/src/call/call_builder.rs
```diff
@@ -61,7 +61,7 @@ where
 
     /// Returns the execution input.
     #[inline]
-    pub(crate) fn exec_input(&self) -> &ExecutionInput<Args> {
+    pub fn exec_input(&self) -> &ExecutionInput<Args> {
```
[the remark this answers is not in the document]
HCastano: It's gonna be a little more noise in the docs, but yeah let's have them documented

HCastano (approved):
Nice!

crates/e2e/src/builders.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn exec_input(&self) -> &[u8] {
+        &self.exec_input
+    }
+}
+
+/// Convenience method for building messages for the default environment.
+///
+/// # Note
+/// This is hardcoded to [`ink_env::DefaultEnvironment`] so the user does not have to specify this
+/// generic parameter, which currently is hardcoded in the e2e test `codegen`.
```
HCastano, line 87:
```suggestion
/// # Note
///
/// This is hardcoded to [`ink_env::DefaultEnvironment`] so the user does not have to specify this
/// generic parameter, which currently is hardcoded in the E2E testing suite.
```

crates/e2e/src/builders.rs
ascjones: :see_no_evil:

crates/e2e/src/builders.rs
```diff
@@ -121,10 +121,34 @@ where
...
+    /// #
+    /// # fn message_builder_doc_test() {
+    /// #     use my_contract::MyContractRef;
+    /// #     let contract_acc_id = ink_primitives::AccountId::from([0x00; 32]);
+    /// ink_e2e::MessageBuilder::<ink::env::DefaultEnvironment, MyContractRef>::from_account_id(
+    ///     contract_acc_id,
+    /// )
+    /// .call(|contract| contract.message());
+    /// # }
+    /// ```
```
ascjones, line 151: :heart:

PR #1519, opened 2022-11-25
Examples CI: ignore lang-err-integration-tests dir :t-rex: 

ascjones:
Fix CI error https://github.com/paritytech/ink/actions/runs/3547091364/jobs/5956837605#step:14:5234.

For now it will just ignore that directory, I don't think we need to attempt to compile those examples for all OSs.

We should attempt to make these CI scripts less brittle and traverse the `examples` dir. I will open an issue for that, https://github.com/paritytech/ink/issues/1520

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -112,6 +112,7 @@ jobs:
            cargo contract ${{ matrix.job }} --verbose --manifest-path examples/upgradeable-contracts/set-code-hash/updated-incrementer/Cargo.toml;
            foreach ($example in Get-ChildItem  -Directory "examples\*") {
                if ($example -Match 'upgradeable-contracts') { continue }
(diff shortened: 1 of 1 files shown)


HCastano:
What's the underlying problem here though? All these examples should (and do) build on a Mac

ascjones:
>     * Why does this break the main examples loop on a Mac?
It's not a specific problem for mac osx, it's just that the `examples.yml` doesn't run on PRs (only runs once merged) so this file had not been updated to take account of the new subdirectory. And mac osx is the first job to run, I guess the windows one would fail too if it got that far

>     * Why are the delegator subcontracts building fine?
Probably because there is a virtual manifest in that directory so the `cargo` commands do not fail. As you allude to in the other issue, that might be a good way to solve this: ensure each subdir has a virtual manifest.

lean-apple (approved):

PR #1523, opened 2022-11-29
`[ink_e2e]` utilize `contract-build` crate

ascjones:
https://github.com/paritytech/cargo-contract/pull/787 introduces the new `contract-build` library, which we can utilize here instead of depending on a user installed version of `cargo-contract`.


crates/e2e/macro/src/codegen.rs
```diff
@@ -174,47 +174,45 @@ impl InkE2ETest {
...
+        manifest_path,
+        verbosity: Verbosity::Default,
+        build_mode: BuildMode::Debug,
+        network: Default::default(),
+        build_artifact: BuildArtifacts::All,
+        unstable_flags: Default::default(),
+        optimization_passes: Default::default(),
+        keep_debug_symbols: false,
+        lint: false,
+        output_type: Default::default(),
```
HCastano, line 198: This was `OutputType::Json` before, I guess changing this is fine now since we don't manually need to parse the build output, right?

HCastano (commented):
I see you had a bigger vision in mind already with https://github.com/paritytech/cargo-contract/pull/787 👌 

I'd approve, but I haven't been able to build this locally on my Mac or using Docker wit the latest production CI image. I keep running into the following issue

```
  --> lib.rs:57:9
   |
57 |         #[ink_e2e::test]
   |         ^^^^^^^^^^^^^^^^
   |
   = help: message: contract build for Cargo.toml failed: Mismatching versions of `parity-scale-codec` were found!
           Please ensure that your contract and your ink! dependencies use a compatible version of this package.

   --> lib.rs:109:9
    |
109 |         #[ink_e2e::test]
    |         ^^^^^^^^^^^^^^^^
    |
    = help: message: Once instance has previously been poisoned
 ```
 
CI passes and local output from `cargo-tree` looks okay, so not sure what's up, but worth looking into before we merge this

File: crates/e2e/macro/src/codegen.rs (hunk 1 above)
Comment (ascjones) [reply crates/e2e/macro/src/codegen.rs:198]: Correct :+1:

Comment (ascjones) [reply crates/e2e/macro/src/codegen.rs]: Based on this  observation I've made all options explicit instead of relying on `Default::default()` which could yield some invisible breaking changes, e.g. I recently changed `OptimizationPasses` to an `Option`.

**Comment by ascjones** [comment]
> I see you had a bigger vision in mind already with [paritytech/cargo-contract#787](https://github.com/paritytech/cargo-contract/pull/787) ok_hand
> 
> I'd approve, but I haven't been able to build this locally on my Mac or using Docker wit the latest production CI image. I keep running into the following issue
> 
> ```
> error: custom attribute panicked
>   --> lib.rs:57:9
>    |
> 57 |         #[ink_e2e::test]
>    |         ^^^^^^^^^^^^^^^^
>    |
>    = help: message: contract build for Cargo.toml failed: Mismatching versions of `parity-scale-codec` were found!
>            Please ensure that your contract and your ink! dependencies use a compatible version of this package.
> 
> error: custom attribute panicked
>    --> lib.rs:109:9
>     |
> 109 |         #[ink_e2e::test]
>     |         ^^^^^^^^^^^^^^^^
>     |
>     = help: message: Once instance has previously been poisoned
> ```
> 
> CI passes and local output from `cargo-tree` looks okay, so not sure what's up, but worth looking into before we merge this

It's a misleading error, it is in fact because of using `cargo +nightly test`. See this issue https://github.com/paritytech/cargo-contract/issues/746.

The difference now is that previously we invoked `cargo +stable contract`, now it uses the lib the `cargo` invocation inherits the `+nightly` so we get this error.

So the question is how to handle this. We could:

1. Force `stable` toolchain on underlying contract build, even if the parent `cargo` invocation has a different toolchain
2. Inherit the parent toolchain but throw a proper error ("must use stable" instead of "mismatching dependencies") 
3. Inherit the parent toolchain and allow to build with nightly (just check the minimum version)

`3.` is the most `cargo` like way to do it, but I can't remember exactly why we are enforcing using the stable toolchain.

**Comment by HCastano** [comment]
Sidenote, I remember why I was running the E2E tests with `nightly`. You can't do `cargo contract test --features e2e-tests` since the `--features` flag isn't supported by `cargo-contract`. I opted to run with `cargo +nightly test` instead (although `cargo +stable test` also would've worked)

File: crates/e2e/macro/Cargo.toml [hunk 2]
```diff
@@ -21,10 +21,11 @@ proc-macro = true
 
 [dependencies]
 ink_ir = { version = "4.0.0-beta", path = "../../ink/ir" }
+contract-build = "2.0.0-beta"
```
HCastano, line 24: So we can either pull in a Git dep here so that we can use `+nightly` again, or wait until a new version of `cargo-contract` is released. I'm okay with waiting for a new version of `cargo-contract`

HCastano (approved):
Nice!

ascjones:
> Sidenote, I remember why I was running the E2E tests with `nightly`. You can't do `cargo contract test --features e2e-tests` since the `--features` flag isn't supported by `cargo-contract`. I opted to run with `cargo +nightly test` instead (although `cargo +stable test` also would've worked)

We should definitely pass the features through then. For me `cargo test --features e2e-tests` works though (without nightly), curious what requires the `nightly` features in this case.

crates/e2e/macro/Cargo.toml
ascjones: Updated to the latest `2.0.0-beta.1` release which allows `+nightly`. Also `--features` can be passed through now anyway from the command line so you can also do `cargo contract test --features e2e-tests`

PR #1524, opened 2022-11-30
`[ink_e2e]` housekeeping

ascjones:
- Remove unused traits and struct, following on from #1518 
- Remove `mut` from signer ref, not required since the signer doesn't increment the nonce anymore.


SkymanOne (approved):
Thanks!

PR #1525, opened 2022-11-30
Clean up `CallBuilder` `return()` type

HCastano:
In #1450 I had to change the return type used by `CallBuilder` callers in order to be
able to encode/decode the new `MessageResult` return type.

This PR bakes the assumption that all messages **must** return a `MessageResult` into the
into the `CallBuilder`'s call stack, allowing callers to specify only the message return
type again. I.e, instead of `returns::<Result<R, LangError>>()` you can again specify just
`returns::<R>()`.

A breaking change that ended up being introduced here is that `ink_env::invoke_contract` now
returns a `MessageResult<R>`, whereas it previously returned just `R`.

I'm not sure if we need to do anything for for the `DelegateCall` implementation, but we can
leave that as a follow-up since there are already examples of people stumbling due to (the
lack of) this: https://github.com/paritytech/ink-workshop/pull/31.


crates/env/src/call/call_builder.rs
```diff
@@ -109,7 +109,28 @@ where
...
+    /// # Panics
+    ///
+    /// This method panics if it encounters an [`ink_primitives::LangError`]. If you want to handle
+    /// those use the [`try_invoke`][`CallParams::try_invoke`] method instead.
     pub fn invoke(&self) -> Result<R, crate::Error> {
+        crate::invoke_contract(self).map(|inner| {
+            inner.expect(
+                "Handling `LangError`s is not supported by `CallParams::invoke`, use \
+                `CallParams::try_invoke` instead.",
+            )
```
athei, line 122: This drops the `LangError`. I think we should just `unwrap` or compose a meaningful error message that includes the error: "Call failed with error {}"

athei, line 122: This drops the `LangError`. I think we should just `unwrap` or compose a meaningful error message that includes the error: "Call failed with error {}"

crates/env/src/api.rs
```diff
@@ -268,7 +268,9 @@ where
 /// - If the called contract execution has trapped.
 /// - If the called contract ran out of gas upon execution.
 /// - If the returned value failed to decode properly.
-pub fn invoke_contract<E, Args, R>(params: &CallParams<E, Call<E>, Args, R>) -> Result<R>
+pub fn invoke_contract<E, Args, R>(
+    params: &CallParams<E, Call<E>, Args, R>,
+) -> Result<ink_primitives::MessageResult<R>>
```
athei, line 273: I know not related to this PR but do we really need these standalone functions? I don't think we want multiple ways to do the same thing.

crates/env/src/call/call_builder.rs
```diff
@@ -109,7 +109,28 @@ where
     /// Invokes the contract with the given built-up call parameters.
     ///
     /// Returns the result of the contract execution.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if it encounters an [`ink_primitives::LangError`]. If you want to handle
+    /// those use the [`try_invoke`][`CallParams::try_invoke`] method instead.
     pub fn invoke(&self) -> Result<R, crate::Error> {
+        crate::invoke_contract(self).map(|inner| {
```
athei, line 118: I think we should take the chance to also unwrap the error from pallet-contracts while we are breaking things. Let's just make the default path as safe and frictionless as possible.

HCastano: Yeah good idea. I'll do this in a small follow-up so that the `CreateBuilder` methods can also be included in the clean-up

crates/env/src/api.rs
HCastano: I'm not sure how much use external developers are making of these specific APIs, but it's
probably important to leave them in. This at least gives developers a chance to optimize
their contracts, and third-party tooling to improve upon our abstractions.

A counter-argument here might be that this API is already opinionated (e.g the use of
`CallParams`) and doesn't give developers enough room to optimize or improve upon it. In
this case, maybe we want to expose even lower level APIs, but I'm not sure we want to go
down that route

athei: I can't see how those stand alone function are more low level. Both operate on the same type `CallParams`. The difference is just inherent vs free function.

athei (approved):

crates/env/src/call/call_builder.rs
```diff
@@ -597,9 +657,23 @@ where
...
+    /// those use the [`try_fire`][`CallBuilder::try_fire`] method instead.
     pub fn fire(self) -> Result<(), Error> {
         self.params().invoke()
     }
+
+    /// Invokes the cross-chain function call.
+    ///
+    /// # Note
+    ///
+    /// On failure this returns an [`ink_primitives::LangError`] which can be handled by the caller.
```
ascjones, line 673: It would be more accurate to clarify that the *inner* Result will be a `LangError`, and the outer Result is an environmental error.

ascjones (approved):

crates/env/src/call/call_builder.rs
HCastano: Yep, I'll fix it in the follow-up that removes the `EnvError` from the default execution path

DoubleOTheven:
The change mentioned above: `returns a MessageResult<R>, whereas it previously returned just R` has been incredibly helpful on the front end. It is probably my favorite feature that ink! offers over Solidity. Nice work 👏🏼

Issue #1422, opened 2022-09-28
E2E testing: Fetch runtime metadata automatically


cmichi:
Follow-up to https://github.com/paritytech/ink/pull/1395.

The current E2E testing framework still contains a hardcoded
```rust
#[subxt::subxt(runtime_metadata_path = "metadata/contracts-node.scale")]
pub(super) mod api {}
```

Instead of requiring this test fixture in the `ink` repository (exported via `subxt metadata > contracts-node.scale` while the node is running), we should fetch it automatically, getting rid of requiring users to do this step.

With https://github.com/paritytech/subxt/commit/d03e599684ade07eb65628ed848ae0844000a547, `subxt` got the possibility of supplying a 
```rust
#[subxt::subxt(runtime_metadata_url = "wss://rpc.polkadot.io:443")]
```

We already support `#[ink_e2e::test(ws_url = "…")]`, so ideally we would use that supplied URL for the `runtime_metadata_url`.

PR #1526, opened 2022-12-01
`[ink_e2e]` delete subxt metadata, use dynamic values

ascjones:
Closes https://github.com/paritytech/ink/issues/1422.

Instead of downloading the metadata from the node, this PR removes the need for compile time metadata entirely.

The extrinsics themselves are already using locally defined types, the only remaining usage of the statically generated runtime types were:

1. Fetching an account balance
2. Detecting `ExtrinsicFailed` events

`1.` can be done using dynamic decoding to extract the `AccountInfo::data`, `AccountData::free` field.
`2.` can be done by just checking for the pallet and event variant name.


HCastano (approved):
Nice 👌

Issue #1530, opened 2022-12-08
Incorrect codegen for trait methods


kvinwang:
**Describe the bug**
I am trying out the ink 4.0-beta.

Given the following trait definition:
```rust
#[ink::trait_definition(namespace = "pink_system")]
pub trait System {
    #[ink(message)]
    fn version(&self) -> (u16, u16);
}
```
The generated forwarder is:
```rust
impl<E> System for __ink_TraitCallBuilderSystem<E>
where
    E: ::ink::env::Environment,
{
    fn version(&self) -> Self::versionOutput {
        ::ink::env::call::build_call::<Self::Env>()
            .call_type(
                ::ink::env::call::Call::new()
                    .callee(::ink::ToAccountId::to_account_id(self)),
            )
            .exec_input(::ink::env::call::ExecutionInput::new(
                ::ink::env::call::Selector::new([0x87_u8, 0xC9_u8, 0x8A_u8, 0x8D_u8]),
            ))
            .returns::<(u16, u16)>()
    }
}
```
And also the metadata.json:
```json
{
  "args": [],
  "docs": [],
  "label": "System::version",
  "mutates": false,
  "payable": false,
  "returnType": {
    "displayName": [],
    "type": 6
  },
  "selector": "0x87c98a8d"
},
...
{
"id": 6,
"type": {
  "def": {
    "tuple": [
      7,
      7
    ]
  }
}
},
```

**Expected behavior**
It should be:
```.returns::<Result<(u16, u16),LangError>>()```
and ditto in the metadata.

ascjones:
I see the handling of `LangError` for traits was left as a follow-up in the @HCastano's original PR: https://github.com/paritytech/ink/pull/1450.

kvinwang:
Hey guys,

I've noticed that `4.0.0-rc` has been released.
Are the follow-ups in #1450 been done? This issue is definitely blocking us to upgrade to ink-4.0.

ascjones:
Yes I believe the issues you have raised here are now fixed via https://github.com/paritytech/ink/pull/1525 (call builder return type) and #1531 (metadata). Can you double check and then we can close this issue?

kvinwang:
Great, it works with `4.0.0-rc`.

PR #1531, opened 2022-12-08
Fix trait message return type metadata

ascjones:
Follow up to https://github.com/paritytech/ink/pull/1450, addresses the metadata part of https://github.com/paritytech/ink/issues/1530.

Messages now [return `Result<_, LangError>`.](https://github.com/paritytech/ink/blob/507d01dcb2848c3deb16691bac9ec0c15bcda900/crates/ink/codegen/src/generator/dispatch.rs#L837) The metadata was updated to reflect this for inherent methods but not for trait methods.

Note that as mentioned in the #1530, this *only* fixes the external metadata, and does not address the codegen for cross contract calling.

**Additionally** this PR copies a similar test for fallible constructor metadata from one of the compile tests, and allows it to use the helpers for extracting result types here.

### DIFF
--- a/crates/ink/tests/return_type_metadata.rs
+++ b/crates/ink/tests/return_type_metadata.rs
@@ -26,8 +26,8 @@ mod contract {
 
     impl Contract {
         #[ink(constructor)]
-        pub fn new() -> Self {
-            Self {}
+        pub fn try_new() -> Result<Self, u8> {
+            Err(1)
         }
     }
 
@@ -58,7 +58,7 @@ mod tests {
     }
 
     #[test]
-    fn trait_message_return_value_is_result() {
+    fn trait_message_metadata_return_value_is_result() {
         let metadata = generate_metadata();
 
         let message = metadata.spec().messages().iter().next().unwrap();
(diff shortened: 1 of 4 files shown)


crates/ink/tests/return_type_metadata.rs
```diff
@@ -0,0 +1,144 @@
...
+        type_id: u32,
+    ) -> &Type<PortableForm> {
+        metadata
+            .registry()
+            .resolve(type_id)
+            .unwrap_or_else(|| panic!("No type found in registry with id {}", type_id))
+    }
+
+    #[test]
+    fn trait_message_metadata_return_value_is_result() {
```
HCastano, line 103: For the sake of completeness, we may also want to add tests for inherent messages, inherent fallible messages, etc.

Right now as long as the tests in`lang-err-integration-tests` work our metadata stuff is probably fine, but it would be nice to be more explicit with our checks

HCastano (approved):
Quick turnaround ⚡

crates/ink/tests/return_type_metadata.rs
ascjones: Agree, we need more tests for the metadata like this. Work for a follow-on PR methinks.

PR #1532, opened 2022-12-08
Unit tests for decoding reverted fallible constructors return type

ascjones:
*WIP* Just extracts that block of code for decoding the `return_value` from a reverted constructor call, and add some unit tests.

For now I have not modified the implementation at all, I have just added some unit tests to try out the different scenarios.

Not necessarily suggesting this is how it should be factored, just extracted the method here for easy unit testing for now.


HCastano:
My decoding logic was wrong for that failing test. I was expecting a single `1`, but forgot that the error type itself was also part of the return buffer so we were getting `1, 1` back instead (which decoded fine)

PR #1536, opened 2022-12-14
Added missed `WhereClosure` for the generics into `storage_item`

xgreenx:
Added missed `WhereClosure` for the generics. Tested this case via documentation


HCastano (commented):
Do you have a contract which shows what the problem is/was? 

If I'd be inclined to add it as a UI test

xgreenx:
Previously it was not possible=)
image

ascjones (approved):
LGTM, the doc tests are good enough to prove this works

PR #1547, opened 2022-12-16
Temporarily add `Determinism` enum from `pallet-contracts`

SkymanOne:
[Recent](https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.23.0) release of substrate-contracts-node introduce new parameter `determinism` in [`upload_code()`](https://github.com/paritytech/substrate-contracts-node/blob/87a3d76c88058667d8c028b886a5e404ef9a70ea/runtime/src/lib.rs#L567) which needs to be specified on the ink! and cargo-contract sides, otherwise no E2E tests can be run with the latest substrate-contracts-node version.
`pallet-contracts` currently do not expose `pallet_contracts::Determinism` publicly.
Therefore, the temporary solution is to copy-paste the enum into the ink! codebase until the next release of `polkadot` branch in substrate and synchronisation of substrate-contracts-node

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -28,7 +28,7 @@ variables:
   CARGO_TARGET_DIR:                "/ci-cache/${CI_PROJECT_NAME}/targets/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
   # CI_IMAGE is changed to "-:staging" when the CI image gets rebuilt
   # read more https://github.com/paritytech/scripts/pull/244
-  CI_IMAGE:                        "paritytech/ink-ci-linux:production"
+  CI_IMAGE:                        "paritytech/ink-ci-linux:staging"
   PURELY_STD_CRATES:               "ink/codegen metadata engine"
   ALSO_WASM_CRATES:                "env storage storage/traits allocator prelude primitives ink ink/macro ink/ir"
   ALL_CRATES:                      "${PURELY_STD_CRATES} ${ALSO_WASM_CRATES}"
--- a/crates/e2e/src/xts.rs
+++ b/crates/e2e/src/xts.rs
@@ -75,11 +75,32 @@ pub struct Call<E: Environment, B> {
     data: Vec<u8>,
 }
 
+#[derive(
+    Debug, Clone, Copy, scale::Encode, scale::Decode, PartialEq, Eq, serde::Serialize,
+)]
+pub enum Determinism {
+    /// The execution should be deterministic and hence no indeterministic instructions are
+    /// allowed.
+    ///
+    /// Dispatchables always use this mode in order to make on-chain execution deterministic.
+    Deterministic,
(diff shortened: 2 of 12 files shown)


examples/flipper/lib.rs
```diff
@@ -16,7 +16,7 @@ pub mod flipper {
 
         /// Creates a new flipper smart contract initialized to `false`.
         #[ink(constructor)]
-        pub fn default() -> Self {
+        pub fn new_default() -> Self {
```
cmichi, line 19: Why are the `new_default` changes necessary?

SkymanOne: New clippy lints do not allow methods with `default()` name as they can be confused with the `Default::default()` implementation. See the the previous clippy output: https://gitlab.parity.io/parity/mirrors/ink/-/jobs/2171790

cmichi (approved):

sergejparity (approved):

examples/flipper/lib.rs
HCastano: It looks like this lint's been around a [very long time](https://github.com/rust-lang/rust-clippy/blob/b3145fea6a5c9f67cf8ac77f1c9ff9bb84b71c18/clippy_lints/src/methods/mod.rs#L334), why is it only being triggered now?

.gitlab-ci.yml
```diff
@@ -28,7 +28,7 @@ variables:
   CARGO_TARGET_DIR:                "/ci-cache/${CI_PROJECT_NAME}/targets/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}"
   # CI_IMAGE is changed to "-:staging" when the CI image gets rebuilt
   # read more https://github.com/paritytech/scripts/pull/244
-  CI_IMAGE:                        "paritytech/ink-ci-linux:production"
+  CI_IMAGE:                        "paritytech/ink-ci-linux:staging"
```
HCastano, line 31: This should've been rolled back before merging

crates/e2e/src/xts.rs
```diff
@@ -75,11 +75,32 @@ pub struct Call<E: Environment, B> {
     data: Vec<u8>,
 }
 
+#[derive(
+    Debug, Clone, Copy, scale::Encode, scale::Decode, PartialEq, Eq, serde::Serialize,
+)]
+pub enum Determinism {
```
HCastano, line 81: This is missing docs

examples/flipper/lib.rs
HCastano: I'm not a fan of this `new_default` method, it made sense to have a `default` method for the contract. Maybe we can work around it using a the `Default` trait impl instead?

SkymanOne: > I'm not a fan of this `new_default` method, it made sense to have a `default` method for the contract. Maybe we can work around it using a the `Default` trait impl instead?

Can you open a follow up issue for this please?

PR #1548, opened 2022-12-16
Apply `cargo clippy` suggestions

SkymanOne:
This PR merely introduces few changes that were suggested by clippy.
It also removed `const_err` from whitelist since it will be a hard error in future: https://github.com/rust-lang/rust/issues/71800


HCastano (approved):
Thanks!

PR #1549, opened 2022-12-16
Bump Substrate and `subxt` dependencies

HCastano:
Bumps `subxt` to `v0.26.0`. Also updates the Substrate dependencies to the matching
versions used by `subxt`.


ascjones:
`subxt` new version released: https://crates.io/crates/subxt/0.26.0

crates/e2e/src/xts.rs
```diff
@@ -176,14 +177,10 @@ pub struct ContractsApi<C: subxt::Config, E: Environment> {
 impl<C, E> ContractsApi<C, E>
 where
     C: subxt::Config,
-    C::AccountId: Into<C::Address> + serde::de::DeserializeOwned,
-    <C::ExtrinsicParams as ExtrinsicParams<C::Index, C::Hash>>::OtherParams: Default,
-
+    C::AccountId: serde::de::DeserializeOwned,
+    C::AccountId: scale::Codec,
```
ascjones, line 181: I'm not sure why they removed the `Parameter` (which is `Codec + Clone` etc. constraint from `type AccountId`

crates/e2e/src/xts.rs
```diff
@@ -118,7 +116,10 @@ pub struct UploadCode<B> {
 /// A struct that encodes RPC parameters required to instantiate a new smart contract.
 #[derive(serde::Serialize, scale::Encode)]
 #[serde(rename_all = "camelCase")]
-struct RpcInstantiateRequest<C: subxt::Config, E: Environment> {
+struct RpcInstantiateRequest<C: subxt::Config, E: Environment>
+where
+    C::AccountId: scale::Codec,
```
ascjones, line 121: I reckon this bound on the struct is not necessary

ascjones (approved):
LGTM!

crates/e2e/src/xts.rs
HCastano: @jsdw I meant to ping you about this last week, but yeah this shouldn't have been removed imo

ascjones: I spoke with @jsdw about this and the outcome was that all bounds which are not required within `subxt` itself are removed. And that if we want glbla bounds we should use our own local trait.

https://github.com/paritytech/subxt/pull/804

PR #1569, opened 2022-12-24
Chain Extension: Evaluation of method return type at compile time

### LINKED ISSUE #952: Follow-up #665: Make ink! chain extensions use is_result_type! utility macro
The https://github.com/paritytech/ink/pull/665 PR introduces some new `is_result_type!` utility macro to check at compile time whether some type is a `Result` using the Rust type system.
This means that we can get rid of the `returns_result: bool` ink! attribute of ink! chain extensions since it is now possible to infer this information via the proc. macro itself.

SkymanOne:
This PR closes #952

It introduces breaking changes to chain extension macro API: **`returns_result` flag has been removed**
Since this PR, `Result<O, E>` return type is detected and handled at the compile time.

If `handle_status` is set to true, the return type will still be wrapped into `Result` as before.

Tasks:
- [x] Code cleanup
- [x] Docs


cmichi (approved):
Not a fan of the > 80 CPL, but I see the surrounding code is using it too.

PR #1571, opened 2022-12-29
Added example dijkstra algorithm

Irene-123:
Hello. I've written a smart contract which is an implementation of the Dijkstra Algorithm to find the shortest path between 2 nodes. 
The smart contract uses another file `graph.in` which has the `vec(source, destination, cost)` and the result is printed. Though this is a very basic implementation of `rust` which is quite academic but if you find it useful and worthy please accept my project.


SkymanOne:
Hi @Irene-123. Thank you for your effort.
I can see that you added a diagram and README for the DNS example which is not mentioned in your PR and goes beyond the scope of it.
Could you please open a separate PR for this one?

With regard to your Dijkstra Algorithm, it is definitely a cool idea. However, it doesn't utilise any ink! API and can not be deployed on chain. I advise you to go to [ink! docs](https://use.ink/4.0.0-alpha.1/) and leverage ink! API to turn your code into a deployable contract.

PR #1589, opened 2023-01-12
Advanced test environment

Artemka374:
This is PR for this issue: https://github.com/Supercolony-net/openbrush-contracts/pull/136#issuecomment-1167236655
that was started in https://github.com/Supercolony-net/ink/pull/4

It allows testing cross-contract calls in off-chain environment

- [x] Add support of call_flags
- [x] Error & revert handling


HCastano:
@Artemka374 thanks for the PR, sorry for leaving it sitting around so long.

Tbh not sure when we'll have time to review this since it's a non-trivial addition

coreggon11:
Hi @HCastano thanks for the reply. We understand that there is so much going on in ink! right now, I will just point out that this is a demanded feature and would make development (and testing especially) much easier for devs. We will keep maintaining this PR and are happy to hear some feedback when you will have time to get to this 😄

cmichi:
@Artemka374 Could you clarify the following:

>It allows testing cross-contract calls in off-chain environment

Is this the main intention for the PR? Because we have released ink! 4.0 in the meantime and E2E testing supports this by default.

The reason I'm asking is because the PR has 1k lines of code and there is some maintenance costs associated with it. We're thinking about what functionality to support in the off-chain testing engine and for which to require usage of the E2E testing.

Artemka374:
@cmichi Actually, it is more like full implementation of off-chain functionality, that wasn't accessible before, but I think off-chain cross-contract calls testing is the most important part of it.

deuszx:
I think this is superseeded by what's being enabled with https://github.com/paritytech/ink/pull/1864 .

ascjones:
> I think this is superseeded by what's being enabled with #1864 .

@xgreenx has pointed out that a strong reason to keep this would be for the ability to do step by step debugging of ink contracts.

deuszx:
> > I think this is superseeded by what's being enabled with #1864 .
> 
> @xgreenx has pointed out that a strong reason to keep this would be for the ability to do step by step debugging of ink contracts.

Using off-chain environment though, right? Not the actual one on which the contract will be ran after deployment.

ascjones:
> > > I think this is superseeded by what's being enabled with #1864 .
> > 
> > 
> > @xgreenx has pointed out that a strong reason to keep this would be for the ability to do step by step debugging of ink contracts.
> 
> Using off-chain environment though, right? Not the actual one on which the contract will be ran after deployment.

Exactly

PR #1591, opened 2023-01-14
Unify fallible and non fallible instantiate methods

ascjones:
Context: https://github.com/paritytech/ink/pull/1512#discussion_r1041508207

Since https://github.com/paritytech/ink/pull/1512 `CreateBuilder` has 4 different `instantiate` methods:

`instantiate`
`instantiate_fallible`
`try_instantiate`
`try_instantiate_faillble`

The use case for this is when we have a "fallible" constructor which returns a `Result` e.g.

```rust
#[ink(constructor)]
pub fn new() -> Result<Self, Error> {
    Ok(Self {})
}
```
In order to invoke this from another contract, the following is required:
```rust
ContractRef::new()
            .code_hash(ink_primitives::Clear::CLEAR_HASH)
            .gas_limit(4000)
            .endowment(25)
            .salt_bytes([0xDE, 0xAD, 0xBE, 0xEF])
            .instantiate_fallible() // must call this because it returns a result.
            .unwrap()
```
Some issues with this approach:

1. Requires the user to know which variant of `instantiate` to call based on the return type of the constructor. Granted it should not compile with the wrong one, but still will cause some friction.
2. Code duplication all the way down, two versions of each `instantiate` method.

This PR unifies the `instantaite` and `instantiate_fallible` versions of each method call, such that the user only needs to call `instantiate` or `instantiate_fallible`, and the `CreateBuilder` method will return either `Result<ContranctRef>` or plain `ContractRef`, depending on the actual return type of the constructor.

See this [UI test](https://github.com/paritytech/ink/blob/698f7e3096c2f2a48a2942d163afa8555efee3e5/crates/ink/tests/ui/contract/pass/constructor-return-result-cross-contract.rs#L38) for the variations of constructor return types.


crates/ink/codegen/src/generator/as_dependency/contract_ref.rs
```diff
@@ -431,7 +432,6 @@ impl ContractRef<'_> {
                             .push_arg(#input_bindings)
                         )*
                     )
-                    .returns::<#ret_type>()
```
HCastano, line 434: You're gonna need to add this back so that the return type is known. You also may have to map `Self -> ()` since that's what the `InstantiateResult` implementations expect

crates/env/src/api.rs
```diff
@@ -321,50 +323,20 @@ where
...
-    ink_primitives::ConstructorResult<core::result::Result<E::AccountId, ContractError>>,
+    ink_primitives::ConstructorResult<<R as InstantiateResult<R>>::Output<ContractRef>>,
 >
 where
     E: Environment,
     Args: scale::Encode,
     Salt: AsRef<[u8]>,
-    ContractError: scale::Decode,
+    R: InstantiateResult<R>,
+    ContractRef: FromAccountId<E>,
```
HCastano, line 336: Why not use the `E::AccountId` type directly?

HCastano (commented):
Looks interesting 👀

crates/ink/codegen/src/generator/as_dependency/contract_ref.rs
ascjones: I just added #ret_type as a generic parameter since it must always be set. Anyway the `CreateBuilder` could do with some refactoring.

crates/env/src/api.rs
ascjones: I did start with that, but would have had to implement some `map` capability that does `id` e.g. `AccountId -> AccountId` and `AccountId -> Result<(), AccountId>` and it was slightly tricky. Maybe possible but thought it was better just to call `FromAccountId` in place to get the thing working.

crates/ink/codegen/src/generator/as_dependency/contract_ref.rs
HCastano: I guess forcing devs to specify the generic is probably nicer from an error message point of view. One thing to keep in mind though is that the API should generally match that of the `call_builder`, which uses a `return()` method instead of a generic. 

We could always remove this method in the future with a bigger refactoring

crates/env/src/engine/mod.rs
```diff
@@ -42,65 +50,118 @@ cfg_if! {
...
+            Ok(Ok(output))
         }
-        ConstructorResult::Ok(ContractResult::Err(contract_error)) => {
-            Ok(ConstructorResult::Ok(ContractResult::Err(contract_error)))
+        Err(ext::Error::CalleeReverted) => {
+            let constructor_result_variant = out_return_value.read_byte()?;
+            match constructor_result_variant {
+                // 0 == `ConstructorResult::Ok` variant
+                0 => {
+                    if <R as InstantiateResult<ContractStorage>>::IS_RESULT {
```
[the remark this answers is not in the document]
ascjones: Yeah it was just to make it work with the blanket impl straight from the constructors return type `Self` or `Result<Self, _>`. But as you allude to possibly we can remove that generic parameter, not very nice to have it propogated everywhere.

crates/env/src/engine/mod.rs
```diff
@@ -42,65 +51,117 @@ cfg_if! {
...
+                        let result_variant = out_return_value.read_byte()?;
+                        match result_variant {
+                            // 0 == `Ok` variant
+                            0 => panic!("The callee reverted, but did not encode an error in the output buffer."),
+                            // 1 == `Err` variant
+                            1 => {
+                                let contract_err = <<R as InstantiateResult<ContractStorage>>::Error
+                                    as scale::Decode>::decode(out_return_value)?;
+                                let err = <R as InstantiateResult<ContractStorage>>::err(contract_err);
+                                Ok(Ok(err))
```
HCastano, line 92: nit: I'd use the typedefs here to make it more clear as to which `Result` is which

 ```suggestion
                                EnvResult::Ok(ConstructorResult::Ok(err))
```

File: crates/env/src/engine/mod.rs [hunk 5]
```diff
@@ -42,65 +51,117 @@ cfg_if! {
...
+                    if <R as InstantiateResult<ContractStorage>>::IS_RESULT {
+                        let result_variant = out_return_value.read_byte()?;
+                        match result_variant {
+                            // 0 == `Ok` variant
+                            0 => panic!("The callee reverted, but did not encode an error in the output buffer."),
+                            // 1 == `Err` variant
+                            1 => {
+                                let contract_err = <<R as InstantiateResult<ContractStorage>>::Error
+                                    as scale::Decode>::decode(out_return_value)?;
+                                let err = <R as InstantiateResult<ContractStorage>>::err(contract_err);
```
HCastano, line 91:
```suggestion
                                let contract_err = <R as InstantiateResult<ContractStorage>>::err(contract_err);
```

HCastano:
@ascjones merging `master` into this messed up the commit history since #1512 hasn't been updated yet

ascjones:
> @ascjones merging `master` into this messed up the commit history since #1512 hasn't been updated yet

Sorry I will fix it.

crates/env/src/contract.rs
```diff
@@ -0,0 +1,134 @@
...
+///
+/// // The following line only compiles successfully if both
+/// // `CustomEnvironment` and `<Contract as ContractEnv>::Env`
+/// // are of the same type.
+/// const _: IsSameType<<Contract as ContractEnv>::Env> =
+///     <IsSameType<CustomEnvironment>>::new();
+///
+/// fn main() {}
+/// ```
+pub trait ContractEnv {
```
ascjones, line 88: These traits have just been copied from `ink::reflect` so we can use them here in `ink_env`

HCastano (commented):
There's still a couple things I want to review, but in general it's looking good 👍

HCastano (approved):
Thanks for cleaning up the API, looks much better now!

:shipit:

PR #1600, opened 2023-01-19
Remove `seal_` prefix from `seal_return`

athei:
We removed the prefixes a while back. This functions seem to be overlooked. Maybe because the name is a keyword.

### DIFF
--- a/crates/env/src/engine/on_chain/ext.rs
+++ b/crates/env/src/engine/on_chain/ext.rs
@@ -240,7 +240,7 @@ mod sys {
         ) -> ReturnCode;
 
         pub fn input(buf_ptr: Ptr32Mut<[u8]>, buf_len_ptr: Ptr32Mut<u32>);
-        pub fn seal_return(flags: u32, data_ptr: Ptr32<[u8]>, data_len: u32) -> !;
+        pub fn r#return(flags: u32, data_ptr: Ptr32<[u8]>, data_len: u32) -> !;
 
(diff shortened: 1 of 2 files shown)


HCastano (approved):
Thanks.

How come this wasn't caught before? Was `pallet-contracts` still exposing this as `seal_return`, or were we just using an old version of the API?

Looks like the CI failure is related:

```
test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.23s
2023-01-19 03:28:11.138 DEBUG tokio-runtime-worker runtime::contracts: failed to instantiate code: cannot find definition for import seal0::return: Func(DedupFuncType(GuardedEntity { guard_idx: EngineIdx(0), entity_idx: DedupFuncTypeIdx(3) }))  
```

Do we need a new node version?

athei:
> How come this wasn't caught before? 

`pallet-contracts` is exposing it with and without the prefix for pre-existing functions. 

> Looks like the CI failure is related:
> 
> ```
> test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.23s
> 2023-01-19 03:28:11.138 DEBUG tokio-runtime-worker runtime::contracts: failed to instantiate code: cannot find definition for import seal0::return: Func(DedupFuncType(GuardedEntity { guard_idx: EngineIdx(0), entity_idx: DedupFuncTypeIdx(3) }))  
> ```
> 
> Do we need a new node version?

No new node should be required. The error message looks as if it is using the correct identifier. I need to investigate.

PR #1602, opened 2023-01-20
Make `CallBuilder` and `CreateBuilder` error handling optional

HCastano:
This is a follow-up related to [this](https://github.com/paritytech/ink/pull/1525#discussion_r1073940032) comment.

We're updating the default APIs to unwrap any errors from the Contracts pallet, and
adding some new APIs in case callers do want to handle the errors.

~~I've also made a drive-by change to rename the `fire()` method to `invoke()`. This keeps
things consistent since we're wrapping a few methods with `invoke` in the name anyways.~~

_Decided to do this in a follow-up. It's gonna make it easier for people to spot the breaking change.
Plus, extra green squares 🧑‍🌾_

The breaking changes have been noted in the `CHANGELOG`.

### DIFF
--- a/crates/ink/codegen/src/generator/trait_def/call_forwarder.rs
+++ b/crates/ink/codegen/src/generator/trait_def/call_forwarder.rs
@@ -355,7 +355,7 @@ impl CallForwarder<'_> {
                         , #input_bindings
                     )*
                 )
-                    .try_invoke()
+                    .try_fire()
                     .unwrap_or_else(|env_err| ::core::panic!("{}: {:?}", #panic_str, env_err))
                     .unwrap_or_else(|lang_err| ::core::panic!("{}: {:?}", #panic_str, lang_err))
             }
--- a/crates/ink/codegen/src/generator/trait_def/call_forwarder.rs
+++ b/crates/ink/codegen/src/generator/trait_def/call_forwarder.rs
@@ -355,7 +355,7 @@ impl CallForwarder<'_> {
                         , #input_bindings
                     )*
                 )
-                    .try_fire()
+                    .try_invoke()
                     .unwrap_or_else(|env_err| ::core::panic!("{}: {:?}", #panic_str, env_err))
                     .unwrap_or_else(|lang_err| ::core::panic!("{}: {:?}", #panic_str, lang_err))
             }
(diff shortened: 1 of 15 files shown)


crates/ink/codegen/src/generator/trait_def/call_forwarder.rs
```diff
@@ -355,8 +355,9 @@ impl CallForwarder<'_> {
                         , #input_bindings
                     )*
                 )
-                    .fire()
-                    .unwrap_or_else(|err| ::core::panic!("{}: {:?}", #panic_str, err))
+                    .try_fire()
+                    .unwrap_or_else(|env_err| ::core::panic!("{}: {:?}", #panic_str, env_err))
+                    .unwrap_or_else(|lang_err| ::core::panic!("{}: {:?}", #panic_str, lang_err))
```
athei, line 360: Why not just call `fire`?

examples/delegator/lib.rs
```diff
@@ -63,7 +63,7 @@ mod delegator {
                 .endowment(total_balance / 4)
                 .code_hash(accumulator_code_hash)
                 .salt_bytes(salt)
-                .instantiate()
+                .try_instantiate()
```
athei, line 66: If we are unwrapping anyways we should use `instantiate`. Otherwise this is setting a bad example.

athei (approved):

ascjones (approved):
LGTM, though we should consider whether/how to unify those errors in a subsequent change.

crates/ink/codegen/src/generator/trait_def/call_forwarder.rs
ascjones: It looks like the only difference would be that the panic string includes extra info:

```
let panic_str = format!(
            "encountered error while calling <{} as {}>::{}",
            forwarder_ident, trait_ident, message_ident,
        );
```
Not sure how useful that is

examples/delegator/lib.rs
ascjones: I agree I have made this change

crates/ink/codegen/src/generator/trait_def/call_forwarder.rs
athei: Yes I think it is useful. If we just `fire` the panic message will just point to the line within the call builder file. Without backtraces (which we don't have) it will be quite hard to find the line which paniced.

PR #1615, opened 2023-01-24
`[ink_e2e]` method to generate and fund unique accounts

ascjones:
Allows individual tests to generate their own unique accounts to avoid the problem of race conditions between tests using the same account, causing failures in #1591.

Can consider integrating this into the environment somehow, but for now providing the method is a good start.


crates/e2e/src/xts.rs
```diff
@@ -45,6 +45,14 @@ use subxt::{
...
 };
 
+/// A raw call to `pallet-balances` `transfer`.
+// #[derive(Debug, scale::Encode)]
+// pub struct Transfer<E: Environment, B> {
+//     dest: sp_runtime::MultiAddress<E::AccountId, ()>,
+//     #[codec(compact)]
+//     value: B,
+// }
+
```
cmichi, line 55: Forgotten?

ascjones:
I suggest to merge #1591 first, then we can merge this into `master` so it has it's own commit for the `CHANGELOG`

ascjones:
> Nice! Would love to see this as part of the environment eventually. Could you create a follow-up so we don't forget about it?

https://github.com/paritytech/ink/issues/1618

PR #1624, opened 2023-01-26
`[ink_e2e]` expose call dry-run method

### LINKED ISSUE #1620: ink_e2e: Return values of `Client` methods are not available in rust docs
The testing framework is hard to use if not available in rustdocs. Those types need to be re-exported from the top crate to be available in docs. rust-analyzer doesn't help as it chokes on the macro.

Someone needs to go through rust docs and make sure everything is available.

ascjones:
As suggested by @athei.

Also closes #1620


athei:
Can you drive by fix https://github.com/paritytech/ink/issues/1620

Should be just a bunch of `pub use`.

cmichi (commented):
API lgtm!

athei:
Looks good. Can you fix https://github.com/paritytech/ink/issues/1620  as a drive by? I think you might be exposing more types now (the dry run type) which might not be exposed.

ascjones:
Completed updating the remaining examples

ascjones:
> Looks good. Can you fix #1620 as a drive by? I think you might be exposing more types now (the dry run type) which might not be exposed.

https://github.com/paritytech/ink/pull/1624/commits/7916ff533e0d59d38be6658439f4cbc097d1a965

athei (approved):
Thanks :)

crates/e2e/src/client.rs
```diff
@@ -172,6 +178,69 @@ where
...
+                "Decoding dry run result to ink! message return type failed: {}",
+                env_err
+            )
+        })
+    }
+
+    /// Returns the return value as raw bytes of the message from the dry-run.
+    ///
+    /// Panics if the dry-run message call failed to execute.
+    pub fn return_data(&self) -> &[u8] {
```
HCastano, line 225: nit: I'd move this after `return_value`. It makes sense that we read this as `MessageResult<V>` -> `V` -> raw data

HCastano (approved):
Nice

PR #1630, opened 2023-01-30
Change CI Runners

alvicsam:
PR changes ci runners for dynamically created ones

cc https://github.com/paritytech/ci_cd/issues/730


oleg-plakida (approved):

sergejparity (approved):

alvicsam:
@cmichi @HCastano @ascjones @SkymanOne 
It seems that I need some help. The test `flipper::e2e_tests::it_works` is flaky on the new runners and I don't know how can I debug it: https://gitlab.parity.io/parity/mirrors/ink/-/jobs/2354938
Sometimes it works: https://gitlab.parity.io/parity/mirrors/ink/-/jobs/2349676 but I don't see correlation between fail and success runs.

In theory there should be no difference between new runners and old runners because everything executes in docker and the environment is the same.

ascjones:
I believe https://github.com/paritytech/ink/pull/1642 will fix this: hope to get it merged soon.

ascjones:
> In theory there should be no difference between new runners and old runners because everything executes in docker and the environment is the same.

The reason it is failing is because some other test/process is running in parallel using the `alice` account, indicated by the following message:

```
"The transaction has too low priority to replace another transaction already in the pool."
```
It means that sometimes another transaction for `alice` gets there first, which is why it fails intermittently.

Having said that, it doesn't explain why it is happening now and not previously.

However, whatever the cause of the interference with the running `substrate-contracts-node`, it should be fixed by #1642, because that launches an instance of `substrate-contracts-node` per-test so they will no longer interfere with each other.

alvicsam:
@ascjones okay, thank you! Waiting for https://github.com/paritytech/ink/pull/1642 then

ascjones:
> @ascjones okay, thank you! Waiting for #1642 then

It's in, so try merging `master` and see what happens!

alvicsam:
bot rebase

alvicsam:
@ascjones now it works, thank you

PR #1632, opened 2023-01-30
[T-Rex]: apply clippy suggestions and fix UI test output

SkymanOne:


ascjones (approved):
This is that weird thing again where the PR CI runs against the old version of Rust `cargo 1.66.1 (ad779e08b 2023-01-10)`, but the `master` CI runs the new version of Rust `1.67`. 

IIRC it is because some docker image is not updated to the latest because the `master` CI fails...

So will merge this despite red PR CI, and hopefully will resolve things.

ascjones:
Damn, it failed on master because now it appears to be using `rustc 1.66.1 (90743e729 2023-01-10)` :thinking:

PR #1634, opened 2023-01-31
Remove custom selectors

ascjones:
Removes custom selectors, now only allows wildcard selectors for fallback methods e.g. `#[ink(selector = _)]`


xermicus:
We use function selector overrides to enable fallback functions too

ascjones:
It is still up for debate whether we want to to this, see https://github.com/paritytech/ink/issues/1643

PR #1635, opened 2023-02-01
Update CODEOWNERS

SkymanOne:


HCastano (approved):
We're gonna need you to swear an oath of allegiance to Squink before merging this though 🦑

PR #1636, opened 2023-02-01
Make cross-contract callee non-optional

ascjones:
Suggestion for how to enforce `callee` in #1255.

This breaks the `CallBuilder` API by introducing two new methods: `call` and `delegate`, which can
be used to streamline the `CallBuilder` set-up for the respective call types.

### DIFF
--- a/crates/env/src/call/call_builder.rs
+++ b/crates/env/src/call/call_builder.rs
@@ -28,7 +28,6 @@ use crate::{
     Error,
 };
 use core::marker::PhantomData;
-use ink_primitives::Clear;
 use num_traits::Zero;
 
 /// The final parameters to the cross-contract call.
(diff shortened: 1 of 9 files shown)


crates/env/src/call/call_builder.rs
```diff
@@ -521,26 +488,43 @@ where
...
+            call_type: Set(Call::new(callee)),
             call_flags: self.call_flags,
             exec_input: self.exec_input,
             return_type: self.return_type,
             _phantom: Default::default(),
         }
     }
 
+    /// Sets the `code_hash` for the current cross-contract delegate call.
+    pub fn code_hash(
```
HCastano, line 510: So @ascjones I guess what you're suggesting is we rename this to `delegate`?

ascjones: Yes exactly, so by calling one method or the other we instantiate the builder type for the type of call we want.

ascjones: Actually maybe the choices are 

1. `callee` & `code_hash`
2. `call` & `delegate`

HCastano: Sure, but then we should rename the `callee` builder method to `call`

ascjones: :brain: :handshake: :brain:

HCastano: More like 🥜 🤝 🥜

ascjones: deez nuts

ascjones:
Not going to wait for examples test to run, tests run locally okay and it is hanging occasionally.

PR #1642, opened 2023-02-03
`[ink_e2e]` spawn a separate contracts node instance per test

ascjones:
Currently the E2E tests depend on a shared instance of `substrate-contracts-node` running in the background. This can cause issues where tests are not independent, and can be non deterministic when different tests share account ids. See https://github.com/paritytech/ink/pull/1615.

This PR spawns a unique instance of `substrate-contracts-node` per test, so the tests are completely isolated. I have copied the [TestNodeProcess](https://github.com/paritytech/subxt/blob/master/testing/integration-tests/src/utils/node_proc.rs) from `subxt` in order to manage the spawning of the node process and retrieving the unique ports which allow them to run in parallel. This also ensures that the processes are killed on drop.

The default is assuming `substrate-contracts-node` is installed on the PATH. The alternative is to specify the `CONTRACTS_NODE` env variable.

This is also a quality of life improvement, because it does not depend on the user or the CI environment starting up a contracts node in the background.


crates/e2e/src/tests.rs
```diff
@@ -1,11 +0,0 @@
-#[tokio::test]
```
SkymanOne, line 1: Maybe worth replacing this with a test on launching the test node

SkymanOne (approved):

SkymanOne (changes requested):
Can you please add CHANGELOG entry?

SkymanOne (approved):
LGTM!

cmichi (approved):
Thanks, it's good!

PR #1646, opened 2023-02-08
Make ink! compile on `no_std`

athei:
Contrary to popular belief `wasm32-unknown-unknown` actually has `std`. This is why we never noticed that we were pulling in std crates from our nostd builds.

For that to never happen again (after fixing) this I added a CI pass which compiles ink! and the examples for a true `nostd` target. I also needed to remove a few `nostd == wasm32` assumptions.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -435,6 +435,8 @@ examples-no-std-check:
   stage:                           examples
   <<:                              *docker-env
   <<:                              *test-refs
+  variables:
+    RUSTC_BOOTSTRAP:               "1"
   script:
     - rustup component add rust-src --toolchain stable
     - cargo contract -V
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -430,12 +430,11 @@ examples-contract-build:
       done
     - cargo +stable contract build --manifest-path ./integration-tests/upgradeable-contracts/set-code-hash/updated-incrementer/Cargo.toml
 
+# we only check a few basic contracts here. no need to check them all.
 examples-no-std-check:
   stage:                           examples
   <<:                              *docker-env
   <<:                              *test-refs
-  variables:
-    RUSTC_BOOTSTRAP:               "1"
   script:
     - rustup component add rust-src --toolchain stable
     - cargo contract -V
@@ -446,14 +445,6 @@ examples-no-std-check:
         cargo +stable check --no-default-features --target $NO_STD_TARGET -Zbuild-std="core,alloc" &&
         popd;
       done
-    - pushd ./integration-tests/delegator/ && ./build-all.sh && popd
-    - for contract in ${UPGRADEABLE_CONTRACTS}; do
-        cargo +stable check --no-default-features --target $NO_STD_TARGET -Zbuild-std="core,alloc" --manifest-path ./integration-tests/upgradeable-contracts/${contract}/Cargo.toml;
-      done
-    - for contract in ${LANG_ERR_INTEGRATION_CONTRACTS}; do
-        cargo +stable check --no-default-features --target $NO_STD_TARGET -Zbuild-std="core,alloc" --manifest-path ./integration-tests/lang-err-integration-tests/${contract}/Cargo.toml;
-      done
-    - cargo +stable check --no-default-features --target $NO_STD_TARGET -Zbuild-std="core,alloc"  --manifest-path ./integration-tests/upgradeable-contracts/set-code-hash/updated-incrementer/Cargo.toml
 
 examples-docs:
   stage:                           examples
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,6 +40,9 @@ variables:
   # this fix not yet in stable: https://github.com/rust-lang/rust-clippy/issues/8895.
   # Remove the following line again as soon as `clippy` on stable succeeds again.
   CLIPPY_ALLOWED:                  "clippy::extra_unused_lifetimes"
+  # This is a target that does not have a std lib in contrast to wasm. We compile against
+  # this target to make sure that we don't pull in std by accident (through dependencies).
+  NO_STD_TARGET:                   "bpfel-unknown-none"
 
 workflow:
   rules:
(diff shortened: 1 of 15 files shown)


athei:
Need this: https://github.com/paritytech/parity-scale-codec/pull/408

pmikolajczyk41:
you might want to use https://github.com/hobofan/cargo-nono in the pipelines

HCastano:
We can follow what [Wasmi did](https://github.com/paritytech/wasmi/issues/374) and build for the  `thumbv7em-none-eabihf` target which has no support for `std`

athei:
`bpfel-unknown-none` es even better. It doesn't have atomic ptr which puts more contraints. Some parts of core and alloc are not available on such platforms.

.gitlab-ci.yml
```diff
@@ -218,6 +218,18 @@ check-wasm:
         cargo check --verbose --no-default-features --target wasm32-unknown-unknown --manifest-path ./crates/${crate}/Cargo.toml;
       done
 
+# We just check on a target which doesn't have std to make sure that we don't pull in std by accident
+check-nostd:
```
HCastano, line 222:
```suggestion
check-no-std:
```

.gitlab-ci.yml
```diff
@@ -415,6 +427,31 @@ examples-contract-build:
       done
     - cargo +stable contract build --manifest-path ./examples/upgradeable-contracts/set-code-hash/updated-incrementer/Cargo.toml
 
+examples-nostd-check:
```
HCastano, line 430:
```suggestion
examples-no-std-check:
```

.gitlab-ci.yml
```diff
@@ -415,6 +427,31 @@ examples-contract-build:
...
+  variables:
+    RUSTC_BOOTSTRAP:               "1"
+  script:
+    - rustup component add rust-src --toolchain stable
+    - cargo contract -V
+    - for example in examples/*/; do
+        if [ "$example" = "examples/upgradeable-contracts/" ]; then continue; fi;
+        if [ "$example" = "examples/lang-err-integration-tests/" ]; then continue; fi;
+        pushd $example &&
+        cargo +stable check --no-default-features --target bpfel-unknown-none -Zbuild-std="core,alloc" &&
```
HCastano, line 443: I'd pull the target out to a variable, maybe something like `NO_STD_TARGET`. Would make it easier to swap in the future

HCastano:
@athei do you want to pick this up again, or should we get someone else to drive it to the finish (maybe @lean-apple or @SkymanOne?)

athei:
I am on it.

.gitlab-ci.yml
```diff
@@ -415,6 +430,24 @@ examples-contract-build:
...
+  variables:
+    RUSTC_BOOTSTRAP:               "1"
+  script:
+    - rustup component add rust-src --toolchain stable
+    - cargo contract -V
+    - for example in integration-tests/*/; do
+        if [ "$example" = "integration-tests/upgradeable-contracts/" ]; then continue; fi;
+        if [ "$example" = "integration-tests/lang-err-integration-tests/" ]; then continue; fi;
+        pushd $example &&
+        cargo +stable check --no-default-features --target $NO_STD_TARGET -Zbuild-std="core,alloc" &&
```
HCastano, line 447: This is still a lot of examples, wonder if we should just spot check a few instead (e.g `flipper`, `erc20`, etc.)

crates/allocator/src/handlers.rs
```diff
@@ -14,5 +14,18 @@
 
 #[alloc_error_handler]
 fn oom(_: core::alloc::Layout) -> ! {
-    core::arch::wasm32::unreachable()
+    #[cfg(target_arch = "wasm32")]
+    core::arch::wasm32::unreachable();
+
+    // For any other nostd architecture we just call an imaginary
```
HCastano, line 20:
```suggestion
    // For any other `no_std` architecture we just call an imaginary
```

.gitlab-ci.yml
```diff
@@ -40,6 +40,9 @@ variables:
   # this fix not yet in stable: https://github.com/rust-lang/rust-clippy/issues/8895.
   # Remove the following line again as soon as `clippy` on stable succeeds again.
   CLIPPY_ALLOWED:                  "clippy::extra_unused_lifetimes"
+  # This is a target that does not have a std lib in contrast to wasm. We compile against
```
HCastano, line 43:
```suggestion
  # This is a target that does not have a standard library in contrast to Wasm. We compile against
```

.gitlab-ci.yml
```diff
@@ -40,6 +40,9 @@ variables:
...
+  # this target to make sure that we don't pull in std by accident (through dependencies).
```
HCastano, line 44:
```suggestion
  # this target to make sure that we don't pull in `std` by accident (through dependencies).
```

.gitlab-ci.yml
```diff
@@ -218,6 +221,18 @@ check-wasm:
         cargo check --verbose --no-default-features --target wasm32-unknown-unknown --manifest-path ./crates/${crate}/Cargo.toml;
       done
 
+# We just check on a target which doesn't have std to make sure that we don't pull in std by accident
```
HCastano, line 224:
```suggestion
# We just check on a target which doesn't have `std` to make sure that we don't pull in `std` by accident
```

.gitlab-ci.yml
```diff
@@ -415,6 +430,24 @@ examples-contract-build:
       done
     - cargo +stable contract build --manifest-path ./integration-tests/upgradeable-contracts/set-code-hash/updated-incrementer/Cargo.toml
 
+# we only check a few basic contracts here. no need to check them all.
```
HCastano, line 433:
```suggestion
# We only check a few basic contracts here, no need to check them all.
```

.gitlab-ci.yml
```diff
@@ -415,6 +430,24 @@ examples-contract-build:
...
+  stage:                           examples
+  <<:                              *docker-env
+  <<:                              *test-refs
+  variables:
+    RUSTC_BOOTSTRAP:               "1"
+  script:
+    - rustup component add rust-src --toolchain stable
+    - cargo contract -V
+    - for example in integration-tests/*/; do
+        if [ "$example" = "integration-tests/upgradeable-contracts/" ]; then continue; fi;
```
HCastano, line 444:
```suggestion
```

I got rid of this in https://github.com/paritytech/ink/pull/1704

HCastano (approved):
One question around the number of examples we want to check, but otherwise looks good

.gitlab-ci.yml
athei: We could but I don't think it causes any harm to just check all of them. Example checks are run in parallel and this is much faster than the rest anyways.

athei:
> @athei CI is failing

We need to skip some example directories. Updated the CI.

HCastano (approved):

PR #1648, opened 2023-02-10
Add `quickcheck` support

### LINKED ISSUE #1647: Add ability to use quickcheck with E2E testing
We should allow for the possibility of using [quickcheck](https://github.com/BurntSushi/quickcheck) with our E2E testing framework. 

The problem here is that the `#[quickcheck]` macro already generates code which contains the `#[test]` attribute (whereas we need `#[ink_e2e::test]`. But I think we can access the quickcheck functionality via some other way directly calling functions on the crate.

## ToDo 1
Please investigate if we can find a way to make this work:
```
#[quickcheck]
#[ink_e2e::test]
```

I think having this :point_up: would be the best API, but I think it's unlikely that we can get this to work. The other way would be sth like
```
#[ink_e2e::test(quickcheck = …)]
```
and then pass the arguments through. So `quickcheck` would be a direct dependency of `ink_e2e` and we would likely call a Rust function directly on that crate.

## ToDo 2: Benchmark

After https://github.com/paritytech/ink/pull/1642 I'm not sure how running quickcheck on top of our E2E testing will behave. We should have some numbers of how long it takes to e.g. fuzz a contract message 10_000 times.

SkymanOne:
closes #1647 

# Notes
- Added `#[ink_e2e::test(quickcheck = true)]` arg
- Developer needs to do add additional argument to a function so `quickcheck` can pump input through. See example for demo.
- The `quickcheck` featured test takes way longer to run (~80s for a single test). Not sure whether it can be optimised, probable need to add notice for devs.
- The tests sometimes fail because of some logger error: 
```
thread 'quickcheck_tests::e2e_tests::default_works' panicked at 'env_logger::init should not be called after logger initialized: SetLoggerError(())', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/env_logger-0.10.0/src/lib.rs:1154:16
```
- I tried to use `#[quickcheck]` but it sits in separate crate `quickcheck_macros` and depends on `quickcheck` crate which [sits behind `dev_dependencies`](https://github.com/BurntSushi/quickcheck/blob/defde6fb0ce20b0c8c4e672aa9ae821f7d1f5b38/quickcheck_macros/Cargo.toml#L25). I did not manage to re-export the attribute macro because it would complain about missing `quickcheck` dependency. If someone can explain to me how I can re-export `quickcheck_macros`, it is trivial to change back to attribute macro.

IMHO, `quicktest` would work better with regular `#[ink::test]` unit tests.


SkymanOne:
Closing PR as it's low on the priority list.

PR #1655, opened 2023-02-15
[T-Rex 🦖 ] use `cargo test` for testing in CI

SkymanOne:


.github/workflows/examples.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         toolchain:
           - stable
         job:
-          - build
+          - contract build
```
athei, line 31: Is this just a label? Or is this a mistake? I think for `build` we still want to use `cargo contract`.

SkymanOne: this is a mistake indeed, thanks for noting

SkymanOne: > Is this just a label? Or is this a mistake? I think for `build` we still want to use `cargo contract`.

No, this is not a mistake. Later down the stage, the job label is concatenated with `cargo` command.
So this way `contract build` job label will be concatenated into `cargo contract build` and `test` label will respectively become just `cargo test`

athei (approved):

PR #1661, opened 2023-02-19
Allow option to opt-out of provided memory allocator

### LINKED ISSUE #1414: Consider add dlmalloc-rs as allocator or opt-out the BumpAllocator

HCastano:
This takes https://github.com/paritytech/ink/pull/1415 one step further and provides the
`no-allocator` feature to the top level `ink` crate, allowing contract authors to opt-out
of the provided memory allocator.

It also adds an example showing how a contract author could provide their own allocator.

Closes #1414.

cc @kvinwang


examples/custom_allocator/lib.rs
```diff
@@ -0,0 +1,180 @@
+//! # Custom Allocator
+//!
+//! This example demonstrates how to opt-out of the ink! provided global memory allocator.
```
cmichi, line 3: Lemme be the guy commenting about something on the order of 80 CPL 😬 .

examples/custom_allocator/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+//! If you don't handle allocations correctly you can introduce security vulnerabilities to your
+//! contracts.
+//!
+//! You may also introduce performance issues. This is because the code of your allocator will
+//! be included in the final contract binary, potentially increasing gas usage significantly.
+//!
+//! ## Why Change the Allocator?
+//!
+//! The default memory allocator was designed to have a tiny size footprint, and made some
+//! compromises to achieve that, e.g it does not free/deallocate memory
```
cmichi, line 21:
```suggestion
//! compromises to achieve that, e.g it does not free/deallocate memory.
```

cmichi (approved):
LGTM. Just the nitpicks about having less CPL throughout.

examples/custom_allocator/lib.rs
ascjones: Can we configure `rustfmt` to catch this? I think it enforces code CPL to 100, not sure about the comments.

cmichi:
Add the example to https://github.com/paritytech/ink-examples as well, please.

examples/custom_allocator/lib.rs
HCastano: We already enforce this to 90 characters, which is what it's at 😉 

https://github.com/paritytech/ink/blob/b8862a1147a8fb577c82304ed1dec9a8c45c386a/.rustfmt.toml#L1

Issue #1520, opened 2022-11-25
Examples CI: improve traversing of examples :t-rex: 


ascjones:
The CI contains many places where we traverse the `examples` directory to run the same operation (e.g. build/check/fmt/test). Previously the structure of the firectory was flat i.e. all examples lived directly in the `examples` dir, so the following simple loop worked: 

```
for example in examples/*/; do
     echo "Processing example: $example";
     cargo test --manifest-path=$example/Cargo.toml;
 done
```
This was flexible because any new example would automatically be handled without any changes to the CI scripts.

However we have started to add more sub directories which contain multiple examples, so we now have the pattern where we ignore the subdirs on the main loop and then hardcode loops over the subdirs:

```
for contract in ${LANG_ERR_INTEGRATION_CONTRACTS}; do
   echo "Processing delegator contract: $contract";
   cargo contract ${{ matrix.job }} --verbose --manifest-path examples/delegator/${contract}/Cargo.toml;
done
for example in examples/*/; do
   if [ "$example" = "examples/lang-err-integration-tests/" ]; then continue; fi;
   echo "Processing example: $example";
   cargo test --manifest-path=$example/Cargo.toml;
done
```
However this has become more brittle because:
1. This code is copy pasted in several places and it is easy to forget to update all copies.
2. Because the names of the individual examples are hardcoded in a variable e.g. `LANG_ERR_INTEGRATION_CONTRACTS`, then it will not automatically handle any new examples added in that subdirectlry.
3. If you add a new subdirectory under `examples` then it will break the main `examples` loop. `See https://github.com/paritytech/ink/pull/1519)

So what we want is a unified way to recursively traverse the `examples`  to run some operation that will automatically handle new subdirectories and new examples, and not break when existing ones are renamed or deleted.

HCastano:
This may be alleviated with https://github.com/paritytech/cargo-contract/pull/647, since we can crate workspaces for the examples and integration tests and build those in one go.

I should definitely get around to looking at that again 😅

ascjones:
Indeed that might be a good way to approach it, to ensure that all subdirectories have a virtual manifest rather than doing anything clever in the scripts themselves.

athei:
I thought multiple times that just having the examples in the same workspace as ink! itself would be nice. But there is a reason why this isn't possible. And I always forget. Can someone remind me? Maybe if we really want it we can make it happen?

PR #1663, opened 2023-02-20
Replacing `cargo test` with `cargo nextest` - Part 2

lean-apple:
Additional tests to conclude if we should use `cargo nextest` instead of `cargo test` 

Related to https://github.com/paritytech/ink/pull/1514. 

3rd tasks of https://github.com/paritytech/ink/issues/1474

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -293,7 +293,24 @@ test:
     - cargo test --verbose --all-features --no-fail-fast --workspace
     - cargo test --verbose --all-features --no-fail-fast --workspace --doc
 
-nextest:
+test-no-doc:
+  stage:                           workspace
+  <<:                              *docker-env
+  <<:                              *test-refs
+  needs:
+    - job:                         check-std
+      artifacts:                   false
+  variables:
+      # Since we run the tests with `--all-features` this implies the feature
+      # `ink-fuzz-tests` as well -- i.e. the fuzz tests are run.
+      # There's no way to disable a single feature while enabling all features
+      # at the same time, hence we use this workaround.
+      QUICKCHECK_TESTS:            0
+  script:
+    - cargo test --verbose --all-features --no-fail-fast --workspace
+
+
+nextest-no-doc:
   stage:                           workspace
   <<:                              *docker-env
   <<:                              *test-refs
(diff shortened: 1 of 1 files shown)


lean-apple:
Time Comparison btw `cargo test` and `cargo nextest` coming from this [job](https://gitlab.parity.io/parity/mirrors/ink/-/pipelines/247781)

| Step - Substep / Tool      | `test` | `nextest`   |
| :---        |    :----:   |          ---: |
| clippy-std - dylint      | 2 min 19       |  2 min 39     |
| workspace - test      | 6 min 43       |  - --    | 
| workspace - codecov      | 8 min 15       |  7 min 54     |
| examples - test      | 19 min 46       |  20 min 40     |

still not really concluant about the benefits of `cargo nextest`

ascjones:
So the conclusion is it doesn't really speed things up, and in some cases is marginally slower?

lean-apple:
I think we reached this same conclusion ... 
The point is for `substrate` or `cumulus` as @alvicsam apparently told me there is a significant difference, 
But because the times for these stages are not really 'high' the difference is not significative, 
Do you think @alvicsam working on [this](https://github.com/paritytech/ink/issues/1520) could improve the use of `cargo nextest` ?

cmichi:
@lean-apple If you mark your PR as draft it can't be merged by accident.

cmichi:
I'm closing this PR and the issue as the results are inconclusive, we might revisit after https://github.com/paritytech/ink/issues/1520.

xermicus:
Since the main thing (regarding performance) of `nextest` is it's improved concurrent execution model, performance gains probably highly depend on the test runner? I'd expect multi-core machines to see significant gains (otherwise should investigate as to why not), but single-core GHA potato-runners probably won't benefit much.

alvicsam:
It seems that for tests in `ink!` repo there is not going to be huge improvement with nextest. There was 3 jobs with nextest in this PR: `examples-nextest`, `codecov-nextest` and `dylint-nextest`. For all of them the longest part was to build the binary.
 - In [dylint-test](https://gitlab.parity.io/parity/mirrors/ink/-/jobs/2423228) there was only one test so it's no much difference between `cargo-test` and `cargo-nextest`.
 - In [codecov-nextest](https://gitlab.parity.io/parity/mirrors/ink/-/jobs/2423236) I see a lot of tests that are super fast (0.01s-0.05s). IMO for these tests there is also no much difference between `cargo-test` and `cargo-nextest`.
 - In [examples-nextest](https://gitlab.parity.io/parity/mirrors/ink/-/jobs/2423238) I see a lot of tests but they are also quick especially if we compare with the time of building the binary.

I see the only way to make the pipeline faster: divide `examples-test` into several jobs

PR #1665, opened 2023-02-20
Add is_contract to test environment

forgetso:
This PR makes it possible for unit tests to specify accounts that are contract accounts.

```
            // Mark the the dapp account as being a contract on-chain
            ink::env::test::set_contract::<ink::env::DefaultEnvironment>(dapp_contract);
```

A boolean is returned when `is_contract` is called.

```
            self.env().is_contract(&contract)
```


ascjones (approved):

HCastano (approved):
Thanks!

Issue #1035, opened 2021-11-23
Investigate Usage of `inline`


HCastano:
This should be a more through version of https://github.com/paritytech/ink/pull/1019. We should dig into our usage of `inline` further since it seems like we can reduce our contract size by being more strategic with its usage.

PR #1666, opened 2023-02-20
Remove all `#[inline.*]`s

agryaznov:
This is a reincarnation of https://github.com/paritytech/ink/pull/1019 to close https://github.com/paritytech/ink/issues/1035 .

These attributes were placed there as a result of some particular test being ran just once, like e.g. [in this PR](https://github.com/paritytech/ink/pull/1012), and probably in some others, all following the same approach. Although those tests could had shown somewhat improvement on contracts sizes, such approach to optimization is in-sustainable in its root.

Both `#[inline]` and `#[inline(always)]` are just hints to the compiler, which inlines functions based on its internal heuristics. As said heuristics change over time, it makes no sense to code against what compiler does in a particular case on a particular run. Rather, we should code against what compiler is allowed to do in general.

Given link-time optimization is [turned on by default](https://github.com/paritytech/cargo-contract/blob/153441356ac944dc31f79d4241f7838ba8e99dbc/crates/build/src/workspace/profile.rs#L34) when building contracts, those attributes should not really affect the compilation result in general.

Therefore it's nothing really wrong with removing all `inline`s with:

```bash
find . -name "*.rs" -type f -exec sed -i '/^.*\#\[inline.*\]/d' {} \;
````

### CODE REVIEW

**Comment by agryaznov** [comment]
Let's check the contracts size change report.

**Comment by HCastano** [comment]
@agryaznov the waterfall isn't running anymore, so you're gonna have to check manually and update us on this

**Comment by agryaznov** [comment]
> the waterfall isn't running anymore, so you're gonna have to check manually and update us on this
 
addressed in https://github.com/paritytech/ink/pull/1667

**Comment by HCastano** [comment]
@agryaznov since the waterfall (still) isn't working, can you provide us with manual sizes?

**Comment by agryaznov** [comment]
>  since the waterfall (still) isn't working, can you provide us with manual sizes?

It's been two years from the last attempt to do what this PR is intended to. I don't think we are in a rush with this and I don't think "manual sizes" is a right thing to do whatsoever. Let's hold on until waterfall is fixed, which seems to be a more urgent issue anyway.

**Comment by HCastano** [comment]
> > since the waterfall (still) isn't working, can you provide us with manual sizes?
> 
> It's been two years from the last attempt to do what this PR is intended to. I don't think we are in a rush with this and I don't think "manual sizes" is a right thing to do whatsoever. Let's hold on until waterfall is fixed, which seems to be a more urgent issue anyway.

Getting the sizes manually is not that hard. If you take a look at what the waterfall is doing for size calculations it's basically running a single script in a loop (one for `master` and once for the PR) and just formatting the results nicely into a table.

**Comment by ascjones** [comment]
Waterfall is back https://github.com/paritytech/ink/pull/1667, merge master

**Comment by agryaznov** [comment]
> Getting the sizes manually is not that hard. If you take a look at what the waterfall is doing for size calculations it's basically running a single script in a loop (one for master and once for the PR) and just formatting the results nicely into a table.

The point is not about how's that hard or not. It's about using representative reproducible benchmarks, to compare a taken effect in a sustainable way. Manual testing was taken to justify adding those attributes. Let's not make it a vicious practice.

**Comment by paritytech-cicd-pr** [comment]
 These are the results when building the `examples/*` contracts from this branch with `cargo-contract 2.0.0-rc-da24337` and comparing them to ink! `master`: 

|                             | Δ Optimized Size | Δ Used Gas | Total Optimized Size | Total Used Gas |
|:-----------------------------|------------------:|------------:|----------------------:|----------------:|
| `accumulator`               | +0.30 K          |            | 1.42 K               |                |
| `adder`                     | +0.33 K          |            | 2.10 K               |                |
| `call-runtime`              | +0.30 K          |            | 2.10 K               |                |
| `contract-terminate`        | +0.39 K          |            | 1.68 K               | 129_990        |
| `contract-transfer`         | +0.44 K          |            | 1.90 K               | 129_990        |
| `custom-environment`        | +0.34 K          |            | 2.91 K               |                |
| `custom_allocator`          | +0.35 K          |            | 8.73 K               |                |
| `delegator`                 | +0.27 K          |            | 6.28 K               | 259_980        |
| `dns`                       | +0.30 K          |            | 9.19 K               | 389_970        |
| `erc1155`                   | +0.37 K          |            | 17.60 K              | 779_940        |
| `erc20`                     | +0.56 K          |            | 8.15 K               | 389_970        |
| `erc721`                    | +0.51 K          |            | 11.92 K              | 1_039_920      |
| `flipper`                   | +0.29 K          |            | 1.70 K               | 129_990        |
| `forward-calls`             | +0.34 K          |            | 2.79 K               | 259_980        |
| `incrementer`               | +0.30 K          |            | 1.53 K               |                |
| `mapping_integration_tests` | +0.37 K          |            | 3.73 K               |                |
| `mother`                    | -0.12 K          |            | 10.32 K              |                |
| `multisig`                  | +0.71 K          |            | 24.89 K              | 779_940        |
| `payment-channel`           | +1.05 K          |            | 7.06 K               |                |
| `psp22-extension`           | +0.31 K          |            | 6.93 K               |                |
| `rand-extension`            | +0.47 K          |            | 3.41 K               |                |
| `set-code-hash`             | +0.30 K          |            | 1.91 K               |                |
| `subber`                    | +0.33 K          |            | 2.12 K               |                |
| `trait-erc20`               | +0.50 K          |            | 8.52 K               | 389_970        |
| `trait-flipper`             | +0.29 K          |            | 1.48 K               | 129_990        |
| `trait-incrementer`         | +0.31 K          |            | 1.66 K               | 259_980        |
| `updated-incrementer`       | +0.30 K          |            | 1.92 K               |                |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/250284) | Last update: Tue Feb 28 15:01:42 CET 2023

**Comment by cmichi** [comment]
So no changes in gas, but the contract sizes got bigger.

**Comment by agryaznov** [comment]
> So no changes in gas, but the contract sizes got bigger.

@cmichi I've tweaked optimization options a little, which resulted in size decrease for most of the example contracts. See the [updated report](https://github.com/paritytech/ink/pull/1666#issuecomment-1448230054) above.

We still get slightly enlarged sizes of the following contracts:
- dns
- erc20
- erc721
- erc1155
- multisig

I have some ideas on how we could try to make them smaller, too. Will propose them in a follow-up PR if succeeded.

File: integration-tests/trait-incrementer/Cargo.toml [hunk 1]
```diff
@@ -26,3 +26,7 @@ std = [
     "traits/std",
 ]
 ink-as-dependency = []
+
+[profile.release]
```
cmichi, line 30: Please remove those from the examples.

agryaznov: why? it overrides default cargo-contract optimization options , to make the contracts smaller

athei: Some background: Those options are already passed by `cargo-contract` by default if you don't override the profile (which you just did).

agryaznov: nope, cargo-contract passes another default val, that's the point.

athei (commented):
It looks like the sizes go down for most contracts and up for some. So it is more or less neutral with a tendency to positive. Which is enough to remove the attributes.

I rather have the optimizer make this decision using its global view of the code than follow gut feeling.

I will approve once you revert your profile changes to the examples.

integration-tests/trait-incrementer/Cargo.toml
agryaznov: it sets it to `z`, while `s`seems to work out better (in most contracts)

agryaznov: I agree that `lto = true` could be removed. But not the `opt-level = 's'`

athei: Yeah I thought we were talking about `z`. lto is enabled, too:
https://github.com/paritytech/cargo-contract/blob/755ea0634d12b4642ab5a2392ea701cdd8da0a4c/crates/build/src/workspace/profile.rs#L31-L39

If we want to change the default we should make the change in cargo-contract and not in the examples. In my eyes it is not a blocker to this PR. Optimizations are chaotic and picking the right level is hard. But if our data suggests that "s" is the better default we should go with that. It is also more conservative which is good.

agryaznov:
> I will approve once you revert your profile changes to the examples.

cargo-contract's default `opt-level = 'z'` does not make always the best results. And that's why it's better to measure it and customize for the cases where the `s` setting works better, which is done here.

athei:
I don't think we should customize here. We should just switch the default to `s` in cargo-contract if data suggests that this is the better default.

Customizing here is brittle. Small changes can make `z` better. We should not care about this here.

agryaznov:
I don't think it should be the same option for all the contracts. Contracts are different in their design, and different optimization options for them is fine. If we change the contract we can do this easy benchmarking to ensure which of the two options (`z` vs `s`) work better for it. The point is, for the 5 contracts listed in this comment, `z` works better. We can of course change the default in cargo-contract, but still why not to customize it for special contracts?

agryaznov:
Alright, the argument that these contracts are not examples anymore, they became the integration tests, never deployed anywhere except a single substrate-contracts-node during CI, is accepted. 

And the clearing the profile config to just have a single non-default option, is fine by me.

So the questions left are:

2. Should we change default of `opt-level` to `s` in cargo-contract?  
    Motivation: most of the contracts have smaller size with this.
1. Should we customize the opt options in the [ink-examples repo](https://github.com/paritytech/ink-examples/), for the [small group of popular contracts](https://github.com/paritytech/ink/pull/1666#issuecomment-1450613304), as those ones exposed to a learning materials? Combined with covering this topic in use.ink docs, of course.

@cmichi @ascjones your opinion is welcomed

athei:
1. Yes. `s` seems to be the more common level and is more balanced. Since it also doesn't perform worse (even better in our examples) we should switch.

2. I think this is fair. But we should make sure that we only override the options that are not set by default (`opt-level`) in this case. This might require a change to `cargo-contract` first (maybe it already works I didn't check). We should also add a comment when overriding.

This is just my opinion. @cmichi @ascjones should whey in here before we make a decision.

ascjones:
1. Agree that if `s` is better overall for sizes of the contracts then we should change the default. However it would be good to prove this as a standalone PR (separate from the `[inline]` changes, which would either involve something like https://github.com/paritytech/cargo-contract/issues/1000 to control the `opt-level` for all contracts, or just a draft PR which simulates it by changing specifying `s` in all example manifests just to generate the size diff.

2. Agree that this is a good idea, 

> But we should make sure that we only override the options that are not set by default

I assume you mean we should only override those values which differ from the defaults.

The behaviour is that the defaults are only used if not specified in `[profile.release]`. 
image
 
So if you only specify `opt-level`, the override will be used, for the rest the defaults will be used.

cmichi (approved):

cmichi:
@agryaznov Could you resolve the merge conflicts? We can merge right after that.

ascjones:
@agryaznov resolve conflicts, merge `master` and you will get the measurements again with https://github.com/paritytech/ink/pull/1916

agryaznov:
> Given link-time optimization is [turned on by default](https://github.com/paritytech/cargo-contract/blob/153441356ac944dc31f79d4241f7838ba8e99dbc/crates/build/src/workspace/profile.rs#L34) when building contracts, those attributes should not really affect the compilation result in general.

so this initial hypothesis proved to be wrong by made measurements, hence closing this PR

PR #1669, opened 2023-02-21
Experiment to improve E2E call builder API

ascjones:
Playing around to see how we can simply remove the existing ugly callback API for building message calls. The requirements are:

1. Absolutely changes and **no** breaking changes to the `ink` call builder codegen.
2. Nice and clean API.

With this idea I have converted `erc20` to the new syntax which removes the unergonomic callback. Thus we can now use `call` on the type returned from `instantiate`:

```rust
// instantiate
let mut erc20 = client
    .instantiate::<Erc20, _, _>(
        "erc20",
        &ink_e2e::bob(),
        constructor,
        0,
        None,
    )
    .await

// build the call (no longer requires callback)
let transfer_from = erc20.call().transfer_from(
    bob_account.clone(),
    charlie_account.clone(),
    500_000_000u128,
);

/// invoke the call
client
  .call(&ink_e2e::charlie(), &transfer_from, 0, None)
  .await
```
To test this run e2e tests on the `erc20` example.

Things to see whether we improve

- [ ] remove requirement for explicit type hint on `instantiate`, see `.instantiate::<Erc20, _, _>` above
- [ ] can we get around the requirement for mutable `let mut` contract ref to access messages with `&mut self` receivers.
- [ ] can we avoid the addidional `Clone` impls added to the call builders (required because many call builder methods consume `self`)


ascjones:
@SkymanOne let me know if this PR is helpful or not, otherwise I can just close it.

PR #1670, opened 2023-02-21
Fixing `ManualKey<0>` to act properly

xgreenx:
Previously the `ManualKey<0>` acted as an `AutoKey` without the possibility of really using the zero storage key. This fix allows using the zero manual key that overrides the auto-generated key.

### DIFF
--- a/crates/storage/traits/src/impls/mod.rs
+++ b/crates/storage/traits/src/impls/mod.rs
@@ -29,6 +29,9 @@ use ink_primitives::{
 
 /// The private trait helping identify the [`AutoKey`] key type.
 trait KeyType {
+    /// It is `true` for [`AutoKey`] and `false` for [`ManualKey`].
+    /// It helps the [`ResolverKey`] select between the user-specified(left key)
+    /// and the auto-generated(right key) keys.
     const IS_AUTO_KEY: bool;
 }
 
--- a/crates/storage/traits/src/impls/mod.rs
+++ b/crates/storage/traits/src/impls/mod.rs
@@ -27,6 +27,11 @@ use ink_primitives::{
     KeyComposer,
 };
 
+/// The private trait helping identify the [`AutoKey`] key type.
+trait KeyType {
+    const IS_AUTO_KEY: bool;
+}
+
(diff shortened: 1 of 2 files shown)


crates/storage/traits/src/impls/mod.rs
```diff
@@ -174,4 +208,24 @@ mod tests {
...
+        assert_eq!(<ResolverKey<AutoKey, AutoKey> as StorageKey>::KEY, 0);
+        assert_eq!(
+            <ResolverKey<AutoKey, ManualKey<123>> as StorageKey>::KEY,
+            123
+        );
+        assert_eq!(
+            <ResolverKey<ManualKey<456>, ManualKey<123>> as StorageKey>::KEY,
+            456
+        );
+        assert_eq!(
```
xgreenx, line 226: Without the fix, it would fail with the `123` storage key=)

cmichi:
Thanks for the PR, @xgreenx!

Can you add a changelog entry to the `Unreleased` section? I guess a patch release would be best, even though it's strictly speaking a breaking change.

ascjones (approved):
LGTM

crates/storage/traits/src/impls/mod.rs
```diff
@@ -27,6 +27,14 @@ use ink_primitives::{
     KeyComposer,
 };
 
+/// The private trait helping identify the [`AutoKey`] key type.
+trait KeyType {
+    /// It is `true` for [`AutoKey`] and `false` for [`ManualKey`].
+    /// It helps the [`ResolverKey`] select between the user-specified(left key)
```
HCastano, line 33:
```suggestion
    /// It helps the [`ResolverKey`] select between the user-specified (left key)
```

crates/storage/traits/src/impls/mod.rs
```diff
@@ -27,6 +27,14 @@ use ink_primitives::{
...
+    /// and the auto-generated(right key) keys.
```
HCastano, line 34:
```suggestion
    /// and the auto-generated (right key) keys.
```

HCastano (approved):
Thanks!

Issue #1672, opened 2023-02-22
E2E tests throw `custom attribute panicked` when under cargo workspace 


wottpal:
As first reported by @ganesh1233456 [here](https://github.com/scio-labs/inkathon/pull/14), and reproduced by me, when putting an e2e tested contract directory (e.g. https://github.com/paritytech/ink-examples/tree/main/flipper) under a `Cargo.toml` with a workspace definition:

```toml
[workspace]
members = ["flipper"]
```

The e2e test command fails and throws the following errors:

```
➜  flipper git:(develop) ✗ cargo test --features e2e-tests 
   Compiling flipper v4.0.0 (/Users/wottpal/Developer/scio-labs/inkathon/packages/contracts/flipper)
error: custom attribute panicked
  --> flipper/lib.rs:62:9
   |
62 |         #[ink_e2e::test]
   |         ^^^^^^^^^^^^^^^^
   |
   = help: message: contract build for Cargo.toml failed: Cannot infer the root project id

error: custom attribute panicked
  --> flipper/lib.rs:94:9
   |
94 |         #[ink_e2e::test]
   |         ^^^^^^^^^^^^^^^^
   |
   = help: message: Once instance has previously been poisoned

warning: unused import: `super::*`
  --> flipper/lib.rs:57:13
   |
57 |         use super::*;
   |             ^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `ink_e2e::build_message`
  --> flipper/lib.rs:58:13
   |
58 |         use ink_e2e::build_message;
   |             ^^^^^^^^^^^^^^^^^^^^^^

warning: `flipper` (lib test) generated 2 warnings
error: could not compile `flipper` due to 2 previous errors; 2 warnings emitted
```

Or are we missing something? Ty in advance!

SkymanOne:
Can confirm the issue:

To reproduce:
1. Checkout https://github.com/paritytech/ink-examples
2. Create Cargo.toml in the root directory and put
```toml
[workspace]
members = ["flipper"]
```
3. run `cargo test --features e2e-tests` in the root directory

armyhaylenko:
any updates here? :(

wottpal:
we would also love to see this one fixed (preferably in the next ~2 weeks – but i know there are also a lot of other things that might need to be prioritized)..

ascjones:
Please try with https://github.com/paritytech/ink/pull/1709

[...]

PR #1673, opened 2023-02-22
Basic support for `dyn Trait` to allow cross-contract calls only with trait

xgreenx:
The change uses the ideas from the [issue](https://github.com/paritytech/ink/issues/631#issuecomment-766931785). But instead of `#[ink(dyn)]` proc macro, it uses the `contract_ref!` macro rule.

The difference between those approaches is that with `contract_ref!`, we cover all possible use cases of the usage but don't have the highlighting and hinting in the IDEs. While with `#[ink(dyn)]`, we would have basic highlighting for methods from the trait in the IDEs, the usage will be specific and limited by the number of supported syntax constructions.

The `contract_ref!` doesn't block us from adding `#[ink(dyn)]` in the future. We can have both because they use the exact mechanism under the hood.

- Added `contract_ref!` macro rule that allows creating a wrapper around the trait defined via `#[ink::trait_definition]`.
- Added `From<AccountId>` for the contract builder and forwarder to simplify the usage. 
- Added `AsRef<AccountId>` and `AsMut<AccountId>` for builder, forwarder, and contract_ref to provide the way modify the wrapper.


crates/ink/codegen/src/generator/as_dependency/call_builder.rs
```diff
@@ -132,6 +132,18 @@ impl CallBuilder<'_> {
                     <AccountId as ::core::clone::Clone>::clone(&self.account_id)
                 }
             }
+
+            impl ::core::convert::AsRef<AccountId> for #cb_ident {
```
deuszx, line 136: <3 such a small change and so much smaller code on my side :)

crates/ink/codegen/src/generator/trait_def/call_builder.rs
```diff
@@ -195,6 +195,20 @@ impl CallBuilder<'_> {
...
             }
+
+            #[cfg(feature = "std")]
+            /// We require this manual implementation since the derive produces incorrect trait bounds.
+            impl<E> ::scale_info::TypeInfo for #call_builder_ident<E>
+            where
+                E: ::ink::env::Environment,
+                <E as ::ink::env::Environment>::AccountId: ::scale_info::TypeInfo + 'static,
+            {
+                type Identity = <E as ::ink::env::Environment>::AccountId;
```
ascjones, line 206: Hm I guess we can't do `type Identity = #call_builder_ident<E>` here because it will complain about `'static` lifetime on `E`?

Downside is the moment there is no way to tell the difference between a plain `AccountId` and this in the contract metadata. Not the end of the world though, as it is still an accurate representation of the actual encoded data.

crates/ink/codegen/src/generator/trait_def/call_builder.rs
```diff
@@ -219,6 +233,16 @@ impl CallBuilder<'_> {
                 }
             }
 
+            impl<E, AccountId> ::core::convert::From<AccountId> for #call_builder_ident<E>
```
ascjones, line 236: :heart:

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,115 @@
...
+/// ```
+#[macro_export]
+macro_rules! contract_ref {
+    // The case of the default `Environment`
+    ( $trait_path:path ) => {
+        $crate::contract_ref!($trait_path, $crate::env::DefaultEnvironment)
+    };
+    // The case of the custom `Environment`
+    ( $trait_path:path, $env:ty ) => {
+        <<$crate::reflect::TraitDefinitionRegistry<$env> as $trait_path>
```
ascjones, line 120:
```suggestion
        <<$crate::reflect::TraitDefinitionRegistry<Environment> as $trait_path>
```
This would use the same environment as the containing contract. Assumes this is always used in the context of `#[ink::contract]` of course.

integration-tests/trait-dyn-cross-contract-calls/lib.rs
```diff
@@ -0,0 +1,106 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+/// We can't run E2E test without the contract. So add `Dummy` contract.
```
ascjones, line 3: Is this the same issue as https://github.com/paritytech/ink/issues/1672

ascjones (commented):
Very nice. 

Just the question mark of whether we need the `$env` arg on the `contract_ref!` macro.

crates/ink/codegen/src/generator/trait_def/call_builder.rs
xgreenx: I did it because I didn't want the appearance of the `__ink_CallBuilderFlipper` in the metadata=D After all, it would look weird to the user. Plus, it is possible to have several different `__ink_CallBuilderFlipper`(when the user defines traits with namespaces or in the different modules), and it also may confuse.

crates/ink/codegen/src/generator/trait_def/call_builder.rs
xgreenx: Later we could add the same for `ContractRef` generated for the contract itself, but it requires adding a generic to add `AccountIdGuard` bound=)

integration-tests/trait-dyn-cross-contract-calls/lib.rs
xgreenx: No, it is another reason. During running of the `cargo test --features e2e-tests` ink! tries to build the current file as a contract via `cargo-contract`. But if it is not a contract, it will fail because it doesn't have `___ink_generate_metadata` function=)

crates/ink/src/contract_ref.rs
xgreenx: You can use `contract_ref!` anywhere. It is not bounded to the `#[ink::contract]` and it is the primary feature. You can use it in any file anywhere=) It returns a plain rust type.

The `contract_ref!` works with `ink::env::DefaultEnvironment` by default, so the user should write `contract_ref!(Erc20, Environment)` to use the environment from the contract.

I did that because `DefaultEnvironment` is always available, and the macro will not throw an error outside of the `#[ink::contract]`.

But we can change the default behavior=) The `contract_ref!` could use `Environment` instead of the `DefaultEnvironment`. Then outside of the `#[ink::contract]` the user should use `contract_ref!(Erc20, DefaultEnvironment)`(some environment).

integration-tests/trait-dyn-cross-contract-calls/lib.rs
ascjones: That seems like a bug, we shouldn't need the metadata anymore for `ink_e2e`

ascjones: Anyway we should attempt to make this more ergonomic for structuring `e2e` tests in this way

crates/ink/src/contract_ref.rs
ascjones: > But we can change the default behavior=) The contract_ref! could use Environment instead of the DefaultEnvironment. Then outside of the #[ink::contract] the user should use contract_ref!(Erc20, DefaultEnvironment)(some environment).

It would be nice, but maybe it's too early in the morning but I can't think of how we would achieve this?

crates/ink/codegen/src/generator/trait_def/call_builder.rs
ascjones: Yes makes sense.

crates/ink/src/contract_ref.rs
xgreenx: Very easy=) 
image

ascjones: Oh I see so you just mean to use `Environment` by default, and if it is outside the contract then the user will just get the compilation error that it can't find this type. I think that is reasonable, as long as it is well documented. It would save users who use their own custom environment for having to specify it everywhere when used inside a contract.

ascjones: > Very easy=)

Yes I was stupidly thinking you would detect the context from within the macro. But yes that makes sense.

If it is well documented I would go this way, assuming that most people will use it inside `#[ink::contract]`

integration-tests/trait-dyn-cross-contract-calls/lib.rs
ascjones: This prompted me to open https://github.com/paritytech/ink/pull/1683 which means it doesn't need to generate the metadata anymore. 

Not sure it will make this structure work without the dummy contract, but at least it is an immprovement.

ascjones (approved):
LGTM

integration-tests/trait-dyn-cross-contract-calls/traits/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//! The trait is extracted into a separate crate to show how to do cross-contract
+//! calls only with traits without importing the contract..
```
HCastano, line 17:
```suggestion
//! calls only with traits without importing the contract.
```

HCastano (commented):
Nice, less code that I would've expected to get something working 💪 

I still want to look through the E2E tests a bit more, so I'll do that tomorrow

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,177 @@
...
+///             self.erc20.total_supply()
+///         }
+///
+///         /// The example of how to use the call builder with `contract_ref!`.
+///         #[ink(message)]
+///         pub fn total_supply_2(&self) -> Balance {
+///             use ink::codegen::TraitCallBuilder;
+///             let erc20_builder = self.erc20.call();
+///             let err: ink::env::Result<ink::primitives::MessageResult<Balance>> =
+///                 erc20_builder.total_supply().transferred_value(0).try_invoke();
```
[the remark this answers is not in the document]
xgreenx: It was done to show that the result is wrapped into two errors. When we started to use ink! it was of one initial questions=)

crates/ink/codegen/src/generator/trait_def/call_builder.rs
```diff
@@ -228,6 +252,24 @@ impl CallBuilder<'_> {
                     <<E as ::ink::env::Environment>::AccountId as ::core::clone::Clone>::clone(&self.account_id)
                 }
             }
+
+            impl<E, AccountId> ::core::convert::AsRef<AccountId> for #call_builder_ident<E>
```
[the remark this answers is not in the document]
xgreenx: I prefer `AccountId`=)

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,177 @@
...
+///         /// The example of how to do common calls.
+///         #[ink(message)]
+///         pub fn total_supply_1(&self) -> Balance {
+///             self.erc20.total_supply()
+///         }
+///
+///         /// The example of how to use the call builder with `contract_ref!`.
+///         #[ink(message)]
+///         pub fn total_supply_2(&self) -> Balance {
+///             use ink::codegen::TraitCallBuilder;
```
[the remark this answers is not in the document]
xgreenx: But there is only one way to get the `CallBuilder` - via this trait. Maybe we need considering moving this trait from the `codegen`=)

crates/ink/src/contract_ref.rs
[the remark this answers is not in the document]
HCastano: That shouldn't be the goal of this set of examples though, we want to focus on the usage of `contract_ref!`. The `CallBuilder` docs should focus on error handling

integration-tests/trait-dyn-cross-contract-calls/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+#[ink::contract]
+pub mod dummy {
```
HCastano, line 5: Instead of having this `dummy` contract can we not have the `caller` contract be the central point of these tests?

It's the one that we're interacting with anyways. This would also simplify the test setup a little bit (one less crate)

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,196 @@
 
 use ink_env::Environment;
 
+/// The macro allows the creation of a wrapper around the trait defined with
+/// [`crate::trait_definition`]. This wrapper can be used for interaction with
+/// the contract.
```
HCastano, line 19:
```suggestion
/// Generates a wrapper which can be use for interaction with the contract.
///
/// This is done by creating a wrapper around the trait defined with the
/// [`ink::trait_definition`](crate::trait_definition) macro.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,196 @@
...
+/// ```rust
+/// #[ink::contract]
+/// mod trait_caller {
+///     use ink::contract_ref;
+///     #[ink::trait_definition]
+///     pub trait Erc20 {
+///        /// Returns the total supply of the ERC-20 smart contract.
+///        #[ink(message)]
+///        fn total_supply(&self) -> Balance;
+///
```
HCastano, line 37:
```suggestion
///
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,196 @@
...
+/// so it can be used in any Rust construction that expects types.
+///
+/// # Usage in the `#[ink::contract]` context
+///
+/// A macro expects the one argument - the path to the trait to create a wrapper around it.
+///
+/// ```rust
+/// #[ink::contract]
+/// mod trait_caller {
+///     use ink::contract_ref;
```
HCastano, line 31:
```suggestion
///     use ink::contract_ref;
///
```

integration-tests/trait-dyn-cross-contract-calls/lib.rs
```diff
@@ -0,0 +1,110 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
HCastano, line 1: A top level `README` or description of the test setup here would be quite helpful

integration-tests/trait-dyn-cross-contract-calls/contracts/incrementer/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// A concrete incrementer smart contract.
+    #[ink(storage)]
+    pub struct Incrementer {
+        value: u64,
+    }
+
+    impl Incrementer {
+        /// Creates a new incrementer smart contract initialized with zero.
+        #[ink(constructor)]
+        pub fn new(init_value: u64) -> Self {
```
HCastano, line 17: No point in making this configurable, just set it to `u64::default()` in the constructor

integration-tests/trait-dyn-cross-contract-calls/contracts/incrementer/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+    #[cfg(test)]
+    mod tests {
+        use super::*;
+
+        #[test]
+        fn default_works() {
+            let incrementer = Incrementer::new(0);
+            assert_eq!(incrementer.get(), 0);
+        }
+
```
HCastano, line 49:
```suggestion
```

This adds extra noise and doesn't test anything related to `contract_ref!`

integration-tests/trait-dyn-cross-contract-calls/contracts/incrementer-caller/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+#![allow(clippy::new_without_default)]
+
+#[ink::contract]
+pub mod caller {
+    use dyn_traits::Increment;
+
+    /// The caller of the incrementer smart contract.
+    #[ink(storage)]
+    pub struct Caller {
+        /// Here we generate and accept a type which implements the `Incrementer` ink! trait.
```
HCastano, line 11:
```suggestion
        /// Here we accept a type which implements the `Incrementer` ink! trait.
```

Looking back at my suggestion this isn't correct, since the macro is only pointing to a type which was already generated by the codegen

integration-tests/trait-dyn-cross-contract-calls/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+                &ink_e2e::alice(),
+                constructor,
+                0,
+                None,
+            )
+            .await
+            .expect("instantiate failed")
+            .account_id;
+
+        // Check throw the caller that the value of the incrementer is zero
```
HCastano, line 78:
```suggestion
        // Check through the caller that the value of the incrementer is zero
```

integration-tests/trait-dyn-cross-contract-calls/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+        let get = build_message::<CallerRef>(caller_account_id.clone())
+            .call(|contract| contract.get());
+        let value = client
+            .call_dry_run(&ink_e2e::bob(), &get, 0, None)
```
HCastano, line 82:
```suggestion
            .call_dry_run(&ink_e2e::alice(), &get, 0, None)
```

We can keep using `alice` throughout, unless you have a good reason not to

crates/env/src/types.rs
```diff
@@ -94,6 +94,15 @@ impl FromLittleEndian for u128 {
     }
 }
 
+/// The trait shows that the type plans to be [`Environment::AccountId`].
+///
+/// If someone wants to use another type than [`AccountId`] in the [`Environment::AccountId`],
+/// he should implement this trait for the corresponding type.
```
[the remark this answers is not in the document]
HCastano: That doesn't really answer my question though. Why would a user want to implement this for a type other an `AccountId`? 

If we don't want this, but don't want to introduce a breaking change, then we shouldn't be highlighting the fact that they _can_ implement this trait on other types

HCastano (commented):
Couples questions around the integration tests

crates/env/src/types.rs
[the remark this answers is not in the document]
xgreenx: The user wants to implement it if he uses another type for the `Environment::AccountId` than `ink::types::AccountId`.

We want this, because it will give the `From<AccountId>` implementation to the user. If it is not a problem to add constrain for `Environment::AccountId`, I prefer to add it.

integration-tests/trait-dyn-cross-contract-calls/lib.rs
ascjones: See #1691 for a possible solution

crates/env/src/types.rs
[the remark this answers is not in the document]
HCastano: Okay I see what you mean now. I'm gonna update the comment to make this more clear

HCastano (approved):
Cool, thanks 💪

HCastano:
@cmichi wants to take a look at this as well, so let's not ape this in just yet

cmichi:
Please don't merge just yet, i'll rake a look this morning as well.

xgreenx:
> Normally, when working with Rust, dyn Trait are discouraged due to dynamic dispatching that affects the performance. 

With dynamic dispatching, the compiler is unable to optimize the code, and at runtime, it requires additional steps to find the function to execute.

> Are there any implications for Wasm programs that use dyn Traits? Is there anything unusual b/c we're interpreting the contracts (wasmi)?

The program will work as usual. Only that can affect the size of the final contract. It may be more or less than without dynamic dispatching(It is based on several factors).

BTW, the current solution doesn't use dynamic dispatching=) it is called `dyn Trait`, but the truth is that we use the type that implements the `Trait` trait.

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
 
 use ink_env::Environment;
 
+/// Generates a wrapper which can be use for interaction with the contract.
```
cmichi, line 17:
```suggestion
/// Generates a wrapper which can be used for interacting with the contract.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///
+/// This is done by creating a wrapper around the trait defined with the
+/// [`ink::trait_definition`](crate::trait_definition) macro.
+///
+/// The macro returns the native Rust type that implements the corresponding trait,
+/// so it can be used in any Rust construction that expects types.
```
cmichi, line 23:
```suggestion
/// so it can be used in any Rust context that expects types.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///
+/// # Usage in the `#[ink::contract]` context
+///
+/// A macro expects the one argument - the path to the trait to create a wrapper around it.
```
cmichi, line 27:
```suggestion
/// The macro expects one argument ‒ the path to the trait, in order to create a wrapper around it.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///     }
+///
+///     #[ink(storage)]
+///     pub struct Caller {
+///         /// The example of `contract_ref!` as a struct type.
+///         erc20: contract_ref!(Erc20),
+///     }
+///
+///     impl Caller {
+///         /// The example of `contract_ref!` as an argument type.
```
cmichi, line 52:
```suggestion
///         /// Example of `contract_ref!` as an argument type.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///         #[ink(constructor)]
+///         pub fn new(erc20: contract_ref!(Erc20)) -> Self {
+///             Self { erc20 }
+///         }
+///
+///         /// The example of converting `AccountId` into `contract_ref!` implicitly.
```
cmichi, line 58:
```suggestion
///         /// Example of converting `AccountId` into `contract_ref!` implicitly.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///         #[ink(message)]
+///         pub fn change_account_id_1(&mut self, new_erc20: AccountId) {
+///             self.erc20 = new_erc20.into();
+///         }
+///
+///         /// The example of converting `AccountId` into `contract_ref!` explicitly.
```
cmichi, line 64:
```suggestion
///         /// Example of converting `AccountId` into `contract_ref!` explicitly.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///         #[ink(message)]
+///         pub fn change_account_id_2(&mut self, new_erc20: AccountId) {
+///             let erc20: contract_ref!(Erc20) = new_erc20.into();
+///             self.erc20 = erc20;
+///         }
+///
+///         /// The example of converting `AccountId` into an alias from `contract_ref!`.
```
cmichi, line 71:
```suggestion
///         /// Example of converting `AccountId` into an alias from `contract_ref!`.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///         #[ink(message)]
+///         pub fn change_account_id_3(&mut self, new_erc20: AccountId) {
+///             type Erc20Wrapper = contract_ref!(Erc20);
+///             let erc20: Erc20Wrapper = new_erc20.into();
+///             self.erc20 = erc20;
+///         }
+///
+///         /// The example of how to do common calls via fully qualified syntax.
```
cmichi, line 79:
```suggestion
///         /// Example of how to do common calls via fully qualified syntax.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///         #[ink(message)]
+///         pub fn total_supply_1(&self) -> Balance {
+///             Erc20::total_supply(&self.erc20)
+///         }
+///
+///         /// The example of how to do common calls without fully qualified syntax.
```
cmichi, line 85:
```suggestion
///         /// Example of how to do common calls without fully qualified syntax.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///         #[ink(message)]
+///         pub fn total_supply_2(&self) -> Balance {
+///             self.erc20.total_supply()
+///         }
+///
+///         /// The example of how to use the call builder with `contract_ref!`.
```
cmichi, line 91:
```suggestion
///         /// Example of how to use the call builder with `contract_ref!`.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///         #[ink(message)]
+///         pub fn total_supply_3(&self) -> Balance {
+///             use ink::codegen::TraitCallBuilder;
+///             // Returns the `CallBuilder` that implements `Erc20` trait.
+///             let erc20_builder = self.erc20.call();
+///             erc20_builder.total_supply().transferred_value(0).invoke()
+///         }
+///
+///         /// The example of how to do common calls and convert the `contract_ref!` into `AccountId`.
```
cmichi, line 100:
```suggestion
///         /// Example of how to do common calls and convert
///         /// the `contract_ref!` into `AccountId`.
```

crates/ink/src/contract_ref.rs
```diff
@@ -14,6 +14,198 @@
...
+///             self.erc20.transfer(amount, erc20_as_account_id)
+///         }
+///     }
+/// }
+/// ```
+///
+/// # Usage outside of the `#[ink::contract]` context
+///
+/// The macro expects two arguments:
+/// - The first argument is the path to the trait, like `Erc20` or `erc20::Erc20`.
```
cmichi, line 113:
```suggestion
/// - The first argument is the path to the trait, e.g. `Erc20` or `erc20::Erc20`.
```

cmichi (approved):
Thank you, Green! Just some doc nitpicks.

cmichi:
Oh and Green, `CHANGELOG.md` entry would be great.

PR #1697, opened 2023-03-01
Clean up upgradeable contract examples

HCastano:
This PR improves the `set_code_hash` example by simplifying the code a little bit, adding
more comments, and adding an E2E test to show the workflow.

It also removes the `forward-calls` example. This example was supposed to demonstrate the
usage of proxies for upgrades, but since it used regular calls instead of delegate calls
it didn't really work as intended.

This is because the state was being stored in the logic/implementation contract as opposed
to the proxy, which would get lost between upgrades.

I'm not sure if we want to replace this with an upgradeable proxy example, or if we
should stick to using `set_code_hash` as the "recommended" upgradeability pattern.


integration-tests/set_code_hash/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        /// Modifies the code which is used to execute calls to this contract address (`AccountId`).
+        ///
+        /// We use this to upgrade the contract logic. We don't do any authorization here, any caller
+        /// can execute this method.
+        ///
+        /// In a production contract you would do some authorization here!
+        #[ink(message)]
+        pub fn set_code(&mut self, code_hash: [u8; 32]) {
+            ink::env::set_code_hash(&code_hash).unwrap_or_else(|err| {
```
ascjones, line 60: I'm wondering whether we should add this to `EnvAccess` so you can do `self.env().set_code_hash()`

ascjones: Like this: https://github.com/paritytech/ink/pull/1698

integration-tests/set_code_hash/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+                .call(&ink_e2e::alice(), inc, 0, None)
+                .await
+                .expect("`inc` failed");
+
+            let get = build_message::<IncrementerRef>(contract_acc_id.clone())
+                .call(|incrementer| incrementer.get());
+            let get_res = client.call_dry_run(&ink_e2e::alice(), &get, 0, None).await;
+
+            // Remember, we updated our incrementer contract to increment by `4`.
+            assert!(matches!(get_res.return_value(), 4));
```
ascjones, line 121: An extremely tiny :mag_right: :pinching_hand: nitpick, we should call `inc` on the original contract to increment by one first, before calling `inc` on the upgraded one. Then test for `5` here. Because it would be possible for this test to pass by having the original contract increment by 4 and not upgrading itself. Not absolutely necessary just pointing it out.

ascjones (approved):
LGTM!

ascjones (approved):
LGTM

sergejparity (approved):

HCastano:
> P.S. `ink-waterfall` needs updating for the new contract structure, see https://github.com/paritytech/ink-waterfall/blob/master/.gitlab-ci.yml#L21

Ah yeah, good point. I'm so used to ignoring the waterfall I was going to merge it anyways 😆

integration-tests/set_code_hash/lib.rs
HCastano: Ya that'd be nice

HCastano:
Hmm okay it looks like the waterfall failure is related to `Flipper`, not any outdated paths

PR #1698, opened 2023-03-02
Add `set_code_hash` to `EnvAccess`

ascjones:
- Add a new method `set_code_hash2` which accepts the `Hash` environment type instead of concrete `[u8; 32]` to match other methods which handle code hashes.
- Add a method to `EnvAccess` to allow calling `self.env.set_code_hash(&code_hash)`, which calls `set_code_hash2`.
- For backwards compatibility we keep the existing `set_code_hash` impl, which will be replaced by the `set_code_hash2` impl in the next MAJOR release


paritytech-cicd-pr:

These are the results when building the `examples/*` contracts from this branch with `cargo-contract 2.0.2-755ea06` and comparing them to ink! `master`: 

|                             | Δ Optimized Size | Δ Used Gas | Total Optimized Size | Total Used Gas |
|:-----------------------------|------------------:|------------:|----------------------:|----------------:|
| `accumulator`               |                  |            | 1.13 K               |                |
| `adder`                     |                  |            | 1.75 K               |                |
| `call-runtime`              |                  |            | 1.79 K               |                |
| `contract-terminate`        |                  |            | 1.28 K               | 129_990        |
| `contract-transfer`         |                  |            | 1.44 K               | 129_990        |
| `custom-environment`        |                  |            | 2.55 K               |                |
| `custom_allocator`          |                  |            | 8.24 K               |                |
| `delegator`                 |                  |            | 6.10 K               | 259_980        |
| `dns`                       |                  |            | 8.84 K               | 389_970        |
| `erc1155`                   |                  |            | 17.03 K              | 779_940        |
| `erc20`                     |                  |            | 7.53 K               | 389_970        |
| `erc721`                    |                  |            | 11.31 K              | 1_039_920      |
| `flipper`                   |                  |            | 1.40 K               | 129_990        |
| `forward-calls`             |                  |            | 2.44 K               | 259_980        |
| `incrementer`               |                  |            | 1.23 K               |                |
| `mapping_integration_tests` |                  |            | 3.35 K               |                |
| `mother`                    |                  |            | 10.49 K              |                |
| `multisig`                  |                  |            | 23.51 K              | 779_940        |
| `payment-channel`           |                  |            | 6.07 K               |                |
| `psp22-extension`           |                  |            | 7.02 K               |                |
| `rand-extension`            |                  |            | 2.93 K               |                |
| `set-code-hash`             |                  |            | 1.60 K               |                |
| `subber`                    |                  |            | 1.77 K               |                |
| `trait-erc20`               |                  |            | 7.95 K               | 389_970        |
| `trait-flipper`             |                  |            | 1.18 K               | 129_990        |
| `trait-incrementer`         |                  |            | 1.34 K               | 259_980        |
| `updated-incrementer`       |                  |            | 1.60 K               |                |



[Link to the run](https://gitlab.parity.io/parity/ink-waterfall/-/pipelines/250941) | Last update: Thu Mar  2 13:59:11 CET 2023

HCastano (changes requested):
I like having `Hash` instead of `[u8; 32]`, but changing the `env/api` function signature would require a `MAJOR` version bump.

It's too early in the life of 4.0 to start keeping release branches around so that `master` can contain breaking changes like this.

While it would be fine to add the function with the new signature to `EnvAccess`, I'd add it with the old signature (`[u8; 32]` ) for consistency.

ascjones:
> I like having `Hash` instead of `[u8; 32]`, but changing the `env/api` function signature would require a `MAJOR` version bump.

You're right, didn't consider that. I wonder if there is a compromise using `AsRef<[u8; 32]>`

ascjones:
> > I like having `Hash` instead of `[u8; 32]`, but changing the `env/api` function signature would require a `MAJOR` version bump.
> 
> You're right, didn't consider that. I wonder if there is a compromise using `AsRef<[u8; 32]>`

Adding a `AsRef<[u8; 32]>` impl to `types::Hash` worked, but was also a breaking change because of the existing `AsRef<[u8]>` impl. e.g. the code `some_hash.as_ref()` would now require a type annotation to pick the correct impl.

Therefore I have gone with adding a `set_code_hash2` method to the api which is called by `EnvAccess::set_code_hash`. Then in the next MAJOR we replace `set_code_hash` with `set_code_hash2`.

crates/env/src/api.rs
```diff
@@ -701,6 +701,106 @@ pub fn set_code_hash(code_hash: &[u8; 32]) -> Result<()> {
     <EnvInstance as OnInstance>::on_instance(|instance| instance.set_code_hash(code_hash))
 }
 
+/// **New version of this function which will replace [`set_code_hash`] in `v5.0`**
```
cmichi, line 704: How not to forget removing this command?

cmichi: Probably best to create a follow-up ticket, can also add the Hernando wishlist there:

> I like having `Hash` instead of `[u8; 32]`, but changing the `env/api` function signature would require a `MAJOR` version bump.

cmichi: I'll create a board for 5.0 soon, will then put the follow-up there.

crates/env/src/api.rs
```diff
@@ -701,6 +701,106 @@ pub fn set_code_hash(code_hash: &[u8; 32]) -> Result<()> {
...
+/// # Note
+///
+/// There are a couple of important considerations which must be taken into account when
+/// using this API:
+///
+/// 1. The storage at the code hash will remain untouched. This means that contract developers
+/// must ensure that the storage layout of the new code is compatible with that of the old code.
+///
+/// 2. Contracts using this API can't be assumed as having deterministic addresses. Said another way,
+/// when using this API you lose the guarantee that an address always identifies a specific code hash.
```
cmichi, line 717: I don't fully get it, the `AccountId` behind which the contract is will never change, but here it sounds like it could.

cmichi (approved):
Clarification of the comment please, besides that LGTM.

Oh, and `CHANGELOG.md` :-).

crates/env/src/api.rs
ascjones: https://github.com/paritytech/ink/issues/1701

crates/env/src/api.rs
ascjones: This whole doc comment is simply copied from the `set_code_hash` impl. 

Anyway I think the comment is accurate, because the `AccountId` is derived from [`hash(deploying_address ++ code_hash ++ salt)`](https://github.com/paritytech/substrate/blob/982f5998c59bd2bd455808345ae1bd2b1767f353/frame/contracts/src/lib.rs#L187). If the `code_hash` of a contract account were immutable you would have the guarantee that you could always derive the `AccountId` from these values. However given the code_hash is mutable, that guarantee no longer exists.

ascjones: FYI I have now just removed the duplicated doc comments, linking to the original `set_code_hash`

athei: > I don't fully get it, the `AccountId` behind which the contract is will never change, but here it sounds like it could.

The comment states that the `code_hash` is mutable **not** the contract address. The consequence is that you can't assume that you can derive the contract address of a contract deterministically that uses this function. But not because the address can change. The code hash changes.

cmichi: >2. Contracts using this API can't be assumed as having deterministic addresses. Said another way,

How does this state immutability?

cmichi: Okay, we have to clarify that in the comment. Because this:

>2. Contracts using this API can't be assumed as having deterministic addresses. 

will definitely cause confusion. The misunderstanding here would be the same train of thought I had: that you cannot necessarily rely on the contract always being behind this address after you use `set_code_hash`.

cmichi (approved):
Thanks, much cleaner!

HCastano (approved):
Thanks 👌

PR #1702, opened 2023-03-03
[don't merge] Demo sizes of examples built with `opt-level="s"`

agryaznov:


HCastano:
Closing because we're cleaning up the open PRs at the moment. We think you can just open a new PR for testing this in the future

PR #1704, opened 2023-03-04
Simplify examples which use `ContractRef`

HCastano:
The `Delegator`, now `MultiContractRef`, example has always been a bit too complicated to
really highlight the basics of how to use `ContractRef`s for cross-contract calls.

This PR introduces a set of contracts which shows the bare minimum needed for performning
cross-contract calls using `ContractRef`s. 

It also renames the `Delegator` contract to `MultiContractRef`. This was done because it
the term "delegate" should be reserved for contracts which use delegate calls. This has
also caused some confusion recently.


integration-tests/multi_contract_ref/README.md
```diff
@@ -1,13 +1,14 @@
-# Delegator Smart Contract
+# `MultiContractRef` Smart Contract
```
SkymanOne, line 1: The name can be confused with the one automatically generated by `ContractRef`.
I would name it smth like 
```suggestion
# `MultiContractCaller` Smart Contract
```
or just
```suggestion
# `MultiContract` Smart Contract
```

integration-tests/basic_contract_ref/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+#[ink::contract]
+mod basic_contract_ref {
+    /// We import the generated `ContractRef` of our other contract.
+    ///
+    /// Note that the other contract must have re-exported it (`pub use OtherContractRef`) for us
+    /// to have access to it.
+    use other_contract::OtherContractRef;
+
+    #[ink(storage)]
+    pub struct BasicContractRef {
```
SkymanOne, line 12: Similarly here `BasicContractRef` is confusing. 
```suggestion
    pub struct BasicContract {
```

SkymanOne (commented):
Looks good, although I don't think naming a contract `<ContractName>Ref` is a good idea as it can be confused with automatically generated `ContractRef` reference.
We should discourage devs from using that naming in their projects.

SkymanOne (approved):
Thanks!

sergejparity (approved):

cmichi:
@HCastano Please transfer to our `ink-examples` repo as well.

Issue #1643, opened 2023-02-07
Custom Selectors and Proxy Selector Clashing Attacks


ascjones:
I have opened https://github.com/paritytech/ink/pull/1634 to remove the custom selector functionality. *Edit*: now closed pending further decision. 

As I understand it, the motivation for doing this is that it enables the following exploit when using the "proxy" contract pattern:

```rust
/// The "fallback" handler which will handle any message with
/// no matching selector in this proxy contract
#[ink(message, payable, selector = _)]
pub fn forward(&self) -> u32 {
  // forward call to the "logic" contract which actually executes the call
}

/// This method uses a selector the same as one in the "logic" contract,
/// thus intercepting a call intended to be forwarded.
#[ink(message, selector = 0xF23A6E61)]
pub fn innocuous_function_name(&self) {
  // do some malicious stuff
}
```
This is an instance of the "Proxy selector clashing" exploit [raised first by Nomic ](https://medium.com/nomic-foundation-blog/malicious-backdoors-in-ethereum-proxies-62629adf3357) and expanded upon in this OZ article: [Beware of the proxy: learn how to exploit function clashing](https://forum.openzeppelin.com/t/beware-of-the-proxy-learn-how-to-exploit-function-clashing/1070).

The main difference is that `ink!` allows custom selectors, whereas in Solidity they can only be derived from function names. The argument is that this makes the attack more likely to succeed because:

1. The malicious contract author does not need to do the extra step of brute forcing a function name with a clashing selector.
2. Can give a malicious function a plausible sounding name with a custom selector.
3. A contract user would be more likely to overlook a malicious function with a plausible sounding name (but a custom selector).

It is objectively easier for the malicious contract author to craft a clashing malicious method with a custom selector, but would not be [too difficult to do](https://medium.com/nomic-foundation-blog/malicious-backdoors-in-ethereum-proxies-62629adf3357) the same with function names:

> Armed with some rust code, we found that clash550254402() has the same selector as proxyOwner(). It took less than 15 minutes to find it in a newish Macbook Pro. A motivated hacker can optimize the process and dedicate much more resources to finding discrete-looking function names.

As for whether a contract user reviewing the proxy contract is more or less likely to notice a malicious method with a clashing selector, that is entirely subjective. One could argue that someone familiar with `ink!` may notice the use of a custom selector in combination with a wildcard selector and immediately be suspicious, at least enough to read what the function is doing.

So now we have to ask: 

[...]

PR #1706, opened 2023-03-06
Remove wildcard selectors

ascjones:
**One of [two](https://github.com/paritytech/ink/pull/1708) possible approaches to solving the  [proxy selector clashing vulnerability](https://github.com/paritytech/ink/issues/1643)**

## Motivation
The [proxy selector clashing vulnerability](https://github.com/paritytech/ink/issues/1643) depends on the usage of wildcard selectors/fallback functions in proxy contracts. By removing wildcard selectors this vulnerability is eliminated. Contract authors can still implement upgradeability using `set_code_hash`.

## Open questions
- What other common patterns depend on this feature? 
- If they cannot be replicated by using `set_code_hash`, are there alternatives? 
- Even if we prevent those patterns being implemented is it worth it to close this vulnerability? 

### Diamonds/Multi-Facet proxy
As described in https://eips.ethereum.org/EIPS/eip-2535, this pattern of delegating to many "facet" contracts routed by the selector would no longer be possible if we removed wildcard selectors.


HCastano:
I'm of the opinion that we don't need to have wildcard/fallback functions in ink!. We
should be able to do everything with explicit message calls.

**Upgradeable Contracts**

The only reason we introduced wildcard selectors was to enable upgradeable contracts
through the proxy pattern (see https://github.com/paritytech/ink/issues/739 for more context/history).

This use case has been made obsolete with `set_code_hash`.

**Accepting Native Token Transfers**

Unlike Solidity, we also don't use the fallback (or in their case `receive`) to handle
incoming native token transfers. If we do decide that we want some special mechanics
around this then we can allow a top level `payable` attribute for contracts.

**Extensibility**

Lastly, using the fallback as a form of extensibility - like in the Diamond Pattern.

Ignoring the fact that Substrate chains can configure a maximum contract size
much more easily than Ethereum's mainnet (albeit with a performance tradeoff),
I think we can replicate this pattern and while also being more explicit.

Instead of having to implicitly pass call data through the fallback function like we do
now with the use of `set_forward_input` we can pass this data in explicitly.

For example, we can do something like this:

```rust
/// Represents an ink! message.
#[derive(scale::Encode, scale::Decode)]
struct Message {
    selector: Selector,
    input: Input,
    value: Value,
    gas_limit: Gas,
}

/// A regular message which will forward the given `message` to the extension address.
#[ink(message)]
fn forward(&mut self, ext_address: AccountId, message: Message) {
    let code_hash = self.env().code_hash(&ext_address).unwrap();

    // Option 1: Use existing `CallBuilder` pattern
    let Message {selector, input, value, gas_limit} = message;

    build_call::<DefaultEnvironment>()
        .delegate(code_hash)
            .exec_input(
                ExecutionInput::new(Selector::new(selector))
                 .push_arg(input)
        )
        .transferred_value(value)
        .gas_limit(gas_limit)
        .call_flags(
            ink::env::CallFlags::default()
                .set_tail_call(false),
        )
        .returns::<()>()
        .invoke();

    // Option 2: Add support for passing an encoded `Message` directly
    //
    // This would be similar to what how we forward input right now. We'd try and run
    // the provided `Message` with the `code_hash`.
    //
    // Would trap if `Message` failed to decode correctly on dispatch.
    build_call::<DefaultEnvironment>()
        .delegate(code_hash)
        .call_buffer(&message.encode())
        .call_flags(
            ink::env::CallFlags::default()
                .set_tail_call(false),
        )
        .invoke();
}
```

Pros:
- No "special" ink! messages
- Can still keep state in single contract
- No selector clashing since the `forward` message is always executed independently of
  the selector in `Message`

Cons:
- UIs would need to manually encode the call they're interested in
- Nesting calls to extension contracts might get messy
- Kinda looks like a poor man's version of actual Library support

ascjones:
> Instead of having to implicitly pass call data through the fallback function like we do
now with the use of set_forward_input we can pass this data in explicitly.

As you note in your *cons* list, this solution means it is no longer a "transparent" proxy, that is clients cannot simply take the metadata of a 'Logic` contract and send messages to the `Proxy` as if it is the `Logic` contract (which as you say is solved by `set_code_hash` for a single `Logic` contract). So this solution would not be easily be compatible with existing clients.

> Ignoring the fact that Substrate chains can configure a maximum contract size
much more easily than Ethereum's mainnet (albeit with a performance tradeoff),

That's true, but for larger contracts you need to pay the `proof_size` cost, which may be disproportionate if the message you are calling is only very small. So there may be some cost saving benefits to splitting a large contract into many smaller pieces.

Another possible way of implementing extensibility without wildcards/fallbacks using `set_code_hash` would be to statically define each proxiable message and just forward the call onto the specified code hash. We might need to build that into the language to avoid the payload being decoded/encoded.

```
#[ink(storage)]
struct Contract {
  logic: Hash,
}

/// generates code to 
#[ink(message(delegate_to = self.logic))]
pub fn transfer(&self, _to: AccountId, _value: Balance) {}
```

ascjones:
Closing in favour of https://github.com/paritytech/ink/pull/1708

Issue #1676, opened 2023-02-23
IIP-2: Limit contracts with a wildcard selector to one other message


ascjones:
## Abstract
Wildcard selectors used in [proxy patterns][1] are vulnerable to malicious or accidental selector clashing, causing the `Proxy` contract to intercept a message intended for a `Logic` contract. This proposal aims to eliminate this vulnerability by restricting an `ink!` contract with a wildcard selector to only having one other message with a well-known reserved selector.

## Motivation

Wildcard selectors can be used to annotate a so called "fallback" function which handles any message with a selector which does not match any of the other selectors in an `ink!` smart contract. These can be used to implement common [proxy patterns][1] for contract upgradeability. Users send messages to a `Proxy` contract which forwards/delegates messages to a `Logic` contract via the fallback message handler e.g.

```rust=
/// Handles any message with no matching selector in this proxy contract
#[ink(message, payable, selector = _)]
pub fn fallback(&self) {
  // forward call to the "logic" contract which actually executes the call
}
```

However this introduces a vulnerability whereby the `Proxy` contract may define a message with a selector identical to one from the `Logic` contract. In this case the `Proxy` would handle the message rather than it being passed on to the `Logic` contract. This is known as [Proxy Selector Clashing][2], described in detail in this [OpenZeppelin post][3].

Because the `Logic` contract can be upgraded dynamically, there is no way to guarantee at compile time whether a selector clash exists.

## Specification

The proposal is to restrict contracts with a wildcard selector to only have one other message with a reserved/well known selector. This guarantees that there are no selector clashes, either by chance or malicious intent, and that the `Proxy` will only handle messages intended for it.

If a contract uses a wildcard selector `#[ink(message, payable, selector = _)]` it MAY define one other message. This message MUST use the reserved selector `::ink::IIP2_HANDLER_SELECTOR`. This selector MUST only be used in conjunction with a wildcard selector. 

```rust=
/// Handles any message with no matching selector in this proxy contract
#[ink(message, selector = _)]
pub fn fallback(&self) {
  // forward call to the "logic" contract which actually executes the call
}

#[derive(Decode)]
pub enum ProxyMessage {
    UpgradeContract(Hash),
}

/// One other message allowed to handle messages.
/// Fails to compile unless `IIP1_HANDLER_SELECTOR` is used.
#[ink(message, selector = ::ink::IIP2_HANDLER_SELECTOR)]
pub fn handler(&self, msg: ProxyMessage) {
    match msg {
[...]

PR #1708, opened 2023-03-10
Restrict wildcard selectors to have exactly one other message

ascjones:
**Draft: a WIP implementation of https://github.com/paritytech/ink/issues/1676, one of the possible solutions to https://github.com/paritytech/ink/issues/1643. Will mark as ready to review once complete and if/when we decide this is the correct course of action.**

The current implementation here differs slightly from the spec, in that it requires a contract to always define the message with the well-known reserved selector when using the wildcard selector. In the spec it is optional. This allows the user to control what happens when a message with that selector when it is sent. The alternatives are to trap, or to pass the message into the wildcard handler.

## todo

- [ ] decide whether to require the user to always define the other message complementing the wildcard, or to default to throwing an error or passing through to the wildcard handler.
- [ ] define well-known reserved constant for the other message required when using a wildcard.
- [ ] prevent this well-known reserved selector being used for normal messages.

### DIFF
--- a/crates/ink/ir/src/ir/item_mod.rs
+++ b/crates/ink/ir/src/ir/item_mod.rs
@@ -275,14 +275,17 @@ impl ItemMod {
                 }
             }
 
+            const WELL_KNOWN_SELECTOR: [u8; 4] = [0x00, 0x00, 0x00, 0x00];
+
             if let Some(wildcard) = wildcard_selector {
                 match other_messages.len() as u32 {
                     0 => (),
                     1 => {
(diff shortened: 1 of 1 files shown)


ascjones:
CI failures fixed in https://github.com/paritytech/ink/pull/1733

SkymanOne (approved):
I personally believe that we should favour this PR in contrast to #1706
The reason is that emerging contracts pattern like diamond proxy pattern may rely on wildcard selector. While we do solve the problem of upgradability with `set_code_hash` by removing the selector we would harm the modularity of ink! smart contracts.

Restricting wildcard selector ensures that we close the vulnerability while maintaining the support for proxy patterns.

cmichi (approved):

PR #1712, opened 2023-03-14
Don't grow memory on first allocation

athei:
Right now the allocator always grows the memory on the first allocation. This is because it isn't aware of how much heap is available on initialization.

Growing the memory is a very expensive operation. I can imagine that this is especially bad for calls which do little work but still allocate some memory. Not sure if we have such a performance profile in our waterfall. But I can imagine that this impacts getter functions.

This solution avoids this by first using the available heap. It is not perfect as it performs a bit more work per allocation. We could probably get rid of that with enough willpower and `unsafe`.


crates/allocator/src/bump.rs
HCastano, line 1: So what this is trying to change is that when we do our first call to `alloc()` we don't need to `request_pages` causing a call to `wasm32::memory_grow()`?

Even with the `inner.next = heap_size` how do we know that we won't need to request a extra page for the allocation?

crates/allocator/src/bump.rs
```diff
@@ -101,13 +120,36 @@ impl InnerAlloc {
...
+                }
+                let heap_start =  unsafe { &__heap_base as *const usize as usize };
+                // if the symbol isn't found it will resolve to 0
+                // for that to happen the rust compiler or linker need to break or change
+                assert_ne!(heap_start, 0);
+                heap_start
+            }
+            fn heap_end() -> usize {
+                // cannot overflow on this arch
+                core::arch::wasm32::memory_size(0) * PAGE_SIZE
```
HCastano, line 138: Does `wasm32::memory_size()` change as we request more pages during execution?

crates/allocator/src/bump.rs
```diff
@@ -101,13 +120,36 @@ impl InnerAlloc {
                     return None;
                 }
 
-                prev_page.checked_mul(PAGE_SIZE)
+                // cannot overflow on this arch
```
HCastano, line 123:
```suggestion
                // Cannot overflow on this architecture
```

crates/allocator/src/bump.rs
```diff
@@ -101,13 +120,36 @@ impl InnerAlloc {
...
+                    static __heap_base: usize;
+                }
+                let heap_start =  unsafe { &__heap_base as *const usize as usize };
+                // if the symbol isn't found it will resolve to 0
+                // for that to happen the rust compiler or linker need to break or change
+                assert_ne!(heap_start, 0);
+                heap_start
+            }
+            fn heap_end() -> usize {
+                // cannot overflow on this arch
```
HCastano, line 137:
```suggestion
                // Cannot overflow on this architecture
```

HCastano (commented):
Couples questions around Wasm instructions

crates/allocator/src/bump.rs
```diff
@@ -101,13 +120,36 @@ impl InnerAlloc {
...
+                Some(prev_page * PAGE_SIZE)
+            }
+            fn heap_start() -> usize {
+                extern "C" {
+                    static __heap_base: usize;
+                }
+                let heap_start =  unsafe { &__heap_base as *const usize as usize };
+                // if the symbol isn't found it will resolve to 0
+                // for that to happen the rust compiler or linker need to break or change
+                assert_ne!(heap_start, 0);
```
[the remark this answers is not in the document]
athei: Yes. But:
1. It is stripped out when using `--release`.
2. The assert generates a default message anyways.

crates/allocator/src/bump.rs
athei: Yes. It it tells you the current size. But since we only call it for the first allocation it doesn't matter.

crates/allocator/src/bump.rs
HCastano: Hey, can you clarify this?

crates/allocator/src/bump.rs
```diff
@@ -101,13 +120,36 @@ impl InnerAlloc {
...
                 }
 
-                prev_page.checked_mul(PAGE_SIZE)
+                // cannot overflow on this arch
+                Some(prev_page * PAGE_SIZE)
+            }
+            fn heap_start() -> usize {
+                extern "C" {
+                    static __heap_base: usize;
+                }
```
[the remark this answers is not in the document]
HCastano: Bumping this

crates/allocator/src/bump.rs
athei: > So what this is trying to change is that when we do our first call to `alloc()` we don't need to `request_pages` causing a call to `wasm32::memory_grow()`?

Correct

> Even with the `inner.next = heap_size` how do we know that we won't need to request a extra page for the allocation?

`inner.next = heap_start` would be correct. We know because `inner.upper_limit = heap_end`.  We know the first and last address of the current heap. So when allocating we know whether we need to grow.

athei:
Merged master and answered questions. Ready for another round.

crates/allocator/src/bump.rs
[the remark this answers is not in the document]
athei: Done

crates/allocator/src/bump.rs
```diff
@@ -101,13 +120,36 @@ impl InnerAlloc {
...
+                let heap_start =  unsafe { &__heap_base as *const usize as usize };
```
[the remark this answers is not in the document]
athei: Done

crates/allocator/src/bump.rs
```diff
@@ -108,12 +115,24 @@ impl InnerAlloc {
...
+                // for that to happen the rust compiler or linker need to break or change
+                assert_ne!(heap_start, 0, "Can't find `__heap_base` symbol.");
+                heap_start
             }
 
-            const fn heap_end() -> usize {
-                0
+            fn heap_end() -> usize {
+                // Cannot overflow on this architecture
+                core::arch::wasm32::memory_size(0) * PAGE_SIZE
```
xgreenx, line 135: I think the `heap_end` is a bad name because otherwise, it should be like:

```rust
                Self::heap_start() + core::arch::wasm32::memory_size(0) * PAGE_SIZE
```

Maybe it is better to rename the function into `upper_limit`.

xgreenx (commented):
Did I get right that there is one available page by default, which is why this change will help?

athei:
> Did I get right that there is one available page by default, which is why this change will help?

Yes. But not necessarily one. Could be more. Depends on the program. `rustc` will request as many initial pages as it needs to place its stack and other data. The point is that the last page will have some free space we can use without doing more allocations.

crates/allocator/src/bump.rs
xgreenx: Got it, `core::arch::wasm32::memory_size(0)` returns the number of pages for the whole memory, including stack and heap. Then yeah, it makes sense.

xgreenx (approved):

PR #1715, opened 2023-03-14
Change `NO_STD_TARGET` to `aarch64-unknown-none` :t-rex:

ascjones:
Since Rust `1.68`, [`check-no-std` fails](https://gitlab.parity.io/parity/mirrors/ink/-/jobs/2515968). Lots of errors missing definitions from [`atomic.rs` ](https://github.com/rust-lang/rust/blob/474ea87943c3077feb7d7f2a6f295bd06654ef82/library/core/src/sync/atomic.rs) e.g.

``cannot find macro `atomic_int` in this scope``

I have confirmed this by running the check with `cargo +1.67.1` and it works okay.

This looks to be an occurrence of https://github.com/rust-lang/rust/issues/106795, which is fixed by https://github.com/rust-lang/rust/pull/106856, but not stabilized yet (I'm not sure exactly how to tell). In any case it works with `cargo +nightly` so I assume this target will be fixed in a upcoming stable release.

In the meantime I have chosen a different target which has the same property of only supporting `no_std`: `aarch64-unknown-none`. This is a tier-2 target which is guaranteed to build, see https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -42,7 +42,7 @@ variables:
   # This is a target that does not have a standard library in contrast to Wasm. We compile against
   # this target to make sure that we don't pull in `std` by accident (through dependencies).
   # See https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2.
-  NO_STD_TARGET:                   "thumbv7m-none-eabi"
+  NO_STD_TARGET:                   "aarch64-unknown-none"
 
 workflow:
   rules:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -41,7 +41,8 @@ variables:
   CLIPPY_ALLOWED:                  "clippy::extra_unused_lifetimes"
   # This is a target that does not have a standard library in contrast to Wasm. We compile against
   # this target to make sure that we don't pull in `std` by accident (through dependencies).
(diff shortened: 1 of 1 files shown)


cmichi (approved):

HCastano:
@athei since the change that broke the `bpf` target seems to be related to atomics, is your [comment from the original PR](https://github.com/paritytech/ink/pull/1646#issuecomment-1424558335) outdated now?

Just wondering if we need to switch back to a more constrained target in the future, or if keeping the `aarch64` one is fine

ascjones:
We can certainly switch back to `bpf` once the fix is stabilized, it seems that works again on `nightly`. Although since atomics are in `core` and we are doing `-Zbuild-std="core,alloc"`, I'm not sure whether the fix is that atomics are enabled again for `bpf`. So we need some clarity.

I'm wondering whether we could/should create our own custom target as per https://docs.rust-embedded.org/embedonomicon/custom-target.html.

athei:
I suggest we use `riscv32imc-unknown-none-elf` as this the current prospect for the next contract byte code we will support.

ascjones:
This is the relevant issue for atomics https://github.com/rust-lang/rust/issues/106795

Issue #1713, opened 2023-03-14
Allow setting block time for off-chain testing


cmichi:
Escalated from this StackExchange question: https://substrate.stackexchange.com/questions/7034/is-it-possible-to-override-ink-envtest-default-block-time.

## ToDo
- [ ] Add a method `Engine::set_block_time(time: BlockTimestamp)`.

CrommVardek:
I'm working on this, but before submitting my PR, shouldn't the env_access (using Env api) also implement a ink_env::set_block_time() function ?

Sorry if I'm missing something here, first time digging into ink! code base.

cmichi:
@CrommVardek Yes, you're correct.

It's best if you look at the function `advance_block` in `crates/engine/src/test_api.rs` and implement `set_block_time` analog to it. So in your IDE you can look for usages of `advance_block` and add `set_block_time` to all these places as well.

CrommVardek:
PR #1721 opened

cmichi:
Closed with https://github.com/paritytech/ink/pull/1721.

PR #1721, opened 2023-03-18
Allow setting block time for off-chain testing

CrommVardek:
Adding a `Engine::set_block_timestamp` function to solve issue #1713 

Added a test accordingly.

### DIFF
--- a/crates/engine/src/exec_context.rs
+++ b/crates/engine/src/exec_context.rs
@@ -65,7 +65,7 @@ impl ExecContext {
         *self = Default::default();
     }
 
-    /// Set the execution context block timestamp
+    /// Set the block timestamp for the execution context.
     pub fn set_block_timestamp(&mut self, block_timestamp: BlockTimestamp) {
         self.block_timestamp = block_timestamp
     }
--- a/crates/engine/src/exec_context.rs
+++ b/crates/engine/src/exec_context.rs
@@ -16,7 +16,7 @@ use super::types::{
     AccountId, 
     Balance, 
     BlockNumber, 
-    BlockTimestamp
+    BlockTimestamp,
 };
 
 /// The context of a contract execution.
@@ -73,7 +73,10 @@ impl ExecContext {
 
 #[cfg(test)]
 mod tests {
-    use super::{AccountId, ExecContext};
+    use super::{
+        AccountId, 
+        ExecContext,
+    };
 
     #[test]
     fn basic_operations() {
--- a/crates/engine/src/exec_context.rs
+++ b/crates/engine/src/exec_context.rs
@@ -13,9 +13,9 @@
 // limitations under the License.
 
 use super::types::{
-    AccountId, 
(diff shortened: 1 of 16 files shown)


crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -240,6 +240,16 @@ where
     })
 }
 
+/// Sets block timestamp for the next advance block.
```
cmichi, line 243:
```suggestion
/// Sets the block timestamp for the next [`advance_block`] invocation.
```

crates/engine/src/exec_context.rs
```diff
@@ -64,6 +64,11 @@ impl ExecContext {
     pub fn reset(&mut self) {
         *self = Default::default();
     }
+
+    /// Set the execution context block timestamp
```
cmichi, line 68:
```suggestion
    /// Set the execution context block timestamp.
```

crates/engine/src/test_api.rs
```diff
@@ -257,6 +258,10 @@ impl Engine {
     pub fn set_value_transferred(&mut self, value: Balance) {
         self.exec_context.value_transferred = value;
     }
+
+    pub fn set_block_timestamp(&mut self, new_block_timestamp: BlockTimestamp) {
```
cmichi, line 262:
```suggestion
    /// Set the block timestamp for the execution context.
    pub fn set_block_timestamp(&mut self, new_block_timestamp: BlockTimestamp) {
```

crates/engine/src/tests.rs
```diff
@@ -272,3 +272,20 @@ fn ecdsa_recovery_with_secp256k1_crate() {
...
+#[test]
+fn setting_getting_block_timestamp() {
+    // given
+    let mut engine = Engine::new();
+    let new_block_timestamp: u64 = 1000;
+    let output = &mut &mut get_buffer()[..];
+
+    // when
+    engine.set_block_timestamp(new_block_timestamp);
+    engine.block_timestamp(output);
```
cmichi, line 285: The test is pretty good and clearly readable. One thing: could you include testing `advance_block` in the test?

cmichi (commented):
The PR is already pretty good, just some nits. 

Could you also update the file `CHANGELOG.md`? There is an `[Unreleased]` section and you can add it under a `## Added` heading there.

crates/engine/src/tests.rs
CrommVardek: Essentially, one might want to use the off_chain engine that way : 

```rust 
engine.set_block_time(new_block_time);
engine.advance_block();
engine.block_timestamp(); // Should return the previous block_timestamp + new_block_time
```

"Incomplete" might not be the best word here, but rather, "could have a complementary implementation".

cmichi (approved):
Thank you, @CrommVardek !

Hope to see more PR's of you :-).

Issue #1703, opened 2023-03-03
Add `default` attribute for constructors and messages


cmichi:
I would like the possibility of optionally annotating a default constructor/message for a contract.

This is inspired by workshops, where we had all kinds of mishaps because people submitted transactions with the first constructor/message selected in the dropdown. Having this functionality for messages/constructors that are supposed to be used during workshops will make the life of everyone easier and reduce the source for confusion.

See here how the default selection looks for `ink-workshop/game`:

![screenshot-contracts-ui substrate io-2023 03 03-19_00_23](https://user-images.githubusercontent.com/241530/222794120-5d9902b2-678f-4d00-938d-6e30486efd2e.png)

We had a lot of people confused by the immediate red error `Contract Trapped` because `destroy()` (a privileged method) is selected by default.

## ToDo

- [x] Add a new ink! attribute `default` that is optional and can be used exactly once for a constructor and once for a message.
- [x] Reflect this information in the metadata. Just add a field `default: bool` for the `message` or `constructor` object:
```json
      {                    
        "args": [],    
        "default": true,      
        "docs": [
          " Flips the current value of the Flipper's boolean."
        ],                 
        "label": "flip",              
        "mutates": true,     
        "payable": false,                                                      
        "returnType": { 
          "displayName": [   
            "ink",           
            "MessageResult"
          ],               
          "type": 1        
        },           
        "selector": "0x633aa551"
      },   
```
- [x] Create a follow-up issue for Contracts UI and `polkadot-js` to choose this entry by default for constructors/messages.
- [x] Create a follow-up issue (or PR, if you're up for it) for [`ink-docs`](https://github.com/paritytech/ink-docs) to extend the documentation of https://use.ink/basics/metadata.

kziemianek:
Hi @cmichi , I would like to take this task.

SkymanOne:
> Hi @cmichi , I would like to take this task.

Hi. No one currently works on this issue. You are more than welcome to tackle it and create a PR. We will be happy to review it and give other assistance if necessary.

kziemianek:
Hi, I've opened pull request with required changes.  I've also resolved few conflicts with commits ugrading syn to version 2. Could you please take a look, there are several changes to be done in downstream projects, the first one is https://github.com/polkadot-js/api .

SkymanOne:
Closing as all the tasks have been completed. Thank you @kziemianek for your contribution!

kziemianek:
@SkymanOne thanks for collaboration!

PR #1724, opened 2023-03-20
Add `#[ink(default)]` attribute for constructors and messages

kziemianek:
Adds `default` attribute for constructors and messages. 
`default` attribute is reflected in metadata so consuming services can for example determine which constructor/message should be selected by default.

#1703

### DIFF
--- a/crates/metadata/src/specs.rs
+++ b/crates/metadata/src/specs.rs
@@ -228,6 +228,10 @@ where
             !self.spec.messages.is_empty(),
             "must have at least one message"
         );
+        assert!(self.spec.constructors.iter().filter(|c| c.default).count() < 2,
+                "only one default constructor is allowed");
+        assert!(self.spec.messages.iter().filter(|m| m.default).count() < 2,
+                "only one default message is allowed");
         self.spec
     }
 }
@@ -275,6 +279,8 @@ pub struct ConstructorSpec<F: Form = MetaForm> {
     pub return_type: ReturnTypeSpec<F>,
     /// The deployment handler documentation.
     pub docs: Vec<F::String>,
+    /// If the constructor is default
+    default: bool
 }
 
 impl IntoPortable for ConstructorSpec {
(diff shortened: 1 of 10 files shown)


kziemianek:
@HCastano i see you are takinkg a 👀 at this PR. A lot checks have failed because of:

`fatal: unable to access 'https://gitlab.parity.io/parity/mirrors/ink.git/': Failed to connect to gitlab.parity.io port 443 after 130941 ms: Operation timed out`

It seems like a one time issue, do you have permission to retry checks ?

crates/metadata/src/specs.rs
```diff
@@ -275,6 +283,8 @@ pub struct ConstructorSpec<F: Form = MetaForm> {
     pub return_type: ReturnTypeSpec<F>,
     /// The deployment handler documentation.
     pub docs: Vec<F::String>,
+    /// If the constructor is default
```
cmichi, line 286:
```suggestion
    /// If the constructor is default.
```

crates/metadata/src/specs.rs
```diff
@@ -498,6 +525,8 @@ pub struct MessageSpec<F: Form = MetaForm> {
     return_type: ReturnTypeSpec<F>,
     /// The message documentation.
     docs: Vec<F::String>,
+    /// If the message is default
```
cmichi, line 528:
```suggestion
    /// If the message is default.
```

HCastano:
```suggestion
    /// If the message is the default for off-chain consumers (e.g UIs).
```

cmichi (approved):
Thank you! Can you create the follow-up issues as mentioned in https://github.com/paritytech/ink/issues/1703?

crates/metadata/src/specs.rs
HCastano:
```suggestion
    /// If the constructor is the default for off-chain consumers (e.g UIs).
```

HCastano (approved):
Nice work, thanks!

crates/metadata/src/specs.rs
```diff
@@ -275,6 +283,8 @@ pub struct ConstructorSpec<F: Form = MetaForm> {
     pub return_type: ReturnTypeSpec<F>,
     /// The deployment handler documentation.
     pub docs: Vec<F::String>,
+    /// If the constructor is default.
```
HCastano, line 286:
```suggestion
    /// If the constructor is the default for off-chain consumers (e.g UIs).
```

PR #1731, opened 2023-03-27
Upgrade to `syn v2`

SkymanOne:
- Bumps all `syn` dependencies to `2`
- Attribute parsing has been reworked by @ascjones


crates/ink/ir/src/ir/utils.rs
```diff
@@ -174,13 +176,12 @@ pub fn extract_cfg_attributes(
...
         .map(|a| {
-            a.tokens
-                .clone()
-                .into_iter()
-                .map(|token| quote::quote_spanned!(span=> #[cfg #token]))
-                .collect()
+            let expr = a
+                .parse_args::<Expr>()
+                .expect("Encountered cfg attribute in wrong format.");
+            quote::quote_spanned!(span=> #[cfg( #expr )])
```
ascjones, line 184: Why parse it? Simply `quote::quote_spanned!(span=> #a)` may work.

If it does we could even take it a step further and make this fn return `Vec<syn::Attribute>` directly.

ascjones (approved):
LGTM :)

crates/ink/ir/src/ir/utils.rs
SkymanOne: The first approach would work better because we want to return an attribute in spanned context.

PR #1737, opened 2023-04-03
Automatically export preludes

SkymanOne:
We've seen a number of questions related to inability to use structures like `Vec` and `String` in smart contract. This is because we require those data types to be explicitly imported from the ink! prelude. 

This PR automatically imports default crates that contain macros and data structures that are present in a typical rust program by default.

### DIFF
--- a/crates/ink/codegen/src/generator/contract.rs
+++ b/crates/ink/codegen/src/generator/contract.rs
@@ -53,7 +53,7 @@ impl GenerateCode for Contract<'_> {
         quote! {
             #( #attrs )*
             #vis mod #ident {
-                use ink::prelude::{borrow::*, boxed::*, format, string::*, vec::*};
+                use ink::prelude::{borrow::*, boxed::*, format, string::*, vec::*, vec};
                 #env
                 #storage
                 #events
--- a/crates/ink/codegen/src/generator/contract.rs
+++ b/crates/ink/codegen/src/generator/contract.rs
@@ -53,7 +53,7 @@ impl GenerateCode for Contract<'_> {
         quote! {
             #( #attrs )*
             #vis mod #ident {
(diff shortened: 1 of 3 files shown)


crates/ink/codegen/src/generator/contract.rs
```diff
@@ -53,6 +53,7 @@ impl GenerateCode for Contract<'_> {
         quote! {
             #( #attrs )*
             #vis mod #ident {
+                use ::ink::prelude::{borrow::*, boxed::*, format, string::*, vec::*};
```
athei, line 56: Why isn't our prelude flat as it is customary for preludes? Then users can just do `prelude::*`. There also seem to be things in the sub modules that we probably don't want to pull in.

SkymanOne: If we do that. The dev still needs to explicitly specify crates: `vec::Vec` & `string::String` which is not ideal.

In "classic" rust program this data structures are available by default.

athei: > If we do that. The dev still needs to explicitly specify crates: vec::Vec & string::String which is not ideal.

If those types are directly inside `prelude` they will be available by just having `use prelude::*`.

> In "classic" rust program this data structures are available by default.

Because the compiler inserts `use std::prelude::v1::*` for you and the types are all in the root (flat):
https://doc.rust-lang.org/stable/std/prelude/v1/index.html


I am just saying we should not have sub modules in our prelude. It should just be a list of re-exports into the root (sorry for shaving the yak here :D).

HCastano:
We talked about this and we figured that there should be a little bit of a barrier to using `String` or `Vec` since they're not idiomatic to use.

If people are struggling to find these, we may need to document their location a bit better in our examples

PR #1738, opened 2023-04-04
Introduces contract demonstrating unit + e2e test event decoding.

blairmunroakusa:
I did my best to satisfy the contributor guidelines. Please let me know if anything needs revision.

### DIFF
--- a/integration-tests/test_events_decode/README.md
+++ b/integration-tests/test_events_decode/README.md
@@ -2,7 +2,7 @@
 
 Capturing events for ink! end-to-end tests in particular is not clearly documented, nor are there clear reference implementations. This contract contains a minimalist flipflop contract that emits two different events when called. It contains both unit and e2e test modules illustrating one way to check for multiple events in both testing environments.
 
-The end-to-end event test/check is awkward and could certainly be improved on. This reference may not be appropriate for it's own example project in the ink-examples repo, so it should be up to this repo's maintainers/seasoned-folk to figure out where to put it, if at all.
+The end-to-end event test/check is awkward and could certainly be improved on.
 
 Notes:
 
--- a/integration-tests/test_events_decode/README.md
+++ b/integration-tests/test_events_decode/README.md
@@ -1,0 +1,11 @@
+# Testing Events Reference Implementation
+
+Capturing events for ink! end-to-end tests in particular is not clearly documented, nor are there clear reference implementations. This contract contains a minimalist flipflop contract that emits two different events when called. It contains both unit and e2e test modules illustrating one way to check for multiple events in both testing environments.
+
+The end-to-end event test/check is awkward and could certainly be improved on. This reference may not be appropriate for it's own example project in the ink-examples repo, so it should be up to this repo's maintainers/seasoned-folk to figure out where to put it, if at all.
+
+Notes:
+
+- e2e: depending on the field number and content of an event struct, the byte range in the `&event` that is decoded may be either `&event[34..]` or `&event[35..]`. This is a point to improve on, dealing with the prepending bytes that don't pertain to the field values.
+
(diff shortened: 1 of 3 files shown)


blairmunroakusa:
@ascjones Hey, I don't understand the reason continuous-integration/gitlab-examples-fmt fails. Do you have the cycles to help with that?

integration-tests/test_events_decode/README.md
```diff
@@ -0,0 +1,11 @@
+# Testing Events Reference Implementation
+
+Capturing events for ink! end-to-end tests in particular is not clearly documented, nor are there clear reference implementations. This contract contains a minimalist flipflop contract that emits two different events when called. It contains both unit and e2e test modules illustrating one way to check for multiple events in both testing environments.
```
SkymanOne, line 3: This contract contains a minimalist flipper contract that emits two different events when called. It contains both unit and e2e test modules illustrating one way to check for multiple events in both testing environments.

integration-tests/test_events_decode/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// This is one way to capture and check event emission in ink unit tests.
+    #[cfg(test)]
+    mod tests {
+
+        use super::*;
+        use ink::env::test::EmittedEvent;
+
+        type Event = <TestingEventDecode as ::ink::reflect::ContractEventBase>::Type;
+
+        fn decode_events(emitted_events: Vec<EmittedEvent>) -> Vec<Event> {
```
SkymanOne, line 73: This is actually a very useful function to be added to e2e create directly. Can you please look into this?

integration-tests/test_events_decode/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+The end-to-end event test/check is awkward and could certainly be improved on.
```
SkymanOne, line 5: Can you please create a follow-up issue for this?

SkymanOne (approved):

SkymanOne:
Please run `cargo +nightly fmt` to format the code and make CI happy

integration-tests/test_events_decode/lib.rs
blairmunroakusa: To clarify, do you mean reimplement this `decode_events` function to work within the ink_e2e test scope? Or do you mean include this particular `decode_events` function as `ink::env::test::decode_events`? Or both?

If the former, this would be part of the followup issue regarding improving event decode to not be so awkward.

In any case, this might be something @ganesh1233456 or @alessandro-baldassarre may wish to address individually. I will see to it that the proper issues are created.

SkymanOne: I think for unit tests, we want to have something like `ink::env::test::recorded_events().decode()` or `ink::env::test::decoded_events()`.  To check whether the events have been emitted, you can use something like  `.contains()` or similar list functions.

With e2e tests it's a little bit trickier, we probably want to have something like `let flip_event = flipflop_result.events.find_decoded::<FlipEvent>()` or similar to automatically check whether the event has been emitted, decode it and then assert on its values.

ascjones:
Hi @blairmunroakusa, since you opened this PR we have changed the way we emit events, see https://github.com/paritytech/ink/pull/1827.

So have a look at the example I added as part of that PR: https://github.com/paritytech/ink/blob/master/integration-tests/events/lib.rs.

I'll close this for now, but feel free to reopen it or open a new one which is compatible with Events 2.0

blairmunroakusa:
Hey @ascjones @SkymanOne sorry for ghosting :/ ; life got me. When I find the cycles, I will look into following through on this. Thank you for the cordial and educational intro to the Parity world; I appreciate your patience and professionalism, and I look forward to properly contributing when the time is right.

Issue #1682, opened 2023-02-24
Provide a `StorageVec` datastructure


xermicus:
Currently, we only provide a `Mapping`. However, storing things in a Vector (Array) on contract storage is also as thing our users need. Using the rust `Vec` from the prelude has a fundamental issue: It exhibits packed layout. This makes it a footgun when used on contract storage, easily leading to various sorts of DoS vulnerabilities.

There once was a dedicated [storage vec data structure](https://github.com/paritytech/ink/blob/v3.0.0-rc8/crates/storage/src/collections/vec/mod.rs). This data structure would still use classical data structures from the `prelude`, but used its own `StorageEntry` struct to wrap it's data in and lazily read or write to/from storage. The approach would be to re-work it to function with the new storage API. On a high level, this should work. It will generate some amount of work implement though (I think it is more involved than just copying over the old code and make some minor changes).

Another thing is, for example, iterating over a Vec with 1000 elements on it, will still cause 1000 storage reads to the contracts pallet, and this will be costly. So I thought about whether we should try to be smart and do some further caching by designing a Lazy data structure that can read and write data in chunks. But it implies some drawbacks as well, e.g. additional complexity and it is not clear to me how we should determine an "optimal" chunk size.

xermicus:
From another discussion; it could also be useful to provide some kind of "reference" data structure (akin to `Box<>`), that just holds a reference (the storage key) to a `Vec<>` or any other data structure on storage. This would specifically allow to store anything in a mapping, without risking to blow up the contract because a mapping value got too large to decode.

So, the approach here would be:
- Implement the reference type
- Implement the storage vector type describe above

@xgreenx WDYT?

xgreenx:
I didn't get how the "reference" should work.  For me, it is similar to the `Lazy` type that we have=) Could you elaborate more, please?

> This would specifically allow to store any packed data in a mapping, without risking to blow up the contract because a mapping value got too large to decode.

Hmm, but `Mapping` already pulls only the data from one storage cell for packed types. It seems we move the problem from `Mapping` to the "reference"

deuszx:
I don't think it's similar to a `Vec`. Access cost for elements of such natively-supported vector would be `O(1)`, while putting it into `Lazy` would load and deserialize the whole thing, right?

A wrapper like `struct<T> StorageVec(Mapping<u32,T>)` would be closer.

xgreenx:
[...]

PR #1740, opened 2023-04-05
Make `set_contract_storage` fallible

xermicus:
Related issue #1682

This is a very naive approach to the problem: If `set_contract_storage` can fail because the data does no longer fit into the encoding buffer, contract authors can account for that possibility. I don't think we need this for decoding because I don't see how something would fit in the buffer while encoding but no longer while decoding it.

A less invasive way would be to just add a `try_set` method to `Lazy`, but I thought this might be useful to have generally available. We also could introduce `try_set_contract_storage` into the API instead so we don't break the existing API. WDYT? Was there was an explicit reason `set_contract_storage` does not return a `Result`?

This is an early WIP. Todo:
- [ ] Don't use `Error::Unknown`, might need a new one
- [ ] Adapt integration tests
- [ ] In the dispatch it just unwraps the result, however now that we have a `Result` we could as well do something about it other than panic
- [ ] Implement tests
- [ ] Probably many more

### DIFF
--- a/crates/storage/traits/src/storage.rs
+++ b/crates/storage/traits/src/storage.rs
@@ -20,6 +20,8 @@ use ink_primitives::Key;
 /// [`scale::Codec`] are storable by default and transferable between contracts.
 /// But not each storable type is transferable.
 pub trait Storable: Sized {
+    fn size_hint(&self) -> usize;
+
     /// Convert self to a slice and append it to the destination.
     fn encode<T: scale::Output + ?Sized>(&self, dest: &mut T);
 
@@ -33,6 +35,11 @@ impl<P> Storable for P
 where
     P: scale::Codec,
 {
+    #[inline]
+    fn size_hint(&self) -> usize {
+        scale::Encode::size_hint(&self)
+    }
+
     #[inline]
     fn encode<T: scale::Output + ?Sized>(&self, dest: &mut T) {
         scale::Encode::encode_to(self, dest)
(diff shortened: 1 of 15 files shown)


xermicus:
@ascjones @SkymanOne WDYT?

SkymanOne:
> I don't think we need this for decoding because I don't see how something would fit in the buffer while encoding but no longer while decoding it.

with #1869 it is now possible for a contract to encode data using a larger buffer size, and then replace its code hash with a smaller buffer size and then fail. This can also pop up during cross-contract execution. So we need the same check for decoding.

crates/storage/traits/src/storage.rs
```diff
@@ -20,6 +20,8 @@ use ink_primitives::Key;
 /// [`scale::Codec`] are storable by default and transferable between contracts.
 /// But not each storable type is transferable.
 pub trait Storable: Sized {
+    fn size_hint(&self) -> usize;
```
SkymanOne, line 23: `size_estimate` perhaps is a better name?

SkymanOne (commented):
Look like a good start I would personally add `try_set_contract_storage()` to avoid breaking changes and be as close to Rust interface (e.g. `from()` and `try_from()`) as possible. 

I would also like to hear more how this work would lead to the implementation of `StorageVec` as you mentioned it as a related issue.

xermicus:
Thanks @SkymanOne for your comments, good points. I'll pick it up again.

xermicus:
Closing in favor of #1910

PR #1741, opened 2023-04-05
Persist `Environment` in metadata

### LINKED ISSUE #1695: Persist `Environment` types and values in metadata
In [this section](https://github.com/paritytech/ink/blob/master/ARCHITECTURE.md#environment-trait) of our `ARCHITECTURE.md` we explain what the `Environment` trait in ink! is used for. The `DefaultEnvironment` which ink! uses is defined [here](https://github.com/paritytech/ink/blob/master/crates/env/src/types.rs#L169-L183).

Right now, we don't persist which `Environment` implementation the contract uses anywhere. This means someone can deploy a contract with types that are different from those a chain uses to that chain. In https://github.com/paritytech/ink/issues/1694 we will investigate the criticality of this issue. But whatever turns out there, we should have a feature that warns contract developers about this. It is a serious oversight.

## ToDo
- [x] Persist the `Environment` types and `MAX_EVENT_TOPICS` value to the metadata. Will have to be a new field `environment`.
- [x] Make sure the schema and formal definition under `metadata/` is up to date.
- [x] Create a follow-up PR to `ink-docs` to also update the documentation on metadata there.
- [x] Talk to @statictype about where to create a follow-up issue for the UI to validate that the chain's types match the environment denoted in the metadata. Either `polkadot-js/api` or both `polkadot-js/apps` and Contracts UI. I'm not sure where this validation should happen.
- [x] Create a follow-up issue for `cargo-contract` to validate that the chain's types match the environment denoted in the metadata.

SkymanOne:
This PR closes #1695 by adding a field `environment` to `metadata.json` which contains the info about the types of the contract's environment.

### DIFF
--- a/crates/env/Cargo.toml
+++ b/crates/env/Cargo.toml
@@ -16,7 +16,6 @@ categories = ["no-std", "embedded"]
 include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 
 [dependencies]
-ink_metadata = { version = "4.1.0", path = "../metadata", default-features = false, features = ["derive"], optional = true }
 ink_allocator = { version = "4.1.0", path = "../allocator", default-features = false }
 ink_storage_traits = { version = "4.1.0", path = "../storage/traits", default-features = false }
 ink_prelude = { version = "4.1.0", path = "../prelude", default-features = false }
@@ -56,7 +55,6 @@ ink = { path = "../ink" }
 [features]
 default = ["std"]
 std = [
-    "ink_metadata/std",
     "ink_allocator/std",
     "ink_prelude/std",
     "ink_primitives/std",
--- a/crates/metadata/src/specs.rs
+++ b/crates/metadata/src/specs.rs
@@ -1340,7 +1340,7 @@ where
     }
 }
 
-/// Describes a contract message.
+/// Describes a contract environment.
 #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(bound(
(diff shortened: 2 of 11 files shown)


crates/env/Cargo.toml
```diff
@@ -16,7 +16,6 @@ categories = ["no-std", "embedded"]
 include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE"]
 
 [dependencies]
-ink_metadata = { version = "4.1.0", path = "../metadata", default-features = false, features = ["derive"], optional = true }
```
SkymanOne, line 19: Removed this because it seems it is not used anywhere.

SkymanOne, line 19: Removed this because it seems it is not used anywhere.

crates/metadata/src/specs.rs
```diff
@@ -59,6 +62,7 @@ pub struct ContractSpec<F: Form = MetaForm> {
     docs: Vec<F::String>,
     /// The language specific error type.
     lang_error: TypeSpec<F>,
+    environment: EnvironmentSpec<F>,
```
cmichi, line 65: Comment plz.

crates/metadata/src/specs.rs
```diff
@@ -115,6 +121,10 @@ where
     pub fn lang_error(&self) -> &TypeSpec<F> {
         &self.lang_error
     }
+
+    pub fn environment(&self) -> &EnvironmentSpec<F> {
```
cmichi, line 125: Comment plz.

cmichi (approved):

PR #1742, opened 2023-04-05
The `call_builder` macro to provide a more developer-friendly way to work with `CallBuilder`

xgreenx:
## Overview

Implemented the `call_builder!` macro from the [proposal](https://github.com/paritytech/ink/issues/631#issuecomment-1475389406).

The macro return the `CallBuilder` type that supports hinting and highlighting in the IDE.
The usage is the same as for standard cross-contract calls, where the call is wrapped into the `call_builder` macro:

```rust
#[ink::trait_definition]
pub trait Increment {
    /// Increments the current value of the implementer by one (1).
    #[ink(message)]
    fn inc_by_delta(&mut self, delta_1: u64, delta_2: u32);

    /// Returns the current value of the implementer.
    #[ink(message)]
    fn get(&self) -> u64;
}

#[ink(message)]
fn get(&self) -> u64 {
    self.incrementer.get();
    call_builder!(self.incrementer.get()).invoke();
    call_builder!(::dyn_traits::Increment::get(&self.incrementer)).invoke();
    call_builder!(<_ as ::dyn_traits::Increment>::get(&self.incrementer))
        .invoke();
    call_builder!(Increment::get(&self.incrementer)).invoke();
    call_builder!(::dyn_traits::Increment::get(&self.incrementer)).invoke();

    call_builder!(self.incrementer().get()).invoke()
}

#[ink(message)]
fn inc_by_delta(&mut self, delta_1: u64, delta_2: u32) -> u64 {
    self.incrementer.inc_by_delta(delta_1, delta_2);
    call_builder!(self.incrementer.inc_by_delta(delta_1, delta_2)).invoke();
    call_builder!(::dyn_traits::Increment::inc_by_delta(&mut self.incrementer, delta_1, delta_2)).invoke();
    call_builder!(<_ as ::dyn_traits::Increment>::inc_by_delta(&mut self.incrementer, delta_1, delta_2))
        .invoke();
    call_builder!(Increment::inc_by_delta(&mut self.incrementer, delta_1, delta_2)).invoke();
    call_builder!(::dyn_traits::Increment::inc_by_delta(&mut self.incrementer, delta_1, delta_2)).invoke();
    let call_builder =
        call_builder!(self.incrementer.inc_by_delta(delta_1, delta_2));

    call_builder.invoke()
}
```

How it looks in the IDE:
image
image

## Follow-up

- In the same way, we can add the `create_builder!` macro for constructors in the follow-up PR. It would simplify instantiation because all functions will be highlighted. The macro would be much simpler because it always requires an explicit type name.
- Add `mesage_descriptor!` macro that returns the `MessageDescription`. Everyone would be able to get the method's selector(and other properties of the message) like in the solidity `message_descriptor!(self.balanec_of).selector()`.
- Remove the [`TraitCallBuilder`](https://github.com/paritytech/ink/blob/master/crates/ink/src/codegen/trait_def/call_builder.rs#L16) and replace its functionality with a new macro. Remove the `{method_ident}Output` associated type from the trait definition.


crates/env/src/call/call_builder.rs
```diff
@@ -777,3 +778,190 @@ where
...
+                call_builder.invoke()
+            };
+            unreachable!();
+        };
+        call_builder
+    }};
+}
+
+/// Returns a [`CallBuilder`] based on the message call signature.
+///
```
cmichi, line 914:
```suggestion
```

xgreenx:
Closed in favor of https://github.com/paritytech/ink/pull/1858

PR #1750, opened 2023-04-13
Update subxt version

pmikolajczyk41:
We bump `subxt` to `0.28.0` version. Main changes are due to new `scale_en/decode` crates used by `subxt` (check https://github.com/paritytech/subxt/pull/842).

### DIFF
--- a/crates/e2e/Cargo.toml
+++ b/crates/e2e/Cargo.toml
@@ -29,14 +29,16 @@ tokio = { version = "1.18.2", features = ["rt-multi-thread"] }
 log = { version = "0.4" }
 env_logger = { version = "0.10" }
 scale = { package = "parity-scale-codec", version = "3.4", default-features = false, features = ["derive"] }
-subxt = "0.27.0"
+scale-decode = { version = "0.5.0", default-features = false, features = ["derive"] }
+scale-encode = { version = "0.1.0", default-features = false, features = ["derive"] }
+subxt = "0.28.0"
 
 # Substrate
-pallet-contracts-primitives = "18.0.0"
-sp-core = "16.0.0"
-sp-keyring = "18.0.0"
-sp-runtime = "18.0.0"
-sp-weights = "14.0.0"
+pallet-contracts-primitives = "23.0.0"
+sp-core = "20.0.0"
+sp-keyring = "23.0.0"
+sp-runtime = "23.0.0"
+sp-weights = "19.0.0"
 
 [dev-dependencies]
 # Required for the doctest of `MessageBuilder::call`
--- a/crates/e2e/Cargo.toml
+++ b/crates/e2e/Cargo.toml
@@ -29,8 +29,6 @@ tokio = { version = "1.18.2", features = ["rt-multi-thread"] }
 log = { version = "0.4" }
 env_logger = { version = "0.10" }
 scale = { package = "parity-scale-codec", version = "3.4", default-features = false, features = ["derive"] }
-scale-decode = { version = "0.5.0", default-features = false, features = ["derive"] }
-scale-encode = { version = "0.1.0", default-features = false, features = ["derive"] }
 subxt = "0.28.0"
 
 # Substrate
--- a/crates/e2e/src/client.rs
+++ b/crates/e2e/src/client.rs
@@ -43,10 +43,14 @@ use subxt::{
     blocks::ExtrinsicEvents,
     config::ExtrinsicParams,
     events::EventDetails,
-    ext::scale_value::{
-        Composite,
-        Value,
-        ValueDef,
+    ext::{
+        scale_decode,
+        scale_encode,
+        scale_value::{
+            Composite,
+            Value,
+            ValueDef,
+        },
     },
     tx::PairSigner,
 };
@@ -323,7 +327,8 @@ where
     scale_decode::DecodeAsType,
     scale_encode::EncodeAsType,
 )]
-#[decode_as_type(trait_bounds = "")]
+#[decode_as_type(trait_bounds = "", crate_path = "subxt::ext::scale_decode")]
+#[encode_as_type(crate_path = "subxt::ext::scale_encode")]
 struct ContractInstantiatedEvent<E: Environment> {
     /// Account id of the deployer.
     pub deployer: E::AccountId,
@@ -347,7 +352,8 @@ where
     scale_decode::DecodeAsType,
     scale_encode::EncodeAsType,
 )]
-#[decode_as_type(trait_bounds = "")]
+#[decode_as_type(trait_bounds = "", crate_path = "subxt::ext::scale_decode")]
+#[encode_as_type(crate_path = "subxt::ext::scale_encode")]
 struct CodeStoredEvent<E: Environment> {
     /// Hash under which the contract code was stored.
     pub code_hash: E::Hash,
--- a/crates/e2e/src/client.rs
+++ b/crates/e2e/src/client.rs
@@ -279,6 +279,8 @@ where
     CallExtrinsic(subxt::error::DispatchError),
     /// Error fetching account balance.
     Balance(String),
+    /// Decoding failed.
+    Decoding(subxt::Error),
 }
 
 // We implement a custom `Debug` here, as to avoid requiring the trait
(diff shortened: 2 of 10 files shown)


crates/e2e/src/client.rs
```diff
@@ -307,12 +310,20 @@ where
...
 /// A contract was successfully instantiated.
-#[derive(Debug, scale::Decode, scale::Encode)]
+#[derive(
+    Debug,
+    scale::Decode,
+    scale::Encode,
+    scale_decode::DecodeAsType,
+    scale_encode::EncodeAsType,
+)]
+#[decode_as_type(trait_bounds = "")]
```
SkymanOne, line 326: Why do you need this if trait bounds are empty?

pmikolajczyk41: because automatically compiler will require bounds for `E` (in general, for every generic parameter)

crates/e2e/src/client.rs
```diff
@@ -31,6 +31,7 @@ use ink_env::Environment;
 use ink_primitives::MessageResult;
 use pallet_contracts_primitives::ExecReturnValue;
 use sp_core::Pair;
+#[cfg(feature = "std")]
```
SkymanOne, line 34: You don't really need this since client is always compiled with `std` feature

crates/e2e/Cargo.toml
```diff
@@ -29,14 +29,16 @@ tokio = { version = "1.18.2", features = ["rt-multi-thread"] }
 log = { version = "0.4" }
 env_logger = { version = "0.10" }
 scale = { package = "parity-scale-codec", version = "3.4", default-features = false, features = ["derive"] }
-subxt = "0.27.0"
+scale-decode = { version = "0.5.0", default-features = false, features = ["derive"] }
+scale-encode = { version = "0.1.0", default-features = false, features = ["derive"] }
```
ascjones, line 33: Instead of importing these, you can use e.g. `subxt::ext::scale_encode`

crates/e2e/src/xts.rs
```diff
@@ -292,12 +304,12 @@ where
...
     ) -> ExtrinsicEvents<C> {
-        let call = subxt::tx::StaticTxPayload::new(
+        let call = subxt::tx::Payload::new_static(
             "Contracts",
             "instantiate_with_code",
-            InstantiateWithCode::<E::Balance> {
+            InstantiateWithCode::<E> {
                 value,
-                gas_limit,
+                gas_limit: subxt::utils::Static(gas_limit),
```
ascjones, line 312:
```suggestion
                gas_limit: gas_limit.into(),
```

crates/e2e/src/xts.rs
```diff
@@ -398,20 +410,20 @@ where
...
-        let call = subxt::tx::StaticTxPayload::new(
+        let call = subxt::tx::Payload::new_static(
             "Contracts",
             "call",
-            Call::<E, E::Balance> {
+            Call::<E> {
                 dest: contract,
                 value,
-                gas_limit,
+                gas_limit: subxt::utils::Static(gas_limit),
```
ascjones, line 426:
```suggestion
                gas_limit: gas_limit.into(),
```

crates/env/Cargo.toml
```diff
@@ -48,6 +48,8 @@ secp256k1 = { version = "0.27.0", features = ["recovery", "global-context"], opt
 #
 # Sadly couldn't be marked as dev-dependency.
 # Never use this crate outside the off-chain environment!
+scale-decode = { version = "0.5.0", default-features = false, optional = true }
+scale-encode = { version = "0.1.0", default-features = false, optional = true }
```
ascjones, line 52: Same here, don't need to import these, just use `subxt::ext` reexports

crates/e2e/src/client.rs
ascjones: Perhaps to avoid having this attribute everywhere we could have a blanket impl of `DecodeAsType` for the `Environment` trait?

crates/e2e/src/xts.rs
```diff
@@ -398,20 +410,20 @@ where
...
-        contract: sp_runtime::MultiAddress<E::AccountId, ()>,
+        contract: MultiAddress<E::AccountId, ()>,
         value: E::Balance,
         gas_limit: Weight,
         storage_deposit_limit: Option<E::Balance>,
         data: Vec<u8>,
         signer: &Signer<C>,
     ) -> ExtrinsicEvents<C> {
-        let call = subxt::tx::StaticTxPayload::new(
+        let call = subxt::tx::Payload::new_static(
```
jsdw, line 420: You could also use `Payload::new()` to use the documented interface (though downside is it would allocate since it doesn't specialise on static strings)

crates/e2e/src/xts.rs
```diff
@@ -33,43 +33,54 @@ use subxt::{
...
+#[derive(Debug, scale::Decode, scale::Encode, scale_encode::EncodeAsType)]
+#[encode_as_type(trait_bounds = "")]
+pub struct Call<E: Environment> {
+    dest: MultiAddress<E::AccountId, ()>,
     #[codec(compact)]
-    value: B,
-    gas_limit: Weight,
-    storage_deposit_limit: Option<B>,
+    value: E::Balance,
+    gas_limit: subxt::utils::Static<Weight>,
```
jsdw, line 60: `Weight` is a super simple struct, so for a little added robustness/simplicity I wonder whether it's worth just writing your own version that impls EncodeAsType et al

crates/env/Cargo.toml
pmikolajczyk41: but we don't have here any access to `subxt`, do we?

crates/e2e/src/xts.rs
pmikolajczyk41: :heavy_check_mark:

crates/e2e/src/xts.rs
pmikolajczyk41: :heavy_check_mark:

crates/e2e/src/client.rs
pmikolajczyk41: a leftover - removed

crates/e2e/src/client.rs
pmikolajczyk41: we could, but then I would have to provide something reasonable here:
```rust
impl DecodeAsType for E {
    fn decode_as_type(input: &mut &[u8], type_id: u32, types: &PortableRegistry) -> Result<Self, Error> {
        todo!()
    }
}

impl EncodeAsType for E {
    fn encode_as_type_to(&self, type_id: u32, types: &PortableRegistry, out: &mut Vec<u8>) -> Result<(), scale_encode::Error> {
        todo!()
    }
}
```

@jsdw what makes sense here?

jsdw: Assuming `Environment` is a type that can't be constructed (eg `enum Environment {}`) I think some like that would be fine (though I'd `panic!("Shouldn't call x on Environment")` on the `DecodeAsType` and maybe `unreachable!()` on the `EncodeAsType` one even so).

If it's at all possible to construct an `Environment` then personally I'd go for the trait bounds everywhere.

(Side note: I'm noting that `Encode` and `Decode` don't require such trait bounds; I wonder how that is possible because maybe we can be smarter with `EncodeAsType` and `DecodeAsType` in this regard also)

crates/e2e/src/xts.rs
ascjones: I'm fine with a local copy of the type with a minimal API, it is likely to be highly stable.

crates/env/src/types.rs
```diff
@@ -106,6 +106,15 @@ pub trait AccountIdGuard {}
...
 impl AccountIdGuard for AccountId {}
 
+#[cfg(feature = "std")]
+pub trait CodecAsType: scale_decode::DecodeAsType + scale_encode::EncodeAsType {}
+#[cfg(feature = "std")]
+impl<T: scale_decode::DecodeAsType + scale_encode::EncodeAsType> CodecAsType for T {}
+#[cfg(not(feature = "std"))]
+pub trait CodecAsType {}
+#[cfg(not(feature = "std"))]
+impl<T> CodecAsType for T {}
```
ascjones, line 116: This crate already uses `cfg_if!` so would be good to use that here.

crates/e2e/src/xts.rs
pmikolajczyk41: done

crates/env/src/types.rs
pmikolajczyk41: sure, I missed that - refactored

ascjones (approved):
LGTM!

crates/e2e/src/client.rs
pmikolajczyk41: hmm... it's a trait and it feels kinda weird to have these panics in the blanket implementation :| I don't have strong opinion; @ascjones (?)

ascjones: Not a good idea. Just leave as is with the `#[decode_as_type(trait_bounds = "")]` for now

jsdw (approved):
Nice!

PR #1752, opened 2023-04-17
delegator build_call example

Daanvdplas:
I wrote a different implementation of the delegator example. In stead of instantiating all the contracts when instantiating the delegator contract, the delegator contract is instantiated with already instantiated contracts of `adder`, `subber` and `accumulator`.

Why I think it could be useful:
* It shows a simple but extended example with multiple contracts interacting with each other through `build_call`.
* Reference material to compare with the other delegator (now called `multi_contract_caller` example).
* Supported by e2e-tests.

Please let me know if it needs any improvements (e.g. the name 😅)!

### DIFF
--- a/integration-tests/delegator/.gitignore
+++ b/integration-tests/delegator/.gitignore
@@ -1,0 +1,9 @@
+# Ignore build artifacts from the local tests sub-crate.
+/target/
+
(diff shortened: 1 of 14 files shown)


SkymanOne:
Closing this PR as #1826 and #1825 need to be addressed first

PR #1756, opened 2023-04-19
Added PR report cron job

Bullrich:
This PR opens a cron job that uses the [paritytech/stale-pr-finder](https://github.com/paritytech/stale-pr-finder) action.

It will run every Tuesday at 7AM UTC (9 AM Germany)

You can also manually run this action by going to https://github.com/paritytech/ink/actions/workflows/pr-notifier.yml

It will fetch PRs from the repositories:
- ink
- cargo-contract
- substrate-contracts-node
- contracts-ui
- ink-docs
- smart-bench
- ink-waterfall
- ink-playground
- nft-marketplace-demo
- pallet-contracts-xcm
- link
- ink-examples

It will only fetch PRs that don't have any reviews (comments don't count).

It will organize them by oldest, and fetch only the first 8 issues and will format them with the format:
`- [(title)]((url)) | (daysStale) days with no review`

In the case that there a no PRs found, it will post the message:
> We have no Pull Requests waiting for reviews.
>
> **Good job everyone!**

I have added a lot of comments to the `yml` file so it is easy to configure if you need to add an extra repository

---

I also created a python script to combine and produce the output because I was having problems with `jq`.

First of all, the `jq` command was becoming way to long and impossible to understand, and second, the length of the array is so long that GitHub fails to use it as a variable, so by having a file that handles all of it in it, we overcome that issue.

And I belive it's more readable than a `jq` bash command:
```bash
jq -r '. | sort_by(.daysStale) | reverse[:8]| to_entries | .[] | "\(.key). [\(.value.title)](\(.value.url)) | \(.value.daysStale) days with no reviews"'
```

So, everything the python script does is:
1. combines all of the jsons in `outputs/*.json`
2. orders all the outputs based on the parameter `daysStale`
3. reverses the list
4. counts the amount of elements and generates an output
5. limits the list to the given number (15 in this PR)
6. formats the strings with the given template (the for loop above)
7. pushes the result as a summary and an output for another github action to use

---

I also limited stale-issues-finder to work with a top of 3 parallel jobs.

This will help with the case where too many jobs are running at the same time and timing out the GitHub API

### DIFF
--- a/.github/scripts/combine-pr.py
+++ b/.github/scripts/combine-pr.py
@@ -1,0 +1,59 @@
+import uuid
+import json
+import glob
+import sys
+import os
+
+
+def set_output(name, value):
+    with open(os.environ['GITHUB_OUTPUT'], 'a') as fh:
+        print(f'{name}={value}', file=fh)
+
+
+def set_multiline_output(name, value):
+    with open(os.environ['GITHUB_OUTPUT'], 'a') as fh:
+        delimiter = uuid.uuid1()
+        print(f'{name}<<{delimiter}', file=fh)
+        print(value, file=fh)
+        print(delimiter, file=fh)
+
+
+if len(sys.argv) > 1:
+    limit = int(sys.argv[1])
+else:
+    limit = 999
+
+result = []
+for f in glob.glob("outputs/*.json"):
+    with open(f, "r") as infile:
+        result += json.load(infile)
+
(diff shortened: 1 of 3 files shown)


Bullrich:
Hi @cmichi!

This is how the message would look like:
image

If you are satisfied with it, please let me know and I'll remove the test step so it is ready to merge.

cmichi (approved):
Thank you, Javier!

PR #1760, opened 2023-04-20
fixed cron job time

Bullrich:
It was set for Thursday, now I changed it for Tuesday

You can check it out here: https://crontab.guru/#0_7_*_*_2

### DIFF
--- a/.github/scripts/combine-pr.py
+++ b/.github/scripts/combine-pr.py
@@ -4,13 +4,24 @@ import glob
 import sys
 import os
 
+# This script does the following:
+# 1. combines all of the jsons in `outputs/*.json`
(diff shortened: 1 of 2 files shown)


alvicsam:
Since it's related to https://github.com/paritytech/ink/pull/1756: could you add a brief description to https://github.com/paritytech/ink/blob/master/.github/scripts/combine-pr.py?

Bullrich:
> Since it's related to #1756: could you add a brief description to https://github.com/paritytech/ink/blob/master/.github/scripts/combine-pr.py?

I documented the code. Would that be enough?

alvicsam:
Awesome, thanks!

PR #1762, opened 2023-04-20
test: integration test storage-types

peetzweg:
The origin of this PR is the need in the `contracts-ui` repo to have a contract to test the proper decoding and display of return values of an ink smart contract.

We currently have integration test using the `mother` contract, however, it's missing some possible types and is relying on inputs for some messages. Testing encoding and sending inputs to an ink! message is should be tested separately.

The original development of this contract happened in a pr for the ink-examples repo before I realised it was a mirror of `ink/intergration-tests` folder.

Related `ink-examples` PR: https://github.com/paritytech/ink-examples/pull/15

Related `contracts-ui` PR: https://github.com/paritytech/contracts-ui/pull/470

cc @statictype


integration-tests/storage-types/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        #[ink(message)]
+        pub fn get_primitive_types(&self) -> PrimitiveTypes {
+            self.primitive_types.clone()
+        }
+
+        #[ink(message)]
+        pub fn get_option_some(&self) -> Option<()> {
+            Some(())
```
statictype, line 236: we need to return something inside `Some` in order to test its output

integration-tests/storage-types/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+        }
+
+        #[ink(message)]
+        pub fn get_option_none(&self) -> Option<()> {
+            None
+        }
+
+        #[ink(message)]
+        pub fn get_result_ok(&self) -> Result<(), CustomError> {
+            Ok(())
```
statictype, line 246: same for `Ok`, it should return something otherwise we get `null` in JS

statictype:
>  Testing encoding and sending inputs to an ink! message is should be tested separately.

the approach of the Mother contract where it returns the input values was ok imo. it allows us to test the encoding of user inputs as well. why do you think it's a problem if we test them together?

statictype:
we could replace Mother with this. wdyt @agryaznov ?

cmichi (approved):

PR #1766, opened 2023-04-27
Add schema formal definition

SkymanOne:
Adds 2 updated schemas that can be used to verify ink! metadata


ascjones (approved):
Assume you have tested out verifying valid and invalid metadata files with these?

SkymanOne:
> Assume you have tested out verifying valid and invalid metadata files with these?

Yep, the example is here https://www.jsonschemavalidator.net/s/2y2Wr4BW

Issue #1644, opened 2023-02-07
Improve E2E interface


SkymanOne:
A thing that would be nice is to simplify the e2e API that at the moment requires sth like this:
```rust
.call(|flipper| flipper.get());

**Comment by ascjones** [comment]
There should be some way to invoke this as an associated function e.g. `FlipperRef::get`. Decouple the call from the account id.

May not be entirely straightforward since `get` is already an inherent method so need to figure out a way around that. Could e.g. generate an entirely new struct alongside `FlipperRef` called e.g. `FlipperCall`, or use different method names.

**Comment by SkymanOne** [comment]
> May not be entirely straightforward since get is already an inherent method so need to figure out a way around that. Could e.g. generate an entirely new struct alongside FlipperRef called e.g. FlipperCall, or use different method names.
Not a fun of using different method names since it harms readability of the tests.
Do I understand correctly that we want to move the message call outside the `.call()` function?

**Comment by SkymanOne** [comment]
# Proposed design
we keep the contract instantiation. Let's take `mapping_integration_tests` as an example:
```rust
let constructor = MappingsRef::new();
 let contract_id = client
    .instantiate(
        "mapping-integration-tests",
        &ink_e2e::charlie(),
        constructor,
        0,
        None,
    )
    .await
    .expect("instantiate failed")
    .account_id;
```
Instead of manually building a message:
```rust
let first_insert = ink_e2e::build_message::<MappingsRef>(contract_id)
    .call(|contract| contract.insert_balance(1_000));
```
and making an explicit call:
```rust
let _ = client
    .call(&ink_e2e::charlie(), first_insert, 0, None)
    .await
    .expect("Calling `insert_balance` failed")
    .return_value();
```

Let's unify the process under the automatically-generated associated message function. i.e we have `mapping_integration_tests` contract. When instantiating the contract as in the first snippet, we generate an associated trait `MappingsRefCal` as proposed by @ascjones which have the same message functions as a contract and incorporates the call dispatch logic.

So we can do something like:
```rust
// get the contracts' constructor
let constructor = MappingsRef::new();

// generate the associated trait for E2E
 let contract = client
    .instantiate(
        "mapping-integration-tests",
        &ink_e2e::charlie(),
        constructor,
        0,
        None,
    )
    .await
    .expect("instantiate failed")
    .account_id;

// we call a contract await the result from `LangError` and return the value automatically
let _ = contract.insert_balance(1_000)
    .call()
    .await
[...]

PR #1782, opened 2023-05-18
`[ink_e2e]` improve call API, remove `build_message` + callback

ascjones:
Resurrecting the idea from https://github.com/paritytech/ink/pull/1669. 

I believe it closes https://github.com/paritytech/ink/issues/1644, since the main thing is removing that ugly callback. This may not be the fina or most sophisticatedl API, but an increment in the right direction for better quality of life for `e2e` test devs.

Replaces `build_message` and callback with a simpler:

```rust
            let erc20 = client
                .instantiate("erc20", &ink_e2e::alice(), constructor, 0, None)
                .await
            let mut call = erc20.call::<Erc20>();

            // dry run
            let total_supply_msg = call.total_supply();
            let total_supply_res = client
                .call_dry_run(&ink_e2e::bob(), &total_supply_msg, 0, None)
                .await;

            // tx
            let transfer_from =
                call.transfer_from(bob_account.clone(), charlie_account.clone(), amount);
            let transfer_from_result = client
                .call(&ink_e2e::charlie(), &transfer_from, 0, None)
                .await;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Schema generation - [#1765](https://github.com/paritytech/ink/pull/1765)
 
+### Changed
+- E2E: improve call API, remove `build_message` + callback - [#1782](https://github.com/paritytech/ink/pull/1782)
+
 ## Version 4.2.0
 
 ### Added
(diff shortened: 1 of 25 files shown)


cmichi (approved):
Thanks for doing this! 

We should reflect the change in a `CHANGELOG.md`. Since we so far always released in lockstep with the rest of ink!, I would just add it there for now.

Issue #1770, opened 2023-05-04
Storage types not output in ABI JSON


forgetso:
Previously (ink version 3.x), if you had storage in your contract like `Mapping<AccountId, MyStruct>` you would get a type like the below in the ABI JSON, which would tell you how to construct the storage type in the client. The below is a `Mapping<Type1, Type4>`.

```JSON
    "types": [
      {
        "id": 0,
        "type": {
          "def": {
            "composite": {
              "fields": [
                {
                  "name": "offset_key",
                  "type": 10,
                  "typeName": "Key"
                }
              ]
            }
          },
          "params": [
            {
              "name": "K",
              "type": 1
            },
            {
              "name": "V",
              "type": 4
            }
          ],
          "path": [
            "ink_storage",
            "lazy",
            "mapping",
            "Mapping"
          ]
        }
      },
```

These types are not present in the ABI generated by version 4.2 of ink!. For example, I have storage like

```rust
        operator_votes: Mapping<AccountId, Vote>,


    pub enum Vote {
        SetCodeHash([u8; 32]),        // accepts the code hash
        Withdraw(AccountId, Balance), // accepts the recipient and the amount
        Terminate(AccountId), // accepts the account to send the remaining balance of this contract to after termination
    }
```
`operator_votes` appears in the `storage` section of the ABI and is shown as an enum but there is no mention of the `Mapping<AccountId` part. How am I meant to infer this from the [ABI](https://github.com/prosopo/captcha/blob/captcha_24_fix_existing_get_storage_functions/packages/contract/src/abi/prosopo.json)?

```JSON
            {
              "layout": {
                "root": {
                  "layout": {
                    "enum": {
                      "dispatchKey": "0x0d1f1c57",
                      "name": "Vote",
                      "variants": {
                        "0": {
                          "fields": [
                            {
                              "layout": {
                                "array": {
                                  "layout": {
                                    "leaf": {
                                      "key": "0x0d1f1c57",
                                      "ty": 3
                                    }
                                  },
                                  "len": 32,
                                  "offset": "0x0d1f1c57"
                                }
                              },
                              "name": "0"
                            }
                          ],
                          "name": "SetCodeHash"
                        },
                        "1": {
                          "fields": [
                            {
                              "layout": {
                                "leaf": {
                                  "key": "0x0d1f1c57",
                                  "ty": 7
                                }
                              },
                              "name": "0"
                            },
                            {
                              "layout": {
                                "leaf": {
[...]

PR #1784, opened 2023-05-19
Expose `ty` for types that use `RootLayout`

xgreenx:
Fixes https://github.com/paritytech/ink/issues/1770

image
image

### DIFF
--- a/crates/metadata/src/layout/mod.rs
+++ b/crates/metadata/src/layout/mod.rs
@@ -37,7 +37,6 @@ use scale_info::{
     },
     meta_type,
     IntoPortable,
-    MetaType,
     Registry,
     TypeInfo,
 };
--- a/crates/metadata/src/layout/mod.rs
+++ b/crates/metadata/src/layout/mod.rs
@@ -37,6 +37,7 @@ use scale_info::{
     },
     meta_type,
     IntoPortable,
+    MetaType,
     Registry,
     TypeInfo,
 };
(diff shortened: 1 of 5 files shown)


crates/metadata/src/layout/mod.rs
```diff
@@ -146,25 +149,37 @@ impl IntoPortable for RootLayout {
...
             layout: Box::new(self.layout.into_portable(registry)),
+            ty: registry.register_type(&self.ty),
         }
     }
 }
 
 impl<F> RootLayout<F>
 where
     F: Form,
+    F::Type: From<MetaType>,
```
ascjones, line 160: With this constraint the impl is no longer generic over `Form`. If we wanted to use the rest of the methods e.g. on a decoded `RootLayout<PortableForm>` then the methods will not be available.

The way we separate the `MetaForm` constructors in other places is to have a separate non generic `MetaForm` impl e.g.

```rust
impl RootLayout<MetaForm> {
    /// Creates a new root layout.
    pub fn new<Root, L>(root_key: LayoutKey, layout: L) -> Self
    where
        Root: TypeInfo + 'static,
        L: Into<Layout<MetaForm>>,
    {
        Self {
            root_key,
            layout: Box::new(layout.into()),
            ty: meta_type::<Root>(),
        }
    }

    /// Creates a new root layout with empty root type.
    pub fn new_empty<L>(root_key: LayoutKey, layout: L) -> Self
    where
        L: Into<Layout<MetaForm>>,
    {
        Self::new::<(), L>(root_key, layout)
    }
}
```

xgreenx: You are right, my bad=D

ascjones (approved):
LGTM

Issue #1785, opened 2023-05-22
Fixed size array support


goastler:
**Questions**

How to include a fixed size array of more than 32 elements in an ink smart contract?

**FAQ**

- Are you using the nightly Rust channel to compile your smart contract?
- Are you using an up-to-date `cargo-contract`? Check via `cargo contract --version`.
    - If you are unsure please tell us your version noted there.
- Are you using an up-to-date ink! version in your smart contract?
    - You can check this by looking into the `[dependencies]` section of your `Cargo.toml`.
    - If you are unsure please tell us your version noted there.

**Describe the bug**
A clear and concise description of what the bug or odd behavior is.

**Expected behavior**

**Screenshots**

**Additional context**

Ink should have support for byte arrays of any length. Currently, out-of-the-box support is arrays of 1-32 elements long. We now have a situation where we need a byte array of length 64. This is unavailable in ink and forces me to implement the `StorageLayout` myself.

Except I can't.

To provide an implementation for `StorageLayout`, I need to import it. I cannot import it into the smart contract package because `ink_storage_traits` has `StorageLayout` marked as `std`-only, (see here)[https://github.com/paritytech/ink/blob/6e87ad042ee9ca9bd4e2e7e2632410ea6b62ab0a/crates/storage/traits/src/lib.rs#L42]. I can't have the `std` feature enabled without it leading to duplicate `ink` libraries during compilation and thus an error.

Further, to implement `StorageLayout` for a 64-length byte array as has been done in `ink_storage_traits`, I would need access to `Layout`, `ArrayLayout` and `LayoutKey`, all of which belong to `ink_metadata`. Again, this cannot be used without `std` enabled leading to duplicate `ink` during compilation.

I'm essentially attempting to replicate this (macro)[https://github.com/paritytech/ink/blob/6e87ad042ee9ca9bd4e2e7e2632410ea6b62ab0a/crates/storage/traits/src/layout/impls.rs#L65] directly instead of as a macro.

To further complicate the issue, I can't implement a `StorageLayout` because I do not own the trait nor the struct, the `[u8; 64]`, which I'm trying to implement for, which is not allowed in rust. Therefore, I have to wrap the array in my own struct like so:
```rust

use ink_primitives::{
    Key,
};
use ink_metadata::{
    ArrayLayout, // doesn't work without std
    Layout, // doesn't work without std
    LayoutKey, // doesn't work without std
};
use ink_storage_traits::{
    StorageLayout, // doesn't work without std
[...]

PR #1787, opened 2023-05-23
Implement `StorageLayout` for any length array

ascjones:
`StorageLayout` impl is currently restricted to a arrays up to 32 in length. Not sure what the original reasoning was for this @xgreenx?

Anyway this PR implements `StorageLayout` for any length array.

Closes https://github.com/paritytech/ink/issues/1785


xgreenx (approved):
It was implemented before the introduction of the const generics=) 
image

PR #1791, opened 2023-05-29
impl From<Hash> for [u8; 32]

kvinwang:
Sometimes, we need to convert the `ink::primitives::Hash` to `[u8; 32]` or to the runtime Hash type in the chain extension. I don't find an elegant way to do so. It should be better to make it able to be converted to `[u8; 32]`.

### DIFF
--- a/crates/primitives/src/types.rs
+++ b/crates/primitives/src/types.rs
@@ -118,6 +118,12 @@ impl AsMut<[u8]> for Hash {
     }
 }
 
+impl From<Hash> for [u8; 32] {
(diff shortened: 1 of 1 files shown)


kvinwang:
Unfortunately, this is a breaking change. What do you think would be the best way to handle this issue?

ascjones:
> impl From<Hash> for [u8; 32

Yes rather implement `impl From<Hash> for [u8; 32]`.

ascjones (approved):

PR #1795, opened 2023-06-01
Update `subxt`, `scale-*` and `sp-*` dependencies

ascjones:
Compatible with https://github.com/paritytech/substrate-contracts-node/releases/tag/v0.27.0

Also updates to latest `pallet-contract-primitives` release, and fixes compilation errors with new `EventRecord` type parameter on the result types.

Note for now we are using `()` as the `EventRecord` type, so the events decoding itself will be skipped, and the dry-run events will not be used. This is because at the moment we don't have a top level `Event` type available to us for decoding. For that we would need to use the `subxt` macro again to get the generated `Event` type which would couple us tightly to a specific version of `substrate-contracts-node`. While the dry-run events would be useful, we will defer this work for a future PR.

### DIFF
--- a/crates/e2e/Cargo.toml
+++ b/crates/e2e/Cargo.toml
@@ -21,7 +21,7 @@ ink_env = { version = "4.2.0", path = "../env" }
 ink_primitives = { version = "4.2.0", path = "../primitives" }
 
 funty = "2.0.0"
-impl-serde = { version = "0.3.1", default-features = false }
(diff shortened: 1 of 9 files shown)


ascjones:
Depends on new `substrate-contracts-node` release https://github.com/paritytech/substrate-contracts-node/pull/192

cmichi (approved):

Issue #1804, opened 2023-06-12
Consider using `decode_all` for `get_contract_storage`


ascjones:
Raised by @DamianStraszak.

At the moment [`get_contract_storage`](https://github.com/paritytech/ink/blob/5991f1418a8e596c575d40aa8188aa6090b8b519/crates/env/src/engine/on_chain/impls.rs#L239) uses `scale::Decode::decode`, which can succeed even if not all bytes are consumed. This would be an issue where corrupted or invalid bytes in a storage cell are successfully decoded, but the expectation is that this should fail.

We should investigate using [`decode_all`](https://github.com/paritytech/parity-scale-codec/blob/master/src/decode_all.rs) instead which will fail if there are bytes remaining in the input after decoding successfully.

Note that this may technically be a "breaking" change if existing contracts are (ab)using this behaviour, so will require some communication if we go ahead with it.

PR #1810, opened 2023-06-15
Use `decode_all` for decoding cross contract call result

ascjones:
Related to https://github.com/paritytech/ink/security/advisories/GHSA-853p-5678-hv8f.

`decode` can still succeed if there are bytes remaining. In the above security advisory the issue was that the bytes for `Result<T>` were successfully decoded into a `T` but with the incorrect value, because `Result` encoding has an extra byte prefix.

If we had been using `decode_all` instead, which fails if there are any bytes remaining in the input, this would have been discovered earlier.

This provides an extra level of safety since it avoids e.g. truncation of values: preventing e.g. an `i32` returned by the callee being decoded into an `i8` (see `integration-test` as part of this PR).

Note there are other uses of `decode` which could be replaced by `decode_all` (see https://github.com/paritytech/ink/issues/1804), but we can tackle those separately.

### DIFF
--- a/integration-tests/call-builder-return-value/lib.rs
+++ b/integration-tests/call-builder-return-value/lib.rs
@@ -1,18 +1,5 @@
-//! # Integration Tests for `LangError`
-//!
-//! This contract is used to ensure that the behavior around `LangError`s works as
-//! expected.
-//!
-//! In particular, it exercises the codepaths that stem from the usage of the
-//! [`CallBuilder`](`ink::env::call::CallBuilder`) and
-//! [`CreateBuilder`](`ink::env::call::CreateBuilder`) structs.
-//!
-//! This differs from the codepath used by external tooling, such as `cargo-contract` or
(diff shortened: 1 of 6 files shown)


integration-tests/call-builder-return-value/lib.rs
```diff
@@ -0,0 +1,293 @@
+//! # Integration Tests for `LangError`
```
SkymanOne, line 1: Why can't his example be incorporated into `call-builder-delegate` example?

SkymanOne (commented):
Looks good, but why can't we have the same generic integration test for the delegate call?

integration-tests/call-builder-return-value/lib.rs
HCastano: Different storage layouts are used in each

SkymanOne (approved):
LGTM!

PR #1813, opened 2023-06-16
Pin inter crate dependencies to specific version

ascjones:
Raised by @deuszx.

Currently dependencies between the `ink` crates are specified with a range i.e. `4.2.0` which means `>= 4.2.0, <5.0`.

This means that when we released `4.2.1` recently, and contracts with an `ink` dependency of `4.2.0` could pull in the new versions sub crates.

This PR pins the versions of the inter-crate dependencies of all the `ink` crates, since they are always released together.

### DIFF
--- a/crates/e2e/Cargo.toml
+++ b/crates/e2e/Cargo.toml
@@ -15,10 +15,10 @@ categories = ["no-std", "embedded"]
 include = ["/Cargo.toml", "src/**/*.rs", "/README.md", "/LICENSE"]
 
 [dependencies]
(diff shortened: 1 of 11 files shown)


deuszx (approved):
If they are _always_ released together - you can consider switching to cargo workspace dependencies features and specify it (version, authors, etc.) in root `Cargo.toml`.

xgreenx (approved):
It creates conflicts for https://github.com/paritytech/ink/pull/1835, but them is not hard to solve. I think we can start with pinning the version without a workspace, and when PR is ready switch to the workspace approach

PR #1815, opened 2023-06-16
[🦖 T-Rex]  Pin `parity-scale-codec-derive` to `3.1.4`

SkymanOne:
Temporary solution until https://github.com/paritytech/parity-scale-codec/issues/454 gets released

### DIFF
--- a/crates/ink/src/env_access.rs
+++ b/crates/ink/src/env_access.rs
@@ -616,8 +616,8 @@ where
     ///         .params();
     ///     self.env()
     ///         .invoke_contract_delegate(&call_params)
-    ///         .unwrap_or_else(|env_err| {
-    ///             panic!("Received an error from the Environment: {:?}", env_err)
+    ///         .unwrap_or_else(|env_error| {
+    ///             panic!("Received an error from the Environment: {:?}", env_error)
     ///         })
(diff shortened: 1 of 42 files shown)


ascjones (commented):
If we are doing it this way we should also pin `parity-scale-codec` to `3.5`.

Also you may have to update all examples.

If the fix is incoming it might be better to disable those clippy warnings, then we don't have the noise changing all the manifests.

ascjones:
> Pinning parity-scale-codec has no effect

True but it is making explicit what is already happening implicitly, i.e. the resolver auto resolves to `3.5`, and it is a reexport of it after all.

> I am against disabling these warnings because they are quite significant to be silenced meaning that these "slips" will not be picked up in future PRs.

They are both just warnings/suggestions to fix redundant code, nothing serious.

ggwpez:
Just FYI: It should only take a day to fix+release SCALE.

ascjones:
> Just FYI: It should only take a day to fix+release SCALE.

:+1:  I reckon just wait for the release then to prevent the churn

SkymanOne:
Okay. In this case I'm closing this PR. The stuff should be automatically resolved once the fix is published.

koute:
Just published 3.6.1 which should fix this (I think). If anything new pops up feel free to ping me and I'll fix it.

PR #1827, opened 2023-06-26
Events `2.0`

### LINKED ISSUE #809: Shared external event definitions
Due to ink!'s design it is not possible to share event definitions between multiple contracts since events can only be defined in the ink! module scope directly.
In order to more extensively use the upcoming trait features we want to make it possible for users to define events to be used in multiple contracts.

For this we will need a new attribute macro `#[ink::event_definition]` or just a derive macro `#[derive(ink::Event)]` that will setup everything required for ink! events. There will be some initial limitations. E.g. events can only be `enum` or `struct` with certain properties.
However, this will allow to define events in some library and use the same event definition in many ink! smart contract or trait definitions.

ascjones:
Rework of event definitions: 
 - decouples events from the `#[ink::contract]` macro and changes the topic calculation, allowing events to be shared between contracts.
 - changes how topics are calculated.
 
Closes #809 and supersedes https://github.com/paritytech/ink/pull/1243.

## New macros
Events can now be defined independently of contracts using the new `event` attribute macro:

```rust
#[ink::event]
pub struct Flipped {
    pub value: bool,
}
```
This event can now be imported and emitted by any contract e.g.
```rust
self.env().emit_event(event_def::Flipped { value: self.value })
```
### Derives
The `ink::event` attribute macro simply expands to underlying `derive` macros for the following traits:
- `Event` (formerly `Topics`): captures the topics of the event when it is emitted at runtime.
- `EventMetadata`: returns the metadata for the event definition.

e.g.

```rust
#[ink::event]
pub struct Flipped { .. }
```
expands to
```rust
#[derive(::ink::Event, ::scale::Encode, ::scale::Decode)]
#[cfg_attr(feature = "std", derive(::ink::EventMetadata))]
pub struct Flipped { .. }
```

### Backwards compatibility
The legacy syntax of "inline" event definitions using the `#[ink(event)]` syntax will continue to work, no code changes in existing contracts are required to update to the new syntax. 

## Topics changes

How topics are calculated has changed, so this is a **breaking change** for any client code which uses topics to filter events.

### Signature topic
Currently, the signature topic of an event was calculated by `<contract_name>::<event_name>` e.g. `"Erc20::Transfer"`, which in turn was hashed if encoded len > 32 bytes.

Now, we mimic the approach of Solidity events of hashing the name of the event and the concatenated field types e.g. given 

```rust
    #[ink::event]
    pub struct Transfer {
        #[ink(topic)]
        from: Option<AccountId>,
        #[ink(topic)]
        to: Option<AccountId>,
        value: Balance,
    }
```
Produces the signature of:

`Transfer(Option<AccountId>, Option<AccountId>, Balance)`

Which is then hashed with `blake2x256` to produce the signature topic. This is calculated by the macro at compile time, and can be accessed statically with `<Transfer as Event>::SIGNATURE_TOPIC`. If the event is annotated with `#[ink(anonymous)]` attribute then the value of this will be `None` and no signature topic is published.

#### Field type name variation

In Rust, the same concrete type can be represented in different ways via type aliases and associated types e.g. `u32, MyAlias, T::AssocType` can all represent the same concrete type but with different names in the source code. This can result in two events with the same name and the same field types having different signature topics. This issue is discussed further with @xgreenx https://github.com/paritytech/ink/pull/1243#discussion_r1045300050.

In the end I decided that it is an acceptable compromise that two binary compatible events would have different signature topics. It will be important to document this fact, where e.g. for block explorers which want to index on common events such as `Transfer` that there is a canonical definition of the event which must have matching field type names.

## Field topics

Previously field topics were calculated by concatenating the path to the field e.g. `Erc20::Transfer::from` with the value of the field, then hashing it (if > 32 bytes).

However with this encoding topic values are often opaque, i.e. from a topic value on it's own there is no way to know the actual value of the field in the topic because it has been hashed. Even if it is < 32 bytes and not been hashed, the path of the field prefixed the value, so the value itself can not easily be separated.

Specifically it is a problem for topics with the `AccountId` type, which is often indexed. Indexers such as block explorers might want a view which shows all events affecting a given `AccountId`. With the current topics encoding, this would be difficult, requiring the metadata for all known events and having to precalculate hashes for topics with known account ids.

With the new topics, the value of the event field itself (if <= 32 bytes) is the topic. In the case of 32 byte account ids this allows subscribing to a topic for a given account id and then it is guaranteed to receive **all** events which relate to that account.

It is still possible to filter those events which are from a specific event type, by combining with the filter with the signature topic. The topics data structure in a block is effectively a map of topics to event indices i.e. `HashSet<Hash, Vec<u32>>>`. So a filter would apply a union to the event indices of the combined topics.

This is all similar to how topics/logs work on EVM chains.

### Fields with the `Option` type

`Option` fields are a special case. If a topic field value is `None` then it pushes the `0x0000...` topic, the same as if `None` was encoded and converted to `[u8; 32]`.

However, the difference is if there is a value present, instead of encoding the `Option<T>` it encodes the value of `T` and uses that as a topic. This is a special case for fields of type `T` and `Option<T>` to have matching topic values. Otherwise a suffix of `1u8` for the encoded `Some` value would prevent the topics being unified. And in the case of the `AccountId` example above it would result in unnecessary hashing because the length of the encoded value would now be 33 bytes.

## Metadata

When events were all defined within the `ink::contract` macro it was easy to gather the metadata for all possible events being raised by a contract.

However with "shared" events now able to be defined and raised anywhere inside or outside of the `ink::contract` definition, the challenge was how to gather all those possible events together and put them in the metadata.

In the end I have used https://github.com/dtolnay/linkme for link time concatenation of event metadata into the `EVENTS` global static. See https://github.com/paritytech/ink/blob/ca861ae7eab19625e3e212be329b185a6219521f/crates/metadata/src/lib.rs#L164

See original discussion https://github.com/paritytech/ink/pull/1243#issuecomment-1116049715

### Max environment topics.

Previously there were compile time checks to ensure that any given event did not have more topics than the max allowed for the given environment. This prevents a possible runtime error at compile time.

However with the event codegen being separated from the contract itself, this compile time check became more difficult, primarily because of the limitation of not being able to assert trait constant values of type parameters (i.e. <E as Environment>::MAX_EVENT_TOPICS)`.

I was able to find a solution to that by using a `const` type parameter and some codegen, but I found the tradeoffs unsatisfactory. So I moved the validation to the metadata generation stage, where it became a simple check at metadata generation time to see whether the number of topics exceeded the environment specification.

### Signature topic collisions

The metadata generation step also ensures that in the scope of the contract, signature topics are unique. This is required (see below) to decide the type of the event to decode when dealing with contract events.

## Consuming Events

Previously the `ink::contract` macro stitched all the `#[ink(event)]` definitions into a top level generated enum. The way to decode any events originating from the contract would be to use the enum encoding of the first byte to determine which event variant to decode to.

Now, the signature topics must be used to determine which event a contract has emitted. See https://github.com/paritytech/cargo-contract/pull/1184.

## Contract Sizes

The simplification of topics codegen should reduce the contract sizes.

With `ink-waterfall` currently not working, the automatic checking of contract size differences is not available. Here are some examples to demonstrate some size reductions:

**erc20** `7.5K` -> `7.1K` = -5.3%
**erc1155** `16.9K` -> `15.8K` = -6.5%

Naturally reductions depend on how many events are being defined in each contract, so would be good to take that into account to.


crates/ink/codegen/src/generator/metadata.rs
```diff
@@ -88,7 +88,7 @@ impl Metadata<'_> {
     fn generate_contract(&self) -> TokenStream2 {
         let constructors = self.generate_constructors();
         let messages = self.generate_messages();
-        let events = self.generate_events();
+        // let events = self.generate_events();
```
SkymanOne, line 91: Is this intentional?

crates/ink/macro/src/event/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+                #topics_builder
+            }
+        }
+     }))
+}
+
+/// The signature topic of an event variant.
+///
+/// Calculated with `blake2b("Event(field1_type,field2_type)")`.
+fn signature_topic(fields: &syn::Fields, event_ident: &syn::Ident) -> syn::LitStr {
```
SkymanOne, line 155: I had an idea of how you can generate a topic for the events with the same field names but different type names (i.e. type aliases).

You can use `TypeId::of::<T>()`. It will infer the type id at the compile type and any aliases and associated types will have the same id. Here is the little demo:
```rust
type Al = u8;

trait Ph {
    type T;
}

struct S;

impl Ph for S {
    type T = u8;
}


fn main() {

    let t1 = TypeId::of::<u8>();
    let t2 = TypeId::of::<Al>();
    let t3 = TypeId::of::<<S as Ph>::T>();
    // all checks out
    assert_eq!(t1, t2);
    assert_eq!(t1, t3);
    assert_eq!(t2, t3);
}
```

TypeId is simply a `u64` value so you can use it to hash. The downside of this approach is that two arbitrary structures that have the same signature (they are identical) but are defined in separate crates will produce different ids.

The only solution to this problem I see is to implement your own version of `TypeId`. Let's call it `EventId`:
* Primitive and built-in types will just use the original id from `TypeId`
* Custom-defined structs which are not events will have to derive `EventId` which will deterministically calculate id based on the fields.
* Any custom-defined structs within that struct should define `EventId` as well.

This is for sure a large additional feature, but this is the only way we can achieve "absolute" determinism in the calculation of the signature topic.

crates/env/src/api.rs
```diff
@@ -175,13 +175,13 @@ where
 }
 
 /// Emits an event with the given event data.
-pub fn emit_event<E, Event>(event: Event)
+pub fn emit_event<E, Evt>(event: Evt)
```
SkymanOne, line 178: Nit: `Evnt` sounds more idiomatic for a generic

crates/ink/macro/src/lib.rs
```diff
@@ -1279,6 +1289,18 @@ pub fn chain_extension(attr: TokenStream, item: TokenStream) -> TokenStream {
     chain_extension::generate(attr.into(), item.into()).into()
 }
 
+synstructure::decl_derive!(
+    [Event, attributes(ink)] =>
+    /// todo
```
SkymanOne, line 1294: What's the todo for?

crates/ink/macro/src/event/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+                        Err(meta
+                            .error(format!("Only `#[ink({path})]` attribute allowed.")))
+                    }
+                }))
+            } else {
+                None
+            }
+        })
+        .collect::<syn::Result<_>>()?;
+    // todo: check only one
```
SkymanOne, line 205: Is this todo still relevant?

SkymanOne (changes requested):
Looks great so far. I put a suggestion on how you can calculate the signature irrespective of the type names. There are also a bunch of places with todos which need to be addressed.

integration-tests/events/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+        pub fn flip(&mut self) {
+            self.value = !self.value;
+            self.env()
+                .emit_event(event_def::Flipped { value: self.value })
+        }
+
+        /// Emit an event with a 32 byte topic.
+        #[ink(message)]
+        pub fn emit_32_byte_topic_event(&mut self, maybe_hash: Option<[u8; 32]>) {
+            self.env().emit_event(event_def::ThirtyTwoByteTopics {
```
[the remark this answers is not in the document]
ascjones: My guess would be that for the new crate you failed to add your new crate to the `std` feature of the contract. Otherwise the `EventMetadata` won't be derived.

https://github.com/paritytech/ink/blob/a78360ad1a3bea244fb686592836eb04fea3470f/integration-tests/events/Cargo.toml#L27

Perhaps we could add a check to the `cargo-contract` companion https://github.com/paritytech/cargo-contract/pull/1184 which would prevent this mistake. It would check that all dependencies are added to the `std` feature.

crates/ink/macro/src/event/mod.rs
ascjones: Main goals for the signature topic are that it can be determined by the macro at codegen time, and that it is reasonably unique and consistent when used across different contracts.

`TypeId`s are only guaranteed to be the same within the context of a single build of a single binary. An event shared across contracts, or even the same contract but with a different version of Rust or some other variable can and does change the value of this. Using the type ids directly is a non starter, because signature topics must be consistent.

That said, I did consider something similar by piggybacking onto `scale_info::TypeInfo` (which uses `TypeId` behind the scenes). The issue with this and similar approaches is that this information is not available in the macro at codegen time, it is only available at runtime. This is why for `ink!` and `cargo-contract` the metadata step creates a binary to execute and build the metadata.

Another idea I had to be agnostic of the field name would be to have some associated consts for types e.g. `trait TopicName { const NAME: &static str; }` then do some const time concatenation. However this was a little tricky and also the best const hash algorithm I could find was https://docs.rs/xxhash-rust/latest/xxhash_rust/const_xxh3/index.html, and topics are 32 bytes.

So for the sake of consistency with topic hashing (using `blake2` hash algorithm), and just keeping things simple I decided to go with using the type name in the code. 

It is true that two otherwise compatible events, one using a type alias the other not would not have the same signature topic, but I believe that is an acceptable trade off.

crates/ink/codegen/src/generator/metadata.rs
ascjones: No, I have removed it now.

crates/ink/macro/src/event/metadata.rs
```diff
@@ -0,0 +1,96 @@
...
+
+            fn event_spec() -> ::ink::metadata::EventSpec {
+               // register this event metadata function in the distributed slice for combining all
+               // events referenced in the contract binary.
+               #[::ink::metadata::linkme::distributed_slice(::ink::metadata::EVENTS)]
+               #[linkme(crate = ::ink::metadata::linkme)]
+               static EVENT_METADATA: fn() -> ::ink::metadata::EventSpec =
+                   <#ident as ::ink::metadata::EventMetadata>::event_spec;
+
+                // todo: check that cfg attributes work here
```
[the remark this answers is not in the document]
ascjones: Thanks. I've just added a UI test for this case

crates/env/src/api.rs
ascjones: I named it `Evt` as I thought it is a common abbreviation of `Event` e.g. https://github.com/abbrcode/abbreviations-in-code#e.

crates/ink/macro/src/lib.rs
ascjones: For docs. Just added those.

integration-tests/events/lib.rs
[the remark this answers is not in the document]
SkymanOne: It has not solved an issue for ,e

crates/env/src/event.rs
```diff
@@ -190,11 +195,15 @@ impl EventTopicsAmount for state::NoRemainingTopics {
...
 ///
 /// Normally this trait should be implemented automatically via the ink! codegen.
-pub trait Topics {
+pub trait Event: scale::Encode {
     /// Type state indicating how many event topics are to be expected by the topics
     /// builder.
     type RemainingTopics: EventTopicsAmount;
 
+    /// The unique signature topic of the event. `None` for anonymous events.
+    /// todo: document how this is calculated
```
SkymanOne, line 204: TODO here

crates/ink/macro/src/event/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+pub use metadata::event_metadata_derive;
+
+use proc_macro2::TokenStream as TokenStream2;
+use quote::{
+    quote,
+    quote_spanned,
+};
+use syn::spanned::Spanned;
+
+/// todo: docs
```
SkymanOne, line 26: TODO here

crates/ink/macro/src/event/mod.rs
SkymanOne: How about allowing to manually specify the signature topic? If for whatever reason, the exact definition of types can not be used, we can at least allow to specify the topic manually. We are doing that with custom selectors, why not do it with the event signatures?

SkymanOne: How about allowing to manually specify the signature topic? If for whatever reason, the exact definition of types can not be used, we can at least allow to specify the topic manually. We are doing that with custom selectors, why not do it with the event signatures?

crates/ink/macro/src/event/mod.rs
ascjones: It was, I have implemented it now

crates/ink/macro/src/event/mod.rs
ascjones: It's a good idea, although needs a little thinking about so we can do it as a follow up: https://github.com/paritytech/ink/issues/1838

crates/ink/macro/src/event/mod.rs
ascjones: Done

integration-tests/events/lib.rs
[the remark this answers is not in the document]
ascjones: Maybe it is something to do with MacOS. I have added some tests to `integration_tests/events` to ensure that all events that are defined and used are included in the metadata. Can you try and run them on your machine?

crates/env/src/event.rs
ascjones: Done

integration-tests/events/lib.rs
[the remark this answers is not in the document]
SkymanOne: Nope. Still has not resolved the issue. I did `cargo clean` then `cargo contract build` and here is the dump of events:
```json
    "events": [
      {
        "args": [
          {
            "docs": [],
            "indexed": false,
            "label": "value",
            "type": {
              "displayName": [],
              "type": 0
            }
          }
        ],
        "docs": [],
        "label": "InlineFlipped",
        "module_path": "events::events",
        "signature_topic": "0xab768ef77e8fd0c0740c523252aebd65d37bb2b5f508d2306f876396ca7e4447"
      },
      {
        "args": [
          {
            "docs": [],
            "indexed": false,
            "label": "value",
            "type": {
              "displayName": [],
              "type": 0
            }
          }
        ],
        "docs": [],
        "label": "ForeignFlipped",
        "module_path": "event_def",
        "signature_topic": "0x7ce035d0d27075652696f71842317eb893587fc339f72377127bb4cfd293cf03"
      },
      {
        "args": [
          {
            "docs": [],
            "indexed": true,
            "label": "hash",
            "type": {
              "displayName": [],
              "type": 6
            }
          },
          {
            "docs": [],
            "indexed": true,
            "label": "maybe_hash",
            "type": {
              "displayName": [],
              "type": 5
            }
          }
        ],
        "docs": [],
        "label": "ThirtyTwoByteTopics",
        "module_path": "event_def",
        "signature_topic": "0xa48606fe432343ce5000004df20ae1efc75fc1dfd9ecb9b99006df6b671faf00"
      }
    ],
 ```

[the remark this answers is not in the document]
Comment (ascjones) [reply integration-tests/events/lib.rs:40]: Please try with https://github.com/paritytech/cargo-contract/pull/1200, it should fix that.

**Review by xgreenx — COMMENTED** [review]
Thank you for implementing it! It looks very nice and easy to use=) And the change itself is clean and with many tests.

I want to highlight that the current implementation also includes all events defined in the crate if we import something from this crate. https://github.com/paritytech/ink/pull/1842 shows it by failing the `collects_specs_for_all_linked_and_used_events` test.

We discussed the potential solution to this problem [here](https://github.com/paritytech/ink/pull/1243#issuecomment-1119838754).

If we plan to fix that, then we need either create a follow-up issue or not close the current one.

**Review by xgreenx — COMMENTED** [review]
Another question, and it sounds more like an additional feature: Do we want to support generics in events? Because right now, the code is not ready for it=)

**Comment by ascjones** [comment]
> Another question, and it sounds more like an additional feature: Do we want to support generics in events? Because right now, the code is not ready for it=)

I think we will want to support it, e.g. to allow for supporting different `Environment`s. But for this first iteration we will disallow them as they were previously.

Thanks for the reminder, I had intended to put in a check for this but forgot. I have added one now in the `#[derive(Event)]` plus a UI test.

File: crates/ink/macro/src/event/mod.rs [hunk 10]
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+}
+
+/// `Event` derive implementation for `struct` types.
+fn event_derive_struct(mut s: synstructure::Structure) -> syn::Result<TokenStream2> {
+    assert_eq!(s.variants().len(), 1, "can only operate on structs");
+    let span = s.ast().span();
+
+    let anonymous = has_ink_attribute(&s.ast().attrs, "anonymous")?;
```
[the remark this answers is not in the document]
ascjones: This is implemented now

ascjones:
> I want to highlight that the current implementation also includes all events defined in the crate if we import something from this crate. #1842 shows it by failing the `collects_specs_for_all_linked_and_used_events` test.
> 
> We discussed the potential solution to this problem [here](https://github.com/paritytech/ink/pull/1243#issuecomment-1119838754).

In the first iteration I implemented this solution, see the [`ink`](https://github.com/paritytech/ink/pull/1243/files#diff-428572beed72394835b75703b95ec8270d48aeb27272ae83a4de759ab70c5578R89) side and the [`cargo-contract`](https://github.com/paritytech/cargo-contract/pull/794/files#diff-776ff6d707c234eea92e74b9c95ec7cb20eafa8bb6053c9fede27bf5f842044aR187) side.

However I decided that this approach was introducing some non obvious coupling between the `ink` and `cargo-contract` implementations, and was slightly too convoluted.

The implementation now is at least "encapsulated" in `ink` (i.e. we could change e.g. to [`inventory`](https://github.com/dtolnay/inventory), similar to your [original idea](https://github.com/xgreenx/ink-metadata)).

Of course the tradeoff is that if you reference a crate with an event but don't emit it, it still gets used in the metadata, as you demonstrate in https://github.com/paritytech/ink/pull/1842. However I would say that having an event in the metadata that *could* possibly be raised but is never raised is not such a big issue. It is even possible with any solution to have event which is emitted in a logically unreachable code branch  it `if balance > MAX_SUPPLY { self.env().emit_event(UnreachableEvent) }` which would be included in the metadata anyway.

So on balance I have chosen the simpler solution.

xgreenx:
> Of course the tradeoff is that if you reference a crate with an event but don't emit it, it still gets used in the metadata, as you demonstrate in https://github.com/paritytech/ink/pull/1842. However I would say that having an event in the metadata that could possibly be raised but is never raised is not such a big issue. It is even possible with any solution to have event which is emitted in a logically unreachable code branch it if balance > MAX_SUPPLY { self.env().emit_event(UnreachableEvent) } which would be included in the metadata anyway.
>
>So on balance I have chosen the simpler solution.

I was happy with this approach from the beginning https://github.com/paritytech/ink/pull/1243#issuecomment-1118720495 =D I only highlighted that the problem is not solved(we discussed it during the previous iteration, and I thought is our final goal to solve it)=)

ascjones:
> I was happy with this approach from the beginning https://github.com/paritytech/ink/pull/1243#issuecomment-1118720495 =D I only highlighted that the problem is not solved(we discussed it during the previous iteration, and I thought is our final goal to solve it)=)

Yes you were right all along :)

crates/ink/macro/src/lib.rs
```diff
@@ -646,6 +650,37 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///
+/// ```
+/// #[ink::event]
+/// pub struct MyEvent {
+///     pub field: u32,
+///     #[ink(topic)]
+///     pub topic: [u8; 32],
+/// }
+///
+/// // The `#[ink(anonymous)]` attribute means no signature topic will be emitted for the event.
```
xgreenx, line 670: I guess we need to update this comment with a new syntax for `anonymous`=)

xgreenx (approved):
@cmichi @SkymanOne The pull request is ready for a final review=) It looks good to me.

ascjones:
While we wait, here is the comparison of contract sizes.

Used script from https://github.com/paritytech/ink/pull/1851, run on `master` and on this branch and then pass the resulting files through `awk`:

```bash
awk 'NR==FNR {a[$1]=$2; next} $1 in a {print "|" $1 "|" a[$1] "|" $2 "|" (($2 - a[$1]) / a[$1]) * 100 "|"}' master shared-events-redux
```

|Contract|master|this PR|diff %|
|------|------|-------|-------|
|trait_erc20|7887|7407|-6.08596|
|call_builder_return_value|8952|8952|0|
|wildcard_selector|2496|2496|0|
|psp22_extension|7012|7012|0|
|rand_extension|2935|2734|-6.84838|
|e2e_call_runtime|1000|1000|0|
|multisig|23387|22074|-5.61423|
|dns|8797|7879|-10.4354|
|erc20|7487|7072|-5.54294|
|conditional_compilation|1176|1176|0|
|trait_incrementer|1274|1336|4.86656|
|erc721|11277|10427|-7.53747|
|call_runtime|1781|1781|0|
|custom_allocator|8216|8216|0|
|erc1155|16893|15804|-6.44646|
|integration_flipper|1599|1599|0|
|call_builder|5179|5179|0|
|call_builder_delegate|2436|2436|0|
|constructors_return_value|1779|1779|0|
|contract_ref|4710|4710|0|
|updated_incrementer|1458|1458|0|
|incrementer|1215|1479|21.7284|
|trait_flipper|1176|1176|0|
|mapping_integration_tests|3346|3346|0|
|accumulator|1052|1052|0|
|adder|1710|1710|0|
|multi_contract_caller|6103|6103|0|
|subber|1727|1727|0|
|contract_transfer|1441|1441|0|
|contract_terminate|1149|1149|0|
|flipper|1403|1403|0|
|payment_channel|5942|5905|-0.622686|
|other_contract|1295|1295|0|
|basic_contract_caller|3117|3117|0|
|trait_incrementer|1274|1274|0|
|trait_incrementer_caller|2569|2569|0|
|custom_environment|2548|1876|-26.3736|
|incrementer|1215|1215|0|
|mother|10410|10390|-0.192123|

SkymanOne (approved):
Looks good to me! Congrats on the PR. Just one last small nitpick

PR #1828, opened 2023-06-26
Show docs for Result type aliases

ascjones:
At the moment where these type aliases are used, you cannot navigate to the definition e.g. https://docs.rs/ink_env/4.2.1/ink_env/fn.invoke_contract.html.

This PR removes the `#[doc(hidden)]` so those types are shown now in the `ink_primitives` crate.


SkymanOne (approved):
Lgtm

PR #1831, opened 2023-07-02
Remove unchecked arithmetic

athei:
We need to remove all unchecked arithmetic from ink! itself so our CI does not fail after https://github.com/paritytech/cargo-contract/pull/1190 is merged.

I trivially transformed all unchecked arithmetic to `checked_* + unwrap` which preserves the current behaviour which was to panic on overflow but makes it explicit. I had to inline the `arrayref` macro so that I were able to remove the unchecked arithmetic from it.


integration-tests/erc1155/lib.rs
```diff
@@ -267,7 +267,10 @@ mod erc1155 {
 
             // Given that TokenId is a `u128` the likelihood of this overflowing is pretty
             // slim.
-            self.token_id_nonce += 1;
+            #[allow(clippy::arithmetic_side_effects)]
```
SkymanOne, line 270: I'm not fun of a this. Is there a reason we can't do checked arithmetic operation here. If the operation is not implemented for the type, then we need to implement it first before introducing the change here and in cargo contract.

Otherwise we are breaking the arithmetic safety guarantees.

athei: We can. But it is not necessary here. Overflowing a u128 is impossible by incrementing by 1 per extrinsic.

SkymanOne: Do I understand correctly that we won't be able to use arithmetic operators (i.e. `+`, `-`, `/`, `*`) on numbers at all? I am just thinking of what impact it will have on the developer experience if we have to tell everyone knows that they need to use `checked_` operations specifically.

athei: > Do I understand correctly that we won't be able to use arithmetic operators (i.e. +, -, /, *) on numbers at all?

Correct.

> I am just thinking of what impact it will have on the developer experience if we have to tell everyone knows that they need to use checked_ operations specifically.

Or `saturating` or `wrapping`. Developers should express the overflow behaviour in-code instead of leaving it open to a compile time switch. We should have done that from the beginning. People **have** to think about this anyways.

athei:
Should be merged before new cargo contract release or the CI will break.

SkymanOne (approved):
Can be merged now

integration-tests/erc1155/lib.rs
koute: > Or `saturating` or `wrapping`.

Or use [`Wrapping`](https://doc.rust-lang.org/core/num/struct.Wrapping.html) and [`Saturating`](https://doc.rust-lang.org/core/num/struct.Saturating.html), so then the arithmetic itself wouldn't become an unreadable soup of function calls, e.g. this compiles:

```rust
#![deny(clippy::arithmetic_side_effects)]
use core::num::Wrapping;

fn foo(a: Wrapping<u32>, b: Wrapping<u32>) -> Wrapping<u32> {
    a + b
}
```

 (`Saturating` is still unstable though, but we could just provide our own wrapper for the time being.)

PR #1837, opened 2023-07-05
Linter: Warn when number primitive is annotated with `#[ink(topic)]`

### LINKED ISSUE #1436: Linter: Warn when number primitive annotated with `#[ink(topic)]`
The `#[ink(topic)]` annotation enables third party tooling to filter for "topics" ‒ discrete events for which it makes sense to filter. Typical examples where this can make sense are `AccountId`, `bool` or `enum` variants.

It typically doesn't make sense to annotate types like `u32` or `i32` as a topic, if those fields can take continuous values that could be anywhere between `::MIN` and `::MAX`. An example of a case where it doesn't make sense at all to have a topic on the storage field is something like `value: Balance`. 

We should provide a linter rule that warns developers when they annotate a number primitive with `#[ink(topic)]`.

jubnzv:
Closes #1436


linting/src/primitive_topic.rs
```diff
@@ -0,0 +1,222 @@
+// Copyright 2018-2023 Parity Technologies (UK) Ltd.
```
SkymanOne, line 1:
```suggestion
// Copyright (C) Parity Technologies (UK) Ltd.
```

linting/src/primitive_topic.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// the [`#[ink(topic)]`](https://use.ink/macros-attributes/topic) annotation with primitive
+    /// number types. Topics are discrete events for which it makes sense to filter. Typical
+    /// examples of fields that should be filtered are `AccountId`, `bool` or `enum` variants.
+    ///
+    /// **Why is this bad?** It typically doesn't make sense to annotate types like `u32` or `i32`
+    /// as a topic, if those fields can take continuous values that could be anywhere between
+    /// `::MIN` and `::MAX`. An example of a case where it doesn't make sense at all to have a
+    /// topic on the storage field is something like `value: Balance` in the examle below.
+    ///
+    /// **Known problems:** Events 2.0 currently are not supported:
```
SkymanOne, line 66: If events are not supported then we should either merge this PR into `4.2` branch or make it compatible with the Events 2.0

linting/src/primitive_topic.rs
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+
+/// Returns `true` if `impl_item` is the `topics` function
+fn is_topics_function(impl_item: &ImplItemRef) -> bool {
+    impl_item.kind == AssocItemKind::Fn { has_self: true }
+        && impl_item.ident.name.as_str() == "topics"
+}
+
+/// Returns `true` if `ty` is a primitive type that should not be annotated with
```
SkymanOne, line 121:
```suggestion
/// Returns `true` if `ty` is a numerical primitive type that should not be annotated with
```

linting/src/primitive_topic.rs
```diff
@@ -0,0 +1,222 @@
...
+/// `#[ink(topic)]`
+fn is_primitive_number_ty(ty: &Ty) -> bool {
+    matches!(ty.kind(), ty::Int(_) | ty::Uint(_))
+}
+
+/// Reports field of the event structure
```
SkymanOne, line 127:
```suggestion
/// Reports a topic-annotated field with a numerical primitive type
```

SkymanOne (changes requested):
Overall, looks very good. We just need to decide whether to merge it as a minor patch of the ink! 4, or make it compatible with Events 2.0 and merge it in ink! 5.

I personally think we should make it work with Event 2.0 as ink! 5.0 might be released soon, and we will still have to add support for the new event definition.

linting/src/primitive_topic.rs
SkymanOne: My bad, it looks like your recent commit: https://github.com/paritytech/ink/pull/1837/commits/f10985bd17393a43f48468be305fd9801a1e1696 added the support for the Events 2.0. If that's the case, can you please update the comment?

jubnzv:
@SkymanOne Thanks for the review. I commited all the suggestions.

SkymanOne (changes requested):
Tried to run the tests locally I got a bunch of compilation errors related to `clippy_utils`. 
 
The requested changes are what helped me to resolve the issue.

linting/Cargo.toml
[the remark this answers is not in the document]
jubnzv: Done.

I updated `clippy_utils` version to `1d334696587ac22b3a9e651e7ac684ac9e0697b2` and set channel to `nightly-2023-07-14` in `rust-toolchain.toml`. `nightly-2023-07-14` is required, since this version [is used](https://github.com/rust-lang/rust-clippy/blob/1d334696587ac22b3a9e651e7ac684ac9e0697b2/rust-toolchain#L2) in `rust-clippy` at commit `1d334696587ac22b3a9e651e7ac684ac9e0697b2`.

How should we choose the `clippy` version for `linting`? Should we always take the latest version from the `clippy` `master` branch?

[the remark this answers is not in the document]
SkymanOne: Not sure what you mean by `clippy` here. If you are referring to `clippy_utils`, I've been following a template https://github.com/trailofbits/dylint/blob/v2.1.12/internal/template/Cargo.toml~ in dylint repo

linting/Cargo.toml
```diff
@@ -18,18 +18,18 @@ include = ["Cargo.toml", "*.rs", "LICENSE"]
 crate-type = ["cdylib"]
 
 [dependencies]
-clippy_utils = { git = "https://github.com/rust-lang/rust-clippy", rev = "1480cea393d0cee195e59949eabdfbcf1230f7f9" }
-dylint_linting = "2.0.0"
+clippy_utils = { path = "/home/georgiy/Sources/rust-clippy/clippy_utils" }
```
SkymanOne, line 21: This points to your home directory instead of the correct commit

SkymanOne (changes requested):

linting/Cargo.toml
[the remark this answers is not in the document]
SkymanOne: AFAIK we only have `rust-toolchain.toml` for the lining crate

PR #1842, opened 2023-07-06
Usage of the crate includes defined there events

xgreenx:
The PR shows that including any code from the crate also imports all defined there events. We discussed about the potential [solution](https://github.com/paritytech/ink/pull/1243#issuecomment-1119838754) during previous PR=)


ascjones (commented):
See comment https://github.com/paritytech/ink/pull/1827#issuecomment-1634490103.

If you agree, could you change this PR to demonstrate that an event defined in a dependency, which is not explicitly used, but other code in the crate is used, **is** expected to be included in the metadata? 

This way the test is proving what our expected behaviour is.

ascjones (approved):

Issue #1834, opened 2023-07-04
Introducing alternative backend to E2E framework


pmikolajczyk41:
# Target

As discussed recently, it would be a beneficial to integrate [`DRink!`](https://github.com/Cardinal-Cryptography/drink/) to the ink! E2E framework. The ultimate goal would be to enable developers to write:
```rust
#[ink::e2e_test(backend = "node")]
fn contract_test() { ... }
// or
#[ink::e2e_test(backend = "runtime-only", runtime = my_chain_lib::Runtime)]
fn contract_test() { ... }
```
i.e. to be able to either:
 - run contract tests on a full stack (as it is done now: running a node in the background, by default `substrate-contracts-node`), or
 - run contract tests directly in the runtime (just like pallet unit tests); this is the approach that `DRink!` library provides (together with a customizable runtime)

# Rationale

While every production contract ought to be tested in the full target environment, it is often very onerous to run such tests during early/local development. This is due to e.g. node/network configuration issues, blocktime/finalization or overhead coming from asynchronous communication between e2e client and RPC chain service. Thus we find it useful, to skip the whole node layer while testing contracts (although the topic applies to generally whole runtime testing).

Interacting directly with runtime might be seen as 'quasi-end-to-end' testing, and hence can't replace old framework, but still gives some benefits:
 - synchronous, instant operations (no need for sending transaction, transaction pool processing, block production, finalization, block/event subscriptions, etc.)
 - direct, full-power access to runtime (skipping million blocks, setting timestamps, balance control)
 - access to runtime internals (e.g. accessing contract's debug buffer or contract execution stack frames in the contracts pallet)
 - no setup apart from choosing runtime (no node-per-test running, opening ports etc.)

# Plan

The work will be done as a series of smaller steps

```[tasklist]
### Tasks
- [x] Make `drink` generic over runtime (to do outside this repository)
- [x] Create a trait for E2E backend - put current `Client`'s capabilities behind a trait + use opaque type in the test body
- [x] Add to `ink::e2e_test` macro a new attribute for choosing backend. In case of `"runtime-only"` one, allow for specifying desired runtime. There should be also an option for choosing custom backend implementation.
- [x] Implement the trait for `Session` from `drink` library
```

PR #1844, opened 2023-07-10
[E2E alternative backend]: Abstract error and result structs

pmikolajczyk41:
First step for introducing common backend trait:
 - move `Error` enum to a separate module and make it backend-agnostic
 - move `CallDryRunResult`, `CallResult`, `InstantiationResult`, `UploadResult` to a separate module and make them backend-agnostic

Fortunately, no changes to the test code is needed (apart from some type annotation for errors).

ref: #1834

### DIFF
--- a/crates/e2e/src/contract_results.rs
+++ b/crates/e2e/src/contract_results.rs
@@ -1,6 +1,8 @@
+use crate::CallDryRunResult;
 use ink::codegen::ContractCallBuilder;
 use ink_env::call::FromAccountId;
 use ink_env::Environment;
+use ink_primitives::MessageResult;
 use pallet_contracts_primitives::{CodeUploadResult, ContractInstantiateResult};
 use std::fmt::Debug;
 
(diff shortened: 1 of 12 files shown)


deuszx (approved):
LGTM!

crates/e2e/src/contract_results.rs
```diff
@@ -0,0 +1,204 @@
...
+
+/// Result of a contract instantiation.
+pub struct InstantiationResult<E: Environment, EventLog> {
+    /// The account id at which the contract was instantiated.
+    pub account_id: E::AccountId,
+    /// The result of the dry run, contains debug messages
+    /// if there were any.
+    pub dry_run: ContractInstantiateResult<E::AccountId, E::Balance, ()>,
+    /// Events that happened with the contract instantiation.
+    pub events: EventLog,
```
ascjones, line 26: So the concrete type for this is `ExtrinsicEvents`, what's the idea for the alternative backend?

Certainly we would want to access the events with a consistent API regardless of the backend being used.

pmikolajczyk41: for 100% drink will support events (it's in progress), but it's not there yet; once it arrives, I will put a common trait over `EventLog` here for both backends; I think it should be at least iterable over some `Event` type with lookup

ascjones (approved):
LGTM!

PR #1848, opened 2023-07-14
Added Lottery contract as an example

niklabh:


SkymanOne:
I am not sure what the educational value behind the example is. Could you elaborate on how the example is different from existing ones? The example also uses a pseudo-random generator which might be misleading for newcomers

niklabh:
Just wanted to add an example where send and receive funds are used. Should i change random number generation. Or if this is not useful should i close this @SkymanOne

PR #1850, opened 2023-07-19
ink-e2e Test Failures in substrate-contracts-node (v0.27.0) due to Unsupported --ws-port Argument

varex83:
# Summary:
This PR addresses an issue with the latest version of substrate-contracts-node (v0.27.0) wherein the --ws-port argument is not being supported, causing the end-to-end tests to fail. The specific error message produced is: "We should find a port before the reader ends".


varex83:
I was doing it from the v4.1.0 tag, and didn't know where to merge, please @SkymanOne can you help me with that?

ascjones (approved):
Thanks! Please merge `master` to fix CI

PR #1857, opened 2023-07-26
[E2E alternative backend]: Backend traits

pmikolajczyk41:
The PR introduces abstraction for all available operations within e2e tests. These actions are broken into two categories:
 - general chain utilities (creating endowed account, checking account's balance, etc.)
 - contract specific operations (code upload, instantiation, calling)

Thus there are two corresponding separate traits: `ChainBackend` and `ContractsBackend` that are combined further into `E2EBackend`.

# Notes / dilemmas

## Mutability in backend methods

While so far, the subxt-based E2E client could work just on `&self`, we have to make all backend methods to operate on `&mut self`. This is because drink! client will have to keep the whole runtime (externalities) by itself and thus any action will mutate its state.

While some workaround might have been done (with some `RefCells` etc), IMHO it's completely not worth the effort and complication (at least at this stage).

## Asynchronicity of backend traits

While drink backend is inherently synchronous, subxt-based one cannot be. Unfortunately I don't know any ethic way of keeping common facade for such two alternatives and thus I made all methods asynchronous. As a result:
 - drink! implementation will be artificially asynchronous, but actually the logic will be instant anyway (there will be no breakpoints in the async state machine)
 - the testcase body stays the same, no matter which backend it uses (this is actually one of the goals)

## Breaking `E2EBackend` into two traits

This is just a matter of taste - for me it is just cleaner to separate general-chain logic from the contract-specific (this is how I did it in drink! as well). However, there are 3 immediate downsides:
 - e2e test module must import two traits
 - there's some redundancy between these traits: they share some associated types, which must be declared twice
 - all the trait bounds must be repeated 3 times

I'm not very dedicated to that support, so if any of the reviewers is in favor of having a single trait, I'm on it.

## Awful associated types bounds repetition

Since there are actually 3 `impl *Backend for Client`, there are 3 long blocks of identical lists of trait bounds for the types within `E` and `C` (environment and subxt config). Unfortunately, due to: https://github.com/rust-lang/rust/issues/20671 I cannot do much here (unless we want to include unstable feature of `trait_alias`).

## `Actor` vs `ActorId`

In the full context (subxt-based client) we usually have to differentiate between user's account and user's keypair: we use the first one to send tokens there or read its state, and the second one for signing transactions, that are sent to a node. However, in drink! backend, there is no need for signing anything, and thus these two concepts can be collapsed to just a user's account.

## Dry-running

I am not sure if dry-running should be a part of the backend trait. It seems useful, but it's not that obvious how it should be implemented for drink!... For now I included it and I will see soon what is possible to do about it in drink!.


pmikolajczyk41:
@ascjones this is now a half-baked PR, but I wanted kindly ask you if you could take a quick look and tell whether the direction is okay with you; if so, then I will continue with `ContractsBackend` along the same line

SkymanOne:
LMK when it is ready for review, I will convert the PR to `ready`

ascjones (commented):
Looks good so far, carry on!

Not sure about the name `Actor`, would be interested to know what the equivalent concrete types are for the `Drink!` backend. That is a minor detail though.

pmikolajczyk41:
okay, I'm done for the first iteration now - @SkymanOne @ascjones please take a look :) notice that the PR description has been significantly expanded, hopefully giving some help/insights into some of the decisions that I made on spot

due to the problems with `contract-build` crate, tests won't work (the version `3.1.0` has been yanked after last commit to main, 4.0.0-alpha gives now `CodeRejected` error while instantiating any contract) - here I'm kindly asking @ascjones for assistance :pray:

crates/e2e/src/subxt_client.rs
```diff
@@ -444,43 +332,125 @@ where
...
+
+    E: Environment,
+    E::AccountId: Debug + Send + Sync,
+    E::Balance: Clone
+        + Debug
+        + Send
+        + Sync
+        + TryFrom<u128>
+        + scale::HasCompact
+        + serde::Serialize,
```
deuszx, line 362: This _could_ be extracted to something like
```rust
trait EnvBalanceT: Clone + Debug + Send + ... {}
impl<B> EnvBalanceT for B where B: Clone + Debug + Send + ... {}
```
and here, simplify to
```rust
E::Balance: EnvBalanceT
```

(others possibly similarly)

deuszx, line 362: I _think_ this can be simplified (but I haven't compiled for this particular case) to:
```rust
trait EnvT: Environment { }
impl<T> EnvT for T where
	T: Environment,
	T::AccountId: Debug + Send + Sync,
	T::Balance: Clone + Debug + Sync + ... { }
```
and here, simplify to
```rust
E: EnvT
```
🤔 
(similarly for others as well)

ascjones (approved):
In general this all LGTM.

Another possibility for the interface would be to make it synchronous and adapt the `subxt` client into sync methods. Though that will require lots of changes to existing tests so let's keep it async for now.

Also if `Actor` is referring to `Account`, couldn't we just make the types `Account` and `AccountId`?

Anyway, nothing from my side to prevent this PR from being merged. Looking forward to the next one!

pmikolajczyk41:
I have just merged master and added licenses to the files that I recently created.

Regarding `Actor` vs `Account`... it's just my hard discomfort when it comes to semantics :P in the subxt case, `type Account = Keypair` would look confusing imho; however, if this `Actor` stings the eyes, I can change it :)

crates/e2e/src/subxt_client.rs
```diff
@@ -46,14 +46,24 @@ use ink_env::{
     },
```
SkymanOne, line 46: I am personally not a fun of renaming the file into `subxt_client.rs`. IMO, it is better either to keep the name `client.rs` or rename it to the `impls.rs` like it's done in other ink! crates. Since, all the create provides an implementation for the backend traits

integration-tests/call-builder-return-value/lib.rs
```diff
@@ -113,6 +113,10 @@ mod call_builder {
     mod e2e_tests {
         use super::*;
         use incrementer::IncrementerRef;
+        use ink_e2e::{
+            ChainBackend,
```
SkymanOne, line 117: Why can't you import `E2EBackend` since it is a subtrait of both backend traits?

SkymanOne (changes requested):

crates/e2e/src/subxt_client.rs
pmikolajczyk41: I was intending to put drink backend into a separate module (like `drink_client.rs`) to keep clear isolation of the very different code and hence this renaming. Would you suggest putting everything into a signle `client.rs`/`impls.rs` module?

integration-tests/call-builder-return-value/lib.rs
pmikolajczyk41: you actually have to import the source trait that provides methods :|

crates/e2e/src/subxt_client.rs
SkymanOne: If you plan to integrate drink backend in the ink! e2e crate directly, then it makes sense to keep `subxt_client.rs` name

integration-tests/call-builder-return-value/lib.rs
SkymanOne: What's the point of having `E2EBackend` trait then?

integration-tests/flipper/lib.rs
```diff
@@ -55,6 +55,10 @@ pub mod flipper {
     #[cfg(all(test, feature = "e2e-tests"))]
     mod e2e_tests {
         use super::*;
+        use ink_e2e::{
+            ChainBackend,
```
SkymanOne, line 59: flipper examples does not need `ChainBackend` import. Can you please double check other examples and keep only the necessary imports?

crates/e2e/src/subxt_client.rs
pmikolajczyk41: that's the plan, so that a developer when writing tests can only specify which backend they need without worrying about dependencies or imports

integration-tests/call-builder-return-value/lib.rs
pmikolajczyk41: your e2e testcase will receive an object implementing this trait (instead of `subxt_client::Client` object as it is done now)

integration-tests/flipper/lib.rs
pmikolajczyk41: good spot - I've removed `ChainBackend` import from overall 14 tests where it wasn't used

SkymanOne (approved):
Looks way better now. Happy to get it merged!

crates/e2e/src/subxt_client.rs
SkymanOne: Then it makes sense to keep the name

PR #1864, opened 2023-08-02
[E2E alternative backend]: Backend choice

pmikolajczyk41:
Finally we provide implementation of `E2EBackend` trait for a drink-based client. Also, we add a new argument for `e2e-test` macro, enabling user to specify target backend architecture.

Since `drink` library is pretty immature yet, we are lacking some functionalities like dry-running or events. Relevant places in code are marked with the corresponding issues in drink repository. We'll fix them in the near future.

Notes:
 - added a new integration test based on `flipper` that uses new backend
 - had to add `wasm-instrument` with `sign_ext` feature because of compilation error: `non-exhaustive patterns: `&Instruction::SignExt(_)` not covered` for `cargo 1.70`
 - removed `Actor` and `ActorId` associated types from backend traits: since we have to support all backends for a single test code, we always have to operate with `Keypair` and `E::AccountId` types
 - moved common code for all backends to a common module `client_utils`
 - error type is for now `()`, I will adapt `DrinkBackend` to the main `Error` soon (required some changes to drink as well)

### DIFF
--- a/crates/e2e/Cargo.toml
+++ b/crates/e2e/Cargo.toml
@@ -20,6 +20,7 @@ ink = { workspace = true, default-features = true }
 ink_env = { workspace = true, default-features = true }
 ink_primitives = { workspace = true, default-features = true }
 
+drink = { workspace = true, features = ["session"] }
 funty = { workspace = true }
 impl-serde = { workspace = true }
 jsonrpsee = { workspace = true, features = ["ws-client"] }
@@ -31,6 +32,7 @@ tracing-subscriber = { workspace = true, features = ["env-filter"] }
 scale = { package = "parity-scale-codec", workspace = true }
 subxt = { workspace = true }
 subxt-signer = { workspace = true, features = ["subxt", "sr25519"] }
+wasm-instrument = { workspace = true }
 
 # Substrate
 pallet-contracts-primitives = { workspace = true }
--- a/crates/e2e/Cargo.toml
+++ b/crates/e2e/Cargo.toml
@@ -20,7 +20,7 @@ ink = { workspace = true, default-features = true }
 ink_env = { workspace = true, default-features = true }
 ink_primitives = { workspace = true, default-features = true }
 
-drink = { workspace = true, features = ["session"] }
+drink = { workspace = true }
 funty = { workspace = true }
 impl-serde = { workspace = true }
 jsonrpsee = { workspace = true, features = ["ws-client"] }
--- a/crates/e2e/Cargo.toml
+++ b/crates/e2e/Cargo.toml
@@ -20,6 +20,8 @@ ink = { workspace = true, default-features = true }
 ink_env = { workspace = true, default-features = true }
 ink_primitives = { workspace = true, default-features = true }
 
+cargo_metadata = { workspace = true }
+contract-build = { workspace = true }
 drink = { workspace = true }
 funty = { workspace = true }
 impl-serde = { workspace = true }
--- a/crates/e2e/macro/src/codegen.rs
+++ b/crates/e2e/macro/src/codegen.rs
@@ -170,7 +170,7 @@ impl InkE2ETest {
                     return ::ink_e2e::tokio::runtime::Builder::new_current_thread()
                         .enable_all()
                         .build()
-                        .unwrap_or_else(|err| panic!("Failed building the Runtime: {}", err))
+                        .unwrap_or_else(|err| panic!("Failed building the Runtime: {err}"))
                         .block_on(run);
                 }
             }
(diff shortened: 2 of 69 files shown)


crates/e2e/macro/src/codegen.rs
```diff
@@ -130,23 +135,10 @@ impl InkE2ETest {
...
-                    "The '{DEFAULT_CONTRACTS_NODE}' executable was not found. Install '{DEFAULT_CONTRACTS_NODE}' on the PATH, \
-                    or specify the `CONTRACTS_NODE` environment variable.",
-                )
-            } else {
-                panic!("The contracts node executable '{contracts_node}' was not found.")
-            }
-        }
+        let client_building = match self.test.config.backend() {
+            Backend::Full => build_full_client(&environment, contracts),
+            Backend::RuntimeOnly => todo!("Runtime-only backend not yet supported"),
```
SkymanOne, line 140: Since the runtime only backend is not yet available. I think it is reasonable to postpone the PR and introduce the changes once the runtime only backend is available.

pmikolajczyk41: Fair point, however this way we will be able to validate easily the next PR - the next code I'll bake is `impl E2EBackend for DrinkClient`. Having already the possibility to run e2e tests with new backend will make testing these changes straightforward.

SkymanOne: 140 lines is not a big change. I think we can squeeze it in the next PR, when you implement the backend for drink. That way we'll have a fuller picture.

SkymanOne (changes requested):
Looks very solid. Just a few nitpicks

crates/e2e/src/client_utils.rs
```diff
@@ -0,0 +1,61 @@
+use crate::log_info;
+use std::{
+    collections::BTreeMap,
+    path::PathBuf,
+};
+
+/// Generate a unique salt based on the system time.
+pub fn salt() -> Vec<u8> {
```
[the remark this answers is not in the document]
pmikolajczyk41: fair point, but I'd go for not changing it in this PR for 2 reasons:
 - it requires adding a new dependency
 - I just copied this code from `subxt_client` module, so actually this PR wasn't aiming at any change in this logic

crates/e2e/src/backend.rs
```diff
@@ -38,10 +39,8 @@ pub trait E2EBackend<E: Environment = DefaultEnvironment>:
 /// General chain operations useful in contract testing.
 #[async_trait]
 pub trait ChainBackend {
-    /// Abstract type representing the entity that interacts with the chain.
-    type Actor: Send;
-    /// Identifier type for an actor.
-    type ActorId;
+    /// Account type.
+    type AccountId;
```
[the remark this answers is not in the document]
pmikolajczyk41: requiring bounds on the associated type here would be a bit implementation-based, so I'm not convinced if it is the correct move

but anyway, even if I put these bounds here, I would have to keep them in the `impl<AccountId, *> *Backend for *`, so nothing to 'gain' here

crates/e2e/src/backend.rs
```diff
@@ -53,12 +52,15 @@ pub trait ChainBackend {
     /// `sender` actor.
```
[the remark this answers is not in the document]
pmikolajczyk41: :white_check_mark:

crates/e2e/macro/src/config.rs
```diff
@@ -18,6 +18,35 @@ use ink_ir::{
...
 };
 
+/// The type of the architecture that should be used to run test.
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Default)]
+pub enum Backend {
+    /// The standard approach with running dedicated single-node blockchain in a
+    /// background process.
+    #[default]
+    Full,
+    /// The lightweight approach skipping node layer.
```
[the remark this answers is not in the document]
pmikolajczyk41: :white_check_mark:

SkymanOne (approved):
LGTM!

crates/e2e/Cargo.toml
```diff
@@ -22,6 +22,7 @@ ink_primitives = { workspace = true, default-features = true }
 
 cargo_metadata = { workspace = true }
 contract-build = { workspace = true }
+drink = { workspace = true }
```
ascjones, line 25: Should we consider having a feature flag for this, since it is bringing in a few heavy substrate dependencies and (at the moment) is not the default backend.

pmikolajczyk41: fair point, I'll hide it tomorrow

pmikolajczyk41: ~70 new dependencies, some of which are substrate crates (for state-machines, externalities, few pallets etc)

pmikolajczyk41: https://github.com/paritytech/ink/pull/1887

PR #1867, opened 2023-08-03
[E2E alternative backend]: Make tests generic over client

pmikolajczyk41:
Small PR making all the e2e integration tests generic over backend trait.

Note: Since the testcase signature is partly abandoned, the generic bound `Client: E2EBackend` is not necessary (even import is not needed). However, I left it anyway for transparency.

Note: These testcases have to be generic, since `E2EBackend` is not object safe trait (nor are `ChainBackend` and `ContractsBackend`).


pmikolajczyk41:
sure thing - new entry added!

Issue #1471, opened 2022-11-05
Add possibility of making static buffer configurable


cmichi:
We should make the size of the static buffer configurable. This is what https://github.com/paritytech/ink/pull/1279 tried to do. See [here](https://github.com/paritytech/ink/blob/master/ARCHITECTURE.md#communication-with-the-pallet) for an explanation what the static buffer is.

The ideal approach for implementing this would be to add a property to our `Environment` trait:

```diff
 /// The environmental types usable by contracts defined with ink!.
 pub trait Environment {
+    /// TODO
+    const CAPACITY: u8;
+
```

This will require adding the `E: Environment` bound in all kinds of places, so that accessing `E::CAPACITY` becomes possible:

```diff
+/// A static buffer with `E::CAPACITY` kB of capacity.
+pub struct StaticBuffer<E: Environment> {
     /// The static buffer with a total capacity of 16 kB.
-    buffer: [u8; Self::CAPACITY],
+    buffer: [u8; E::CAPACITY],
 }
```

Some necessary modifications:

```diff
impl OnInstance for EnvInstance {
-    fn on_instance<F, R>(f: F) -> R
+    fn on_instance<E, F, R>(f: F) -> R
     where
         F: FnOnce(&mut Self) -> R,
+        E: Environment,
     {
```

```diff
 /// The on-chain environment.
-pub struct EnvInstance {
+pub struct EnvInstance<E: Environment> {
```

There's a ton more places where we'll need to pass the type.

SkymanOne:
@cmichi rust stable does not support generic parameters to be used in const generics. I have to fallback to passing the constant explicitly across the whole codebase:
```rust
pub struct StaticBuffer<const CAPACITY: usize>
```

SkymanOne:
The problem doesn't seem to be trivial because [`generic_const_exprs`](https://github.com/rust-lang/rust/issues/76560) hasn't landed in stable. Might need to use this crate: https://crates.io/crates/generic_singleton as a workaround

SkymanOne:
After digging, what we can do to tackle the issue is to use constant generics for `StaticBuffer`
```diff
+/// A static buffer with `CAPACITY` kB of capacity.
+pub struct StaticBuffer<const CAPACITY: usize> {
     /// The static buffer with a total capacity of 16 kB.
-    buffer: [u8; Self::CAPACITY],
+    buffer: [u8; CAPACITY],
 }
```
[...]

PR #1869, opened 2023-08-07
Modify static buffer size via env var

SkymanOne:
This is a trivial solution addressing #1471 by introducing a build script to `ink_env` crate which read the `STATIC_BUFFER_SIZE` env var and appends the constant to a file. It has been tested to work.

Because the value is baked into the `ink_env` binaries, `cargo clean` must be run every time the buffer size is modified.

**Checklist:**

- [ ] Introduce `.env` files in a contract project and propagate the env values to the core `ink` crates
- [x] Add integration test
- [ ] Persist the size in the metadata

**Follow-ups:**

- Allow `cargo-contract` to pass read contract's local `.env` file and append the env vars to the build command.
- Persist, the static buffer size in the metadata. Should also be done via cargo-contract by reading the supplied env var.


kvinwang:
> This is a trivial solution addressing #1471 by introducing a build script to `ink_env` crate which read the `STATIC_BUFFER_SIZE` env var and appends the constant to a file.

`INK_STATIC_BUFFER_SIZE` would be a better name.

crates/env/build.rs
```diff
@@ -0,0 +1,23 @@
...
+};
+
+fn main() {
+    // Default size of the buffer: 16 kB.
+    let mut size: usize = 1 << 14;
+    // if environmental variable is present we update the size.
+    if let Ok(size_str) = std::env::var("STATIC_BUFFER_SIZE") {
+        if let Ok(new_size) = size_str.parse::<usize>() {
+            size = new_size;
+        }
```
xgreenx, line 15: Maybe it is worth printing an error or throwing it, because the user specified the `STATIC_BUFFER_SIZE`, but the value was incorrect

integration-tests/static-buffer/README.md
```diff
@@ -0,0 +1,18 @@
...
+cargo clean
+STATIC_BUFFER_SIZE=30 cargo test -F e2e-tests
+```
+
+This will configure the buffer to have enough space to instantiate the contract,
+but not enough space to retrieve the caller's address as it is of 32 bytes,
+but we only allocated 30 bytes to the contract.
+
+## Note
+You must run `cargo clean` every time you want to modify the buffer size
```
xgreenx, line 17: It is a sad property of so nice solution=)

We need to highlight that in the documentation

xgreenx (approved):

xgreenx:
Do you want to implement `.env` during this PR or in follow-up PR?

SkymanOne:
> Do you want to implement `.env` during this PR or in follow-up PR?

This is a `cargo-contract` related issue

xgreenx:
But it seems you try to use the same idea with the build script but use the `CARGO_MANIFEST_DIR` env variable that stores the `.env` file.

integration-tests/static-buffer/README.md
xgreenx: Also, we need to notify all projects that provide deployment of the contracts to the network(like Swanky, Patron, etc). Because they should somehow support `STATIC_BUFFER_SIZE` during deployment, and they need to be sure that the binary uses the correct size for the buffer.

SkymanOne:
`CARGO_MANIFEST_DIR` will point to the `ink_env` folder because that's where the build is happening. Since the order of dependencies is `contract -> ink -> ink_env`, then `ink_env` build script will execute first. So, the idea is that we need to read `.env` file from `cargo-contract` and source it globally, so that the var is available during `ink_env` build

integration-tests/static-buffer/README.md
SkymanOne: That's why I am personally in favour of #1872 since it is easier and more idiomatic for ppl to use, i.e. it eliminated any issues associated with the host machine.

cmichi:
My concern is that with the env variable we introduce another means of configuring the ink! environment, besides the trait. And that mean is outside of the usual Rust primitives.

It will be easy to overlook that the (hidden) `.env` file is missing during a build. This file will also need to be uploaded when using third party tools (such as block explorers contract verification or [patron.works](https://patron.works)). 

I'm in favor of merging https://github.com/paritytech/ink/pull/1872. It's also not an ideal solution, but in my opinion given the circumstances the best choice: setting this variable in `Cargo.toml` won't be overlooked or forgotten.

ascjones (approved):
On balance I think this is the lesser of two evils. 

I don't think using an `.env` file is a good idea though. Instead we should take a leaf out of [`cargo`s book](https://doc.rust-lang.org/cargo/reference/config.html#command-line-overrides) and allow it to be configured by (in order of precedence):

- Command line arg e.g. `cargo contract build --static-buffer-size 16384`
- Config in the contracts `Cargo.toml` using either a custom section or the [`[env]` section](https://doc.rust-lang.org/cargo/reference/config.html#env)
- Just using the env var itself if set

>  cargo clean must be run every time the buffer size is modified.

This can be solved in `cargo-contract` by first checking the existing metadata for the existing buffer size, and running `cargo clean` if the new value is different.

ascjones:
Note that a possible alternative implementation of this avoiding the requirement of the `build.rs`file would be to use [`const_env`](https://crates.io/crates/const_env) and do 

```
#[from_env("INK_STATIC_BUFFER_SIZE")]
const BUFFER_SIZE: usize = 1 << 14;
```

SkymanOne:
> On balance I think this is the lesser of two evils.
> 
> I don't think using an `.env` file is a good idea though. Instead we should take a leaf out of [`cargo`s book](https://doc.rust-lang.org/cargo/reference/config.html#command-line-overrides) and allow it to be configured by (in order of precedence):
> 
> * Command line arg e.g. `cargo contract build --static-buffer-size 16384`
> * Config in the contracts `Cargo.toml` using either a custom section or the [`[env]` section](https://doc.rust-lang.org/cargo/reference/config.html#env)
> * Just using the env var itself if set
> 
> > cargo clean must be run every time the buffer size is modified.
> 
> This can be solved in `cargo-contract` by first checking the existing metadata for the existing buffer size, and running `cargo clean` if the new value is different.

Agreed. `cargo-contract` support is essential here.

> Note that a possible alternative implementation of this avoiding the build.rs would be to use [const_env](https://crates.io/crates/const_env) and do

While it looks cleaner, it doesn't gives us much flexibility in terms of error reporting. However, I will test it out and see

upendra-eth:
I am reaching out to seek an update on the status of the issue at hand. Could you kindly confirm if it has been successfully resolved? If so, I would be most grateful for your guidance on the specific changes required in my Ink contract. Additionally, I would like to know whether it is necessary to create a .env file in my contract folder. Your assistance and clarification on these matters would be greatly appreciated.

PR #1872, opened 2023-08-09
Modify static buffer size via features

SkymanOne:
An alternative to the #1869 which addresses #1471 

Instead of allowing developers to specify the static buffer size, we let them choose the size from the given set of features that we offer:

```toml
2GB-buffer = []
1GB-buffer = []
512MB-buffer = []
128MB-buffer = []
16MB-buffer = []
1MB-buffer = []
512kB-buffer = []
128kB-buffer = []
```
This range of values should cover most of the use cases.

The benefit over the other PR is that devs don't need to mess with the environmental variables of `.env` files, and the configuration of the static buffer size is trivially simple eliminating any possibility of human error.

We also avoid using build scripts and messing up with the codegen.


crates/env/Cargo.toml
```diff
@@ -79,3 +79,13 @@ ink-debug = []
 
 # Disable the ink! provided global memory allocator.
 no-allocator = ["ink_allocator/no-allocator"]
+
+# Configurable sizes of the static buffer
+2GB-buffer = []
```
xgreenx, line 84: It says that memory is to huge:
image

crates/env/Cargo.toml
```diff
@@ -79,3 +79,13 @@ ink-debug = []
...
+1GB-buffer = []
+512MB-buffer = []
+128MB-buffer = []
+16MB-buffer = []
+1MB-buffer = []
```
xgreenx, line 89: Starting `1MB` it returns another error:
image

SkymanOne:
Closing in favour of #1869

Issue #1875, opened 2023-08-14
End-to-end tests contracts fails despite correct node setup


0xf333:
## Description:
End-to-end tests (`e2e-tests` feature) for the Ink contracts fail to run even when `substrate-contracts-node` is installed and setup properly.

<br>

---

## Steps to Reproduce:

1. Make sure you have the latest `substrate-contracts-node` and `cargo-contract` installed; if you don't please run:
```
cargo install contracts-node --git https://github.com/paritytech/substrate-contracts-node.git
```
and
```
cargo install cargo-contract --force
```

2. clone this repo
```
git clone https://github.com/0xf333/scenario_for_E2E_test_issue
```

3. Navigate to the `flipper` contract directory:
```
cd scenario_for_E2E_test_issue/issue/flipper
```

4. First, run the following command to build:
```
cargo contract build --release
```

5. And then run this command for the E2E test:
```
cargo test --features e2e-tests
```

<br>

---

## Expected Behavior:
The E2E test should run successfully.

## Actual Behavior:
The E2E test fails regardless of having `substrate-contracts-node` installed. Even if you are running the node in the background and then run the E2E test, it will still fail.

https://github.com/paritytech/ink/assets/100840345/1ab31641-4a8a-4a5d-816f-0e2f6cb2ca3e

<br>

---

## Environment:

`OS`: Debian
`cargo-contract` version **3.0.1**
`ink!` **version**: **4.2.1**
`ink_e2e` version: **4.2.1**
`substrate-contracts-node` version: **0.30.0-72e68577688**
`Rustc` version: **1.71.1** `stable`

.

SkymanOne:
Related to https://github.com/paritytech/cargo-contract/issues/1239#issuecomment-1672202373

ascjones:
I believe the problem here is that the latest `substrate-contracts-node` is *not* compatible with the released `4.2.1` release of `ink_e2e` on crates.io. This is because of changes to the port discovery string.

See https://github.com/paritytech/ink/pull/1876#pullrequestreview-1578317652

ascjones:
This should be fixed in `4.3.0` and also `5.0.0-alpha`

PR #1876, opened 2023-08-14
fix: fix E2E test failures in Ink contracts

0xf333:
## Description
The e2e-tests for ink contracts were failing due to an issue with port parsing. This PR introduces: 
- `tokio::task::spawn_blocking` to offload the task to a separate thread. 
- `std::sync::mpsc` channel to efficiently relay the extracted port information. 
- Continuous reading of process output to prevent any potential blockages.

<br>

---

## What This Enables
You can now reliably execute e2e-tests for ink contracts. 

## Impacts
This pull request addresses issue #1875

<br>

---

## Testing The Implemented Fix
`Environment ==> **Linux**`
1. Clone this repo:
```
git clone https://github.com/0xf333/scenario_for_E2E_test_issue
```

2. Navigate to `/scenario_for_E2E_test_issue/test_PR_fix` and make `substrate-contracts-node` and 
`env_setup.sh` executable
```
chmod +x substrate-contracts-node env_setup.sh
```

3. Navigate to `/scenario_for_E2E_test_issue/test_PR_fix/flipper`:
```
cd flipper
```

4. Compile the contract; you need to do this to pull all the necessary dependencies. 
(Wait for the compilation to finish before proceeding to the next step.)
```
cargo contract build --release
``` 

5. Navigate back to the `test_PR_fix` directory:
```
cd ..
``` 

6. Run the `env_setup.sh` bash script to:
    - Set the **CONTRACTS_NODE** environment variable.
    - Get the path to the crate file where you need to apply this PR fix.
    - Get the path string required to accordingly set `ink_e2e` inside 
    the `Cargo.toml` file in order to test this PR fix.
```
source env_setup.sh
```
*Note:* Make sure to use `source` instead of `./` to ensure the **CONTRACTS_NODE** environment variable is properly set.
    
7. Copy and paste the changes from this PR into the .rs file ( you'll get the file path after running the bash script ). 
Save the change and then go back to the `flipper` directory, and run these commands in the same order:
```
cargo clean
``` 
```
cargo cargo contract build --release
``` 
```
cargo test --features e2e-tests
``` 
`clean` to clean the previous build
`conrtact build --release` to compile the contract and include changes made inside the ink_e2e dependency. 
`test --features e2e-tests` to run the E2E test

Finally, after executing `cargo test --features e2e-tests`, make sure that the compilation completes entirely. If you encounter an interruption at:
```
 [2/4] Post processing wasm file
```
Run the command again. Occasionally in Rust, compilation may prematurely exit during heavy CPU load.

<br>

---

## Result

**Before the fix**

https://github.com/paritytech/ink/assets/100840345/2b642999-edc8-43fd-9474-bf59ab5d1edd

**After the fix**

https://github.com/paritytech/ink/assets/100840345/36569096-a229-45a4-8b2d-272e9bf72999

### DIFF
--- a/crates/e2e/src/node_proc.rs
+++ b/crates/e2e/src/node_proc.rs
@@ -1,4 +1,4 @@
-// Copyright (C) Parity Technologies (UK) Ltd.
+// Copyright 2018-2022 Parity Technologies (UK) Ltd.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -24,11 +24,14 @@ use std::{
         Read,
     },
     process,
+    sync::mpsc,
 };
 use subxt::{
     Config,
     OnlineClient,
 };
+
+use tokio::task;
 
 /// Spawn a local substrate node for testing.
 pub struct TestNodeProcess<R: Config> {
@@ -134,8 +137,12 @@ where
 
         // Wait for RPC port to be logged (it's logged to stderr):
         let stderr = proc.stderr.take().unwrap();
-        let port = find_substrate_port_from_output(stderr);
-        let url = format!("ws://127.0.0.1:{port}");
+        let port_future =
+            tokio::task::spawn_blocking(move || find_substrate_port_from_output(stderr));
+        let port = port_future
+            .await
+            .map_err(|_| "Failed to spawn blocking task".to_string())?;
+        let url = format!("ws://127.0.0.1:{}", port.await);
 
         // Connect to the node with a `subxt` client:
         let client = OnlineClient::from_url(url.clone()).await;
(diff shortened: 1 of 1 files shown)


0xf333:
Hi guys, it's now ready for review.

ascjones (commented):
Looking at your reproduction, in the `issue/flipper/Cargo.toml` you are using the `4.2.1` released version of `ink_e2e` on crates.io. This is *not* compatible with the latest `substrate-contracts-node` release.

I would bet that if you change that to use `ink` master then it will resolve the issue without these changes.

Our CI ensures that the `master` branch can run all E2E tests against the latest `substrate-contracts-node`

0xf333:
> Looking at your reproduction, in the `issue/flipper/Cargo.toml` you are using the `4.2.1` released version of `ink_e2e` on crates.io. This is _not_ compatible with the latest `substrate-contracts-node` release.
> 
> I would bet that if you change that to use `ink` master then it will resolve the issue without these changes.
> 
> Our CI ensures that the `master` branch can run all E2E tests against the latest `substrate-contracts-node`

Hi @ascjones,
Thanks for the feedback, this case is present in ink_e2e 4.2.0 too; the reason I used 4.2.1 is that I thought it would make more sense to solve it in the latest version of ink_e2e rather than in an older one.
Please let me know what you think.

ascjones:
> > Looking at your reproduction, in the `issue/flipper/Cargo.toml` you are using the `4.2.1` released version of `ink_e2e` on crates.io. This is _not_ compatible with the latest `substrate-contracts-node` release.
> > I would bet that if you change that to use `ink` master then it will resolve the issue without these changes.
> > Our CI ensures that the `master` branch can run all E2E tests against the latest `substrate-contracts-node`
> 
> Hi @ascjones, Thanks for the feedback, this case is present in ink_e2e 4.2.0 too; the reason I used 4.2.1 is that I thought it would make more sense to solve it in the latest version of ink_e2e rather than in an older one. Please let me know what you think.

None of the `crates.io` releases `4.2.0`, `4.2.1` or any others are compatible with the latest `substrate-contracts-node`, because they were released *before* the node was updated. They will only work with an older version of `substrate-contracts-node`, and have already been released so can't be changed.

The `ink!` `master` branch is already compatible with the latest `substrate-contracts-node`: you can prove it by navigating to any `integration_tests` example and running the e2e tests (this is also checked by a CI step).

If you want to use e2e tests with any of the released `ink_e2e` versions, then use an older release of `substrate-contracts-node`

If you want to use the latest `substrate-contracts-node` (which you are doing) then use the `ink` `master` branch via a `git` dependency.

Thank you for the detailed report and reproduction steps, but I don't think these changes are required.

0xf333:
> None of the `crates.io` releases `4.2.0`, `4.2.1`, or any others are compatible with the latest `substrate-contracts-node`, because they were released _before_ the node was updated. They will only work with an older version of `substrate-contracts-node`, and have already been released so can't be changed.
> 
> The `ink!` `master` branch is already compatible with the latest `substrate-contracts-node`: you can prove it by navigating to any `integration_tests` example and running the e2e tests (this is also checked by a CI step).
> 
> If you want to use e2e tests with any of the released `ink_e2e` versions, then use an older release of `substrate-contracts-node`
> 
> If you want to use the latest `substrate-contracts-node` (which you are doing) then use the `ink` `master` branch via a `git` dependency.
> 
> Thank you for the detailed report and reproduction steps, but I don't think these changes are required.

Hi @ascjones,
Thanks for the feedback.

ascjones:
@0xf333 we are considering a `4.x` release with a backport to fix this issue. Stay tuned.

0xf333:
> @0xf333 we are considering a `4.x` release with a backport to fix this issue. Stay tuned.

Hi @ascjones , I sent you an email, please check and let me know.

Issue #1826, opened 2023-06-24
`DelegateCall` does not mutate caller's layoutful storage


SkymanOne:
Similar background as in #1825.

This time, if the storage is layoutful (it contains at least one non-`Lazy` or non-`Mapping` filed with), the delegated callee does not mutate the caller's storage.

# Detailed description

If the storage is layoutful, delegate call indeed transfers the context of the caller's storage to the callee. So all the mutation within callee's code appear to be valid. However, when the call is return to the caller, this mutations are not reflected in the caller's storage. 

You can see two reproducers based on the ones from #1825 
* https://github.com/paritytech/ink/blob/8e183d0c2d8c03a9fec47c1956dec052eb723e12/integration-tests/key-reproducer/lib.rs#L7 
* https://github.com/paritytech/ink/blob/8e183d0c2d8c03a9fec47c1956dec052eb723e12/integration-tests/delegatecall-bug/lib.rs#L14

You can see from e2e tests that `value: bool` field does not get updated from delegate call. Furthermore, the `Mapping` field with a manually set key does not also get updated via delegate call.

The only way I got it to work is to wrap primitive types into `Lazy` and specify `ManualKey` for every single field, that way the delegate call behaves as expected.

This issue has not been raised before because the only integration test we have has only tested for reading the value via delegate call: https://github.com/paritytech/ink/blob/97f90ca6d032ecea1b0ff173ddde5ba0c18e1b30/integration-tests/lang-err-integration-tests/call-builder-delegate/lib.rs#L122

# Potential Solution

With manual keys specified, the behaviour is correct. So I don't think we need to raise "code red". However, the way it works in solidity and our lack of documentation can introduce a footgun for new-coming devs.
Therefore, I propose two solutions:
* Enforce the use of manual keys when using delegate call at the type checking, and document this (fast)
* Investigate the causes of this behaviour and allow the mutation of layoutful storages via `delegate_call()` (slow)

deuszx:
In your second test you're still calling `flip_delegate` https://github.com/paritytech/ink/blob/8e183d0c2d8c03a9fec47c1956dec052eb723e12/integration-tests/delegatecall-bug/lib.rs#L149 even though it seems you intended to call `add_value_delegate`.

SkymanOne:
Yes, good spot. But it doesn't change the point of this issue.

deuszx:
[...]

PR #1889, opened 2023-08-24
Upgradeable contracts

SkymanOne:
Addresses #1826 

Turns out there are a few hiccups when using `delegate_call` that a develop needs to know. This PR re-introduces the examples and documentation for the upgradeable contracts.


integration-tests/upgradeable-contracts/delegator/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+                // See https://substrate.stackexchange.com/questions/3336/i-found-set-allow-reentry-may-have-some-problems/3352#3352
+                .call_flags(CallFlags::default().set_tail_call(true))
+                .exec_input(ExecutionInput::new(Selector::new(selector)))
+                .returns::<()>()
+                .try_invoke();
+        }
+
+        /// Adds entry to `addresses` using delegate call.
+        /// Note that we don't need `set_tail_call(true)` flag
+        /// because `Mapping` updates the storage instantly on-demand.
```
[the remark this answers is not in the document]
xgreenx: No, but maybe we want to add something like this=)

integration-tests/upgradeable-contracts/README.md
```diff
@@ -0,0 +1,36 @@
...
+ink! provides an ability to replace the code under the given contract's address.
+This is exactly what `set_code_hash()` function does.
+
+However, developers needs to be mindful of storage compatibility.
+You can read more about storage compatibility on [use.ink](https://use.ink/basics/upgradeable-contracts#replacing-contract-code-with-set_code_hash)
+
+## [Delegator](delegator/)
+
+Delegator patter is based around a low level cross contract call function `delegate_call`.
+It allows a contract to delegate its execution to some on-chain uploaded code.
```
deuszx, line 18: I think it's a mistake to present these as two separate ways of upgrading storage - `set_code_hash` is almost always used together with `delegate_call` to migrate the contract's storage after upgrading its code.

integration-tests/upgradeable-contracts/README.md
```diff
@@ -0,0 +1,36 @@
...
+First of all, as demonstrated in the example, if the delegated code intends to mutate the caller's storage,
+a developer needs to be mindful. If the delegated code modifies layout-full storage
+(i.e. it contains at least non-`Lazy`, non-`Mapping` field), the `.set_tail_call(true)` flag of `CallFlags` needs to be specified and the storage layouts must match.
+This is due to the way ink! execution call stack is operated
+(see [Stack Exchange Answer](https://substrate.stackexchange.com/a/3352/3098) for more explanation).
+
+If the delegated code only modifies `Lazy` or `Mapping` field, the keys must be identical and `.set_tail_call(true)` is optional.
+This is because `Lazy` and `Mapping` interact with the storage directly instead of loading and flushing storage states.
+
+If your storage is completely layoutless (it only contains `Lazy` and `Mapping` fields), the order of fields and layout do not need to match for the same reason as mentioned above.
```
deuszx, line 35: I'd avoid usage of layoutfull and layoutless since they don't have well-defined meaning. Storage under `Lazy` and `Mapping` also has a layout, it's just different.

SkymanOne: Well, it is hard to explain in terms of Packed and Non-Packed because storage can be Non-Packed by still layoutful. Here the "layoutfulness" plays an important role.

integration-tests/upgradeable-contracts/README.md
deuszx: `delegate_call` doesn't upgrade the contract - i.e. it doesn't change it's logic (one could argue that if all calls delegate then it is but I'd say it's a proxy then) it can only change its data. Hence the only way to upgrade the contract is to:
1. forward to a different contract - proxy pattern
2. replace the code - `set_code_hash`. 


Also, you can only delegate if that logic has been there since the beginning - i.e. since the moment the contract was uploaded. I'd argue that the contract isn't _upgraded_ .

I think the nomenclature is important here b/c if used loosely it will only add to devs confusion.

SkymanOne: `delegate_call` allows to delegate an execution of logic to another on-chain. If used correctly, you can introduce the upgradeability functionality where delegated code hash can be replaced. The reason why it belongs to upgradeability is because that's how it is implemented in the EVM world - via proxies. 

`set_code_hash` is the "Polkadot" way of upgradability. 

Nonetheless, these nuances should be elaborated in the documentation. The examples just provide a concrete illustration of how these "blocks" can be used individually.

PR #1891, opened 2023-08-29
Drink backend: support runtime call

pmikolajczyk41:
Since https://github.com/Cardinal-Cryptography/drink/issues/36 is done, we can finally support arbitrary runtime calls in e2e tests. However, the solution is kinda tricky.

## Problem

In drink! we have strongly typed runtime calls. This is because we keep the runtime directly, in-memory, and thus we have access to the `RuntimeCall` enum family. And thus there's no need of dynamic dispatch.

In standard e2e backend (i.e. node+subxt) we must have only dynamic dispatch. This is because we support running a node with an arbitrary runtime, and therefore we cannot statically generate runtime types, including `RuntimeCall` enum family.

But in order to support using both backends with the same test code, we have to provide a common interface...

## Solution

We can't make `ChainBackend::runtime_call` statically typed, since for `subxt_client::Client` there is no way of providing correct types. Hence, we have to leave the API as it is (i.e. dynamic dispatch with pallet and call as strings and call data as `subxt::Value` wrappers).

For the drink! implementation we do following mumbo-jumbo:
 - we get metadata from the runtime
 - using subxt we create a valid call object using arguments and metadata
 - we encode the call object
 - we decode it to the `RuntimeCall` object
:man_shrugging: 

## Note

Since `drink::MinimalRuntime` and `substrate-contracts-node`'s runtime use different addressing, the testcases that relate to runtime calling are compatible with only one backend. However, notice that this is the incompatibility of used runtimes, not backends, and thus it's okay.

### DIFF
--- a/crates/e2e/src/drink_client.rs
+++ b/crates/e2e/src/drink_client.rs
@@ -132,21 +132,31 @@ impl<AccountId: AsRef<[u8; 32]> + Send, Hash> ChainBackend for Client<AccountId,
         call_name: &'a str,
         call_data: Vec<Value>,
     ) -> Result<Self::EventLog, Self::Error> {
+        // Since in general, `ChainBackend::runtime_call` must be dynamic, we have to
+        // perform some translation here in order to invoke strongly-typed drink!
+        // API.
+
+        // Get metadata of the drink! runtime, so that we can encode the call object.
+        // Panic on error - metadata of the static im-memory runtime should always be
+        // available.
         let raw_metadata: Vec<u8> = MinimalRuntime::metadata().into();
         let metadata = subxt_metadata::Metadata::decode(&mut raw_metadata.as_slice())
             .expect("Failed to decode metadata");
 
+        // Encode the call object.
         let call = subxt::dynamic::tx(pallet_name, call_name, call_data);
-        let encoded_call = call
-            .encode_call_data(&metadata.into())
-            .expect("Failed to encode call data");
+        let encoded_call = call.encode_call_data(&metadata.into()).map_err(|_| ())?;
 
+        // Decode the call object.
+        // Panic on error - we just encoded a validated call object, so it should be
+        // decodable.
+        let decoded_call =
+            RuntimeCall::<MinimalRuntime>::decode(&mut encoded_call.as_slice())
+                .expect("Failed to decode runtime call");
+
+        // Execute the call.
         self.sandbox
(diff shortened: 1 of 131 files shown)


crates/e2e/src/drink_client.rs
```diff
@@ -121,12 +127,39 @@ impl<AccountId: AsRef<[u8; 32]> + Send, Hash> ChainBackend for Client<AccountId,
...
     ) -> Result<Self::EventLog, Self::Error> {
-        todo!("https://github.com/Cardinal-Cryptography/drink/issues/36")
+        // Since in general, `ChainBackend::runtime_call` must be dynamic, we have to
+        // perform some translation here in order to invoke strongly-typed drink!
+        // API.
+
+        // Get metadata of the drink! runtime, so that we can encode the call object.
+        // Panic on error - metadata of the static im-memory runtime should always be
+        // available.
+        let raw_metadata: Vec<u8> = MinimalRuntime::metadata().into();
```
deuszx, line 142: What if I want to write an e2e test for a runtime that is different than `MinimalRuntime`? For example Aleph Zero liminal one? I won't be able to use the drink! as backend?

pmikolajczyk41: I will add this feature soon

pmikolajczyk41: @deuszx https://github.com/paritytech/ink/pull/1892

SkymanOne (approved):
You could potentially do some hacky type inference here, but current solutions looks good enough to me.

ascjones (approved):
> In standard e2e backend (i.e. node+subxt) we must have only dynamic dispatch. This is because we support running a node with an arbitrary runtime, and therefore we cannot statically generate runtime types, including RuntimeCall enum family.

Yes and no, I think it would be beneficial to define a minimal set of hardcodeed common calls that are useful and supported by both `substrate-contracts-node` and `drink`. Balances::transfer etc... (in fact we already do this for the `pallet_contracts` calls `instantiate`, `call` etc.

Of course people can still target their own custom node and then if they don't satisfy that API then calls simply return an error "Not implemented for target node".

Either way this dynamic API is required and very useful.

pmikolajczyk41:
> Either way this dynamic API is required and very useful.

Sure it is, I'm not attacking it. Just saying that it is not (and won't be) natively supported in drink

PR #1892, opened 2023-08-30
Drink backend: allow for arbitrary runtime

pmikolajczyk41:
Make `drink_client::Client` generic over runtime and allow choosing runtime in the e2e test macro.

### DIFF
--- a/crates/e2e/src/drink_client.rs
+++ b/crates/e2e/src/drink_client.rs
@@ -57,7 +57,10 @@ pub struct Client<AccountId, Hash, Runtime: RuntimeT> {
 // While it is not necessary true that `Client` is `Send`, it will not be used in a way
 // that would violate this bound. In particular, all `Client` instances will be operating
 // synchronously.
-unsafe impl<AccountId, Hash, Runtime: RuntimeT> Send for Client<AccountId, Hash, Runtime> {}
+unsafe impl<AccountId, Hash, Runtime: RuntimeT> Send
+    for Client<AccountId, Hash, Runtime>
+{
+}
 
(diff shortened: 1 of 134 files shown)


integration-tests/e2e-runtime-only-backend/lib.rs
```diff
@@ -73,5 +73,22 @@ pub mod flipper {
 
             Ok(())
         }
+
+        /// Just instantiate a contract using non-default runtime.
+        #[ink_e2e::test(backend = "runtime-only", runtime = drink::runtime::MinimalRuntime)]
```
SkymanOne, line 78: Can't you do?
```suggestion
        #[ink_e2e::test(backend = "runtime-only", runtime = ::ink_e2e::MinimalRuntime)]
```

SkymanOne (approved):

integration-tests/e2e-runtime-only-backend/lib.rs
pmikolajczyk41: oh sure, that's why I reexported it (but somehow I forgot about it here...); fixed

crates/e2e/src/drink_client.rs
```diff
@@ -221,10 +227,12 @@ impl<
             }
         };
 
+        let code_hash_raw: [u8; 32] = result.code_hash.as_ref().try_into().expect("");
```
ascjones, line 230: Empty string in `expect("")`

ascjones (approved):

crates/e2e/src/drink_client.rs
pmikolajczyk41: must have missed it; fixed

PR #1893, opened 2023-08-31
Script to check if a `integration-test` manifest is a contract

ascjones:
I was using this script locally to check all examples e.g.
```
CARGO_TARGET_DIR=/tmp/integration-tests-target/ find ./integration-tests/ -name "Cargo.toml" \
       -exec scripts/is_contract.sh {} \; \
       -exec cargo contract build --release --manifest-path {} \;^
```
I will try and update some of the CI jobs to use it instead of the current filtering we do, to make it more concise and robust.


SkymanOne (approved):
Yes. That's what we need

PR #1897, opened 2023-09-07
Fail when decoding from storage and not all bytes consumed

### LINKED ISSUE #1804: Consider using `decode_all` for `get_contract_storage`

ascjones:
Closes #1804.

:warning: **BREAKING CHANGE** :warning: 

If a contract previously relied on successful decoding which does not consume all bytes, then recompiling with a version of `ink!` which includes this change will cause that contract to trap at runtime when attempting to decode.

See the new `fn decode_all` for `Storable` impls. It just copies the `DecodeAll` impl from `parity-scale-codec` which is very simple.

## todo

- [x] check that this still allows for contract migration (see https://github.com/paritytech/ink/pull/1909)
- [x] look at integration-test failures
- [x] test for `take_contract_storage`
- [x] measure changes in gas usage and contract sizes.


crates/storage/traits/src/storage.rs
```diff
@@ -44,6 +44,30 @@ where
...
+    ///
+    /// If not all data is consumed, an error is returned.
+    fn decode_all(input: &mut &[u8]) -> Result<Self, scale::Error>;
+}
+
+impl<P> StorableDecodeAll for P
+where
+    P: Storable,
+{
+    fn decode_all(input: &mut &[u8]) -> Result<Self, scale::Error> {
```
xgreenx, line 60: Why do we want a separate `StorableDecodeAll` trait with blanket implementation when we can use this code directly in the `get_contract_storage` and `take_contract_storage`? (If you want to reuse the code, then you can have a procedure method)

ascjones: Yes, that might be a better solution instead of the extra trait :+1:

ascjones: Done

crates/engine/src/ext.rs
```diff
@@ -241,32 +241,26 @@ impl Engine {
     }
 
     /// Returns the decoded contract storage at the key if any.
-    pub fn get_storage(&mut self, key: &[u8], output: &mut &mut [u8]) -> Result {
+    pub fn get_storage(&mut self, key: &[u8]) -> core::result::Result<&[u8], Error> {
```
SkymanOne, line 244: Can't you just import the full path?

crates/storage/traits/src/storage.rs
```diff
@@ -44,6 +44,19 @@ where
...
+/// Decode and consume all of the given input data.
+///
+/// If not all data is consumed, an error is returned.
+pub fn decode_all<T: Storable>(input: &mut &[u8]) -> Result<T, scale::Error> {
+    let res = <T as Storable>::decode(input)?;
+
+    if input.is_empty() {
+        Ok(res)
+    } else {
+        Err("Input buffer has still data left after decoding!".into())
```
SkymanOne, line 56: Perhaps would be useful to give the length of the remaining bytes left in a buffer for debugging purposes

SkymanOne (approved):

crates/engine/src/ext.rs
ascjones: There was a type alias here `type Result = core::result::Result<(), Error>;`. However I've just removed that and replaced its usages with the explicit type.

crates/storage/traits/src/storage.rs
ascjones: I'm not sure it is worth bringing in all the string formatting machinery in for that...we need to be mindful of code size. 

It's possible to debug this without this information by fetching the data at the given key and attempting to decode it into the respective type.

paritytech-cicd-pr:

These are the results when building the `integration-tests/*` contracts from this branch with `cargo-contract 4.0.0-alpha-acc02b1` and comparing them to ink! `master`:

| Contract                                                | Upstream Size (kB) | PR Size (kB) | Diff (kB) | Diff (%) | Change                     |
| ------------------------------------------------------- | ------------------ | ------------ | --------- | -------- | -------------------------- |
| basic-contract-caller                                   | 2.998              | 2.998        | 0         | 0        | :heavy_minus_sign:         |
| basic-contract-caller/other-contract                    | 1.331              | 1.338        | 0.007     | 0.52592  | :chart_with_upwards_trend: |
| call-builder-return-value                               | 8.734              | 8.74         | 0.006     | 0.068697 | :chart_with_upwards_trend: |
| call-runtime                                            | 1.77               | 1.774        | 0.004     | 0.225989 | :chart_with_upwards_trend: |
| conditional-compilation                                 | 1.204              | 1.21         | 0.006     | 0.498339 | :chart_with_upwards_trend: |
| contract-terminate                                      | 1.087              | 1.092        | 0.005     | 0.459982 | :chart_with_upwards_trend: |
| contract-transfer                                       | 1.446              | 1.449        | 0.003     | 0.207469 | :chart_with_upwards_trend: |
| custom-allocator                                        | 7.39               | 7.429        | 0.039     | 0.52774  | :chart_with_upwards_trend: |
| dns                                                     | 7.313              | 7.313        | 0         | 0        | :heavy_minus_sign:         |
| e2e-call-runtime                                        | 1.057              | 1.062        | 0.005     | 0.473037 | :chart_with_upwards_trend: |
| e2e-runtime-only-backend                                | 1.633              | 1.64         | 0.007     | 0.428659 | :chart_with_upwards_trend: |
| erc1155                                                 | 14.107             | 14.125       | 0.018     | 0.127596 | :chart_with_upwards_trend: |
| erc20                                                   | 6.572              | 6.729        | 0.157     | 2.38892  | :chart_with_upwards_trend: |
| erc721                                                  | 9.547              | 9.716        | 0.169     | 1.77019  | :chart_with_upwards_trend: |
| events                                                  | 4.816              | 4.823        | 0.007     | 0.145349 | :chart_with_upwards_trend: |
| flipper                                                 | 1.387              | 1.394        | 0.007     | 0.504686 | :chart_with_upwards_trend: |
| incrementer                                             | 1.217              | 1.223        | 0.006     | 0.493016 | :chart_with_upwards_trend: |
| lang-err-integration-tests/call-builder-delegate        | 2.321              | 2.327        | 0.006     | 0.258509 | :chart_with_upwards_trend: |
| lang-err-integration-tests/call-builder                 | 4.875              | 4.88         | 0.005     | 0.102564 | :chart_with_upwards_trend: |
| lang-err-integration-tests/constructors-return-value    | 1.772              | 1.783        | 0.011     | 0.620767 | :chart_with_upwards_trend: |
| lang-err-integration-tests/contract-ref                 | 4.363              | 4.363        | 0         | 0        | :heavy_minus_sign:         |
| lang-err-integration-tests/integration-flipper          | 1.565              | 1.572        | 0.007     | 0.447284 | :chart_with_upwards_trend: |
| mapping-integration-tests                               | 2.975              | 3.216        | 0.241     | 8.10084  | :chart_with_upwards_trend: |
| mother                                                  | 9.499              | 9.555        | 0.056     | 0.589536 | :chart_with_upwards_trend: |
| multi-contract-caller                                   | 5.977              | 5.993        | 0.016     | 0.267693 | :chart_with_upwards_trend: |
| multi-contract-caller/accumulator                       | 1.09               | 1.096        | 0.006     | 0.550459 | :chart_with_upwards_trend: |
| multi-contract-caller/adder                             | 1.672              | 1.672        | 0         | 0        | :heavy_minus_sign:         |
| multi-contract-caller/subber                            | 1.693              | 1.693        | 0         | 0        | :heavy_minus_sign:         |
| multisig                                                | 21.406             | 21.628       | 0.222     | 1.03709  | :chart_with_upwards_trend: |
| payment-channel                                         | 5.544              | 5.559        | 0.015     | 0.270563 | :chart_with_upwards_trend: |
| sr25519-verification                                    | 0.865              | 0.87         | 0.005     | 0.578035 | :chart_with_upwards_trend: |
| static-buffer                                           | 1.411              | 1.418        | 0.007     | 0.496102 | :chart_with_upwards_trend: |
| trait-dyn-cross-contract-calls                          | 2.491              | 2.491        | 0         | 0        | :heavy_minus_sign:         |
| trait-dyn-cross-contract-calls/contracts/incrementer    | 1.3                | 1.306        | 0.006     | 0.461538 | :chart_with_upwards_trend: |
| trait-erc20                                             | 6.956              | 7.113        | 0.157     | 2.25704  | :chart_with_upwards_trend: |
| trait-flipper                                           | 1.204              | 1.21         | 0.006     | 0.498339 | :chart_with_upwards_trend: |
| trait-incrementer                                       | 1.365              | 1.371        | 0.006     | 0.43956  | :chart_with_upwards_trend: |
| upgradeable-contracts/delegator                         | 2.902              | 2.914        | 0.012     | 0.413508 | :chart_with_upwards_trend: |
| upgradeable-contracts/delegator/delegatee               | 1.368              | 1.374        | 0.006     | 0.438596 | :chart_with_upwards_trend: |
| upgradeable-contracts/set-code-hash                     | 1.46               | 1.466        | 0.006     | 0.410959 | :chart_with_upwards_trend: |
| upgradeable-contracts/set-code-hash/updated-incrementer | 1.439              | 1.445        | 0.006     | 0.416956 | :chart_with_upwards_trend: |
| wildcard-selector                                       | 2.619              | 2.625        | 0.006     | 0.229095 | :chart_with_upwards_trend: |

[Link to the run](https://gitlab.parity.io/parity/mirrors/ink/-/pipelines/404073) | Last update: Mon Oct 23 16:34:45 CEST 2023

smiasojed (approved):

PR #1907, opened 2023-09-12
Update `schnorrkel` to `0.11.2`

ascjones:
https://crates.io/crates/schnorrkel/0.11.1 has been yanked, initially I downgraded to `0.11.0`, but since I opened this PR a new version `0.11.2` has been released so have upgraded to that :see_no_evil:


SkymanOne (approved):
Should we keep it `0.11` for convenience?

ascjones:
> Should we keep it `0.11` for convenience?

It's the same as `0.11.0`, it means >= 0.11 && < 0.12.

I'm okay with that format but @dependabot always comes along and adds the patch version anyway :shrug:

Issue #1388, opened 2022-09-07
Proposal to initialize the contract on demand


xgreenx:
With the introduction of `delegate_call`, we can have situations when one contract `X` uses another contract `Y` as a blueprint. The blueprint may have totally another storage layout. Then the `X` do the first `delegate_call` to `Y`, `Y` can't load the contract's storage from the `X`'s storage because it is not initialized(`Y`'s constructor was not called).

It creates a situation when we need support initializing the storage during the execution of the `#[ink(message)]`. The user can always wrap his storage into another structure and implement the `Storable` trait for this struct that supports initialization on demand. But we can add native support of it into the ink!.

The proposal is to introduce the `OnCallInitialize` trait.

```rust
/// The contract can implement that trait to support initialization on the runtime
 /// if it is unable to pull from the storage.
 ///
 /// It can be in several cases:
 /// - The contract doesn't have a constructor. That initializer can be an alternative for the constructor.
 /// - The constructor was not called due to upgradeability, `Proxy` or `Diamond` pattern.
 /// - The storage was moved or corrupted.
 ///
 /// If the trait is not implemented, the storage behavior is the default.
 /// It should be first initialized by the constructor.
 pub trait OnCallInitializer: Default {
     /// `Default::default` creates the instance of the contract.
     /// After the `initialize` method is called on that instance.
     /// The developer can do everything that he wants during initialization or do nothing.
     fn initialize(&mut self);
 }
```

If the contract implements `Default` and `OnCallInitialize` traits, the ink! will try to load the contract from the storage, and in the case of fail, it will initialize it.

```rust
 impl<T: OnCallInitializer + Storable> PullOrInit<T> {
     #[allow(dead_code)]
     pub fn pull_or_init(key: &Key) -> T {
         let maybe_instance = ink_env::get_contract_storage::<Key, T>(key);
         match maybe_instance {
             Ok(None) | Err(_) => {
                 let mut instance = Default::default();
                 <T as OnCallInitializer>::initialize(&mut instance);
                 instance
             }
             Ok(Some(storage)) => storage,
         }
     }
 }
```

This idea was implemented in https://github.com/paritytech/ink/pull/1331. But we decided to discuss it separately because it is an important API change. After implementation "initialization on demand" we need to return back `delegate-call` example, [removed in the mentioned PR](https://github.com/paritytech/ink/pull/1331#discussion_r959861595).

ascjones:
Some downsides of this are 
[...]

PR #1909, opened 2023-09-13
Add `integration-test` for possible migration pattern

ascjones:
Here I add an example that performs a storage migration with the following flow:

1. Upload a `migration` contract with a message `migrate` which performs the storage migration.
2. Set code hash to the `migration` contract.
3. Upload a new version of the `incrementer` contract.
4. Call `migrate` on the `migration` contract, passing the code hash of the new updated incrementer contract from `3.` This must happen as a single message, because following the storage migration, the contract would not be able to be called again, since it will fail to load the migrated storage.

I imagine that a similar pattern could be applied using `delegate_call` instead of `set_code_hash`?

The `migrate` message looks like:

```rust
#[ink(message)]
pub fn migrate(&self, inc_by: u8, code_hash: Hash) {
    let incrementer_new = IncrementerNew {
        count: self.count as u64,
        inc_by,
    };
    const STORAGE_KEY: u32 = 0x00000000;
    ink::env::set_contract_storage(&STORAGE_KEY, &incrementer_new);

    ink::env::set_code_hash::<<Self as ink::env::ContractEnv>::Env>(&code_hash)
        .unwrap_or_else(|err| {
            panic!("Failed to `set_code_hash` to {code_hash:?} due to {err:?}")
        })
}
```
Note the use of `&self` instead than `&mut self`, ensuring that the storage will not be written to automatically following the invocation.

I have added this to demonstrate how we might perform a migration as it stands, without having to add any extra hooks in the language, as mentioned here https://github.com/paritytech/ink/issues/1388#issuecomment-1259595641.

Also to demonstrate that it would still be possible to do a migration without using `_reserved: Option<()>` trick for schema evolution as suggested by @athei in relation to https://github.com/paritytech/ink/pull/1897


SkymanOne:
My personal take on the migration pattern:

While this approach works, it is only feasible for trivial storage migrations. As shown in the example, we are migrating a simple, Packed, layoutful storage with 1-2 primitive fields. In production contracts, storage is much more complex, it can have multiple custom data types, Mappings, Lazy fields. Migrating this type of storage using raw operations on the storage cells is tedious and error-prone. 

While storage migration is an "advanced" concept, we should still discourage developers from raw storage operations, and leave it for use only for some niche use cases, similar to how solidity devs are not advised to use EVM opcodes on a regular basis. 
The whole contract gets bricked if a developer gets at least one storage cell migration wrong. It is also extremely hard to audit this kind of storage migration.

What I am proposing instead is to look into #1388 and define a migration API for the developers to work with the concrete storage layouts instead. The concept is very similar to the database migration in ORM libraries like this: https://github.com/paritytech/ink/issues/1388

While this obviously more work, I strongly believe, that if we "market" the "native" contract's upgradeability, we should also provide a safe workflow with it to lower the entry point of using the feature.

paritytech-cicd-pr:

These are the results when building the `integration-tests/*` contracts from this branch with `cargo-contract 4.0.0-alpha-e8c2cfc` and comparing them to ink! `master`:

| Contract                                                | Upstream Size (kB) | PR Size (kB) | Diff (kB) | Diff (%)   | Change                       |
| ------------------------------------------------------- | ------------------ | ------------ | --------- | ---------- | ---------------------------- |
| basic-contract-caller                                   | 2.967              | 2.967        | 0         | 0          | :heavy_minus_sign:           |
| basic-contract-caller/other-contract                    | 1.337              | 1.337        | 0         | 0          | :heavy_minus_sign:           |
| call-builder-return-value                               | 8.688              | 8.688        | 0         | 0          | :heavy_minus_sign:           |
| call-runtime                                            | 1.769              | 1.769        | 0         | 0          | :heavy_minus_sign:           |
| conditional-compilation                                 | 1.209              | 1.209        | 0         | 0          | :heavy_minus_sign:           |
| contract-storage                                        | 7.124              | 7.124        | 0         | 0          | :heavy_minus_sign:           |
| contract-terminate                                      | 1.092              | 1.092        | 0         | 0          | :heavy_minus_sign:           |
| contract-transfer                                       | 1.444              | 1.444        | 0         | 0          | :heavy_minus_sign:           |
| custom-allocator                                        | 7.428              | 7.428        | 0         | 0          | :heavy_minus_sign:           |
| dns                                                     | 7.249              | 7.249        | 0         | 0          | :heavy_minus_sign:           |
| e2e-call-runtime                                        | 1.058              | 1.058        | 0         | 0          | :heavy_minus_sign:           |
| e2e-runtime-only-backend                                | 1.635              | 1.635        | 0         | 0          | :heavy_minus_sign:           |
| erc1155                                                 | 14.026             | 14.02        | -0.006    | -0.0427777 | :chart_with_downwards_trend: |
| erc20                                                   | 6.687              | 6.687        | 0         | 0          | :heavy_minus_sign:           |
| erc721                                                  | 9.64               | 9.64         | 0         | 0          | :heavy_minus_sign:           |
| events                                                  | 4.763              | 4.763        | 0         | 0          | :heavy_minus_sign:           |
| flipper                                                 | 1.393              | 1.393        | 0         | 0          | :heavy_minus_sign:           |
| incrementer                                             | 1.221              | 1.221        | 0         | 0          | :heavy_minus_sign:           |
| lang-err-integration-tests/call-builder-delegate        | 2.317              | 2.317        | 0         | 0          | :heavy_minus_sign:           |
| lang-err-integration-tests/call-builder                 | 4.847              | 4.847        | 0         | 0          | :heavy_minus_sign:           |
| lang-err-integration-tests/constructors-return-value    | 1.773              | 1.773        | 0         | 0          | :heavy_minus_sign:           |
| lang-err-integration-tests/contract-ref                 | 4.328              | 4.328        | 0         | 0          | :heavy_minus_sign:           |
| lang-err-integration-tests/integration-flipper          | 1.571              | 1.571        | 0         | 0          | :heavy_minus_sign:           |
| mapping-integration-tests                               | 3.203              | 3.203        | 0         | 0          | :heavy_minus_sign:           |
| mother                                                  | 9.508              | 9.508        | 0         | 0          | :heavy_minus_sign:           |
| multi-contract-caller                                   | 5.924              | 5.924        | 0         | 0          | :heavy_minus_sign:           |
| multi-contract-caller/accumulator                       | 1.095              | 1.095        | 0         | 0          | :heavy_minus_sign:           |
| multi-contract-caller/adder                             | 1.669              | 1.669        | 0         | 0          | :heavy_minus_sign:           |
| multi-contract-caller/subber                            | 1.689              | 1.689        | 0         | 0          | :heavy_minus_sign:           |
| multisig                                                | 21.471             | 21.471       | 0         | 0          | :heavy_minus_sign:           |
| payment-channel                                         | 5.501              | 5.501        | 0         | 0          | :heavy_minus_sign:           |
| sr25519-verification                                    | 0.865              | 0.865        | 0         | 0          | :heavy_minus_sign:           |
| static-buffer                                           | 1.405              | 1.405        | 0         | 0          | :heavy_minus_sign:           |
| trait-dyn-cross-contract-calls                          | 2.466              | 2.466        | 0         | 0          | :heavy_minus_sign:           |
| trait-dyn-cross-contract-calls/contracts/incrementer    | 1.305              | 1.305        | 0         | 0          | :heavy_minus_sign:           |
| trait-erc20                                             | 7.063              | 7.063        | 0         | 0          | :heavy_minus_sign:           |
| trait-flipper                                           | 1.209              | 1.209        | 0         | 0          | :heavy_minus_sign:           |
| trait-incrementer                                       | 1.37               | 1.37         | 0         | 0          | :heavy_minus_sign:           |
| upgradeable-contracts/delegator                         | 2.908              | 2.908        | 0         | 0          | :heavy_minus_sign:           |
| upgradeable-contracts/delegator/delegatee               | 1.369              | 1.369        | 0         | 0          | :heavy_minus_sign:           |
| upgradeable-contracts/set-code-hash                     | 1.464              | 1.464        | 0         | 0          | :heavy_minus_sign:           |
| upgradeable-contracts/set-code-hash/updated-incrementer | 1.443              | 1.443        | 0         | 0          | :heavy_minus_sign:           |
| wildcard-selector                                       | 2.622              | 2.622        | 0         | 0          | :heavy_minus_sign:           |

[Link to the run](https://gitlab.parity.io/parity/mirrors/ink/-/pipelines/405387) | Last update: Thu Oct 26 11:34:47 CEST 2023

ascjones:
See IIP-3 https://github.com/paritytech/ink/issues/1981 for a suggestion to make this better, specifically to allow multi-step data migration and contract upgrades, without the risk of bricking the contract.

integration-tests/upgradeable-contracts/set-code-hash-migration/lib.rs
```diff
@@ -0,0 +1,69 @@
+#![cfg_attr(not(feature = "std"), no_std, no_main)]
+
+//! Demonstrates how to use [`set_code_hash`](https://docs.rs/ink_env/latest/ink_env/fn.set_code_hash.html)
+//! to swap out the `code_hash` of an on-chain contract.
+//!
+//! We will swap the code of our `Incrementer` contract with that of the an `Incrementer`
```
cmichi, line 6:
```suggestion
//! We will swap the code of our `Incrementer` contract with that of the `Incrementer`
```

integration-tests/upgradeable-contracts/set-code-hash-migration/migration/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+        /// In a production contract you would do some authorization here!
+        ///
+        /// # Note
+        ///
+        /// This function necessarily accepts a `&self` instead of a `&mut self` because
+        /// we are modifying storage directly for the migration. Using `&mut self`
+        /// would overwrite our migration changes with the contents of the
+        /// original `Incrementer`.
+        #[ink(message)]
+        pub fn migrate(&self, inc_by: u8, code_hash: Hash) {
```
cmichi, line 50:
```suggestion
        pub fn migrate(&mut self, inc_by: u8, code_hash: Hash) {
```

Strange that it seems to work without `mut`, despite setting the contract storage. Any idea why?

cmichi (approved):
This is great example!

cmichi:
Would be good to add an entry here: https://github.com/paritytech/ink/tree/master/integration-tests/upgradeable-contracts.

integration-tests/upgradeable-contracts/set-code-hash-migration/migration/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+        /// upgraded to the supplied code hash.
+        ///
+        /// In a production contract you would do some authorization here!
+        ///
+        /// # Note
+        ///
+        /// This function necessarily accepts a `&self` instead of a `&mut self` because
+        /// we are modifying storage directly for the migration. Using `&mut self`
+        /// would overwrite our migration changes with the contents of the
+        /// original `Incrementer`.
```
cmichi, line 48: Why is this the case? If you would have `&mut self` and do `self.inc_by = inc_by`, wouldn't this just get flushed to the original `Incrementer` contract state?

PR #1917, opened 2023-09-16
[E2E] Call builders and extra gas margin option

SkymanOne:
## Summary
- [y] y/n | Is it dependent on the specific version of `cargo-contract` or `pallet-contracts`? - `ink! 5.0.0-alpha`

This PR reworks the E2E API by introducing the builder API. As part of this PR, the user can specify gas margin in percentage as part of the on-chain call.

### Builder API

`instantiate`, `call` and `upload` will now return a builder instance. The user can specify optional arguments with builder methods, and submit the call for on-chain execution with `.submit()` method, or dry-run it with `dry_run()`.

```rust
let contract = client
    .instantiate("flipper", &ink_e2e::alice(), &mut constructor)
    .submit()
    .await
let mut call = contract.call::<Flipper>();

let get = call.get();
let get_res = client.call(&ink_e2e::bob(), &get).dry_run().await;
assert!(matches!(get_res.return_value(), false));
```

### Extra gas margin

There are cases when gas estimates may not necessarily be accurate enough due to the complexity of the smart contract logic that adds additional overhead and gas consumption. Therefore, it is helpful to allow to specify an extra portion of the gas to be added to the limit (i.e. 5%, 10%). This functionality was achieved with `.extra_gas_portion(margin: u64)` method as part of the builder API.

### Internal changes

`Environment` trait now extends `Clone` in order to  make `CreateBuilder` cloneable. There are other structs that add `Clone` trait bound as well.

Extract from flipper:
```rust
let constructor = FlipperRef::new(false);
let contract = client
    .create_instantiate_call("flipper", &ink_e2e::alice(), constructor)
    .submit()
    .await
let mut call = contract.call::<Flipper>();

let get = call.get();
let get_res = client
    .create_call(&ink_e2e::bob(), &get)
    .submit_dry_run()
    .await;
```

@pmikolajczyk41 Pinging for review as well


integration-tests/call-builder-return-value/lib.rs
```diff
@@ -148,7 +148,7 @@ mod call_builder {
             let selector = ink::selector_bytes!("get");
             let call = call_builder_call.delegate_call(code_hash, selector);
             let call_result = client
-                .call(&origin, &call, 0, None)
+                .call(&origin, &call, 0, None, None)
```
ascjones, line 151: Maybe now is the time to finally convert to a builder style API, since the default values `0, None, None` are the most common.

SkymanOne: Maybe it is 😄

ascjones: We should do a builder API, but maybe as a separate task where we can design it properly

I think a more minimal approach to this problem (applying the extra gas portion without breaking the `call` API) would be:

1. Have the `Client` store a value for the "extra portion"
2. Set it to some default low value e.g. 2%
3. Allow overriding the value via e.g. `client.set_extra_gas(2u8)` (this could also be configured via the e2e test attribute which initialises the client)

crates/e2e/src/subxt_client.rs
```diff
@@ -527,6 +530,7 @@ where
         caller: &Keypair,
         message: &CallBuilderFinal<E, Args, RetType>,
         value: E::Balance,
+        extra_gas_portion: Option<usize>,
```
ascjones, line 533: > the old instantiate and call are left unchanged so this PR doesn't introduce breaking changes

This looks like a breaking change.

Not a huge problem for it to be a breaking change, just the fact that in 99% of cases the value passed will be `None`.

One approach would be to have a new method e.g. `bare_call` which just passes raw arguments to (in this case) `self.api.call()`. The `gas_limit` would be an absolute value and the part of adding the "portion" could just be done in the builder exec method.

Then we could even keep this method as is so it is a non-breaking change if we wanted. But this is not strictly necessary since we are doing a major release...it will be just annoying for users to have to update all their tests.

integration-tests/contract-transfer/lib.rs
```diff
@@ -252,7 +254,7 @@ pub mod give_me {
             let transfer = call.give_me(120);
 
             let call_res = client
-                .call(&ink_e2e::eve(), &transfer, 0, None)
+                .call(&ink_e2e::eve(), &transfer, 0, None, None)
```
ascjones, line 257: As far as I can see we are not using this API anywhere in the examples.

If we are going to move to this API and have it a breaking change anyway, we can just change `call` to return the call builder, so the default becomes in this case:

```
let call_res = client
  .call(&ink_e2e::eve(), &transfer)
  .submit()
```

crates/e2e/src/subxt_client.rs
SkymanOne: >  The gas_limit would be an absolute value and the part of adding the "portion" could just be done in the builder exec method.

I am not fun of an idea of having 3 different ways of building a call. We are going to end up with `call()`, `bare_call()` and call builder which serve similar purposes. This will just confuse people

SkymanOne: Additionally, there is no way of specifying gas limit in E2E anyway, so it will be a breaking change nonetheless

SkymanOne:
I think the best course of action is to have `bare_call()` where you can specify all args explicitly (including the gas limit), and we are going to have a builder with a gas limit margin.

crates/e2e/src/backend_calls.rs
```diff
@@ -0,0 +1,229 @@
...
+    storage_deposit_limit: Option<E::Balance>,
+}
+
+impl<'a, E, Args, RetType, CB> CallBuilder<'a, E, Args, RetType, CB>
+where
+    E: Environment,
+    Args: Sync + Encode,
+    RetType: Send + Decode,
+    E::Balance: Clone,
+
```
pmikolajczyk41, line 41: nit: unnecessary empty line

pmikolajczyk41 (approved):
looks nice, we might want a similar excess to be specified in cargo-contract as well

crates/e2e/src/backend.rs
```diff
@@ -42,7 +44,7 @@ pub trait ChainBackend {
     /// Account type.
     type AccountId;
     /// Balance type.
-    type Balance: Send;
+    type Balance: Send + From<u32>;
```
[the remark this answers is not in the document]
SkymanOne: I had to put something there, otherwise I cannot provide the default value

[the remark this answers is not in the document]
pmikolajczyk41: I get it, but I'm wondering how `From<u32>` relates to `From<u128>` and possibly other numeric trait from e.g. `num_traits`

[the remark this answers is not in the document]
SkymanOne: Tbf, it makes sense to change to `From<u128>` since it is less restrictive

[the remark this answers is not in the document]
pmikolajczyk41: oh, right, so cool then

crates/e2e/src/backend_calls.rs
SkymanOne: That was on purpose to logically separate generics

SkymanOne:
Converting to draft as `Instantiatebuilder` needs to be reworked. This is due to `submit()` taking `self` and a move occurs which is because `CreateBuilder<>` does not implement `Clone`

crates/e2e/src/backend_calls.rs
```diff
@@ -0,0 +1,320 @@
...
+    /// Increases the gas limit marginally by a specified percent.
+    /// Useful when the message's gas usage depends on the runtime state
+    /// and the dry run does not produce an accurate gas estimate.
+    ///
+    /// # Example
+    ///
+    /// With dry run gas estimate of `100` units and `5`% extra gas portion specified,
+    /// the set gas limit becomes `105` units
+    pub fn extra_gas_portion(&mut self, per_cent: u64) -> &mut Self {
```
smiasojed, line 215: Is it possible to predict the extra gas portion based on the block saturation level?

SkymanOne: If by block saturation you mean how much block weight is consumed for a given block, then I don't think it is related to the gas estimate. The reason why we want to add extra gas margin is because dry-run does not produce an accurate estimate. This can happend when the state changes when we submit an on-chain call and the contract performs some extra computation consuming extra gas

smiasojed: I thought that this extra gas consumption could be related to https://spec.polkadot.network/id-weights#id-fee-multiplier. I encountered issues while conducting contract benchmarks on Moonbeam, where the same call on the same contract could consume varying amounts of gas due to network load.

SkymanOne:
@pmikolajczyk41 This PR heavily relies on dry-running. As drink! currently does not implement those. I had to put contract runtime call inside dry-run function. When do you think you'll be able to implement dry-running?

crates/e2e/src/backend_calls.rs
SkymanOne: This is part of polkadot runtime, not the contract gas consumption which is completely different thing

pmikolajczyk41 (changes requested):
**dry-running in drink!**: it's already there, ready to be used at any time

**the new API**: I really like the new flow with `client.action().modifier1().modifier2().submit/dry_run()`. However, IMHO, the current traiting and API needs some encapsulation. In particular:

1. For the user (e2e test author) I would expose only a single API for interacting with a contract (upload/instantiate/call). Preferably, I would remove `bare_*` and `bare_*_dry_run` from `ContractsBackend`. These methods do not add any more expressiveness, but instead may introduce some confusion + it makes us to support and care for broader API.

2. I would move things like `bare_instantiate` and `bare_instantiate_dry_run` into a crate private trait, like `Submitter` or whatever you like. This trait would be completely internal, used only for callbacks from `CallBuilder`/`InstantiateBuilder`/`UploadBuilder` implementations. Of course, this trait can and probably should be implemented by clients, but anyway, shouldn't be accessible by the end user. Please notice, that you don't need both `bare_instantiate` and `instantiate_with_gas_margin` (and thus `BuilderClient` is not needed).

Side note/question: I don't know if it is intended, but the E2E backend is missing instantiation without upload. I mean the scenario where we upload the code and then create contract instance in separate transactions. Should we use then `upload` and `instantiate` calls? Probably it should work... This is why there's some discrepancy with drink! where I used `deploy` term for packing both operations into a single action.

SkymanOne:
> For the user (e2e test author) I would expose only a single API for interacting with a contract (upload/instantiate/call). Preferably, I would remove bare_* and bare_*_dry_run from ContractsBackend. These methods do not add any more expressiveness, but instead may introduce some confusion + it makes us to support and care for broader API.

That's a fair point. Th initial idea behind having `bare_` calls is that allow more "precise" control over execution such as running dry-run and submitting extrinsics in separate calls and specifying gas limit separately. 

I'll see if I can encapsulate this login into the builder API as well

crates/e2e/src/backend.rs
```diff
@@ -101,58 +149,122 @@ pub trait ContractsBackend<E: Environment> {
...
+        contract_name: &'a str,
+        caller: &'a Keypair,
+    ) -> UploadBuilder<E, Self>
+    where
+        Self: BuilderClient<E> + Sized,
+    {
+        UploadBuilder::new(self, contract_name, caller)
+    }
+
     /// The function subsequently uploads and instantiates an instance of the contract.
```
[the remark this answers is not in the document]
SkymanOne: Can you put the correct one here plz?

crates/e2e/src/backend.rs
```diff
@@ -101,58 +149,122 @@ pub trait ContractsBackend<E: Environment> {
...
+    async fn bare_instantiate_dry_run<
+        Contract: Clone,
+        Args: Send + Sync + Encode + Clone,
+        R,
+    >(
         &mut self,
         contract_name: &str,
         caller: &Keypair,
-        constructor: CreateBuilderPartial<E, Contract, Args, R>,
+        constructor: &mut CreateBuilderPartial<E, Contract, Args, R>,
```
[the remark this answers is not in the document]
SkymanOne: Required as part of the builder. See `constructor_exec_input()`

integration-tests/call-builder-return-value/lib.rs
```diff
@@ -132,23 +131,26 @@ mod call_builder {
                 .await;
 
             let expected_value = 42;
-            let constructor = CallBuilderReturnValueRef::new(expected_value);
+            let mut constructor = CallBuilderReturnValueRef::new(expected_value);
             let call_builder = client
-                .instantiate("call_builder_return_value", &origin, constructor, 0, None)
+                .instantiate("call_builder_return_value", &origin, &mut &mut constructor)
```
ascjones, line 136:
```suggestion
                .instantiate("call_builder_return_value", &origin, &mut constructor)
```

crates/e2e/src/backend.rs
```diff
@@ -92,67 +103,173 @@ pub trait ContractsBackend<E: Environment> {
...
+    async fn call_with_gas_limit<Args: Sync + Encode + Clone, RetType: Send + Decode>(
+        &mut self,
+        caller: &Keypair,
+        message: &CallBuilderFinal<E, Args, RetType>,
+        value: E::Balance,
+        gas_limit: Weight,
+        storage_deposit_limit: Option<E::Balance>,
+    ) -> Result<CallResult<E, RetType, Self::EventLog>, Self::Error>
+    where
+        CallBuilderFinal<E, Args, RetType>: Clone;
```
ascjones, line 248: Can't we get rid of all these and just use `bare_call` and `bare_instantiate` from the builders?

crates/e2e/src/subxt_client.rs
```diff
@@ -607,9 +571,162 @@ where
...
+            .api
+            .instantiate_with_code_dry_run(
+                value,
+                storage_deposit_limit,
+                code.clone(),
+                data.clone(),
+                salt(),
+                caller,
+            )
+            .await;
```
ascjones, line 600: If the user wants the gas portion (or explicitly specifies a gas limit) we should just perform the dry-run from the builder itself via `bare_call_dry_run`

ascjones (commented):
Main thing would be to simplify the `BuilderClient` trait, could potentially remove a lot of code, and some duplication between the client implementations (moved to the builder itself).

Builder API itself looks good :+1:

crates/e2e/src/backend_calls.rs
```diff
@@ -0,0 +1,366 @@
...
+            B::call_with_gas_margin(
+                self.client,
+                self.caller,
+                self.message,
+                self.value,
+                self.extra_gas_portion,
+                self.storage_deposit_limit,
+            )
+            .await
+        }
```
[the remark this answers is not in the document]
ascjones: And do the dry-run if necessary here (only iff the `self.gas_limit` is not set)

crates/e2e/src/backend.rs
```diff
@@ -92,67 +103,173 @@ pub trait ContractsBackend<E: Environment> {
...
+    ///     .storage_deposit_limit(100)
+    ///     // Submit the call for on-chain execution.
+    ///     .submit()
+    ///     .await
+    ///     .expect("instantiate failed");
+    /// ```
+    fn instantiate<'a, Contract: Clone, Args: Send + Clone + Encode + Sync, R>(
+        &'a mut self,
+        contract_name: &'a str,
+        caller: &'a Keypair,
```
pmikolajczyk41, line 129: just wondering... maybe `alice()` should be moved to default as well (similarly to value, storage deposit and margin)?

crates/e2e/src/backend_calls.rs
```diff
@@ -0,0 +1,366 @@
...
+            message,
+            value: 0u32.into(),
+            extra_gas_portion: None,
+            gas_limit: None,
+            storage_deposit_limit: None,
+        }
+    }
+
+    /// Provide value with a call
+    pub fn value(&mut self, value: E::Balance) -> &mut Self {
```
pmikolajczyk41, line 69: why not `pub fn value(mut self, value: E::Balance) -> Self`?

pmikolajczyk41 (approved):
the new API looks very nice to me; I agree with Andrew's suggestions, but apart from them, it's good for me

I guess some minor implementation parts might be changed or refactored in the future, but the public interface seems to be final

crates/e2e/src/backend_calls.rs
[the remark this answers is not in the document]
SkymanOne: The main problem is that `type Error` is different in both `subxt_client` and `drink` and we need to return a concrete error when dry-run fails

crates/e2e/src/backend.rs
SkymanOne: Nope. Specifically because of https://github.com/paritytech/ink/blob/3d8f9ba4f6681db51a9cce3c599110d78d1c9606/crates/e2e/src/subxt_client.rs#L602

crates/e2e/src/backend_calls.rs
SkymanOne: We don't want to move values out of references, and, generally speaking, it is a good practise to implement non-consuming builders where possible:
https://doc.rust-lang.org/1.0.0/style/ownership/builders.html#non-consuming-builders-(preferred):

crates/e2e/src/backend.rs
SkymanOne: IMHO, the caller should be specified explicitly. I don't want to hide this information from the developer per se

SkymanOne:
@ascjones To address your comments with regard to using `bare_call` and `bare_instantiate` and handling dry-run results from the builder: the main problem lies in error handling from the dry-run result as I pointed out above. To address this issue, there needs to be a standardised error handling between `drink` and `subxt`. Specifically, we need to have

`type Error: ExecutionErrors` where `ExecutionError` is a trait with method like `return_dry_run_error` and so on. Then I can do something like:
```
if dry_run.result.is_err() {
    return Err(return_dry_run_error(dry_run))
}
```

But then we are going to end up with the same amount of code as `drink` and `subxt` clients will have to implement `ExectionErrors` trait.

crates/e2e/src/backend.rs
ascjones: My suggestion: https://github.com/paritytech/ink/pull/1938

ascjones:
> @ascjones To address your comments with regard to using `bare_call` and `bare_instantiate` and handling dry-run results from the builder: the main problem lies in error handling from the dry-run result as I pointed out above. To address this issue, there needs to be a standardised error handling between `drink` and `subxt`. Specifically, we need to have
> 
> `type Error: ExecutionErrors` where `ExecutionError` is a trait with method like `return_dry_run_error` and so on. Then I can do something like:
> 
> ```rust
> if dry_run.result.is_err() {
>     return Err(return_dry_run_error(dry_run))
> }
> ```
> 
> But then we are going to end up with the same amount of code as `drink` and `subxt` clients will have to implement `ExectionErrors` trait.

See https://github.com/paritytech/ink/pull/1938 for an approach of how to solve this.

paritytech-cicd-pr:

These are the results when building the `integration-tests/*` contracts from this branch with `cargo-contract 4.0.0-alpha-acc02b1` and comparing them to ink! `master`:

| Contract                                                | Upstream Size (kB) | PR Size (kB) | Diff (kB) | Diff (%)  | Change                       |
| ------------------------------------------------------- | ------------------ | ------------ | --------- | --------- | ---------------------------- |
| basic-contract-caller                                   | 2.998              | 2.998        | 0         | 0         | :heavy_minus_sign:           |
| basic-contract-caller/other-contract                    | 1.331              | 1.331        | 0         | 0         | :heavy_minus_sign:           |
| call-builder-return-value                               | 8.734              | 8.734        | 0         | 0         | :heavy_minus_sign:           |
| call-runtime                                            | 1.77               | 1.77         | 0         | 0         | :heavy_minus_sign:           |
| conditional-compilation                                 | 1.204              | 1.204        | 0         | 0         | :heavy_minus_sign:           |
| contract-terminate                                      | 1.087              | 1.087        | 0         | 0         | :heavy_minus_sign:           |
| contract-transfer                                       | 1.446              | 1.446        | 0         | 0         | :heavy_minus_sign:           |
| custom-allocator                                        | 7.39               | 7.39         | 0         | 0         | :heavy_minus_sign:           |
| dns                                                     | 7.313              | 7.313        | 0         | 0         | :heavy_minus_sign:           |
| e2e-call-runtime                                        | 1.057              | 1.057        | 0         | 0         | :heavy_minus_sign:           |
| e2e-runtime-only-backend                                | 1.633              | 1.633        | 0         | 0         | :heavy_minus_sign:           |
| erc1155                                                 | 14.113             | 14.107       | -0.006    | -0.042514 | :chart_with_downwards_trend: |
| erc20                                                   | 6.572              | 6.572        | 0         | 0         | :heavy_minus_sign:           |
| erc721                                                  | 9.547              | 9.547        | 0         | 0         | :heavy_minus_sign:           |
| events                                                  | 4.816              | 4.816        | 0         | 0         | :heavy_minus_sign:           |
| flipper                                                 | 1.387              | 1.387        | 0         | 0         | :heavy_minus_sign:           |
| incrementer                                             | 1.217              | 1.217        | 0         | 0         | :heavy_minus_sign:           |
| lang-err-integration-tests/call-builder-delegate        | 2.321              | 2.321        | 0         | 0         | :heavy_minus_sign:           |
| lang-err-integration-tests/call-builder                 | 4.875              | 4.875        | 0         | 0         | :heavy_minus_sign:           |
| lang-err-integration-tests/constructors-return-value    | 1.772              | 1.772        | 0         | 0         | :heavy_minus_sign:           |
| lang-err-integration-tests/contract-ref                 | 4.363              | 4.363        | 0         | 0         | :heavy_minus_sign:           |
| lang-err-integration-tests/integration-flipper          | 1.565              | 1.565        | 0         | 0         | :heavy_minus_sign:           |
| mapping-integration-tests                               | 2.975              | 2.975        | 0         | 0         | :heavy_minus_sign:           |
| mother                                                  | 9.499              | 9.499        | 0         | 0         | :heavy_minus_sign:           |
| multi-contract-caller                                   | 5.977              | 5.977        | 0         | 0         | :heavy_minus_sign:           |
| multi-contract-caller/accumulator                       | 1.09               | 1.09         | 0         | 0         | :heavy_minus_sign:           |
| multi-contract-caller/adder                             | 1.672              | 1.672        | 0         | 0         | :heavy_minus_sign:           |
| multi-contract-caller/subber                            | 1.693              | 1.693        | 0         | 0         | :heavy_minus_sign:           |
| multisig                                                | 21.406             | 21.406       | 0         | 0         | :heavy_minus_sign:           |
| payment-channel                                         | 5.544              | 5.544        | 0         | 0         | :heavy_minus_sign:           |
| sr25519-verification                                    | 0.865              | 0.865        | 0         | 0         | :heavy_minus_sign:           |
| static-buffer                                           | 1.411              | 1.411        | 0         | 0         | :heavy_minus_sign:           |
| trait-dyn-cross-contract-calls                          | 2.491              | 2.491        | 0         | 0         | :heavy_minus_sign:           |
| trait-dyn-cross-contract-calls/contracts/incrementer    | 1.3                | 1.3          | 0         | 0         | :heavy_minus_sign:           |
| trait-erc20                                             | 6.956              | 6.956        | 0         | 0         | :heavy_minus_sign:           |
| trait-flipper                                           | 1.204              | 1.204        | 0         | 0         | :heavy_minus_sign:           |
| trait-incrementer                                       | 1.365              | 1.365        | 0         | 0         | :heavy_minus_sign:           |
| upgradeable-contracts/delegator                         | 2.902              | 2.902        | 0         | 0         | :heavy_minus_sign:           |
| upgradeable-contracts/delegator/delegatee               | 1.368              | 1.368        | 0         | 0         | :heavy_minus_sign:           |
| upgradeable-contracts/set-code-hash                     | 1.46               | 1.46         | 0         | 0         | :heavy_minus_sign:           |
| upgradeable-contracts/set-code-hash/updated-incrementer | 1.439              | 1.439        | 0         | 0         | :heavy_minus_sign:           |
| wildcard-selector                                       | 2.619              | 2.619        | 0         | 0         | :heavy_minus_sign:           |

[Link to the run](https://gitlab.parity.io/parity/mirrors/ink/-/pipelines/402137) | Last update: Wed Oct 18 13:45:51 CEST 2023

PR #1921, opened 2023-09-26
:t-rex: remove `RUSTFLAGS` for `examples-contract-build`

ascjones:
[Latest CI image](https://hub.docker.com/layers/paritytech/ci-unified/latest/images/sha256-425118b2700387f330f19ded56f03e4493ba1f61630d665b6a7c743839c7de27?context=explore) should bring in https://github.com/paritytech/cargo-contract/pull/1344/ which adds `-Clink-dead-code` to the metadata gen `RUSTFLAGS`. CI should be fixed.

The `RUSTFLAGS` remain set for the `test` and `examples-test` steps because those are not using `cargo-contract` directly but building via the `contract-build` dependency which has not yet been released with the fix from https://github.com/paritytech/cargo-contract/pull/1344

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -244,7 +244,7 @@ test:
       artifacts:                   false
   variables:
       # Fix for linking of `linkme` for `cargo test`: https://github.com/dtolnay/linkme/issues/49
-      RUSTFLAGS: "-Clink-arg=-z -Clink-arg=nostart-stop-gc"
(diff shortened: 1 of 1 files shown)


cmichi (approved):

ascjones:
> Attempt to fix linking errors with latest ci image e.g. https://gitlab.parity.io/parity/mirrors/ink/-/jobs/3792660/raw

This does not work :(

PR #1922, opened 2023-09-26
E2E: Clean config parsing

pmikolajczyk41:
- [y] y/n | Does it introduce breaking changes? (although an extremely cosmetic one)

## Description
Instead of parsing e2e configuration arguments, checking for errors and handling all situations manually, we can use `darling` library that is dedicated for exactly such boilerplate.

I have removed most of the testcases (leaving only the positive scenario) - they seem to be completely unnecessary now, since all the error detection and handling is now outsourced. We may however bring them back (preferably in another PR) as UI tests for e2e crate.

The only breaking change is that we won't accept `runtime-only` literal (only `runtime_only` is now valid). Additionally, we accept path arguments (like `runtime`) both in raw form and as literals (default in `darling`).

Note: we had to change `additional_contracts` field type from `Vec<String>` to `String`, since `darling` doesn't support parsing `Vec<String>` in the intended way (only by multiple argument occurrences). However, this is not visible from outside struct internal implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Changed
 - Make `set_code_hash` generic - [#1906](https://github.com/paritytech/ink/pull/1906)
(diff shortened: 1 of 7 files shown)


ascjones (approved):
Love it :heart:

PR #1930, opened 2023-10-03
Fix remaining unchecked arithmentic in examples

ascjones:
Follow up to https://github.com/paritytech/ink/pull/1831, using checked arithmetic


agryaznov (approved):

smiasojed (approved):

ascjones:
> why not saturating_add instead checked_add + unwrap?

In the unlikely event of an overflow we definitely want to panic instead of wrapping around to `0`

PR #1938, opened 2023-10-17
Suggestions for removing code duplication in #1917

ascjones:
Moves implementation detail of how to calculate the gas limit up front from the client impls up to the builder. It should not be a concern of individual client implementations, they are just adapters.


crates/e2e/src/error.rs
```diff
@@ -44,3 +44,20 @@ pub enum Error<AccountId, Balance, CodeHash, DispatchError> {
     /// Decoding failed.
     Decoding(String),
 }
+
+impl<AccountId, Balance, CodeHash, DispatchError>
```
SkymanOne, line 48: we might as well implement this for other errors

SkymanOne (approved):

crates/e2e/src/error.rs
```diff
@@ -44,3 +44,20 @@ pub enum Error<AccountId, Balance, CodeHash, DispatchError> {
...
+    From<ContractInstantiateResult<AccountId, Balance, ()>>
+    for Error<AccountId, Balance, CodeHash, DispatchError>
+{
+    fn from(value: ContractInstantiateResult<AccountId, Balance, ()>) -> Self {
+        Self::InstantiateDryRun(value)
```
SkymanOne, line 53: And there can be other errors returned. It's not always `InstantiateDryRun`

SkymanOne: Ideally we need to implement this for each error variant

Issue #1650, opened 2023-02-13
Fix E2E workflow for use with long-running contracts node


ascjones:
In https://github.com/paritytech/ink/pull/1642, the E2E testing framework will now automatically spawn a contracts node instance per test. The primary motivation is to make the tests independent from one another, making the CI test runs more robust.

However, for developer workflows this assumes that there is a locally accessible pre-built binary. Previously it was possible to 
1. Run a `substrate-contracts-node` on a remote machine and run the tests against that
2. Easily run a node in the background from source via `cargo run`, and then run the E2E tests against that. 

We should make these workflows possible again, which both assume a long running contracts node rather than the auto spawn per test.

One way to achieve this would be via an environment variable e.g. `E2E_BACKGROUND_CONTRACTS_NODE_URL` or something shorter and equally descriptive. Setting this would disable the default functionality of spawning a test per node, and connect with a client to the assumed-top-be-running node at the supplied URL. e.g.

`E2E_BACKGROUND_CONTRACTS_NODE_URL=ws://localhost:9944 cargo test --features e2e-tests`

Admittedly that is a bit verbose, so here are some ideas off the top of my head:

1. Short term: user can set that env variable in their shell.
2. Allow setting this in the `Cargo.toml` of the target contract? Downside is this could be committed by mistake to source control in shared repos.
3. Bring back `cargo contract test` (which I just [removed](https://github.com/paritytech/cargo-contract/pull/958)), and then we can provide CLI options for this e.g. `cargo contract test --node-url ws://localhost:9944 `

cmichi:
The reason to originally switch to one node process per test was to work around a failure in the CI, which we couldn't reproduce locally, correct?

I think it would be better to have one process in the background be the default behavior of our E2E tests. It could also be spawned automatically, like we do now, but only once. This will make things go faster and also help with the performance of https://github.com/paritytech/ink/pull/1648.

SkymanOne:
Can we also investigate why the CI was hanging on a single running instance? AFAIK it works just fine locally

ascjones:
> The reason to originally switch to one node process per test was to work around a failure in the CI, which we couldn't reproduce locally, correct?

This failure was another manifestation of the fundamental issue of the individual tests running in parallel against the same node. In this configuration it is impossible to prevent race conditions between the tests. Previously it was required that each test in a suite was using a different account for interacting with contracts, otherwise the tests could interfere with one another. 

[...]

PR #1949, opened 2023-10-25
[E2E] Allow testing with live-chain state

### LINKED ISSUE #1947: Allow E2E testing with live-chain state
We got it working by now and @brunopgalvao wrote the instructions down here: https://github.com/paritytech/ink-docs/pull/283.

For this to work though, my custom `ink` branch https://github.com/paritytech/ink/commits/cmichi-use-sc-chopstick-compatible-2 has to be used. It contains only two commits though.

Please create a PR against `master` that implements those two changes properly. My branch is hacky and contains some unnecessary changes (you'll see them).

* Allow specifying the `WS_PORT` of the RPC node.
* Add a feature `unstable-light-client` to ink!. If passed on the command-line the effect should be [this](https://github.com/paritytech/ink/commit/7e1328363e49f53969f58f9a66f3aedb859e8b39).

Please test if the instructions from Bruno's PR still work with your PR.

SkymanOne:
## Summary
Closes #1947
- [n] y/n | Is it dependent on the specific version of `cargo-contract` or `pallet-contracts`?

Allows to run E2E tests against the live chain leveraging its live state

## How to test

Due to the inability to run the test on GH workflow, here are instructions on how to run the test locally:

_Copied from https://github.com/paritytech/ink-docs/pull/283_

1. Clone chopsticks:
```
git clone https://github.com/AcalaNetwork/chopsticks
```

2. Modify the `dev.yml` config file in the repo or create one from scratch that you can reference later:
```
endpoint: ws://127.0.0.1:9944
mock-signature-host: true
block: 1
db: ./db.sqlite
```
> Note: In the example above chopsticks will be mirroring up until block 1 from the substrate-contracts-node. For real world use case you would want to use a different block number and this is the place where you can configure other variables such as a sudo key. Read the chopsticks docs for more info.

Basically produce two empty blocks by submitting 2 `system.remark(0)` extrinsics on polkadot.js

3. Run chopsticks locally:
```
npx @acala-network/chopsticks@latest --config=configs/dev.yml
```

4. Go to `integration-tests/e2e-call-runtime` and modify the test attribute:
```rust
...
#[ink_e2e::test(node_url = "ws://127.0.0.1:8000")]
...
```

5. Run `cargo test -F e2e-tests`


crates/e2e/src/node_proc.rs
```diff
@@ -150,9 +159,6 @@ where
             Err(err) => {
                 let err = format!("Failed to connect to node rpc at {url}: {err}");
                 tracing::error!("{}", err);
-                proc.kill().map_err(|e| {
-                    format!("Error killing substrate process '{}': {}", proc.id(), e)
-                })?;
```
cmichi, line 155: Accidental removal?

SkymanOne: Yep

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- [E2E] Allow testing with live-chain state - [#1949](https://github.com/paritytech/ink/pull/1949)
+- Stabilize `call_runtime` ‒ [#1749](https://github.com/paritytech/ink/pull/1749)
+- Schema generation - [#1765](https://github.com/paritytech/ink/pull/1765)
+- Add `set_block_number` to off-chain test api `Engine` - [#1806](https://github.com/paritytech/ink/pull/1806)
```
cmichi, line 13: Was it intentional to add all those unrelated ones here?

cmichi:
On my branch I had added the feature to more `subxt` deps than you. Did your tests with Chopsticks work with your PR?

SkymanOne:
> On my branch I had added the feature to more `subxt` deps than you. Did your tests with Chopsticks work with your PR?

Yes. I have run tests with `WS_PORT=8000 cargo test -F e2e-tests` on `e2e-call-runtime` and `flipper. Both succeeded, and logs were produced in chopsticks node.

CHANGELOG.md
SkymanOne: No, just a mistake during merge

cmichi:
> Yes. I have run tests with `WS_PORT=8000 cargo test -F e2e-tests` on `e2e-call-runtime` and `flipper`. Both succeeded, and logs were produced in chopsticks node.

Did you also run them with `-F live-state-test`?

SkymanOne:
> > Yes. I have run tests with `WS_PORT=8000 cargo test -F e2e-tests` on `e2e-call-runtime` and `flipper`. Both succeeded, and logs were produced in chopsticks node.
> 
> Did you also run them with `-F live-state-test`?

Yes, as you can see in git diff in this PR, `e2e-call-runtime` has `live-state-test` enabled by default

crates/e2e/src/node_proc.rs
```diff
@@ -161,11 +161,20 @@ where
                 e
             )
         })?;
-
-        // Wait for RPC port to be logged (it's logged to stderr):
         let stderr = proc.stderr.take().unwrap();
-        let port = find_substrate_port_from_output(stderr);
-        let url = format!("ws://127.0.0.1:{port}");
+        // Wait for RPC port to be logged (it's logged to stderr):
+        let port_str: String = match option_env!("WS_PORT") {
```
ascjones, line 166: Is it the case that the node that is spawned here does not print out the port to `stderr`, or are we just attaching to a different already running process (chopsticks?)

ascjones (commented):
Can you give a brief explanation of the changes and the setup, and how it works to enable testing live-chain state?

crates/e2e/src/node_proc.rs
ascjones: Ok so you have clarified that chopsticks is an already running process.

In that case, I don't think we should be modifying `node_proc` in this way. It is intended to spin up a node from scratch and then return a client which can communicate with the node. So in this case it is spinning up a node for the test which is never used (because requests are sent to the chopsticks node via the overriding of the port).

What we could do instead is allow the user to specify the full url of a running process in the config. This would override the default behaviour of spawning a node for each test. e.g.

`#[ink_e2e::test(node_urk = "ws://127.0.0.1:8000")]`

Then in this case it wouldn't spawn the node_proc https://github.com/paritytech/ink/blob/6fa4343e5df8fd025179bceeec527ef8874ad78f/crates/e2e/macro/src/codegen.rs#L116

ascjones: Alternatively can use an environment variable, as per https://github.com/paritytech/ink/issues/1650.

However it may be good to have it in the attribute config, because this changes how e2e tests are written: usually they are expecting clean state on a chain and they don't affect each other. With a long running node the state is shared between tests so is not deterministic.

crates/e2e/Cargo.toml
```diff
@@ -52,6 +52,10 @@ scale-info = { workspace = true, features = ["derive"] }
 [features]
 default = ["std"]
 std = []
+live-state-test = [
+    "subxt/unstable-light-client",
```
ascjones, line 56: What's the purpose of this, do we need to use `smoldot` to communicate with chopsticks? It looks like we just use the normal RPC?

SkymanOne: I assume yes, we need a light client to communicate with a node.

crates/e2e/macro/src/config.rs
```diff
@@ -46,6 +46,9 @@ pub struct E2EConfig {
     /// The type of the architecture that should be used to run test.
     #[darling(default)]
     backend: Backend,
+    /// The URL to the running chopsticks node.
+    #[cfg(any(test, feature = "live-state-test"))]
+    chopsticks_url: Option<String>,
```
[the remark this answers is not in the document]
ascjones: I would say so, yes.

crates/e2e/Cargo.toml
ascjones: The initialisation code uses a regular `RpcClient` https://github.com/paritytech/ink/pull/1949/files#diff-4f8810a386176bf775fb827b07ad2d012aef2f57638a97ffad1fc7d898e61395R125, so no usage of the light client rpc AFAIU.

Does it still work without this feature enabled?

SkymanOne: Yes, but I have very limited knowledge whether there is the edge case when it won't work

SkymanOne: > The initialisation code uses a regular `RpcClient` https://github.com/paritytech/ink/pull/1949/files#diff-4f8810a386176bf775fb827b07ad2d012aef2f57638a97ffad1fc7d898e61395R125, so no usage of the light client rpc AFAIU.
> 
> Does it still work without this feature enabled?

Apparently, it is possible to `LightClient` because it shares the same interface as `RpcClient` but I don't see the difference for local testing. Need to consult @cmichi

crates/e2e/macro/src/config.rs
```diff
@@ -46,6 +46,8 @@ pub struct E2EConfig {
     /// The type of the architecture that should be used to run test.
     #[darling(default)]
     backend: Backend,
+    /// The URL to the running node.
```
ascjones, line 49:
```suggestion
    /// The URL to the running node. If not set then a default node instance will be spawned per test.
```

PR #1971, opened 2023-11-03
Migrating the measurement and publish workflows to GitHub Actions

smiasojed:
## Summary
Related #1454
- [n] y/n | Does it introduce breaking changes?
- [n] y/n | Is it dependant on the specific version of `cargo-contract` or `pallet-contracts`?
Migrating the measurement and publish workflows to GitHub Actions

## Description
Depends on https://github.com/paritytech/ink/pull/1977
Adds the measurement and publish-docs workflows

## Checklist before requesting a review
- [y] My code follows the style guidelines of this project
- [n] I have added an entry to `CHANGELOG.md`
- [y] I have commented my code, particularly in hard-to-understand areas
- [n] I have added tests that prove my fix is effective or that my feature works
- [y] Any dependent changes have been merged and published in downstream modules

Issue #1454, opened 2022-10-29
CI: Migration away from GitLab to GitHub Actions


cmichi:

smiasojed:
Open points:
1. Decide what to do with Ink-waterfall.
2. Enable the fuzz job and add an issue creation job.
3. Optimize storage consumption for the `examples-test` and `test` jobs; it currently takes almost 19GB, whereas on GitHub's standard machine, we have 14GB of free space.
4. Improve the Clippy job to include comments in the PR code review.

PR #1977, opened 2023-11-07
Add submit-contract-sizes GHA workflow

smiasojed:
## Summary
Related #1454
Add submit-contract-sizes workflow

## Description
Adds submit-contract-sizes workflow.
After merging, this workflow will fail because 'ARTIFACT_NAME: contract-sizes-data' does not exist. This won't impact the review process, as it is not marked as a required workflow, but it will allow for testing the measurement functionality in #1971

- [y] My code follows the style guidelines of this project
- [n] I have added an entry to `CHANGELOG.md`
- [y] I have commented my code, particularly in hard-to-understand areas
- [n] I have added tests that prove my fix is effective or that my feature works
- [y] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/.github/workflows/submit-contract-sizes.yml
+++ b/.github/workflows/submit-contract-sizes.yml
@@ -28,7 +28,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           ARTIFACT_NAME: contract-sizes-data
 
-
       - name: Submit Comment
         env:
           GITHUB_PR_TOKEN:         ${{ secrets.github_token }}
--- a/.github/workflows/submit-contract-sizes.yml
+++ b/.github/workflows/submit-contract-sizes.yml
@@ -1,0 +1,46 @@
+# The workflow has write access, so it needs to be isolated for security reasons from pull request-based workflows,
+# which may be triggered from forked repositories.
+on:
+  workflow_run:
+    workflows:
+      - continuous-integration
+    types:
+      - completed
+
+jobs:
+  submit-contract-sizes:
+    permissions:
+      pull-requests: write
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+        github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 100
+
+      - name: Download Contract Sizes
+        uses: ./.github/download-artifact
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ARTIFACT_NAME: contract-sizes-data
+
+
+      - name: Submit Comment
+        env:
+          GITHUB_PR_TOKEN:         ${{ secrets.github_token }}
+          GITHUB_PR_WORKFLOW_ID:   ${{ github.event.workflow_run.id }}
+        run: |
+          # context.out is considered as an untrusted file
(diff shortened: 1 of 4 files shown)


.github/workflows/submit-contract-sizes.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: |
+          # context.out is considered as an untrusted file
+          PR_NUMBER=$(grep -oE 'PR_NUMBER="[^"]+"' context.out | awk -F '"' '{print $2}')
+          CARGO_CONTRACT_VERSION=$(grep -oE 'CARGO_CONTRACT_VERSION="[^"]+"' context.out | awk -F '"' '{print $2}')
+          PR_COMMENTS_URL="${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments"
+          WORKFLOW_URL="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_PR_WORKFLOW_ID}"
+
+          # Submit the comparison table as a comment to the PR
+          echo "Submitting contract sizes diff to ${PR_COMMENTS_URL}"
+          GITHUB_PR_TOKEN=${GITHUB_PR_TOKEN} scripts/contract_sizes_submit.sh ${PR_COMMENTS_URL} ${WORKFLOW_URL} ${CARGO_CONTRACT_VERSION} < ./contract_sizes_diff.md
```
ascjones, line 45: So `contract_sizes_diff.md` comes from the downloaded `contract-sizes-data`? It's not 100% clear and relying on the filename is brittle. Are we able to paramerize this somehow?

smiasojed: contract_sizes_diff.md along with context.out is part of contract-sizes-data artifact.
I can add comment or put it to env - but env is not shared between workflows.

.github/workflows/submit-contract-sizes.yml
```diff
@@ -0,0 +1,46 @@
...
+      github.event.workflow_run.event == 'pull_request' &&
+        github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 100
+
+      - name: Download Contract Sizes
+        uses: ./.github/download-artifact
```
ascjones, line 26: So this action is part of https://github.com/paritytech/ink/pull/1971, why is this a standalone PR if it is depending on stuff in the other PR?

.github/workflows/submit-contract-sizes.yml
ascjones: Would it not be possible to structure this `contract-sizes` as a single job, and then `submit` as a step, and then use [`$GITHUB_OUTPUT`](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) to send values between the individual steps, in this case it would contain the contents of `contract_sizes_diff.md`

smiasojed: It is possible, I could user here pull_request_target trigger but it introduces security risk.
"GitHub’s standard pull_request workflow trigger by default prevents write permissions and secrets access to the target repository. However, in some scenarios such access is needed to properly process the PR. To this end the pull_request_target workflow trigger was introduced." 
More informations: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

.github/workflows/submit-contract-sizes.yml
smiasojed: I missed it, thanks. I will update this PR

ascjones (approved):

PR #1982, opened 2023-11-09
Balance integration fix

faculerena:
## Summary
- [N] y/n | Does it introduce breaking changes?
- [N] y/n | Is it dependant on the specific version of `cargo-contract` or `pallet-contracts`?

Updated the default balance in integration tests.

## Description
We updated the default balance in integration tests to be the same as in e2e-tests. This is not definitive, as we want opinions if this should be like this.

- [X] My code follows the style guidelines of this project
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/crates/env/src/engine/off_chain/test_api.rs
+++ b/crates/env/src/engine/off_chain/test_api.rs
@@ -14,14 +14,8 @@
 
 //! Operations on the off-chain testing environment.
 
-use super::{
-    EnvInstance,
-    OnInstance,
-};
-use crate::{
-    Environment,
-    Result,
-};
+use super::{EnvInstance, OnInstance};
+use crate::{Environment, Result};
 use core::fmt::Debug;
 use ink_engine::test_api::RecordedDebugMessages;
 use std::panic::UnwindSafe;
@@ -306,7 +300,8 @@ where
         instance.engine.set_callee(encoded_alice.clone());
 
         // set up the funds for the default accounts
-        let substantial = 1_000_000;
+        // the 1_000_000_000 is the same value as in the e2e tests
+        let substantial = 1_000_000_000;
         let some = 1_000;
         instance.engine.set_balance(encoded_alice, substantial);
         instance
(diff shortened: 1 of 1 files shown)


crates/env/src/engine/off_chain/test_api.rs
```diff
@@ -306,7 +306,8 @@ where
         instance.engine.set_callee(encoded_alice.clone());
 
         // set up the funds for the default accounts
-        let substantial = 1_000_000;
+        // the 1_000_000_000 is the same value as in the e2e tests
+        let substantial = 1_000_000_000;
```
ascjones, line 310: I think the original `1_000_000` value was somewhat arbitrary anyway, so makes sense to change it to be the same as the `substrate-contracts-node` used for testsin.

Is it only the Alice account that is endowed with this amount? I see bob and charlie are still set to `1_000` below - does that match up with the endowed amounts for the e2e tests too? Seems kind of low.

faculerena: That's true, i will check the behaviour and make it equal for all name accounts (at least). The accounts created with `::From([1;32])` should have a balance of 0 or the existential minimum?

faculerena:
I changed the value to be the same as [the testnet genesis](https://github.com/paritytech/substrate-contracts-node/blob/0c25210b445d444ccab5656236d0be62e32df33f/node/src/chain_spec.rs#L142). This is the same value that this test gives in an e2e test.

`     
  let alice_balance = client
      .balance(ink_e2e::account_id(ink_e2e::AccountKeyring::Alice))
      .await
      .expect("Failed to get account balance");
`     
In this case, `alice_balance` is (1 << 60), and for each account in the keyring this is true too, except for `one` and `two`, that is 0.

cmichi:
@faculerena Could you merge `master` into your PR? There are some conflicts that have to be resolved.

PR #1990, opened 2023-11-14
Add e2e tests with zombienet_sdk

pgherveou:
This adds end-to-end tests with the Zombienet SDK.
The syntax is the same as for writing tests with a single node. The first parameter is a  client to the contracts parachain, and the second optional argument lets you access the Zombienet SDK to manipulate, for example, the relay chain.

```rust
        #[ink_e2e::test(backend = "network")]
        async fn default_network_works<Client: E2EBackend>(mut client: Client, /* network: zombienet_sdk::Network */) -> E2EResult<()> {
            // given
            let mut constructor = FlipperRef::new_default();

            // when
            let contract = client
                .instantiate("flipper", &ink_e2e::bob(), &mut constructor)
                .submit()
                .await
            let call = contract.call::<Flipper>();

            // then
            let get = call.get();
            let get_res = client.call(&ink_e2e::bob(), &get).dry_run().await;
            assert!(matches!(get_res.return_value(), false));

            Ok(())
        }
    }
    }
```


pgherveou:
cc @pepoviola 
Interesting part for you is `crates/e2e/src/zombienet.rs` and `crates/e2e/macro/src/codegen.rs`

crates/e2e/src/zombienet.rs
```diff
@@ -0,0 +1,42 @@
...
+        "polkadot-execute-worker",
+        "polkadot-prepare-worker",
+    ]
+    .into_iter()
+    .filter(|binary| which::which(binary).is_err())
+    .collect::<Vec<_>>();
+
+    if !missing_polkadot_binaries.is_empty() {
+        panic!("The following binaries were not found in the PATH: {missing_polkadot_binaries:?}");
+    }
```
pgherveou, line 21: @pepoviola do we have these checks when we configure the Network with "polkadot" command? This could be removed here if zombienet_sdk panic with similar error messages

integration-tests/flipper/lib.rs
```diff
@@ -110,5 +113,29 @@ pub mod flipper {
 
             Ok(())
         }
+
+        #[ink_e2e::test(backend = "network")]
```
pgherveou, line 117: this test will take over 60s to complete. Should we gate it so it does not run in CI for now?

crates/e2e/src/zombienet.rs
pepoviola: If the `polkadot` is not in the path the spawn will fail but with a different error, you are asserting on the error?

integration-tests/flipper/lib.rs
pepoviola: Those 60s includes the network spawning?

crates/e2e/src/zombienet.rs
pgherveou: I just want the error to be easy to find for users trying to run their tests.
I'll try to hit the error to see how it looks when these binaries are missing.

integration-tests/flipper/lib.rs
pgherveou: yes that is the biggest chunk.

crates/e2e/src/zombienet.rs
pepoviola: IIRC if `polkadot` is missing the spawn will fail quickly but if one of the workers is missing the failure will be a little bit more hidden in the `node`'s logs.

PR #1991, opened 2023-11-14
Caller is origin implementation

ramirez7358:
## Summary
Closes #_

### Function

`caller_is_origin()`

### Initial status

Not implemented in integration test

### Issue documentation & test case

https://github.com/CoinFabrik/on-ink-integration-tests/tree/main/test-cases/caller-is-origin

### Implementation note

We have added the `depth`field to the `ExecContext` struct in order to be able to detect caller chages avery time calls are made between contracts.

Basically every time the callee changes in the `instantiate_contract` or `invoke_contract` function calls, one is aded to `depth`. And when the callee returns to the previous one, 1 is subtracted.

The `caller_is_origin()`function compares the value of `depth` to zero to check if the current contract caller is the origin of the entire call stack.

### Testing guide

In the directory `integration-tests/set_contract_storage` of the [target directory](https://github.com/paritytech/ink), we include in our pull request a test case showing that the observed implementation difference has been corrected. Note that this test is different from the [original test case in our repository](https://github.com/CoinFabrik/on-ink-integration-tests/tree/main/test-cases/caller-is-origin), which showed the difference.<br><br>In order to run the integration tests run:<br>`cargo test`<br><br>In order to run the e2e tests run:<br>`cargo test -–features e2e-tests`<br><br>These tests are run separately because we do not need to compare their results.


cmichi:
@ramirez7358 Sorry for pinging you only now about all those old PR's. Could you possibly merge `master` and resolve the conflicts here as well?

PR #1994, opened 2023-11-15
`[ink_e2e]` resolve `DispatchError` error details for dry-runs

ascjones:
Currently for dry runs we only get raw dispatch errors e.g. `ModuleError { index: 21, error: [22, 0, 0, 0],`. This PR will resolve the error message to a more useful e.g.`InstantiateDryRun(DryRunError { debug_message: "", error: Module(ModuleError(<Contracts::StorageDepositNotEnoughFunds>)) })`

The API for executing dry runs will be aligned with the transaction execution methods by returning a top level `Result`. The signatures for the `dry_run` methods are now:

```rust
pub async fn dry_run(&mut self) -> Result<CallDryRunResult<E, RetType>, B::Error>;
pub async fn dry_run(&mut self) -> Result<InstantiateDryRunResult<E>, B::Error>
```
`InstantiateDryRunResult` has been added in this PR to encapsulate the returned result, similar to the existing `CallDryRunResult`.

- [x] y/n | Does it introduce breaking changes?

### DIFF
--- a/crates/e2e/Cargo.toml
+++ b/crates/e2e/Cargo.toml
@@ -35,7 +35,7 @@ scale = { package = "parity-scale-codec", workspace = true }
 subxt = { workspace = true }
 subxt-metadata = { workspace = true, optional = true }
 subxt-signer = { workspace = true, features = ["subxt", "sr25519"] }
-thiserror = "1.0.50"
+thiserror = { workspace = true }
 wasm-instrument = { workspace = true }
 which = { workspace = true }
 
--- a/crates/e2e/Cargo.toml
+++ b/crates/e2e/Cargo.toml
@@ -35,6 +35,7 @@ scale = { package = "parity-scale-codec", workspace = true }
 subxt = { workspace = true }
 subxt-metadata = { workspace = true, optional = true }
 subxt-signer = { workspace = true, features = ["subxt", "sr25519"] }
+thiserror = "1.0.50"
 wasm-instrument = { workspace = true }
 which = { workspace = true }
 
--- a/crates/e2e/src/contract_results.rs
+++ b/crates/e2e/src/contract_results.rs
@@ -204,18 +204,6 @@ impl<E: Environment, V: scale::Decode> CallDryRunResult<E, V> {
         self.exec_result.result.is_err()
     }
 
-    /// Converts the dry-run result into a `Result` type.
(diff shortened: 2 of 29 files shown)


crates/e2e/Cargo.toml
```diff
@@ -35,6 +35,7 @@ scale = { package = "parity-scale-codec", workspace = true }
 subxt = { workspace = true }
 subxt-metadata = { workspace = true, optional = true }
 subxt-signer = { workspace = true, features = ["subxt", "sr25519"] }
+thiserror = "1.0.50"
```
pgherveou, line 38: move to workspace?

pgherveou (approved):

crates/e2e/src/subxt_client.rs
```diff
@@ -268,6 +267,44 @@ where
             events: tx_events,
         })
     }
+
+    /// Transforms a [`ContractResult`] from a dry run into a [`Result`] type, containing
+    /// details of the [`DispatchError`] if the dry run failed.
+    fn contract_result_to_result<V>(
```
SkymanOne, line 274: Can't this be more generically be implemented with `From` trait?

SkymanOne (approved):
LGTM! Just some nits

crates/e2e/src/subxt_client.rs
ascjones: In this case we need the `&self` to get the metadata for decoding the `DispatchError`

crates/e2e/src/contract_results.rs
```diff
@@ -265,3 +258,73 @@ impl<E: Environment, V: scale::Decode> CallDryRunResult<E, V> {
...
+        let data = &self.instantiate_return_value().result.data;
+        scale::Decode::decode(&mut data.as_ref()).unwrap_or_else(|env_err| {
+            panic!("Decoding dry run result to constructor return type failed: {env_err}")
+        })
+    }
+
+    /// Returns the return value as raw bytes of the message from the dry-run.
+    ///
+    /// Panics if the dry-run message call failed to execute.
+    pub fn return_data(&self) -> &[u8] {
```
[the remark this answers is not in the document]
ascjones: This can already be done `constructor_result` above. That said I copied this from the `CallDryRunResult`, and I don't think this is super useful for constructors, so have deleted it.

PR #1995, opened 2023-11-15
Provide a `StorageVec` datastructure

xermicus:
## Summary
Provide a vector like data structure, built on top of `Mapping`.

This allows to retrieve elements from the vector and grow the vector without loading and pushing all elements.

## Description
Alternative: #1997

I favor this implementation slightly over #1997. It is less code and adds less contract size. The drawback is that here we have no iterators and no `Index`. However, I don't think iterators are really required. Also, #1997 provides binary search, however the vector needs to be sorted and this is kind of impossible if it grows too large. 


crates/storage/src/lazy/vec.rs
```diff
@@ -0,0 +1,677 @@
...
+        }
+    }
+
+    /// Returns the number of elements in the vector, also referred to as its length.
+    ///
+    /// The length is cached; subsequent calls (without writing to the vector) won't
+    /// trigger additional storage reads.
+    #[inline]
+    pub fn len(&self) -> u32 {
+        self.len_cached
```
ascjones, line 209: Could add a `debug_assert!(self.len_cached.is_none() || Some(self.len) == self.len_cached)` here for safety.

crates/storage/src/lazy/vec.rs
```diff
@@ -0,0 +1,677 @@
...
+    ///
+    /// Panics if `index` exceeds the length of the vector.
+    pub fn clear_at(&mut self, index: u32) {
+        assert!(index < self.len());
+
+        self.elements.remove(index);
+    }
+}
+
+impl<'a, V, KeyType> From<&'a [V]> for StorageVec<V, KeyType>
```
ascjones, line 422: Could it be a a `FromIterator<V>` instead?

crates/storage/src/lazy/vec.rs
```diff
@@ -0,0 +1,677 @@
...
+
+impl<V, KeyType> ::core::fmt::Debug for StorageVec<V, KeyType>
+where
+    V: Packed,
+    KeyType: StorageKey,
+{
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        f.debug_struct("StorageVec")
+            .field("key", &KeyType::KEY)
+            .field("len", &self.len)
```
ascjones, line 446: For debugging we may want the `len_cached` too

integration-tests/lazyvec-integration-test/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+        pub fn default() -> Self {
+            Self {
+                proposals: Default::default(),
+            }
+        }
+
+        /// Checks whether given account is allowed to vote and didn't already
+        /// participate.
+        fn is_eligible(&self, _voter: AccountId) -> bool {
+            // todo
```
ascjones, line 43: Still planning on implementing this?

ascjones (commented):
I prefer this proposal over #1997 too. Simpler, smaller code size, and in line with the philosophy of `Mapping` and `Lazy` (thin wrappers around storage access).

xgreenx:
I agree with Andrew, and I like this approach more than a Rust-like vector with caching inside.

crates/storage/src/lazy/vec.rs
xermicus: Sure, added it

crates/storage/src/lazy/vec.rs
xermicus: Yeah, can do that too :) Implement it and added a simple test

crates/storage/src/lazy/vec.rs
xermicus: Added it

crates/storage/src/lazy/vec.rs
ascjones: Is this `From<&'a [V]>` not redundant now then? Can just call `collect()` instead of `into()`? Impl looks identical.

ascjones (approved):
LGTM

xermicus:
Yeah why not, :axe: it

crates/ink/tests/ui/contract/fail/constructor-return-result-non-codec-error.stderr
ascjones, line 1: Just revert this whole file and it should pass. I think you are running locally with a different version (1.74?) where CI is on `1.73`

crates/storage/src/lazy/vec.rs
```diff
@@ -0,0 +1,689 @@
...
+
+    /// Returns the number of elements in the vector, also referred to as its length.
+    ///
+    /// The length is cached; subsequent calls (without writing to the vector) won't
+    /// trigger additional storage reads.
+    #[inline]
+    pub fn len(&self) -> u32 {
+        debug_assert!(self.len_cached.is_none() || self.len.get() == self.len_cached);
+
+        self.len_cached
```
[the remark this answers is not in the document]
ascjones: For that, and all other getters it would need a `&mut self` receiver.

Maybe could get around that with interior mutability to remove the inefficiency.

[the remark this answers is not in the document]
xermicus: Yeah I missed that. Easy to do with a `Cell` though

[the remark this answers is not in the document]
xermicus: @SkymanOne added it together with a test

SkymanOne (approved):
LGTM!

crates/storage/src/lazy/vec.rs
```diff
@@ -0,0 +1,747 @@
...
+    /// This iterates through all elements in the vector; complexity is O(n).
+    /// It might not be possible to clear large vectors within a single block!
+    pub fn clear(&mut self) {
+        for i in 0..self.len() {
+            self.elements.remove(i);
+        }
+        self.set_len(0);
+    }
+
+    /// Remove the element at `index`.
```
xgreenx, line 441:
```suggestion
    /// Clears the value of the element at `index`. It doesn't change the length of the vector.
```

xgreenx (commented):
Maybe it makes sense to add a new method like `swap_and_remove(index: usize)` that swaps the element at `index` with the last element and removes it(via `pop`). It is very popular method to remove elements from the vector in Solidity

xermicus:
@xgreenx I see, Just merged it to get it into the RC, but can do a follow up for adding a swap remove API  anytime :)

PR #2001, opened 2023-11-20
Add `no_main` lint

### LINKED ISSUE #1976: Add a lint for missing `no_main`
Since https://github.com/paritytech/cargo-contract/pull/1076 we require contracts to be annotated with a `no_main`outer attribute. We added a feature in `cargo_contract` to check for the error https://github.com/paritytech/cargo-contract/pull/1113, but it involves some brittle code to intercept and redirect the `stderr` output of `cargo-contract` in order to catch the specific error.

If it is possible to check for attributes, it would be preferable to do this as a lint.

jubnzv:
## Summary
Closes #1976
- [y] y/n | Is it dependent on the specific version of `cargo-contract` or `pallet-contracts`?

## Description
That PR adds the `no_main` lint.
The implementation of the lint is straightforward, and it could be merged as it is.

However, its purpose is to enhance the `cargo-contract` compilation pipeline and it should run with every `build` command. Therefore, `cargo-contract` needs to be patched to execute dylint-based lints, including `no_main`, on each execution of `cargo contract build`.


linting/src/no_main.rs
```diff
@@ -0,0 +1,98 @@
...
+    declare_lint_pass,
+};
+use rustc_span::sym;
+
+declare_lint! {
+    /// ## What it does
+    /// Checks if a contract is annotated with the `no_main` outer attribute.
+    ///
+    /// ## Why is this necessary?
+    /// Contracts must be annotated with `no_main` outer attribute when compiled to WebAssembly for
```
ascjones, line 39: Not for Wasm, in fact, the `no_main` is required to compile to `riscv` targets. So you could just say e.g. to allow for compilation to all supported targets.

linting/src/no_main.rs
```diff
@@ -0,0 +1,98 @@
...
+            if !in_external_macro(cx.sess(), attr.span);
+            if let AttrStyle::Inner = attr.style;
+            if attr.has_name(sym::no_main);
+            then { false } else { true }}
+        }) {
+            span_lint_and_help(
+                cx,
+                NO_MAIN,
+                krate.spans.inner_span,
+                "contract must be annotated with the `no_main` outer attribute",
```
ascjones, line 92:
```suggestion
                "contract must be annotated with the `no_main` inner attribute",
```

linting/src/no_main.rs
```diff
@@ -0,0 +1,98 @@
...
+                None,
+                "consider annotating contract with `#![cfg_attr(not(feature = \"std\"), no_std, no_main)]` or `#[no_main]`"
```
ascjones, line 94:
```suggestion
                "consider annotating contract with `#![cfg_attr(not(feature = \"std\"), no_std, no_main)]` or `#![no_main]`"
```
I think `no_main` is an inner attribute so needs a `!`

ascjones:
@jubnzv please resolve the conflict and we can merge this

jubnzv:
@ascjones Thanks! I'd like to suggest a small fix here first.

linting/src/no_main.rs
```diff
@@ -0,0 +1,98 @@
...
+    matches!(
+        cx.sess().target.llvm_target.to_string().as_str(),
+        "wasm32-unknown-unknown" | "riscv32i-unknown-none-elf"
+    )
+}
+
+impl EarlyLintPass for NoMain {
+    fn check_crate(&mut self, cx: &EarlyContext<'_>, krate: &Crate) {
+        // Disable when building for e2e tests
+        if !is_contract_build(cx) {
```
jubnzv, line 76: Can we execute this lint on all the builds, even for e2e tests, making `no_main` mandatory for every ink! contract?

The point is that `cargo-contract` always runs linter for the host architecture. It is possible to hack it a bit, but should we? Why should we skip this lint for tests?

ascjones: E2E tests will still run the linter because they compile the contract code blob.

Maybe what we really mean/want is that when building the contract "natively" i.e. not `no_std`, which happens for unit tests and metadata generation. But those are usually executed on a different path so maybe we don't need to make the distinction here.

jubnzv: Yes, I see. Let's try to introduce this lint for all contracts. I think it won't make any problems.

ascjones (approved):

PR #2004, opened 2023-11-23
Allow mutable parameters in messages

### LINKED ISSUE #826: Cannot add mut to message arguments
```
#[ink(message)]
pub fn do_something(&mut self, mut token: Erc20) {}
```
This causes the following error:
```
error: expected expression, found keyword `mut`
```

SkymanOne:
## Summary
Closes #826 

Allows message to have `mut` parameters.

## Description
Constructor already provides the ability to have `mut` parameters as the parameter idents are replaced with `ink_binding_{n}` during codegen. The main problem with messages occurring before had to do with the `contract_ref` call builder. when extracting the ident from `pat` for input binding during codegen, `syn` automatically includes `mut` keyword. The solution is to go down the AST of `Pat` and extract `ident` from there. This is what `input_message_idents()` function does.


integration-tests/incrementer-mut/lib.rs
ascjones, line 1: I'm not sure if this needs its own `integration-test`. For this particular change I think a `ui` test in `crates/ink/tests/ui/contract/pass` will suffice?

Adding more and more integration tests is not for free, in terms of keeping them all up to date and increasing CI times.

I think we discussed it before but perhaps a compromise would be to small changes like this to e.g. `mother`?

SkymanOne: Sure. Adding to `mother` looks sensible to me

ascjones (approved):
LGTM. Would be nice to have a `ui/pass` test too for the specification, but I suppose it is covered by the unit test + integration test.

SkymanOne:
> LGTM. Would be nice to have a `ui/pass` test too for the specification, but I suppose it is covered by the unit test + integration test.

The unit test already covers that `must` is parsed correctly. Everything else stays the same.

ascjones:
> > LGTM. Would be nice to have a `ui/pass` test too for the specification, but I suppose it is covered by the unit test + integration test.
> 
> The unit test already covers that `must` is parsed correctly. Everything else stays the same.

Indeed it does, so fine without it. Just pointing out that for the other cases in the unit test there is also a `ui` test. Maybe overkill, up to you.

PR #2005, opened 2023-11-24
`[ink_e2e]` update to new `drink` API

ascjones:
Updates the `drink` backend to be compatible with the latest API, specifically with changes done as part of https://github.com/inkdevhub/drink/pull/84 by @pgherveou 

:warning: **breaking** change for e2e tests :warning:

### DIFF
--- a/crates/e2e/src/drink_client.rs
+++ b/crates/e2e/src/drink_client.rs
@@ -240,7 +240,7 @@ where
             salt(),
             keypair_to_account(caller),
             gas_limit,
-            storage_deposit_limit.map(ContractsBalanceOf::<Runtime>::from),
+            storage_deposit_limit,
         );
 
         let account_id_raw = match &result.result {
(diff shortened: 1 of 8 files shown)


crates/e2e/src/drink_client.rs
```diff
@@ -199,11 +209,17 @@ where
...
         AccountId: Clone + Send + Sync + From<[u8; 32]> + AsRef<[u8; 32]>,
         Hash: Copy + From<[u8; 32]>,
-        Runtime: RuntimeT,
-        E: Environment<AccountId = AccountId, Balance = u128, Hash = Hash> + 'static,
+        Runtime: RuntimeT + pallet_balances::Config + pallet_contracts::Config,
+        E: Environment<
+                AccountId = AccountId,
+                Balance = ContractsBalanceOf<Runtime>,
+                Hash = Hash,
+            > + Send
```
SkymanOne, line 217: Why do you need `Send` here?

SkymanOne (approved):

crates/e2e/src/drink_client.rs
ascjones: We don't, I mistakenly added it. Removed!

pgherveou (approved):

Issue #1820, opened 2023-06-20
Persist hash function in `Environment`


SkymanOne:
Originally raised by @cmichi 

By default, ink! and corresponding tooling assumes that `blake2` hasher is used. This is because it is default hasher set in `pallet-system` in substrate. It is quite possible that a chain can use a different hasher (like `sha256`or `poseidon` for ZKP).

This will cause an incorrect calculation of the metadata.

We need to introduce additional field in `Environment` trait that will persist a specific hasher used across the toolstack

```[tasklist]
### Tasks
- [ ] Introduce a hasher type/field in `Environment` trait
- [ ] Open issue in `cargo-contract` to use custom hasher or use the one specified in `Environment`
- [ ] Open issue in `contracts-ui` to validate that the correct hasher has been used
```

PR #2031, opened 2023-12-21
Custom signature topic

### LINKED ISSUE #1838: Events `2.0`: manually specify signature topic
> How about allowing to manually specify the signature topic? If for whatever reason, the exact definition of types can not be used, we can at least allow to specify the topic manually. We are doing that with custom selectors, why not do it with the event signatures?

_Originally posted by @SkymanOne in https://github.com/paritytech/ink/pull/1827#discussion_r1252290352_

How it might look:

```rust
#[derive(ink::Event)]
#[ink(signature_topic = 0x000...]
pub struct Event {}
```
The type of the signature topic should be `[u8; 32]`, although we may need to consider making this configurable to `Environment::Hash`.

SkymanOne:
## Summary
Closes #1838 and follows up #1827

Introduces optional attribute `signature_topic` to the `#[ink::event]` and `#[ink(event)]`  that takes 32-byte hash string to specify a signature topic for the specific event manually.

## Description
It introduces a new macro attribute `#[ink(signature_topic = _)]`, part of `Event`, and the same argument can be used with inline events, but subject to discussion in #2046.

### Example 

```rust
#[derive(ink::Event, scale::Encode)]
#[ink(signature_topic = "1111111111111111111111111111111111111111111111111111111111111111")]
pub struct MyCustomSignatureEvent {
     pub field: u32,
     pub topic: [u8; 32],
}

assert_eq!(Some([17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17]),
     <MyCustomSignatureEvent as ink::env::Event>::SIGNATURE_TOPIC)
```

If `signature_topic` is not specified, then topic will be calculated automatically, otherwise the hash string will be used.
This attribute is appended automatically in `#[ink::event]`. 

<details closed>
<summary> <del>Breaking change</del> </summary>
<be>

## As part of old discussion

However, when we manually use `#[derive(Event, scale::Encode)]`, we must also implement `GetSignatureTopic` as implied above. Therefore, we must also use `[ink::signature_event]`.
```rust
#[derive(::ink::Event, ::scale::Encode, ::scale::Decode)]
#[cfg_attr(feature = "std", derive(::ink::EventMetadata))]
#[ink::signature_topic]
pub struct Flipped { .. }
```

However, if `#[ink(anonymous)]` is used, then `#[derive(event)` will derive a blank implementation of `GetSignatureTopic` that returns `None`, and `[ink::signature_event]` should not be used.

```rust
#[derive(::ink::Event, ::scale::Encode, ::scale::Decode)]
#[cfg_attr(feature = "std", derive(::ink::EventMetadata))]
#[ink(anonymous)]
pub struct Flipped { .. }
```

### Rationale

This approach was followed due to the philosophy behind the `[ink::event]` macro that simply appends necessary proc macro attributes. Therefore, the user should achieve the same result manually.

Earlier, I have attempted to generate an implementation for `GetSignatureTopic` [inside `[ink::event]`](https://github.com/paritytech/ink/blob/48cec1f4a4679e2f8e851c991dad744450b88071/crates/ink/codegen/src/generator/event.rs#L56) but this prevented from using `#[derive(Event)]` independently.  Therefore, if use `signature_topic = S` arg inside `#[ink::event]` macro that simply passes the string arg into `#[ink::signature_topic(hash = S)]`

Specifically, `GetSignatureTopic` provides the flexibility to provide custom calculation of `SIGNATURE_TOPIC` without the need to implement `Event` trait.

 ```rust
impl GetSignatureTopic for Flipper {
    const SIGNATURE_TOPIC: Option<[u8; 32]> = <literal value or const fn>
}

#[derive(::ink::Event, ::scale::Encode, ::scale::Decode)]
#[cfg_attr(feature = "std", derive(::ink::EventMetadata))]
pub struct Flipped { .. }
```

So signature topic can be defined in 6 ways:
1. Automatically calculated using `#[ink(event)` or `#[ink::event]` attributes
2. Automatically calculated with `#[ink::signature_topic]` when `#[derive(Event)` is used directly
3. Manually specified with `#[ink(event, signature_topic = s)` for local events
4. Manually specified with `#[ink::event(signature_topic = s)]` for shared events
5. Manually specified with `#[ink::signature_topic(hash = s)]` for shared events with `#[derive(Event)]`
6. Manually specified by implementing `GetSignatureTopic` for the event struct directly for shared events

</details>

No breaking changes have been introduced. The `signature_topic` argument is totally optional.

### CODE REVIEW

File: crates/ink/ir/src/ir/event/signature_topic.rs [hunk 1]
```diff
@@ -0,0 +1,157 @@
...
+        Self { topic: *value }
+    }
+}
+
+impl TryFrom<&ast::MetaValue> for SignatureTopicArg {
+    type Error = syn::Error;
+
+    fn try_from(value: &ast::MetaValue) -> Result<Self, Self::Error> {
+        if let ast::MetaValue::Lit(lit) = value {
+            if let syn::Lit::Str(s) = lit {
```
ascjones, line 54: Since we are decoding the hex string here, we should rather use `Lit::BytStr` as we do for [selector](https://github.com/paritytech/ink/blob/bcbfa6697b5a6f59c4b27f3082dd6f6a20b0c236/crates/ink/ir/src/ir/selector.rs#L194).

Then can do `signature_topic = 0x1111111111111111111111111111111111111111111111111111111111111111`.

So we only support the byte literal and not strings.

crates/env/src/event.rs
```diff
@@ -215,3 +209,16 @@ pub trait Event: scale::Encode {
...
+
+/// Getter that returns the signature topic for the specific event.
+///
+/// It can be automatically calculated or manually specified.
+///
+/// The unique signature topic of the event. `None` for anonymous events.
+///
+/// Usually this is calculated using the `#[derive(ink::Event)]` derive, which by
+/// default calculates this as `blake2b("Event(field1_type,field2_type)")`
+pub trait GetSignatureTopic {
```
ascjones, line 221: I'm not convinced of the need for this additional trait, it is not useful on its own.

crates/ink/codegen/Cargo.toml
```diff
@@ -30,6 +30,7 @@ blake2 = { workspace = true }
 heck = { workspace = true }
 scale = { workspace = true }
 impl-serde = { workspace = true, default-features = true }
+hex = { workspace = true }
```
ascjones, line 33: We already use `impl_serde` (renamed to serde_hex`) for hex encoding/decoding, no need to have two different libs for this.

ascjones (commented):
Instead of adding a new attribute macro `#[::ink::signature_topic]`, we should look at extending the mechanism we already have for configuring the signature topic.

At the moment we have `#[ink(anonymous)]` which allows configuring the `#[derive(Event)]`. This is an idiomatic way to configure a derive macro, so imagine we introduce a new derive configuration attribute  `#[ink(signature_topic = 0x...)]`. To follow this idea further, `#[ink(anonymous)]` is just an alias for `#[ink(signature_topic = None)]` (or whatever syntax we choose).

By combining these we might be able to simplify the implementation here, certainly we won't need the additional trait.

crates/env/src/event.rs
SkymanOne: It is useful when devs want to specify custom logic for implementing the generation of the topic signature (e.g. using a different hashing scheme or prepending it with some index). They can implement only the `GetSignatureTopic` trait instead of the whole `Event` trait.

It is also hard to support both otherwise `derive(Event)` proc macro, and nested `ink(event)` attr with the configuration options. 

The alternative would be to eliminate individual derive macros in favour of a single `#[ink::event]` proc macro that would allow the generation of the custom signature topic "in-place".

The other approach is to duplicate logic, and parse nested `#ink(signature_topic)` inside `event_derive`, similar to how it's done with `anonymous`.

ascjones: > It is useful when devs want to specify custom logic for the generation of the topic signature (e.g. using a different hashing scheme or prepending it with some index). They can implement only the GetSignatureTopic trait instead of the whole Event trait.

As per your example in the description:
```
impl GetSignatureTopic for Flipper {
   const SIGNATURE_TOPIC: Option<[u8; 32]> = <literal value or const fn>
}
```
I don't see how this adds anything over and above specifying it directly in the `signature_topic = <literal>`. It is a `const` so in all cases it is calculated at compile time. Sure it is a bit more complicated to add a `const fn` there because of attribute parsing, but I don't think we need to necessarily support that to start with, the user can just perform the equivalent in a custom macro if they really want.

> It is also hard to support both otherwise derive(Event) proc macro, and nested ink(event) attr with the configuration options.

I don't think we need to add anything to the legacy `#[ink(event)]` inline macro, I left that there for backwards compatibility, if they want to `signature_topic` they can just switch to `ink::event`. 

> Nonetheless, I am in favour of multiple small traits serving specific logic. GetSignatureTopic generates the signature topic in the event generation pipeline and, therefore, is self-sufficient. 

I agree with the principle of having multiple smaller traits where clear responsibilities can be separated. In this case, however, they coherently belong together. The trait was originally called `Topics` (which I renamed to `Event` (mistakenly?) because it is responsible for the topics of an event, and as part of that I moved the signature topic from being returned from `topics` to a `const`.

> Since we want to allow developers to customise the topic, we want to allow them to customise its generation as well

I argue above that with the `signature_topic = <literal>`  that they are able to customise it's generation, without having to understand about underlying traits.

Also be mindful that these traits are a public API and we want to keep it as simple as possible, and only add new things if they are absolutely necessary.

ascjones:
> So signature topic can be defined in 6 ways:

I think it should be:just 3 ways:

Automatically calculated using #[ink(event) or #[ink::event] attributes
~~Automatically calculated with #[ink::signature_topic] when #[derive(Event) is used directly~~
~~Manually specified with #[ink(event, signature_topic = s) for local events~~
Manually specified with #[ink::event(signature_topic = s)] for shared events
~~Manually specified with #[ink::signature_topic(hash = s)] for shared events with #[derive(Event)]~~
^^ Instead: `#[derive(Event)]` with `#[ink(signature_topic = 0x)]`
~~Manually specified by implementing GetSignatureTopic for the event struct directly for shared events~~

So in essence we have the 

`#[ink::event(signature_topic = 0x...)]]`

Which expands to 

```
#[derive(Event)]
#[ink(signature_topic = 0x...)]
```
The derive then picks up the `ink` attribute, so exactly how `anonymous` currently works (IIRC).

crates/ink/ir/src/ir/event/signature_topic.rs
SkymanOne: This wouldn't work since `0x` format is only for integers, and the max value is `0xffffffffffffffffffffffffffffffff`.
using `b"0x1111111111111111111111111111111111111111111111111111111111111111"` wouldn't also work because it parses characters as ASCI. Therefore, using raw strings allows to specify hashes directly, which can also be copy-pasted from the metadata of the the other contract.

crates/ink/tests/ui/event/fail/conficting_attributes_inline.rs
```diff
@@ -0,0 +1,7 @@
+#[ink::event]
+#[ink(anonymous, signature_topic = "1111111111111111111111111111111111111111111111111111111111111111")]
```
ascjones, line 2: This hits a different error `Invalid `#[ink(anonymous)]` attribute: multiple arguments not allowed`, and not the clause you added which has a more useful error message. Maybe separate `ink` attributes?

crates/ink/tests/ui/event/fail/conficting_attributes.stderr
```diff
@@ -0,0 +1,13 @@
+error: Only a single `#[ink(anonymous)]` attribute allowed.
```
ascjones, line 1: Here again hitting a misleading error message

crates/ink/macro/src/lib.rs
```diff
@@ -1399,13 +1410,33 @@ synstructure::decl_derive!(
...
+    /// assert_ne!(<MyEvent as ink::env::Event>::SIGNATURE_TOPIC, <other_event::MyEvent as ink::env::Event>::SIGNATURE_TOPIC);
     /// ```
     ///
+    /// ## Custom Signature
+    ///
+    /// Sometimes it is useful to specify the custom signature topic.
+    /// For example, when the event definition from the other contract is not accessible.
+    ///
+    /// The macro provides `#[ink(signature_topic = _)]` nested macro that allows to provide
+    /// 32 byte hash string of the custom signature topic.
```
ascjones, line 1422: It doesn't have to be a hash string, it can be any 32 bytes.

crates/ink/macro/src/lib.rs
```diff
@@ -677,6 +680,15 @@ pub fn trait_definition(attr: TokenStream, item: TokenStream) -> TokenStream {
 ///     #[ink(topic)]
 ///     pub topic: [u8; 32],
 /// }
+/// // Setting `signature_topic = <hash_string>` specifies custom signature topic.
```
ascjones, line 683:
```suggestion
/// // Setting `signature_topic = <hex_string>` specifies custom signature topic.
```
Ah maybe you mean hex and not hash string?

crates/ink/ir/src/ir/event/signature_topic.rs
```diff
@@ -0,0 +1,125 @@
...
+                        format_err_spanned!(
+                            lit,
+                            "`signature_topic` has invalid hex string",
+                        )
+                    })?
+                    .try_into()
+                    .map_err(|e: Vec<u8>| {
+                        format_err_spanned!(
+                            lit,
+                            "`signature_topic` is expected to be 32-byte hash. Found {} bytes",
```
ascjones, line 56: Hex instead of hash?

crates/ink/ir/src/ir/event/config.rs
```diff
@@ -44,27 +48,56 @@ impl TryFrom<ast::AttributeArgs> for EventConfig {
                         "expected a bool literal for `anonymous` ink! event item configuration argument",
                     ));
                 }
+            } else if arg.name.is_ident("signature_topic") {
+                if anonymous.is_some() {
+                    return Err(format_err_spanned!(
+                        arg,
+                        "cannot specify `signature_topic` with `anonymous` in ink! event item configuration argument",
```
ascjones, line 55: I guess this is the error we want to see in those ui tests

crates/ink/codegen/src/generator/event.rs
```diff
@@ -15,27 +15,36 @@
...
     fn generate_code(&self) -> TokenStream2 {
         let item = self.item.item();
         let anonymous = self
             .item
             .anonymous()
             .then(|| quote::quote! { #[ink(anonymous)] });
+        let signature_topic = self
+            .item
+            .signature_topic_hash()
+            .map(|hash| quote::quote! { #[ink(signature_topic = #hash)] });
```
ascjones, line 38: `hash` again here

crates/ink/ir/src/ir/attrs.rs
```diff
@@ -278,7 +278,17 @@ impl InkAttribute {
...
             if let ir::AttributeArg::Selector(selector) = arg.kind() {
-                return Some(*selector)
+                return Some(*selector);
+            }
+            None
+        })
+    }
+
+    /// Returns the signature topic of the ink! attribute if any.
+    pub fn signature_topic_hash(&self) -> Option<String> {
```
ascjones, line 288: `hash`

crates/ink/ir/src/ir/attrs.rs
```diff
@@ -363,6 +373,8 @@ pub enum AttributeArgKind {
     /// `#[ink(selector = _)]`
     /// `#[ink(selector = 0xDEADBEEF)]`
     Selector,
+    /// `#[ink(signature_topic = "DEADBEEF")]`
```
ascjones, line 376: This is an invalid `signature_topic`, should be 32 bytes? We could do with a ui test for that.

crates/ink/ir/src/ir/attrs.rs
```diff
@@ -418,6 +430,8 @@ pub enum AttributeArg {
     /// - `#[ink(selector = _)]` Applied on ink! messages to define a fallback messages
     ///   that is invoked if no other ink! message matches a given selector.
     Selector(SelectorOrWildcard),
+    /// `#[ink(signature_topic = "DEADBEEF")]`
```
ascjones, line 433: Also invalid

crates/ink/ir/src/ir/event/config.rs
```diff
@@ -44,27 +48,56 @@ impl TryFrom<ast::AttributeArgs> for EventConfig {
...
     }
 
     /// Returns the anonymous configuration argument.
     pub fn anonymous(&self) -> bool {
         self.anonymous
     }
+
+    /// Returns the manually specified signature topic.
+    pub fn signature_topic_hash(&self) -> Option<String> {
+        self.signature_topic_hash.clone()
```
ascjones, line 101: Nitpick: not a big deal at all but couldn't this return `Option<&str>` and then just do `.as_ref()`.

crates/ink/ir/src/ir/event/signature_topic.rs
```diff
@@ -0,0 +1,125 @@
...
+                    return Err(format_err!(
+                        arg.span(),
+                        "encountered duplicate ink! event configuration argument"
+                    ));
+                }
+                signature_topic = Some(SignatureTopicArg::try_from(&arg.value)?);
+            } else {
+                return Err(format_err_spanned!(
+                    arg,
+                    "encountered unknown or unsupported ink! storage item configuration argument",
```
ascjones, line 103:
```suggestion
                    "encountered unknown or unsupported ink! event item configuration argument",
```

crates/ink/macro/src/lib.rs
SkymanOne: Yes, I meant hex

crates/ink/ir/src/ir/attrs.rs
SkymanOne: This is just a comment. The string length should be checked. But I agree that it wouldn't hurt adding a UI test for this

crates/ink/tests/ui/event/fail/conficting_attributes_inline.rs
SkymanOne: Well, technically, we don't allow to have multiple arguments inside `#[ink()]`. So this is a valid error. And what clause are you talking about specifically?

crates/ink/tests/ui/event/fail/conficting_attributes.stderr
SkymanOne: What's misleading here? You can't have multiple nested `#[ink()]` attributes. it doesn't matter if they are the same or different

deuszx:
A general question - why do we need it? I'm not convinced original argument "_we have this for selectors so we can have it for events_" . For selectors it makes sense for backwards compatibility - imagine upgrading a contract and want to maintain compatibility with the rest of the ecosystem. For events, signature is based on the event's structure only, right? So there's no problem with backwards compatibility - if event has the same structure, old deserializers will still work.

The way I see it, this feature can actually trick clients to observe, fetch, deserialize, etc. events that are not the ones they're interested in - example I'm observing `PSP22::Transfer` events with signature `XYZ` and there are contracts using custom signature topic `XYZ` even though they're not `PSP22` contracts or not emitting `PSP22::Transfer` events.

SkymanOne:
> A general question - why do we need it? I'm not convinced original argument "_we have this for selectors so we can have it for events_" . For selectors it makes sense for backwards compatibility - imagine upgrading a contract and want to maintain compatibility with the rest of the ecosystem. For events, signature is based on the event's structure only, right? So there's no problem with backwards compatibility - if event has the same structure, old deserializers will still work.
> 
> The way I see it, this feature can actually trick clients to observe, fetch, deserialize, etc. events that are not the ones they're interested in - example I'm observing `PSP22::Transfer` events with signature `XYZ` and there are contracts using custom signature topic `XYZ` even though they're not `PSP22` contracts or not emitting `PSP22::Transfer` events.

This is useful for two reasons:
1. Backward compatibility as you said. The hash function will be configurable in the future, there is an issue #1820 
2. The event signature topic is calculated by hashing `Ident(field_types)`. If the develop wants to use ident definition from the other contract that is publically accessible, and the event definition is conflicting with the existing one, they can just rename it and specify the signature topic of the original one.

crates/ink/ir/src/ir/attrs.rs
ascjones: Just this method name now which should not have `hash` in it.

ascjones (approved):
LGTM.

(see my final nitpick above, also conflict needs resolving)

PR #2038, opened 2024-01-02
Use pallet-contracts-uapi

pgherveou:
We now have a `pallet-contracts-uapi` crate (see issue https://github.com/paritytech/polkadot-sdk/issues/2189)  that expose all the host functions that a contract can import.

This PR remove the functions that were originally defined in [https://github.com/paritytech/ink/blob/master/crates/env/src/engine/on_chain/ext.rs](ext.rs) and use the one defined in this crate instead.

Some shared structs like Return and CallFlags are also now imported from this crate instead of being redefined in ink!

Note: Depends on the next crates.io release of `pallet-contracts-uapi`


crates/env/src/engine/mod.rs
```diff
@@ -193,7 +195,11 @@ mod decode_instantiate_result_tests {
             DefaultEnvironment,
             TestContractRef,
             Result<TestContractRef, ContractError>,
-        >(Err(Error::CalleeReverted), out_address, out_return_value)
+        >(
+            Err(Error::ReturnError(ReturnErrorCode::CalleeReverted)),
```
ascjones, line 199: And here

crates/env/src/lib.rs
```diff
@@ -102,16 +102,18 @@ mod tests;
 #[doc(inline)]
 pub use self::engine::off_chain::test_api as test;
 
+pub use pallet_contracts_uapi::{
```
ascjones, line 105: How will the docs look, should we do `#[doc(inline)]` for these types?

ascjones (commented):
All in alllooks pretty good. 

It would still be interesting to know in a couple of cases (e.g. `call-builder-return-value` and `ontract-storage`) why the contracts are 250 bytes or so bigger. I remember you looking into it previosly and overall the diffs don't seem as big now, and in many cases small reductions in size.

crates/env/src/lib.rs
pgherveou: ah did not know we could do that, let me  test to see how the gen docs looks like

crates/env/src/engine/mod.rs
```diff
@@ -89,7 +91,7 @@ where
             let output = <R as ConstructorReturnType<ContractRef>>::ok(contract_ref);
             Ok(Ok(output))
         }
-        Err(EnvError::CalleeReverted) => {
+        Err(EnvError::ReturnError(ReturnErrorCode::CalleeReverted)) => {
```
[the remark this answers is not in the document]
pgherveou: you can't do that in a pattern matching can you?

ascjones (approved):
LGTM

pgherveou:
mm pallet-contracts-uapi@5.0.0 is actually a buggy version, thus the tests are not passing here we will have to wait for the next release of the crate

pgherveou:
@ascjones, I published the package under `ink-pallet-contracts-uapi` so we can merge it now if that's ok for you

PR #2045, opened 2024-01-05
Linter: Warn about using non-fallible APIs for `Lazy` and `Mapping`

### LINKED ISSUE #2017: Linter: Warn about using non-fallible APIs for `Lazy` and `Mapping`
#1910 has introduced a safe API for reading and writing from `Mapping` and `Lazy` to and from storage. This API is intended to use by default when there are no explicit checks for encoded size of the object.

A new lint should check these cases and raise warnings when the following functions are used:

* [`Mapping.get`](https://paritytech.github.io/ink/ink_storage/struct.Mapping.html#method.get)
* [`Mapping.insert`](https://paritytech.github.io/ink/ink_storage/struct.Mapping.html#method.insert)
* [`Mapping.take`](https://paritytech.github.io/ink/ink_storage/struct.Mapping.html#method.take)
* [`Lazy.get`](https://paritytech.github.io/ink/ink_storage/struct.Lazy.html#method.get)
* [`Lazy.set`](https://paritytech.github.io/ink/ink_storage/struct.Lazy.html#method.set)

And suggest to replace them to their safe `try_` alternatives. The warnings should be raised only if there are no explicit checks for encoded size of the inserted key/value in user's code.

jubnzv:
## Summary
Closes #2017
- [n] y/n | Is it dependent on the specific version of `cargo-contract` or `pallet-contracts`?

Add a new lint for `ink_linting_extra`. 

## Description
This PR adds a new lint for `ink_linting_extra`. It warns about usage of the non-fallible API for `Mapping` and `Lazy` which potentially can lead to overflow of the static buffer.


linting/extra/src/non_fallible_api.rs
```diff
@@ -0,0 +1,305 @@
...
+    ///
+    /// ```rust
+    /// #[ink(storage)]
+    /// pub struct Example { map: Mapping<String, AccountId> }
+    /// ```
+    ///
+    /// The following usage of the non-fallible API is unsafe:
+    ///
+    /// ```rust
+    /// pub fn test(&mut self, a: String, b: AccountId) {
```
xermicus, line 88: I think this exact case can actually not panic.

The input for this looks like this.
`[u8; 4]` + `String` + `[u8; 32]` = 36 plus the length of the encoded String (first 4 bytes are the message selector)

The `map.insert()` will encode:
`mapping base key` + `String` + `[u8; 32]`)  = 36 plus the length of the encoded String (the base key is 4 bytes)

Thus, if the message input can be decoded using the static buffer, that the encoded data is guaranteed to fit into the static buffer. IOW, any input to this message that would fail the storage operation will revert prior to that anyways (during the decoding of the input), rendering this usage of the fallible API unnecessary (and adding overhead).

To make the example valid you can for example push to the supplied string to grow it beyond the capacity of the static buffer. Or you could remove the `b: AccountID` function argument and just use the `AccountId` of the caller.

xermicus (commented):
I think this goes into the right direction.

linting/extra/src/non_fallible_api.rs
jubnzv: That's a good point. Fixed.

xermicus (approved):
LGTM

PR #2047, opened 2024-01-08
Use existing node to run e2e tests in dev

pgherveou:
Add an option to run e2e tests with an existing node in dev.
This is useful when you want to easily scan the debug logs of pallet-contracts to debug an e2e test

### DIFF
--- a/crates/e2e/macro/src/config.rs
+++ b/crates/e2e/macro/src/config.rs
@@ -76,7 +76,7 @@ impl E2EConfig {
         self.backend.clone()
     }
 
-    /// The URL to the running node, default value can be overriden with
+    /// The URL to the running node, default value can be overridden with
     /// `CONTRACTS_NODE_URL`. If no URL is provided, then a default node instance will
     /// be spawned per test.
     pub fn node_url(&self) -> Option<String> {
--- a/crates/e2e/macro/src/config.rs
+++ b/crates/e2e/macro/src/config.rs
@@ -76,7 +76,7 @@ impl E2EConfig {
         self.backend.clone()
     }
 
-    /// The URL to the running node, default to `CONTRACTS_NODE_URL` env is not set.
+    /// The URL to the running node, default to `CONTRACTS_NODE_URL` if not set.
     /// If no URL is provided, then a default node instance will be spawned per test.
     pub fn node_url(&self) -> Option<String> {
         self.node_url
--- a/crates/e2e/macro/src/config.rs
+++ b/crates/e2e/macro/src/config.rs
@@ -46,8 +46,7 @@ pub struct E2EConfig {
     /// The type of the architecture that should be used to run test.
     #[darling(default)]
     backend: Backend,
-    /// The URL to the running node. If not set then a default node instance will be
-    /// spawned per test.
+    /// The URL to the running node. See [`Self::node_url()`] for more details.
     node_url: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


crates/e2e/src/node_proc.rs
```diff
@@ -33,7 +33,7 @@ use subxt::{
 
 /// Spawn a local substrate node for testing.
 pub struct TestNodeProcess<R: Config> {
-    proc: process::Child,
+    proc: Option<process::Child>,
```
pgherveou, line 36: 🤔  maybe we just rename TestNodeProcess -> TestNode if this becomes optional

crates/e2e/src/test_node.rs
```diff
@@ -32,14 +32,14 @@ use subxt::{
 };
 
 /// Spawn a local substrate node for testing.
-pub struct TestNodeProcess<R: Config> {
-    proc: process::Child,
+pub struct TestNode<R: Config> {
```
pgherveou, line 35: Renamed TestNodeProcess -> TestNode, since proc is now optional this seems like a better name

ascjones (commented):
This functionality already exists, albeit not configured by an environment variable. The node url comes from the attribute config `#[e2e_test(node_url = "ws://127.0.0.1:8000")]`.

What we could do here is to modify [this line ](https://github.com/paritytech/ink/blob/fa6dd8ed731ae05385e0836e6a29e1783623343b/crates/e2e/macro/src/codegen.rs#L67) and check for the env var which would take precedence over the config?

pgherveou:
> This functionality already exists, albeit not configured by an environment variable. The node url comes from the attribute config `#[e2e_test(node_url = "ws://127.0.0.1:8000")]`.
> 
> What we could do here is to modify [this line ](https://github.com/paritytech/ink/blob/fa6dd8ed731ae05385e0836e6a29e1783623343b/crates/e2e/macro/src/codegen.rs#L67) and check for the env var which would take precedence over the config?

Ah nice did not know that will look into your suggestion

crates/e2e/macro/src/config.rs
```diff
@@ -77,10 +76,12 @@ impl E2EConfig {
         self.backend.clone()
     }
 
-    /// The URL to the running node. If not set then a default node instance will be
-    /// spawned per test.
+    /// The URL to the running node, default to `CONTRACTS_NODE_URL` env is not set.
```
pgherveou, line 79:
```suggestion
    /// The URL to the running node, default to `CONTRACTS_NODE_URL` if not set.
```

crates/e2e/macro/src/config.rs
```diff
@@ -77,10 +76,12 @@ impl E2EConfig {
...
 
-    /// The URL to the running node. If not set then a default node instance will be
-    /// spawned per test.
+    /// The URL to the running node, default to `CONTRACTS_NODE_URL` if not set.
+    /// If no URL is provided, then a default node instance will be spawned per test.
     pub fn node_url(&self) -> Option<String> {
-        self.node_url.clone()
+        self.node_url
+            .clone()
+            .or_else(|| std::env::var("CONTRACTS_NODE_URL").ok())
```
ascjones, line 84: Should it not be the other way around, the env var has precendence so you can override the hardcoded URL?

ascjones (approved):

crates/e2e/macro/src/config.rs
```diff
@@ -77,10 +76,13 @@ impl E2EConfig {
         self.backend.clone()
     }
 
-    /// The URL to the running node. If not set then a default node instance will be
-    /// spawned per test.
+    /// The URL to the running node, default value can be overriden with
```
pgherveou, line 79:
```suggestion
    /// The URL to the running node, default value can be overridden with
```

PR #2048, opened 2024-01-08
Serialize storage keys in metadata as SCALE encoded

ascjones:
Storage keys are [SCALE encoded for storage access](https://github.com/paritytech/ink/blob/27407b5f59f24233cddc3cea59cc2213ba2c6674/crates/env/src/engine/on_chain/impls.rs#L224), so the layout keys in the metadata should also be serialized in the same way. At the moment they are BE encoded.

:warning: **BREAKING metadata change** :warning: 

Client tools which use the layout keys will need to adapt to this change by decoding from LE bytes instead of BE bytes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2767,6 +2767,7 @@ dependencies = [
  "ink_prelude 5.0.0-rc",
  "ink_primitives 5.0.0-rc",
  "linkme",
+ "parity-scale-codec",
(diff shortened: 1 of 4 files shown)


xgreenx (approved):
It seems like a breaking metadata change, but it's nice that we found it before releasing ink! 5=)

PR #2052, opened 2024-01-09
Fix the `StorageVec` type by excluding the `len_cached` field from its type info

smiasojed:
## Summary
- [n y/n | Is it dependant on the specific version of `cargo-contract` or `pallet-contracts`?

Fix the `StorageVec` type by excluding the `len_cached` field from its type info

## Description
There is an issue with the `len_cached` field of the `StorageVec` type. It is not stored in the storage but is defined in the `StorageVec` as `TypeInfo`  `Option<u32>`. This is causing a decoding error when the `StorageVec` type is decoded from storage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ascjones (commented):
I've just thought of an alternative way to do it: add `#[codec(skip)]` to the `len_cached` field. Then `CachedLen` should not require its own `TypeInfo` implementation.

ascjones (approved):

xermicus (approved):

xermicus:
Good catch

PR #2067, opened 2024-01-19
Linter: Update toolchain

jubnzv:
## Summary

Update the toolchain version and the uses of internal rustc APIs.

## Description
It would be beneficial to compile both the linter and `cargo-contract` using the same toolchain version. This approach would prevent the need to build a second toolchain [in the CI](https://github.com/paritytech/cargo-contract/actions/runs/7577624398/job/20638789124?pr=1412#step:6:152) and on user's machine when compiling a contract.

The necessity of this update because `subxt`, a dependency of `cargo-contract`, requires rustc version 1.74.0 or higher, therefore `cargo-contract` cannot be built with the toolchain used in the linter before this PR.

### DIFF
--- a/linting/extra/src/strict_balance_equality.rs
+++ b/linting/extra/src/strict_balance_equality.rs
@@ -42,7 +42,8 @@ use rustc_middle::{
         BinOp,
         Body,
         BorrowKind,
-        Constant,
+        CallReturnPlaces,
+        ConstOperand,
         HasLocalDecls,
         Local,
         Location,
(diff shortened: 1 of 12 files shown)


linting/extra/src/strict_balance_equality.rs
```diff
@@ -224,8 +224,6 @@ impl<'a, 'tcx> AnalysisDomain<'tcx> for StrictBalanceEqualityAnalysis<'a, 'tcx>
 
     const NAME: &'static str = "strict_balance_equality";
 
-    type Direction = Forward;
```
jubnzv, line 227: The default `Direction` is `Forward`, see: https://github.com/rust-lang/rust/issues/120130

cmichi (approved):

PR #2075, opened 2024-01-25
[E2E] Add `create_call_builder` for testing on-chain contracts

cmichi:
## Why?

Allows for easy writing of an E2E test that runs against an already deployed contract. Our typical examples always have the whole instantiation workflow with creating a contract instance, constructor & co.

So you can e.g. take a snapshot of Aleph Zero, upgrade your contract there to a new ink! version, and then run your E2E test suite against this updated contract to assert that everything will still work.


ascjones (commented):
Maybe we can make a simple free function which constructs from an account id, extracting this code here

```
<<Contract as ContractCallBuilder>::Type as FromAccountId<E>>::from_account_id(
    self.account_id.clone(),
)
```

https://github.com/paritytech/ink/blob/cmichi-reexport-contractcallbuilder/crates/e2e/src/contract_results.rs#L52

To replace

```
    type CallBuilderType = <Flipper as ink_e2e::ContractCallBuilder>::Type;
    let call_builder = CallBuilderType {
        account_id: acc_id
    };
```

with e.g.

```
let call_builder = create_call_builder::<Flipper>(acc_id);
```

ascjones (approved):

.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ on:
       - 'FILE_HEADER'
 
 env:
-  IMAGE:                           paritytech/ci-unified:bullseye-1.74.0
+  IMAGE:                           paritytech/ci-unified:bullseye-1.75.0
```
smiasojed, line 29: I believe we should keep the image aligned across workflows. It is also used in .github/workflows/measurements.yml.

.github/workflows/ci.yml
```diff
@@ -491,6 +491,24 @@ jobs:
...
+            cargo contract build --release --manifest-path integration-tests/flipper/Cargo.toml
+            export CONTRACT_ADDR_HEX=$(cargo contract instantiate \
+              --manifest-path integration-tests/flipper/Cargo.toml \
+              --suri //Alice --args true -x -y --output-json | \
+              jq -r .contract | xargs subkey inspect | grep -o "0x.*" | head -n1)
+            CONTRACTS_NODE_URL=ws://127.0.0.1:9944 cargo test \
+              --features e2e-tests \
+              --manifest-path integration-tests/flipper/Cargo.toml \
+              e2e_test_deployed_contract \
+              -- --ignored --nocapture
```
smiasojed, line 510: I'm just thinking if this logic could be moved to an e2e/integration test for Ink instead of the CI workflow. This change would enable us to run it locally.

cmichi: You mean as an ink! e2e test that directly uses the `cargo-contract` crates to build and instantiate?

cmichi: So my idea with the e2e test in `flipper` was to illustrate how to test a contract on-chain, this doesn't necessarily require having to build the contract beforehand.

Since the `integration-tests/` end up on `ink-examples`, I'd like to avoid putting non-user relevant things there.

PR #2077, opened 2024-01-29
`call_v2` cross-contract calls with additional limit parameters

ascjones:
**Requires version of `pallet-contracts` including `call_v2` host function. Introduced in [this commit](https://github.com/paritytech/polkadot-sdk/commit/60310de7d6402b6e44624aaa9b5db2dd848545e7)**

There is a new host function [`call_v2`](https://github.com/paritytech/polkadot-sdk/blob/bd80dcf6858a0ec9e61b5aceab933979f7258d2b/substrate/frame/contracts/uapi/src/host.rs#L184) which allows passing both `Weight` parts: `ref_time_limit` and `proof_time_limit` and the `storage_deposit_limit`. The legacy `call` function only provides the single `gas_limit` parameter, which is used as the value for `ref_time_limit`.

The [`call` extrinsic](https://github.com/paritytech/polkadot-sdk/blob/0e124a05597b5575f79c90aae212fcf48a87e2ab/substrate/frame/contracts/src/lib.rs#L702) already requires these, so this just brings the cross-contract call API into line with the extrinsic.

These parameters can be set on a call builder instance, by first calling `call_v2()` e.g.

```rust
call_builder
  .ref_time_limit(ref_time_limit)
  .proof_time_limit(proof_time_limit)
  .storage_deposit_limit(storage_deposit_limit)
  .invoke();
```
Calling `invoke` will cause the host function `ext::call_v2` to be invoked, with the supplied parameters.

To call the original `call` host function with just the `gas_limit` parameter, call the `call_v1` method on the call builder:

```rust
call_builder
  .call_v1()
  .gas_limit(ref_time_limit)
  .invoke();
```
Calling `invoke` will cause the host function `ext::call` to be invoked, with the supplied parameters.


SkymanOne:
Can you please specify the commit or tag from which the `call_v2` host function is available?

SkymanOne (commented):
Looks good, but can we incorporate the e2e test into the existing one? It also looks that the new contracts are not backward compatible with the old version of pallet?

pgherveou:
will take a look at this tomorrow 
I would have thought that we could just replace all call_v1 by call_v2 since they do essentially the same thing.

pgherveou (approved):
looks good to me. 
I just wonder if we should rename call -> call_v1 and maybe mark it as deprecated.
Should we also remove it altogether in upcoming versions of ink!?

ascjones:
> Can you please specify the commit or tag from which the `call_v2` host function is available?

It's https://github.com/paritytech/polkadot-sdk/commit/60310de7d6402b6e44624aaa9b5db2dd848545e7 I believe. I've added that commit to the description.

ascjones:
> Looks good, but can we incorporate the e2e test into the existing one?

I originally had this as part of `basic-contract-caller`, but extracted it as it didn't seem very "basic". However can do it the other way around and merge `basic-contract-caller` into `cross-contract-calls`.

cmichi:
> I just wonder if we should rename call -> call_v1 and maybe mark it as deprecated.
> Should we also remove it altogether in upcoming versions of ink!?

I thought about that too, it requires stabilizing `call_v2` in `pallet-contracts` though and a new release of Substrate. This would delay the usability of ink! quite a bit, most production chains are quite a bit behind. 

I'd still opt for it, people have to migrate code anyway and we can just add an entry to our migration guide. That way we can easier remove `.call_v1` in ink! 6.

crates/env/src/call/call_builder.rs
```diff
@@ -566,6 +702,61 @@ where
...
+            call_type: Set(CallV2 {
+                ref_time_limit,
+                ..call_type
+            }),
+            ..self
+        }
+    }
+
+    /// Sets the `proof_time_limit` part of the weight limit for the current
+    /// cross-contract call.
```
cmichi, line 723: I wonder if we can tell people here that they can ignore setting the proof limit for standalone chains.

crates/env/src/api.rs
```diff
@@ -286,6 +287,37 @@ where
...
+/// This is a low level way to evaluate another smart contract.
+/// Prefer to use the ink! guided and type safe approach to using this.
+///
+/// # Errors
+///
+/// - If the called account does not exist.
+/// - If the called account is not a contract.
+/// - If arguments passed to the called contract message are invalid.
+/// - If the called contract execution has trapped.
+/// - If the called contract ran out of gas upon execution.
```
cmichi, line 306:
```suggestion
/// - If the called contract ran out of gas, proof time, or storage deposit upon execution.
```

crates/env/src/call/call_builder.rs
cmichi, line 723:
```suggestion
    /// Sets the `proof_time_limit` part of the weight limit for the current
    /// cross-contract call.
    ///
    /// `proof_time` refers to the amount of storage in bytes that a transaction
    /// is allowed to read. You can find more info
    /// [here](https://use.ink/basics/cross-contract-calling/).
```

crates/env/src/call/call_builder.rs
```diff
@@ -566,6 +702,61 @@ where
...
+        CallBuilder {
+            call_type: Set(CallV2 {
+                proof_time_limit,
+                ..call_type
+            }),
+            ..self
+        }
+    }
+
+    /// Sets the `storage_deposit_limit` for the current cross-contract call.
```
cmichi, line 735:
```suggestion
    /// Sets the `storage_deposit_limit` for the current cross-contract call.
    ///
    /// The `storage_deposit_limit` specifies the amount of user funds that
    /// can be charged for creating storage. You can find more info
    /// [here](https://use.ink/basics/cross-contract-calling/).
```

crates/env/src/call/call_builder.rs
```diff
@@ -566,6 +702,61 @@ where
...
+        CallBuilder {
+            call_type: Set(CallV2 {
+                storage_deposit_limit: Some(storage_deposit_limit),
+                ..call_type
+            }),
+            ..self
+        }
+    }
+
+    /// Sets the `transferred_value` for the current cross-contract call.
```
cmichi, line 747:
```suggestion
    /// Sets the `transferred_value` for the current cross-contract call.
    ///
    /// This value specifies the amount of user funds that are transferred 
    /// to the other contract with this call.
```

integration-tests/cross-contract-calls/e2e_tests.rs
```diff
@@ -0,0 +1,84 @@
...
+        .instantiate("cross-contract-calls", &ink_e2e::alice(), &mut constructor)
+        .submit()
+        .await
+        .expect("basic-contract-caller instantiate failed");
+    let mut call_builder = contract.call_builder::<CrossContractCalls>();
+    let call = call_builder.flip_and_get_v1();
+
+    // when
+    let result = client
+        .call(&ink_e2e::alice(), &call)
```
SkymanOne, line 26: Can we get the actual ref time and proof time used for the call to test against?

SkymanOne (approved):
LGTM. Need also to add migration notice in ink-docs

integration-tests/cross-contract-calls/e2e_tests.rs
ascjones: Please clarify what you mean

SkymanOne: In cargo contract, when executing a call, you get the response of how much gas (ref and proof time) was actually used. Can we use that info in e2e test as well?

ascjones: The e2e `call` already does this transparently by doing a dry run first and passing the estimated limits to the actual tx.

Do you mean for the other test which passes explicit limits for the sub call? So dry run the sub call and use those values instead of the constants? We could do that, what would be the benefit though?

ascjones:
One unrelated test currently failing, should be fixed in https://github.com/paritytech/ink/pull/2101

integration-tests/cross-contract-calls/e2e_tests.rs
SkymanOne: > We could do that, what would be the benefit though?

Purely for illustrative purposes so we test that ref and proof times are set and estimated correctly.

PR #2078, opened 2024-01-29
[E2E] Rename `.call` to `.call_builder`

cmichi:
## Summary

I always stumble over this:

```rust
let mut call = contract.call::<Flipper>();
let get = call.get();
let get_res = client.call(&ink_e2e::bob(), &get).dry_run().await?;
```

To me this makes more sense:

```rust
let mut call_builder = contract.call_builder::<Flipper>();
let get = call_builder.get();
let get_res = client.call(&ink_e2e::bob(), &get).dry_run().await?;
```

### DIFF
--- a/integration-tests/call-builder-return-value/lib.rs
+++ b/integration-tests/call-builder-return-value/lib.rs
@@ -137,7 +137,7 @@ mod call_builder {
                 .submit()
                 .await
                 .expect("instantiate failed");
-            let mut call_builder = call_builder.call_builder::<CallBuilderReturnValue>();
+            let mut call_builder = contract.call_builder::<CallBuilderReturnValue>();
 
             let code_hash = client
                 .upload("incrementer", &origin)
--- a/integration-tests/call-builder-return-value/lib.rs
+++ b/integration-tests/call-builder-return-value/lib.rs
@@ -137,7 +137,8 @@ mod call_builder {
                 .submit()
                 .await
                 .expect("instantiate failed");
-            let mut call_builder_call = call_builder.call_builder::<CallBuilderReturnValue>();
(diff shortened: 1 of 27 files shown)


integration-tests/call-builder-return-value/lib.rs
```diff
@@ -137,7 +137,8 @@ mod call_builder {
                 .submit()
                 .await
                 .expect("instantiate failed");
-            let mut call_builder_call = call_builder.call::<CallBuilderReturnValue>();
+            let mut call_builder_call =
+                call_builder.call_builder::<CallBuilderReturnValue>();
```
ascjones, line 141: All these instances could do with some renaming to make it clear

ascjones: Maybe rename the `call_builder` to `contract` like we do in other examples.

SkymanOne (approved):
LGTM. Just need changes mentioned by Andrew

smiasojed (approved):

integration-tests/call-builder-return-value/lib.rs
```diff
@@ -137,8 +137,7 @@ mod call_builder {
                 .submit()
                 .await
                 .expect("instantiate failed");
-            let mut call_builder_call =
-                call_builder.call_builder::<CallBuilderReturnValue>();
+            let mut contract = call_builder.call_builder::<CallBuilderReturnValue>();
```
ascjones, line 140:
```suggestion
            let mut call_builder = contract.call_builder::<CallBuilderReturnValue>();
```

And change `let call_builder` on L135 to `let contract = `

ascjones (approved):

PR #2079, opened 2024-01-29
Reflect all possible `Mapping` failure cases in comments

cmichi:
## Summary

### DIFF
--- a/crates/storage/src/lazy/mapping.rs
+++ b/crates/storage/src/lazy/mapping.rs
@@ -138,7 +138,7 @@ where
     ///
     /// # Panics
     ///
-    /// Traps if the encoded `value` doesn't fit into the static buffer.
+    /// Traps if the the encoded `key` or `value` doesn't fit into the static buffer.
     #[inline]
     pub fn insert<Q, R>(&mut self, key: Q, value: &R) -> Option<u32>
     where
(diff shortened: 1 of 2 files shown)


crates/storage/src/lazy/mapping.rs
```diff
@@ -138,7 +138,7 @@ where
     ///
     /// # Panics
     ///
-    /// Traps if the encoded `value` doesn't fit into the static buffer.
+    /// Traps if the the encoded `key` or `value` doesn't fit into the static buffer.
```
xermicus, line 141:
```suggestion
    /// Traps if the the encoded `key` and `value` doesn't fit into the static buffer.
```

Good catch! However they are encoded together so the total size must fit to not trap, WDYT?

crates/storage/src/lazy/mapping.rs
```diff
@@ -150,13 +150,13 @@ where
...
     ///
-    /// Fails if `value` exceeds the static buffer size.
+    /// Fails if `key` or `value` exceeds the static buffer size.
     ///
     /// Returns:
     /// - `Ok(Some(_))` if the value was inserted successfully, containing the size in
     ///   bytes of the pre-existing value at the specified key if any.
     /// - `Ok(None)` if the insert was successful but there was no pre-existing value.
-    /// - `Err(_)` if the encoded value exceeds the static buffer size.
+    /// - `Err(_)` if the encoded `key` or `value` exceeds the static buffer size.
```
xermicus, line 159: Same should apply here

PR #2090, opened 2024-01-31
Restore display name for event metadata arg type

ascjones:
Restore the `displayName` property of the type of an event argument. 

It should have been ported across as part of #1827.


SkymanOne (approved):
I assume the changes in `generator/metadata.rs` are purely cosmetic?

ascjones:
> I assume the changes in `generator/metadata.rs` are purely cosmetic?

Indeed: the code is simply moved to a free function so it can be exported and used in the `ink_macro` crate.

cmichi (approved):
Thanks!

PR #2092, opened 2024-02-02
ERC-721 example: fix panic re: `approve_for` with nonexistent token

zgrannan:
## Summary

In the current ERC-721 example used in integration tests (which is the source of the erc-721 example in ink-examples), the `approve_for` function will panic with the text '"Error with AccountId"' if the token id does not exist. This PR changes the function to return an `Error::TokenNotFound` in this case instead.

I am using [Prusti](https://www.pm.inf.ethz.ch/research/prusti.html) to verify correctness properties of the contract; the presence of this bug prevents the verification of a "panic-freedom" property. 

## Description

Updated the logic and added a corresponding test.

- [x] Any dependent changes have been merged and published in downstream modules **No dependent changes**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fix the `StorageVec` type by excluding the `len_cached` field from its type info - [#2052](https://github.com/paritytech/ink/pull/2052)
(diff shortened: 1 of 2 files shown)


ascjones (approved):
LGTM

ascjones:
Need to run `cargo fmt` to fix CI

ascjones (commented):
There are a couple of other instances of calling `self.owner_of` which panic when `None`, could you fix those here too?

zgrannan:
@ascjones done

ascjones (approved):

PR #2093, opened 2024-02-02
ERC-721: Check ownership in `transfer_from`

zgrannan:
## Summary

The [ERC-721 Specification](https://eips.ethereum.org/EIPS/eip-721#specification) states that `transferFrom` should throw if `from` is not the owner of the token. The current implementation doesn't check that. As a result, it's possible for a caller `A` who is allowed to transfer `B`'s token `T`, can call `transferFrom(C, A, T)` without any problem. A side-effect is that the balance of `C` will be decremented instead of `B`. 

This PR modifies the `transferFrom()` function to perform the check and adds a corresponding test case.

- [x] Any dependent changes have been merged and published in downstream modules **No dependent changes**

### DIFF
--- a/integration-tests/erc721/lib.rs
+++ b/integration-tests/erc721/lib.rs
@@ -188,6 +188,9 @@ mod erc721 {
             to: AccountId,
             id: TokenId,
         ) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


ascjones (approved):
LGTM

PR #2100, opened 2024-02-06
Align the start pointer in ink allocator

### LINKED ISSUE #1535: Mis-aligned allocation in ink
I am trying to port the quickjs runtime into ink. It always crashes because it rely on the heap allocation alignment.

Looks like the ink's bump allocator doesn't obey Rust's alignment requirements.
Given the follow code:
```rust
let buf1 = Vec::<u8>::with_capacity(3);
let buf2 = Vec::<u64>::with_capacity(1);
info!("ptr to buf1 :{:?}", buf1.as_ptr());
info!("ptr to buf2 :{:?}", buf2.as_ptr());
info!("align of u64:{}", align_of::<u64>());
```
Runing in ink would get somthing like:
```
 ptr to buf1 :0x30022
 ptr to buf2 :0x30025
 align of u64:8
```
Where the ptr of buf2 is mis-aligned.

**Expected behavior**
It should be aligned as Rust's default allocator do. See [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d360e8a505d68acb41aadd985de0e063) which meets the [Allocator requirements](https://doc.rust-lang.org/beta/core/alloc/trait.Allocator.html#required-methods).

SkymanOne:
## Summary
Closes #1535 

Fixes alignment rules on the ink bump allocator.

## Description
Before this PR, the allocator allocated types of different order in sequential order after each other without respecting the alignment and padding rules. While this saved up tiny bits of memory usage, it misaligned the types in memory which can potentially increase direct access.

This PR ensure the valid alignment of datatypes, by correctly aligning the end pointer of the linear memory buffer.


integration-tests/flipper/lib.rs
```diff
@@ -160,5 +171,27 @@ pub mod flipper {
             assert!(matches!(get_res.return_value(), true));
             Ok(())
         }
+
+        #[ink_e2e::test]
+        async fn buffer<Client: E2EBackend>(mut client: Client) -> E2EResult<()> {
```
SkymanOne, line 176: This will be replaced by a proper E2E test

crates/allocator/src/bump.rs
```diff
@@ -194,6 +195,11 @@ impl InnerAlloc {
             Some(alloc_start)
         }
     }
+
+    /// Aligns the start pointer of the next allocation.
+    fn get_aligned_ptr(&self, layout: &Layout) -> usize {
```
athei, line 200: This is not Java.

```suggestion
    fn aligned_ptr(&self, layout: &Layout) -> usize {
```

crates/allocator/src/bump.rs
```diff
@@ -174,9 +174,10 @@ impl InnerAlloc {
     /// Note: This implementation results in internal fragmentation when allocating across
     /// pages.
     fn alloc(&mut self, layout: Layout) -> Option<usize> {
-        let alloc_start = self.next;
+        let alloc_start = self.get_aligned_ptr(&layout);
 
         let aligned_size = layout.pad_to_align().size();
```
athei, line 179: We don't need to align the size as me make sure that the `alloc_start` is always aligned.

crates/allocator/src/bump.rs
```diff
@@ -194,6 +195,11 @@ impl InnerAlloc {
             Some(alloc_start)
         }
     }
+
+    /// Aligns the start pointer of the next allocation.
+    fn align_ptr(&self, layout: &Layout) -> usize {
+        (self.next + layout.align() - 1) & !(layout.align() - 1)
```
cmichi, line 201: Please add a comment with the reasoning behind the calculation and the `AND` here.

ascjones (commented):
As a reviewer coming to this for the first time, I'd like to see at least a brief description in the PR of the problem and how this solves it.

crates/allocator/src/bump.rs
```diff
@@ -194,6 +195,18 @@ impl InnerAlloc {
...
     }
+
+    /// Aligns the start pointer of the next allocation.
+    ///
+    /// We inductively calculate the start index
+    /// of a layout in the linear memory.
+    /// - Initially `self.next` is `0`` and aligned
+    /// - `layout.align() - 1` accounts for `0` as the first index.
+    /// - the binary with the inverse of the align ensures
+    /// that the next allocated pointer address is of the power of 2.
```
smiasojed, line 206: The inverse of the expression (align - 1) creates a bitmask that is used to zero out bits, ensuring alignment according to type requirements.

crates/allocator/src/bump.rs
```diff
@@ -174,9 +174,10 @@ impl InnerAlloc {
     /// Note: This implementation results in internal fragmentation when allocating across
     /// pages.
     fn alloc(&mut self, layout: Layout) -> Option<usize> {
-        let alloc_start = self.next;
+        let alloc_start = self.align_ptr(&layout);
+
+        let aligned_size = layout.size();
 
-        let aligned_size = layout.pad_to_align().size();
```
smiasojed, line 179: Just out of curiosity, how is padding performed for fields in a struct, considering they field types have their own alignment requirements?

athei: That is not really the concern of the allocator. The allocator will just create an allocation that adheres to the `Layout` passed. For a struct the alignment will be set to the alignment for the whole struct (AFAIK the alignment of the largest type in the struct). The padding within the struct is given implicitly by the size of the `Layout`. Rust will just assume the padding when accessing the struct.

crates/allocator/src/bump.rs
athei: - The expression `self.next + layout.align() - 1` ensures that the pointer is at least as large as the next alignment boundary.
- The bitwise AND with `!(layout.align() - 1)` clears the lower bits of the result to ensure that the address is a multiple of the alignment. The `!` operator creates a mask that zeroes out the lower bits that should not contribute to the final aligned address.

athei (approved):

crates/allocator/src/bump.rs
pgherveou: maybe some concrete example will help here

integration-tests/static-buffer/lib.rs
```diff
@@ -75,5 +86,34 @@ pub mod static_buffer {
...
+            // then
+            let get = call_builder.buffer();
+            let get_res = client.call(&ink_e2e::bob(), &get).submit().await?;
+
+            let debug_msg = get_res.debug_message();
+            let msgs: Vec<&str> = debug_msg.split('\n').collect();
+            let ptr1 = u64::from_str_radix(msgs[0].trim_start_matches("0x"), 16).unwrap();
+            let ptr2 = u64::from_str_radix(msgs[1].trim_start_matches("0x"), 16).unwrap();
+            let align = u64::from_str_radix(msgs[2], 10).unwrap();
+
```
pgherveou, line 112: This seems a bit hacky to rely on the debug_message here. 
Can't we just do that in the ink! message itself and assert it does not panic, or return the values and assert here?

pgherveou:
stumbled upon this 
https://github.com/paritytech/ink/blob/da598d3bb04acd145b246f01061165a636494018/crates/allocator/src/bump.rs#L17-L18

maybe we can remove the reference to wee_alloc in this PR since you are touching this file

PR #2101, opened 2024-02-07
 `[E2E]` change node url backend config

ascjones:
Modifies the e2e configuration for backend node from `node_url = "<url>"` to `backend(node(url = "<url>"))`.

The previous configuration allowed `node_url` to be set even if using `backend(runtime_only)` which uses no node at all so the url is not used at all.

Follows up https://github.com/paritytech/ink/pull/2047

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [E2E] Rename `.call` to `.call_builder` ‒ [#2078](https://github.com/paritytech/ink/pull/2078)
 - Improve syntax for ink! e2e `runtime_only` attribute argument - [#2083](https://github.com/paritytech/ink/pull/2083)
 - [E2E] Remove `additional_contracts` parameter [#2098](https://github.com/paritytech/ink/pull/2098)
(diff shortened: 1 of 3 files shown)


SkymanOne:
Maybe we can make the url an optional param?
```diff
- backend(node(url = "<URL>"))
+ backend(node = "<URL>")
```

ascjones:
> Maybe we can make the url an optional param?
> 
> ```diff
> - backend(node(url = "<URL>"))
> + backend(node = "<URL>")
> ```

I'm not sure exactly how to achieve that using Darling. Would have to make `node` a property of `Backend`, but that is an enum not a struct hence the derive produces the `enum(variant)` syntax.

I've played around a little bit and can't immediately find a nice way to do that. I'm sure it is possible but I will leave it for now.

SkymanOne:
> > Maybe we can make the url an optional param?
> > ```diff
> > - backend(node(url = "<URL>"))
> > + backend(node = "<URL>")
> > ```
> 
> I'm not sure exactly how to achieve that using Darling. Would have to make `node` a property of `Backend`, but that is an enum not a struct hence the derive produces the `enum(variant)` syntax.
> 
> I've played around a little bit and can't immediately find a nice way to do that. I'm sure it is possible but I will leave it for now.

It is not critical ofc, just nice to have less verbose syntax

smiasojed (approved):

PR #2102, opened 2024-02-07
chore: remove archived projects from workflows

peetzweg:
## Description
Due to deprecation of `useink` we should remove all references to it in this code base.


ascjones (approved):

cmichi:
CI currently fails because of missing endowments for Charlie, Dave, etc. in `substrate-contracts-node`. Will be fixed soon.

PR #2103, opened 2024-02-07
:t-rex: pin `curve25519-dalek` to `4.1.1` to fix linting CI

ascjones:
CI currently broken. See https://github.com/dalek-cryptography/curve25519-dalek/issues/618 and https://github.com/dalek-cryptography/curve25519-dalek/pull/619


smiasojed (approved):
LGTM

jubnzv (approved):
Thanks! For now it's the best workaround. We cannot simply update `clippy_utils`, since it [depends](https://github.com/rust-lang/rust-clippy/blob/62dcbd672bbf1ae6fc5df793e63b849692cfa3c4/rust-toolchain#L2) on the previous nightly toolchain.

PR #2098, opened 2024-02-05
[E2E] Remove `additional_contracts` parameter

SkymanOne:
## Summary
Closes #_
- [ ] y/n | Does it introduce breaking changes?
- [ ] y/n | Is it dependant on the specific version of `cargo-contract` or `pallet-contracts`?

## Description
Removed `additiona_contracts` parameters from `ink_e2e::test` macro as this info is implied from the contract's manifest

## Checklist before requesting a review
- [x] My code follows the style guidelines of this project
- [x] I have added an entry to `CHANGELOG.md`
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules

PR #2083, opened 2024-01-30
Improve syntax for ink! e2e `runtime_only` attribute argument

davidsemakula:
## Summary
Closes #_
- [n] y/n | Does it introduce breaking changes?
- [n] y/n | Is it dependant on the specific version of `cargo-contract` or `pallet-contracts`?

Use `#[darling(word)]` to enable writing `#[ink_e2e::test(backend(runtime_only))]` instead of `#[ink_e2e::test(backend(runtime_only()))]`

## Description

- Bump darling to 0.20.4
- Update parser for ink! e2e `runtime_only` attribute argument by using `#[darling(word)]` to enable writing `#[ink_e2e::test(backend(runtime_only))]` instead of `#[ink_e2e::test(backend(runtime_only()))]`

See <https://github.com/TedDriggs/darling/pull/260> and <https://github.com/TedDriggs/darling/blob/master/examples/heterogeneous_enum_and_word.rs> for details.

## Checklist before requesting a review
- [x] My code follows the style guidelines of this project
- [x] I have added an entry to `CHANGELOG.md`
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules

PR #2060, opened 2024-01-12
Linter: Prepare the `ink_linting` crates for publishing

jubnzv:
## Summary
- [n] y/n | Does it introduce breaking changes?
- [n] y/n | Is it dependant on the specific version of `cargo-contract` or `pallet-contracts`?

Various improvements in `ink_linting` that allow to publish the its crates on crates.io.
This is necessary for https://github.com/paritytech/cargo-contract/pull/1412.

## Description
To publish `ink_linting` on crates.io, we need to prepare the Cargo workspace in `linting` and resolve all local dependencies to ensure they are also publishable.

The summary to the changes present here:
1. Published the [`parity_clippy_utils`](https://crates.io/crates/parity_clippy_utils) crate on crates.io and added it as a dependency to `ink_linting`. We had to do this, because all the dependencies must be present on crates.io, but the original `clippy_utils` are not. For more context, see: https://github.com/rust-lang/rust-clippy/discussions/12111#discussioncomment-8052237 and https://github.com/rust-lang/rust-clippy/pull/6746#issuecomment-780747522.
2. Created the separate `ink_linting_utils` crate that basically contains the existing utilities functions shared between lints and [wraps](https://github.com/paritytech/ink/pull/2060/files#diff-9f54a3c6d90b7b080f7c6a4e15a5688664421bd3a9ff2982814ab9316cce012eR33) the `parity_clippy_utils` trait.
3. Various changes in `Cargo.toml` files to make `cargo publish` happy.

The [`ink_linting`](https://crates.io/crates/ink_linting) and [`ink_linting_mandatory`](https://crates.io/crates/ink_linting_mandatory) crates are currently reserved on crates.io and could be replaced with the real ones as soon as this PR is reviewed.

## Checklist before requesting a review
- [x] My code follows the style guidelines of this project
- [x] I have added an entry to `CHANGELOG.md`
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules

PR #2106, opened 2024-02-08
Release `5.0.0-rc.1`

SkymanOne:
### Added
- Custom signature topic in Events - [#2031](https://github.com/paritytech/ink/pull/2031)
- [Linter] `non_fallible_api` lint - [#2004](https://github.com/paritytech/ink/pull/2004)
- [Linter] Publish the linting crates on crates.io - [#2060](https://github.com/paritytech/ink/pull/2060)
- [E2E] Added `create_call_builder` for testing existing contracts - [#2075](https://github.com/paritytech/ink/pull/2075)
- `call_v2` cross-contract calls with additional limit parameters - [#2077](https://github.com/paritytech/ink/pull/2077)

### Changed
- `Mapping`: Reflect all possible failure cases in comments ‒ [#2079](https://github.com/paritytech/ink/pull/2079)
- [E2E] Rename `.call` to `.call_builder` ‒ [#2078](https://github.com/paritytech/ink/pull/2078)
- Improve syntax for ink! e2e `runtime_only` attribute argument - [#2083](https://github.com/paritytech/ink/pull/2083)
- [E2E] Remove `additional_contracts` parameter [#2098](https://github.com/paritytech/ink/pull/2098)
- [E2E] change node url backend config - [#2101](https://github.com/paritytech/ink/pull/2101)

### Fixed
- Fix the `StorageVec` type by excluding the `len_cached` field from its type info - [#2052](https://github.com/paritytech/ink/pull/2052)
- Fix panic in `approve_for` in the ERC-721 example - [#2092](https://github.com/paritytech/ink/pull/2092)
- ERC-721: `transfer_token_from` now ensures the token owner is correct - [#2093](https://github.com/paritytech/ink/pull/2093)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -174,16 +174,145 @@ dependencies = [
 
 [[package]]
 name = "aquamarine"
-version = "0.3.3"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d1da02abba9f9063d786eab1509833ebb2fac0f966862ca59439c76b9c566760"
+checksum = "074b80d14d0240b6ce94d68f059a2d26a5d77280ae142662365a21ef6e2594ef"
 dependencies = [
  "include_dir",
  "itertools 0.10.5",
  "proc-macro-error",
  "proc-macro2",
  "quote",
+ "syn 2.0.48",
+]
+
+[[package]]
+name = "ark-bls12-377"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fb00293ba84f51ce3bd026bd0de55899c4e68f0a39a5728cebae3a73ffdc0a4f"
+dependencies = [
+ "ark-ec",
+ "ark-ff",
+ "ark-std",
+]
+
+[[package]]
+name = "ark-bls12-381"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c775f0d12169cba7aae4caeb547bb6a50781c7449a8aa53793827c9ec4abf488"
+dependencies = [
+ "ark-ec",
+ "ark-ff",
+ "ark-serialize",
+ "ark-std",
+]
+
+[[package]]
+name = "ark-ec"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "defd9a439d56ac24968cca0571f598a61bc8c55f71d50a89cda591cb750670ba"
+dependencies = [
+ "ark-ff",
+ "ark-poly",
+ "ark-serialize",
+ "ark-std",
+ "derivative",
+ "hashbrown 0.13.2",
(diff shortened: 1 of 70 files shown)


smiasojed (approved):
LGTM
tests failed:
error: the lock file /__w/ink/ink/Cargo.lock needs to be updated but --locked was passed to prevent this

Cargo.lock
```diff
@@ -1016,8 +1016,8 @@ dependencies = [
  "escape8259",
  "hex",
  "indexmap 2.1.0",
- "ink_env 5.0.0-rc (registry+https://github.com/rust-lang/crates.io-index)",
- "ink_metadata 5.0.0-rc (registry+https://github.com/rust-lang/crates.io-index)",
+ "ink_env 5.0.0-rc",
+ "ink_metadata 5.0.0-rc",
```
ascjones, line 1020: We should first release these two packages, then release a new `contract-transcode`, then update to that version. Also will need to release and update `contract-build` at the same time, though that has no `ink` deps itself.

We should probably add that to the release process.

ascjones (approved):

RELEASES_CHECKLIST.md
```diff
@@ -54,53 +54,58 @@ in the future.
...
-1. Do a dry run with `cargo release [new_version] -v --no-tag --no-push`
-    - `[new_version]` should be the **exact** SemVer compatible version you are attempting 
-      to release e.g. `4.0.0-alpha.3`
-      - It is possible to supply a SemVer level here e.g. `major`, `minor`, `patch` or 
-        `<pre-release-name>`, however this will automatically bump and commit the changes 
-        to the `Cargo.toml` manifests. We have already done that in an earlier step so it 
+1. Before releasing, take note of the release order:
+    - `ink_e2e` is dependent on the `cargo-contract` crates, which are dependent on the `ink` "core" crates.
+    - Therefore, first release all `ink` crates except `ink_e2e`.
+    - If there are new versions of the `cargo-contract` crates, release them.
```
ascjones, line 60: In fact, there will necessarily be a release of `cargo-contract` crates which depend on the new version of `ink`. So those should be released first before `e2e`, then we can use the released version of `contract-build` there.

It will probably work without this for the RC release, but we will need to do this for the main release.

integration-tests/e2e-runtime-only-backend/lib.rs
```diff
@@ -121,7 +121,7 @@ pub mod flipper {
 
             // when
             let call_data = vec![
-                Value::from_bytes(&contract.account_id),
+                Value::unnamed_variant("Id", [Value::from_bytes(contract.account_id)]),
```
smiasojed, line 124: I think it may have sense to understand why and when it has changed.
I also noticed that the drink client returns metadata in version V14. Shouldn't it be V15?

SkymanOne: This is the issue that we should resolve before the final release. For now, I'm just gonna push the release gracefully.

PR #2107, opened 2024-02-08
ERC-721: Failing transfer should preserve contract state

zgrannan:
## Summary

The error handling logic for the ERC-721 token implementation states:

> //! Any function that modifies the state returns a `Result` type and does not changes the
> //! state if the `Error` occurs.

Calling `transfer()` or `transfer_from` with a recipient address of `0x0` fails as expected, but it also removes the original token from its owner, contradicting the error-handling logic.

This PR resolves the issue by performing the recipient address check prior to making any changes to contract state in `transfer_token_from`. A regression test is also added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix the `StorageVec` type by excluding the `len_cached` field from its type info - [#2052](https://github.com/paritytech/ink/pull/2052)
 - Fix panic in `approve_for` in the ERC-721 example - [#2092](https://github.com/paritytech/ink/pull/2092)
 - ERC-721: `transfer_token_from` now ensures the token owner is correct - [#2093](https://github.com/paritytech/ink/pull/2093)
+- ERC-721: `transfer_token_from` fixed to preserve contract state on failure 
+[#2107](https://github.com/paritytech/ink/pull/2107)
 
 ## Version 5.0.0-rc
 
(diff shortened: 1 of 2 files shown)


DamianStraszak:
Zero address is not special in any sense in substrate, so I don't see why would we want to treat is specially.

The issues from solidity/EVM do not apply in this setting.

Can you explain why do you think the previous behavior is buggy and the current is correct?

zgrannan:
@DamianStraszak Thanks, I wasn't aware of that. I suppose one argument for having transfers to `0x0` fail is because this is what is expected behaviour of `transferFrom` in the [ERC721 spec](https://eips.ethereum.org/EIPS/eip-721#specification). I don't actually have a strong preference either way.

Regardless of the above, even without this PR, a transfer to `0x0` *already* returns an `Error::NotAllowed`. This is due to the existing code in the `add_token_to()` function:

https://github.com/paritytech/ink/blob/4fdef89a723edd411add0600bfcf5e759fef75aa/integration-tests/erc721/lib.rs#L303-L305

The problem with the current implementation is that even though `transfer` / `transferFrom` returns an `Error` when sending to `0x0`, it nonetheless modifies the contract state (by removing the token from the `from` address before calling `add_token_to()`), which seems to contradict the error handling logic described in the comments:

https://github.com/paritytech/ink/blob/4fdef89a723edd411add0600bfcf5e759fef75aa/integration-tests/erc721/lib.rs#L14-L17

The current PR just moves the check earlier, I suppose another option is to remove it entirely.

DamianStraszak:
Thanks -- indeed I missed that `add_token_to` already returns an error on `0` address. 

But if this is the case then your PR actually does not change anything, if error is returned in `add_token_to` then the call with fail and rollback anyway.

zgrannan:
Ah okay I didn't realize that returning `Err` will cause the contract to revert. I'll close the PR then.

PR #2123, opened 2024-02-21
`instantiate_v2` with additional limit parameters

ascjones:
**Requires version of `pallet-contracts` including `instantiate_v2` host function. Introduced in [this commit](https://github.com/paritytech/polkadot-sdk/commit/60310de7d6402b6e44624aaa9b5db2dd848545e7)**

Following up on https://github.com/paritytech/ink/pull/2077.

There is a new host function [`instantiate_v2`](https://github.com/paritytech/polkadot-sdk/blob/bd80dcf6858a0ec9e61b5aceab933979f7258d2b/substrate/frame/contracts/uapi/src/host.rs#L533) which allows passing both `Weight` parts: `ref_time_limit` and `proof_time_limit` and the `storage_deposit_limit`. The legacy `instantiate` function only provides the single `gas_limit` parameter, which is used as the value for `ref_time_limit`.

The [`instantiate` extrinsic](https://github.com/paritytech/polkadot-sdk/blob/0e124a05597b5575f79c90aae212fcf48a87e2ab/substrate/frame/contracts/src/lib.rs#L819) already requires these, so this just brings the cross-contract instantiation API into line with the extrinsic.

### Migration

If your target node does not yet support `instantiate_v2`, then you can continue to use the original `v1` host function, but it requires calling the `instantiate_v1` method on the builder e.g.

```
let create_params = build_create::<OtherContractRef>()
  .instantiate_v1()
  .code_hash(Hash::from([0x42; 32]))
  .gas_limit(500_000_000)
```

Otherwise if you **do not change your code, it will attempt to call the new instantiate_v2 function**, with no compiler warnings unless you are using `gas_limit` which is replaced by `ref_time_limit`.

### DIFF
--- a/crates/env/src/engine/on_chain/impls.rs
+++ b/crates/env/src/engine/on_chain/impls.rs
@@ -543,9 +543,9 @@ impl TypedEnvBackend for EnvInstance {
         }
     }
 
-    fn instantiate_contract<E, ContractRef, Args, Salt, RetType>(
+    fn instantiate_contract<E, ContractRef, Limits, Args, Salt, RetType>(
         &mut self,
-        params: &CreateParams<E, ContractRef, Args, Salt, RetType>,
+        params: &CreateParams<E, ContractRef, Limits, Args, Salt, RetType>,
     ) -> Result<
         ink_primitives::ConstructorResult<
             <RetType as ConstructorReturnType<ContractRef>>::Output,
--- a/crates/env/src/engine/on_chain/impls.rs
+++ b/crates/env/src/engine/on_chain/impls.rs
@@ -25,6 +25,8 @@ use crate::{
         CreateParams,
         DelegateCall,
         FromAccountId,
+        LimitParamsV1,
+        LimitParamsV2,
     },
     event::{
         Event,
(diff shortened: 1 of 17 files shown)


crates/env/src/engine/on_chain/impls.rs
```diff
@@ -546,7 +548,57 @@ impl TypedEnvBackend for EnvInstance {
...
+        let storage_deposit_limit = params
+            .storage_deposit_limit()
+            .map(|limit| &*scoped.take_encoded(limit));
+        let enc_code_hash = scoped.take_encoded(params.code_hash());
+        let enc_endowment = scoped.take_encoded(params.endowment());
+        let enc_input = scoped.take_encoded(params.exec_input());
+        // We support `AccountId` types with an encoding that requires up to
+        // 1024 bytes. Beyond that limit ink! contracts will trap for now.
+        // In the default configuration encoded `AccountId` require 32 bytes.
+        let out_address = &mut scoped.take(1024);
```
pgherveou, line 576: We can't make use of `AccountId::max_encoded_len()` here instead of hard coding 1024?

pgherveou (approved):

crates/env/src/engine/on_chain/impls.rs
ascjones: It's a good idea. We will need to:

1. Enable `max-encoded-len` feature
2. Add some bounds to the traits and impls
3. Derive `MaxEncodedLen` for `AccountId`

It's not that much to do, but I'm inclined to make that a separate PR to this, so we can properly test it and see e.g. contract size differences if any.

ascjones: https://github.com/paritytech/ink/pull/2128

PR #2126, opened 2024-02-26
Bump metadata version to 5

SkymanOne:
## Summary
Bumps ink! metadata version to 5

## Description
Since the introduction of Event 2.0 (#1827), and static buffer size in metadata (#1880), we need to increase the ABI metadata version to 5.

### Note
**This PR temporarily disables e2e on-chain test step in CI until the new version of `cargo-contract` including new `ink_metadata` crate is released.**

### DIFF
--- a/crates/metadata/src/lib.rs
+++ b/crates/metadata/src/lib.rs
@@ -99,7 +99,7 @@ impl InkProject {
         let mut registry = Registry::new();
 
         Self {
-            version: Default::default(),
+            version: METADATA_VERSION,
             layout: layout.into().into_portable(&mut registry),
             spec: spec.into().into_portable(&mut registry),
             registry: registry.into(),
--- a/crates/metadata/src/lib.rs
+++ b/crates/metadata/src/lib.rs
@@ -75,30 +75,12 @@ use serde::{
 ///
 /// The serialized metadata format (which this represents) is different from the
 /// version of this crate or the contract for Rust semantic versioning purposes.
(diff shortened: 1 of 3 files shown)


ascjones (commented):
What do you think about just making it a `const` uint instead of the enum?

ascjones (approved):

smiasojed (approved):

crates/metadata/src/lib.rs
```diff
@@ -131,15 +115,15 @@ impl InkProject {
...
+            version: METADATA_VERSION,
             layout,
             spec,
             registry,
         }
     }
 
     /// Returns the metadata version used by the contract.
-    pub fn version(&self) -> &MetadataVersion {
+    pub fn version(&self) -> &u64 {
```
peetzweg, line 126: Will this change result in the version being output as a number?

So `{version: 5}` instead of `{version: '5'}`?

SkymanOne: Yes, this changes the string into the integer. All good there

Rockettty: $DOP (Data Ownership Protocol) dop.org - Currently in the last rounds of their private sale ending in a day and few hours, listing on binance this month of march, the private sale price is $0.08 but don't be surprised that it will push to $5 in few days after listing because is a high demand project.
Access the private sale here: privatesale.dop.org
  access code:3656924464

PR #1359, opened 2022-08-19
Pin Github Actions Versions

sergejparity:
In order to improve our security posture with GitHub Actions usage. I've made a version pinning ether to commit hash or to specific version.

Related issues and policy:
https://github.com/paritytech/ci_cd/issues/464
https://github.com/paritytech/ci_cd/wiki/Policies-and-regulations:-GitHub-Actions-usage-policies

PR #2127, opened 2024-02-27
Pin unsafe GHA to commit SHA

smiasojed:
## Summary
Closes #_

Align to CI security policy

## Description
Align to CI security policy
https://github.com/paritytech/ink/pull/1359

- [ ] My code follows the style guidelines of this project

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,7 +128,7 @@ jobs:
           fetch-depth: 100
 
       - name: Cache
-        uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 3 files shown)


alvicsam:
It still makes sense because the threat is the same as it was 1,5 years ago. Dependabot can update actions versions even if they are pinned with sha.

alvicsam (approved):

smiasojed:
> Do you think this is a sensible policy? The pinning was done nearly 1.5 years ago and I'm not sure if it still makes sense.

@cmichi It makes sense if someone will review an action before updating it. However, if we're just updating the action to the newest SHA without reviewing it first, it doesn't.

cmichi (approved):

PR #2131, opened 2024-03-01
Split examples into public and internal

smiasojed:
## Summary
Closes #_

Split examples located in `integration-tests` directory into `public` and `internal`

## Description
Split examples to easily distinguish which are `public` and `internal`.
Public examples will be synchronised with `ink-examples` repository


ascjones (commented):
Could you describe the criteria to decide which contracts are `public` and which ones are `internal`?

cmichi:
> Could you describe the criteria to decide which contracts are `public` and which ones are `internal`?

DM's with Michi 🙈 .

Basically stuff that either was already in `ink-examples` or was added since 4.x with an educational benefit. It's quite a lot of examples now though. I think we're getting to a state where it's overwhelming. Maybe we should structure `ink-examples` better into subfolders.

smiasojed:
> Could you describe the criteria to decide which contracts are `public` and which ones are `internal`?

Internal contracts for CI should be created solely for testing purposes and should not demonstrate any use cases or best practices that cannot be observed in other public contracts. In contrast, public contracts are those that provide value for users by demonstrating API usage with best practices.

These contracts, which were already in `the ink-examples` repo, I assumed are public. This is the theory, but it is not so black and white. For example, I am not sure about `call-builder-return-value`. I have added this as an example of `call-builder` API usage, but it is also used in other contracts. Now I think that I should exclude it from public contracts. What do you think?

cmichi:
> For example, I am not sure about `call-builder-return-value`. I have added this as an example of `call-builder` API usage, but it is also used in other contracts. Now I think that I should exclude it from public contracts. What do you think?

Looking at it again, I agree. I would remove it from the public examples.

ascjones:
> Internal contracts for CI should be created solely for testing purposes and should not demonstrate any use cases or best practices that cannot be observed in other public contracts. In contrast, public contracts are those that provide value for users by demonstrating API usage with best practices.

By that criteria we should include the `events` example in the public folder since that demonstrates using the new `#[ink::event]` syntax for events defined across multiple crates.

ascjones (approved):
LGTM

PR #2132, opened 2024-03-04
Fix Typo: Use 'Proof Size' Instead of 'Proof Time'

pgherveou:


crates/env/src/call/call_builder.rs
```diff
@@ -106,8 +106,8 @@ where
 
     /// Returns the chosen proof time limit for the called contract execution.
     #[inline]
-    pub fn proof_time_limit(&self) -> u64 {
-        self.call_type.proof_time_limit
+    pub fn proof_size_limit(&self) -> u64 {
```
ascjones, line 109: Code comment needs updating too

ascjones (approved):
Should have noticed this when I was adding this API :see_no_evil:

SkymanOne:
Why do you need to bump ink version? We can release it as part of stable 5.0.0 release

ascjones:
Can we have a more descriptive title for this PR?

smiasojed (approved):

cmichi (approved):

PR #2149, opened 2024-03-11
Add Hash trait to Selector struct

0xLucca:
## Summary
Closes #_

Add Hash trait to Selector struct

## Description

With this implementation, you can now use Selector as a key in hash maps or hash sets. The hash method hashes the underlying bytes of the selector array, making it suitable for hashing.

### DIFF
--- a/crates/metadata/src/specs.rs
+++ b/crates/metadata/src/specs.rs
@@ -1094,7 +1094,8 @@ where
 }
 
 /// The 4 byte selector to identify constructors and messages
-#[derive(Debug, Default, PartialEq, Eq, derive_more::From, JsonSchema, Hash)]
+#[cfg_attr(feature = "std", derive(Hash))]
+#[derive(Debug, Default, PartialEq, Eq, derive_more::From, JsonSchema)]
 pub struct Selector(#[schemars(with = "String")] [u8; 4]);
 
 impl serde::Serialize for Selector {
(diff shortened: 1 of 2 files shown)


crates/metadata/src/specs.rs
```diff
@@ -1091,7 +1094,7 @@ where
 }
 
 /// The 4 byte selector to identify constructors and messages
-#[derive(Debug, Default, PartialEq, Eq, derive_more::From, JsonSchema)]
+#[derive(Debug, Default, PartialEq, Eq, derive_more::From, JsonSchema, Hash)]
```
ascjones, line 1097: Shouldn't the `Hash` derive also be behind the `std` feature?

0xLucca: Yes you are right. Something like this may work:
```rust
/// The 4 byte selector to identify constructors and messages
#[cfg_attr(feature = "std", derive(Hash))]
#[derive(Debug, Default, PartialEq, Eq, derive_more::From, JsonSchema)]
pub struct Selector(#[schemars(with = "String")] [u8; 4]);
...
```

ascjones (approved):

cmichi:
>With this implementation, you can now use Selector as a key in hash maps or hash sets. 

Was the intention to have this only available for testing contracts in `std` and not on-chain?

0xLucca:
> >With this implementation, you can now use Selector as a key in hash maps or hash sets. 
> 
> 
> 
> Was the intention to have this only available for testing contracts in `std` and not on-chain?

At least for my use case it is. Im using the ink_metadata crate directly

PR #2155, opened 2024-03-13
Add parallel tests execution to CI

smiasojed:
## Summary
Closes #_

## Description
Add `nextest` to the workflow where parallel test execution makes sense.
In cases where parallel execution of tests cannot be used, split the tests between different machines.

- [ ] My code follows the style guidelines of this project


smiasojed:
Time for `test` job decreased by 50%, from 18min to 9 min

smiasojed:
Each end-to-end test takes between 50 to 90 seconds. Currently, they are executed iteratively because end-to-end tests build contracts during test execution, and the same contracts are rebuilt by multiple tests instead of reusing them, causing a race condition.

smiasojed:
The `examples-test` job is split into 4 parallel jobs: linear duration decreased from 36 to 12 minutes, but when we sum up the 4 parallel jobs, we get 48 minutes (the cost of launching and compilation of common crates on each machine)

smiasojed:
Total time has change from 70 to 50 minutes

.github/workflows/ci.yml
```diff
@@ -444,7 +444,7 @@ jobs:
           git config --global --add safe.directory '*'
           # RUSTFLAGS are the cause target cache can't be used here
           cargo build --all-features --workspace
-          cargo test --all-features --no-fail-fast --workspace --locked
+          cargo test --all-features --no-fail-fast --workspace --locked -- --skip ui_tests
```
cmichi, line 447: Why the `--skip ui_tests` here?

cmichi, line 447: Good idea to skip it for `codecov`!

cmichi (approved):
This is some good work! Thanks, Sebastian!

ascjones (approved):
:rocket:

PR #2158, opened 2024-03-14
Replace drink sandbox by internal ink_sandbox

pgherveou:
Remove the drink dependency, and move the drink::Sandbox to it's own package.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,7 @@ jobs:
 
       - name: Check Formatting
         run: |
+          zepter run check
           cargo +nightly fmt --all -- --check
           # For the UI tests we need to disable the license check
           cargo +nightly fmt --all -- --check ./crates/ink/tests/ui/contract/{pass,fail}/*.rs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,6 @@ jobs:
 
       - name: Check Formatting
         run: |
-          zepter run check
           cargo +nightly fmt --all -- --check
           # For the UI tests we need to disable the license check
           cargo +nightly fmt --all -- --check ./crates/ink/tests/ui/contract/{pass,fail}/*.rs
--- a/RELEASES_CHECKLIST.md
+++ b/RELEASES_CHECKLIST.md
@@ -60,7 +60,7 @@ in the future.
     - Release `cargo-contract` crates.
     - Request update of `drink` client, which depends on the `cargo-contract` crates.
     - Release `ink_e2e`.
-2. Do a dry run with `cargo release [new_version] -v --no-tag --no-push --exclude ink_sandbox`
+2. Do a dry run with `cargo release [new_version] -v --no-tag --no-push`
     - `[new_version]` should be the **exact** SemVer compatible version you are attempting
       to release e.g. `5.0.0`
       - It is possible to supply a SemVer level here e.g. `major`, `minor`, `patch` or
@@ -69,7 +69,7 @@ in the future.
         is not necessary.
     - We don't want `cargo-release` to create any releases or push any code, we'll do
       that manually once we've actually published to `crates.io`.
-    - Note that `ink_sandbox` is released separetly and is excluded via the `--exclude` flag.
+    - Note that `ink_sandbox` has its own version number and can be released separately of the other crates.
 3. Check the output of the dry run:
    - Does not show any automatic bumping of crate versions.
    - Runs without error.
@@ -83,7 +83,7 @@ in the future.
        `master` during the release.
      - The above are to ensure that the bundled code pushed to crates.io is the same as
        the tagged release on GitHub.
-5. Publish with `cargo release [new_version] -v --no-tag --no-push --execute --exclude ink_sandbox`
+5. Publish with `cargo release [new_version] -v --no-tag --no-push --execute`
     - Ensure the same `[new_version]` as the dry run, which should be the **exact** SemVer
       compatible version you are attempting to release e.g. `4.0.0-alpha.3`.
     - We add the grace period since crates depend on one another.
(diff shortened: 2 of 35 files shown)


.github/workflows/ci.yml
pgherveou, line 1: Removing all the --verbose flags as part of these PR, they make GHA logs super hard to read, 
Happy to rollback if people disagree

ascjones: I agree, could always set `CARGO_TERM_VERBOSE` if we want to see the logs.

I would like to see this as a separate PR though.

cmichi: Agree with both of you. Separate PR would be better, you already have it as a separate commit so you can just `git revert` it here and `cherry-pick` to a new PR.

pgherveou: sure will do so, honestly did it because I thought I had some flaky results in the tests and I wanted to get a clear view on the CI logs. now that it's green I think I will just cherry pick the commit in a new PR

crates/e2e/sandbox/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "ink_sandbox"
+version.workspace = true
+authors.workspace = true
```
cmichi, line 4: Part of this code is extracted, right? Then we should put:

```suggestion
authors = ["Cardinal Cryptography", "Parity Technologies <admin@parity.io>"]
```

The code licenses for ink! and DRink! are both Apache 2.0, so no issue there.

crates/env/Cargo.toml
```diff
@@ -61,22 +61,24 @@ ink = { path = "../ink" }
...
+	"ink_primitives/std",
+	"ink_storage_traits/std",
+	"ink_engine/std",
+	"scale/std",
+	"scale-decode",
+	"scale-encode",
+	"scale-info/std",
+	"secp256k1",
+	"schnorrkel",
+	"num-traits/std",
```
cmichi, line 75: Please revert, seems like your IDE converted spaces to tabs.

deuszx:
Won't this become a maintenance problem? Previously, when we wanted to extend capabilities of drink! we had to do it only for that one crate - drink! itself - now, there are two sandboxes - drink's and ink's. How do you envision the future of the tool if one of its biggest users is now ink! itself? `drink` is used not only as ink's native e2e test backend but standalone as well.

crates/env/Cargo.toml
pgherveou: that's me using zepter to auto fix the missing std propagation, I think we should keep this, so we can eventually use this tool in CI to fix issues

pgherveou:
> Won't this become a maintenance problem? Previously, when we wanted to extend capabilities of drink! we had to do it only for that one crate - drink! itself - now, there are two sandboxes - drink's and ink's. How do you envision the future of the tool if one of its biggest users is now ink! itself? `drink` is used not only as ink's native e2e test backend but standalone as well.

@deuszx to be clear, the goal is to move the shared part to ink!, not to duplicate the trait. The follow up will be to remove the code in drink! repo and use ink_sandbox as a dependency instead.

I agree that this is moving the maintenance burden, from one repo to the other, but because drink! already has (indirect) dependency on the ink! repo, it would make the overall maintenance much easier if we re-organize the dependency as suggested here.

What do you think?

crates/e2e/src/sandbox_client.rs
```diff
@@ -320,7 +319,7 @@ where
             Ok(result) => result,
             Err(err) => {
                 log_error(&format!("Upload failed: {err:?}"));
-                return Err(DrinkErr); // todo: make a proper error type
+                return Err(SandboxErr); // todo: make a proper error type
```
pgherveou, line 322: @ascjones  not to sure what were the expectation here for this TODO

crates/e2e/sandbox/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "ink_sandbox"
+version.workspace = true
```
pgherveou, line 3: this should probably be un-pegged to the workspace version so that drink! can push update independently of ink! releases

crates/env/Cargo.toml
pgherveou: actually you are right the tool should be more flexible, will fix this up

cmichi (commented):
Making the comment style consistent with the codebase.

pgherveou:
I have both side of the drink! sandbox migration ready to review.
This is kind of annoying to rebase and maintain as a branch so ideally we can merge that quickly.
cc @cmichi @ascjones @deuszx

deuszx:
From my perspective looks good but let's consider separating release cycles of Sandbox and ink! as you mentioned in the comment.

crates/e2e/sandbox/Cargo.toml
pgherveou: I don't see any down side doing it. It also does not depend on other local crate defined in the repo so I don't see that as being problematic.

ascjones: I'm on board with giving this its own independent version. Would need to update the release checklist for that.

ascjones (approved):
LGTM

Thank you @pmikolajczyk41 @deuszx :heart:

pgherveou:
> LGTM
> 
> Thank you @pmikolajczyk41 @deuszx ❤️

I have excluded ink_sandbox from the release and added a note to the release, not sure if there is a better way to do it

See relevant commit here
https://github.com/paritytech/ink/pull/2158/commits/11757409c9b05fdb1ab7988355585eb4d03d38ef

RELEASES_CHECKLIST.md
```diff
@@ -69,6 +69,7 @@ in the future.
         is not necessary.
     - We don't want `cargo-release` to create any releases or push any code, we'll do
       that manually once we've actually published to `crates.io`.
+    - Note that `ink_sandbox` is released separetly and is excluded via the `--exclude` flag.
```
ascjones, line 72: We could still include ink_sandbox in the ink release (sometimes even it will be necessary because e2e depends on it). The difference is that it can be released on its own independent of the other crates, and therefore has its own version number.

*This is a repost of a comment I added on the commit, but it doesn't show on this PR*

pgherveou: Thks Andrew, I have updated the checklist changes, with your comment

PR #2161, opened 2024-03-15
Use zepter to format Cargo.toml files

pgherveou:
See
https://github.com/paritytech/polkadot-sdk/blob/master/substrate/docs/STYLE_GUIDE.md?plain=1#L148-L172
https://github.com/paritytech/polkadot-sdk/blob/master/docs/contributor/CONTRIBUTING.md?plain=1#L146


ascjones (approved):
Nice!

PR #2162, opened 2024-03-15
Fix outdated docs for `[ink_e2e::test]`

cmichi:
Some of the docs related to `[ink_e2e::test]` were outdated. 

This PR updates the syntax and adds doc tests.

### DIFF
--- a/crates/e2e/macro/src/lib.rs
+++ b/crates/e2e/macro/src/lib.rs
@@ -34,63 +34,106 @@ use syn::Result;
 ///   and install it on your PATH, or provide a path to an executable using the
 ///   `CONTRACTS_NODE` environment variable.
 ///
-/// Before the test function is invoked the contract will have been build. Any errors
-/// that occur during the contract build will prevent the test function from being
-/// invoked.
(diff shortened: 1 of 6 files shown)


crates/e2e/macro/src/lib.rs
```diff
@@ -34,63 +34,107 @@ use syn::Result;
 ///   and install it on your PATH, or provide a path to an executable using the
 ///   `CONTRACTS_NODE` environment variable.
 ///
-/// Before the test function is invoked the contract will have been build. Any errors
-/// that occur during the contract build will prevent the test function from being
-/// invoked.
+/// Before the test function is invoked the contract will be build. Any errors that occur
```
ascjones, line 37:
```suggestion
/// Before the test function is invoked the contract will be built. Any errors that occur
```

ascjones (approved):

crates/e2e/macro/src/lib.rs
cmichi: Happens every time!!

Issue #2169, opened 2024-03-13
Add links to linter warning details on use.ink


cmichi:
The warnings that our linter outputs look like this:

<img width="938" alt="image" src="https://github.com/paritytech/cargo-contract/assets/241530/be9312ac-8e82-4503-9e80-d087c186752b">

Each lint is documented in more detail at e.g. https://use.ink/linter/rules/strict_balance_equality.

Atm of writing this we haven't switched the default content of use.ink to 5.x, but that'll happen today and then the link will work. Right now it's https://use.ink/5.x/linter/rules/strict_balance_equality.

Please put a link to the details for each linting rule on use.ink into the warning message.

PR #2170, opened 2024-03-20
Linter: Add links to detailed lint description 

cmichi:
## Summary
Closes #_
- [y] y/n | Is it dependant on the specific version of `cargo-contract` or `pallet-contracts`?

I'll create an accompanying `cargo-contract` PR so that the new version of these lints is used.

## Description

Closes https://github.com/paritytech/ink/issues/2169.

For the linter warnings, I've added a link to the detailed description of the lint on use.ink.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- [Linter] Add links to detailed lint description ‒ [#2170](https://github.com/paritytech/ink/pull/2170)
+
 ### Fixed
 - Fix outdated docs for `[ink_e2e::test]` ‒ [#2162](https://github.com/paritytech/ink/pull/2162)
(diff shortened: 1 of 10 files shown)


smiasojed (approved):

jubnzv (commented):
What do you think about implementing a more generic approach similar to `clippy`?

This could be achievable by adding a function like [`docs_link`](https://github.com/rust-lang/rust-clippy/blob/3e3a09e2bb26ba3b8207019ea53fd9d7da4831d3/clippy_utils/src/diagnostics.rs#L17) and creating `ctx.tcx.node_span_lint` wrappers like [this one](https://github.com/rust-lang/rust-clippy/blob/3e3a09e2bb26ba3b8207019ea53fd9d7da4831d3/clippy_utils/src/diagnostics.rs#L171).

cmichi:
> What do you think about implementing a more generic approach similar to `clippy`?

That sounds good! I'd merge this to have something in right now and create a follow-up ticket.

PR #2173, opened 2024-03-22
Split meassurements

smiasojed:
## Summary
Closes #_

## Description
Splits measurements into 2 matrix jobs, 
Removes the need for using 8-core custom runners: increases the number of example-tests jobs to 6, and performs runner cleanup before starting.


.github/workflows/ci.yml
```diff
@@ -480,20 +507,38 @@ jobs:
...
 
-      - name: Rust Info
-        uses: ./.github/rust-info
-
       - name: Test Examples
         env:
           # Fix linking of `linkme`: https://github.com/dtolnay/linkme/issues/49
           RUSTFLAGS:                     -Clink-arg=-z -Clink-arg=nostart-stop-gc
-        run: |
+        uses: docker://paritytech/ci-unified:bullseye-1.75.0
```
smiasojed, line 506: How to use here ${{ env.Image }} any hints?

cmichi: Can you elaborate on what you want to know?

smiasojed: I would like to avoid duplicating image name and using the definition from the environment as `env.IMAGE`, but I cannot use the environment variable in this scope. Additionally, the workaround with `needs.set-image.outputs.IMAGE` does not work. Maybe there is another option which I do not know

cmichi: @alvicsam Do you know?

ascjones (approved):
LGTM

cmichi (approved):

PR #2196, opened 2024-04-10
Bump polkadot-sdk deps

pgherveou:


.github/workflows/ci.yml
```diff
@@ -41,7 +41,7 @@ env:
   # this target also makes sure that we don't pull in `std` by accident (through dependencies).
   # RISC-V is a modular architecture. We might switch to a different flavor with more features
   # later. For example, `riscv32imc-unknown-none-elf`.
-  RISCV_TARGET:                    riscv32i-unknown-none-elf
+  RISCV_TARGET:                    riscv32ema-unknown-none-elf
```
xermicus, line 44: AFAICT planned is `rv32ecm` for PolkaVM (no atomics). I thought it shouldn't be too limiting. However if it's really needed for `ink!` we should think about getting it into PolkaVM and surrounding specs?

xermicus (approved):

PR #2219, opened 2024-04-23
Environment agnostic contract invocation API

ascjones:
The next foundational layer towards https://github.com/use-ink/ink/issues/1674.

```rust
 let executor =
    executor::PalletContractsExecutor::<ink::env::DefaultEnvironment, T> {
        origin: who.clone(),
        contract: contract.clone(),
        value: 0.into(),
        gas_limit,
        storage_deposit_limit,
        marker: Default::default(),
    };

let mut flipper = ink::message_builder!(Flip);
let result = flipper.flip().exec(&executor)?;
```

This PR:
- decouples the generation of the message execution payload (selector + input args + return type) from the other call parameters: the callee account id and the other gas and storage limit parameters.
- introduces an `Executor` trait which can be implemented in different environments (e.g. `pallet-contracts`, see example). The `Executor` implementation has the responsibility of gathering the other call parameters and performing the environment specific invocation, as well as decoding the result.
- modifies the existing call builder codegen to use the message builder codegen to reduce duplication.
- adds the `message_builder!` macro which instantiates the generated message builder instance for use to generate and invoke messages. 

This PR does **not**:
- provide an opinionated API for how to perform calls into contracts from `pallet-contracts`, it requires a boilerplate impl of `Executor` (see `integration-tests`). The reason this is not provided as part of the lib is that it would require matching versions of `pallet-contracts` etc. with the target runtime. Need to think carefully about how best do this for maximum convenience for the user and minimum maintenance cost on the `ink` side.
- allow access to non trait messages, only works for trait impls (for now). This makes sense in the `pallet-contracts` context as it is better to depend on a trait rather than a contract impl, since you are working against an interface.
- offer a similar API for constructors, the existing create builder API and `ContractRef` would have to be used for that,
- integrate with the existing `CallBuilder` API which provides a nice interface for applying the weight and storage limit parameters etc.

These can be resolved in subsequent work. The goal here is to provide a relatively simple non-breaking stepping stone to a possible universal contract invocation API, can use in the cross-contract context, from `pallet-contracts` and from `e2e` tests.


cmichi:
Please also add a changelog entry.

PR #2222, opened 2024-04-25
Update owners

cmichi:
Changes are in line with the CLA and IP transfer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ exclude = [
 ]
 
 [workspace.package]
-authors = ["Parity Technologies <admin@parity.io>"]
+authors = ["Use Ink <ink@r0gue.io>"]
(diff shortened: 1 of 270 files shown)


brunopgalvao (approved):
LGTM. I noticed that Robin Freyler and George Oastler were removed from a few. Just want to make sure that was intentional.

smiasojed (approved):

peterwht (approved):
Approved to keep things moving. Cross-posting: To second Frank, it would be good to get an `ink@use.ink` or similar (e.g. `hi@use.ink`) for the author.

cmichi:
> LGTM. I noticed that Robin Freyler and George Oastler were removed from a few. Just want to make sure that was intentional.

Yeah intentional, the CLA and IP transfer allows this, I wanted to make the authors consistent everywhere.

PR #2228, opened 2024-04-30
Refactor: separate different types of builders into different files

ascjones:
A step towards providing the call builder style API for calling contracts from pallets, following up on https://github.com/use-ink/ink/pull/2219 as part of https://github.com/use-ink/ink/issues/1674.

This also separates the `CallFlags` from the `CallBuilder` (since call flags are not used when calling into `pallet_contracts` directly)


cmichi (approved):
Changelog update plz. And could you check if change to `ink-docs` are necessary?

ascjones:
> Changelog update plz. And could you check if change to `ink-docs` are necessary?

This is an internal refactoring, the external API remains unchanged so I believe no changelog entry or docs required.

PR #2345, opened 2024-12-12
Call `ink!` contracts from Solidity: support RLP encoding

ascjones:


peterwht:
Not necessary for the first version, but would it make sense to give devs the ability to do a per-message RLP encoding. That way an ink! dev can use Rust for most of the messages, and then expose a couple messages for use by Solidity. They can do this without having to support the entire contract as RLP.

peterwht:
Just documenting here: currently there is a bug when the abi_encoding is "all" and selector is provided by the user. 

This results in 
```
conflicting implementations of trait `DispatchableMessageInfo<3235826430>` for type `Contract`
```

crates/ink/ir/src/ir/item_impl/callable.rs
```diff
@@ -324,13 +332,34 @@ where
...
+    ir::Selector::compute(&preimage)
+}
+
+/// Returns the composed selector of the ink! callable.
+fn compose_selector_rlp<C>(item_impl: &ir::ItemImpl, callable: &C) -> ir::Selector
+where
+    C: Callable,
+{
+    // todo: handle user provided RLP selector...
+    if let Some(selector) = callable.user_provided_selector() {
```
peterwht, line 345: a user provided selector does not rely on encoding, right (equivalent for both scale and RLP)?

If this is the case, this commit should have resolved this TODO: https://github.com/use-ink/ink/pull/2375/commits/5d2f64d40b90b72ada1d8673e615272e4395fbd1

cmichi (approved):
Although incomplete, we are merging this, as it makes our work on RLP in separate PRs easier.

PR #2359, opened 2024-12-22
fix: Fix issues with path handling in grep command Update is_contract.sh

ursulabauer:
## Summary
Closes #_

## Description

I’ve updated the script to address two issues that were causing unexpected behavior when working with file paths. 

1. I’ve added quotes around the `$SOURCE_PATH` variable in the `grep` command to ensure it handles paths that contain spaces or special characters correctly. This ensures that paths with spaces, for example, are processed as a single entity rather than being split into multiple arguments.

2. I also included a check to verify that the `SOURCE_PATH` variable is not empty. This prevents errors when the source file path is either not found or wasn't properly extracted.


cmichi (approved):
Hey @ursulabauer, thanks a lot for this patch! And happy new year :-).

May I ask how you ran into it?

PR #2397, opened 2025-02-10
Add Solidity compatible address and byte types

davidsemakula:
## Summary
Closes - N/A

Adds `Address` and `Byte` types that can be unambiguously mapped to Solidity's `address` and `bytes<N>` (and `bytes`) types.

## Description

See [this discussion](https://github.com/use-ink/cargo-contract/pull/1930#pullrequestreview-2597912123) for rationale and additional details.


cmichi:
As discussed in Element, the PR should also change the function signature of e.g. `self.env().address` to return `Address`. I think there's more functions that return an `H160` right now, when it's actually an `Address`.

What should also be done is go through the examples and see where a `H160` is used, when it semantically should be an `Address`.

PR #2401, opened 2025-02-14
feat: use solidity ABI encoding instead of RLP

peterwht:
wip. Switches from RLP encoding to Solidity ABI encoding using alloy.


peterwht:
superseded by https://github.com/use-ink/ink/pull/2406 (clean base)

PR #2406, opened 2025-02-19
feat: use solidity ABI encoding instead of RLP

peterwht:
wip. Switches from RLP encoding to Solidity ABI encoding using alloy.


crates/ink/tests/ui/contract/pass/01-message-encoding-inputs-sol.rs
```diff
@@ -0,0 +1,51 @@
...
+            _input_2: i16,
+            _input_3: i32,
+            _input_4: i64,
+            _input_5: i128,
+            _input_6: u8,
+            _input_7: u16,
+            _input_8: u32,
+            _input_9: u64,
+            _input_10: u128,
+        ) {
```
peterwht, line 27: passes when `input_6` is excluded.

crates/ink/tests/ui/contract/pass/01-message-encoding-inputs-sol.rs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        #[ink(message)]
+        pub fn message_1(
+            &mut self,
+            _input_0: bool,
+            _input_1: String,
+            _input_2: Vec<u8>,
+            _input_3: [u8; 32],
```
peterwht, line 36: each of these parameters pass when they are the only parameter.

crates/ink/tests/ui/contract/pass/01-message-encoding-inputs-sol.rs
```diff
@@ -0,0 +1,51 @@
...
+        ) {
+        }
+
+        #[ink(message)]
+        pub fn message_2_primitives(
+            &mut self,
+            _input_0: AccountId,
+            _input_1: Hash,
+            _input_2: ink::H160,
+        ) {
```
peterwht, line 46: AccountId and Hash pass when they are the only parameter.

H160 fails because SolValue is not implemented.

cmichi (approved):

PR #2408, opened 2025-02-25
feat: ink! can call Solidity contracts

peterwht:
Allows ink! to do cross-contract calls to contracts that use Solidity ABI Encoding.

The create and call builder each have a new method with a `_solidity` postfix.

The following code snippet executes a message to a Solidity ABI encoded contract, encoding the provided arguments and decoding the return type with Solidity ABI encoding.
``` rust
let result = build_call_solidity::<<Self as ::ink::env::ContractEnv>::Env>()
    .call(callee)
    .ref_time_limit(1000000000)
    .transferred_value(U256::zero())
    .call_flags(CallFlags::empty())             
    .exec_input(ExecutionInput::new(selector.into()).push_arg(true))
    .returns::<()>()
    .try_invoke();
```

## Details

To achieve this, three new traits are introduced:
``` rust
// crates/primitives/src/reflect/dispatch.rs

/// Trait for ABI-specific encoding with support for both slice and vector buffers.
pub trait AbiEncodeWith<Abi> {
    /// Encodes the data into a fixed-size buffer, returning the number of bytes written.
    fn encode_to_slice(&self, buffer: &mut [u8]) -> usize;

    /// Encodes the data into a dynamically resizing vector.
    fn encode_to_vec(&self, buffer: &mut Vec<u8>);
}

/// Trait for ABI-specific decoding.
pub trait AbiDecodeWith<Abi>: Sized {
    /// The error type that can occur during decoding.
    type Error: core::fmt::Debug;
    /// Decodes the data from a buffer using the provided ABI.
    fn decode_with(buffer: &[u8]) -> Result<Self, Self::Error>;
}
```

``` rust
// crates/env/src/call/common.rs

/// A trait for decoding the output of a message based on different ABIs.
/// This is necessary as contracts with different ABIs have different return types.
/// For example, Solidity contracts return the output directly without `MessageResult`.
pub trait DecodeMessageResult<Abi>: Sized {
    /// Decodes the output of a message call, requiring the output
    /// to be wrapped with `MessageResult` (if not included in the output).
    fn decode_output(buffer: &[u8]) -> crate::Result<MessageResult<Self>>;
}
```

The `Abi` generic can be either of these market types
``` rust
/// Marker type for SCALE encoding. Used with [`AbiEncodeWith`], [`AbiDecodeWith`] and
/// [`DecodeMessageResult`].
#[derive(Default, Clone)]
pub struct ScaleEncoding;
/// Marker type for Solidity ABI encoding. Used with [`AbiEncodeWith`],
/// [`AbiDecodeWith`] and [`DecodeMessageResult`].
#[derive(Default, Clone)]
pub struct SolEncoding;
```

When a contract is used as a dependency, the code generation will generate the call builder with the selected ABI. If the ABI is `all`, then two call builders will be generated.

### Miscellaneous 
- Removes the unused file `call_v1` from the call builder.


crates/env/src/error.rs
```diff
@@ -19,10 +19,12 @@ use crate::engine::off_chain::OffChainError;
 use pallet_revive_uapi::ReturnErrorCode;
 
 /// Errors that can be encountered upon environmental interaction.
-#[derive(Debug, From, PartialEq, Eq)]
```
peterwht, line 22: `Eq` was removed as `alloy_sol_types::Error` does not derive it. Removing it does not seem to cause any errors in the code.

cmichi (approved):
LGTM, Peter!

PR #2441, opened 2025-03-21
Abstractions for mapping arbitrary Rust types to Solidity ABI compatible types

### LINKED ISSUE #2428: Abstractions for user-controlled mapping of `u8` and `u8` sequences/collections to Solidity types
## Primer

Is `&[u8]` a byte slice or an 8-bit unsigned integer slice? what about `Vec<8>`?

In Rust/ink!, there isn't any semantic difference.
This can be seen in various `<from/to/into/as>_bytes` conversions (including from the standard library) that either take or return some `u8` sequence/collection (e.g. [`String::as_bytes`](https://doc.rust-lang.org/std/string/struct.String.html#method.as_bytes)).

**NOTE:** Newtype wrappers are commonly used to add higher-level semantics to sequences/collections of bytes/`u8`s, but fundamentally, these are zero-cost abstractions that are cheap/free to convert back into the underlying sequences/collections (e.g. `struct Bytes(pub Vec<u8>)`).

However, unlike Rust/ink!, Solidity has primitives for both 8-bit unsigned integers (i.e. [`uint8`](https://docs.soliditylang.org/en/latest/types.html#integers)), and bytes sequences (i.e. [`bytes`](https://docs.soliditylang.org/en/latest/types.html#bytes-and-string-as-arrays) and [`bytes1`, `bytes2` ... `bytes32`](https://docs.soliditylang.org/en/latest/types.html#fixed-size-byte-arrays)).
As such, there are meaningful differences between how `bytes` (and `bytes<N>`) sequences, and `uint8` sequences (i.e. fixed-size and dynamic `uint8` arrays) are represented/encoded in calldata (i.e. [Solidity ABI encoding](https://docs.soliditylang.org/en/latest/abi-spec.html#formal-specification-of-the-encoding)), but also in [memory](https://docs.soliditylang.org/en/latest/internals/layout_in_memory.html).

One meaningful difference for us is that in Solidity ABI encoding, `bytes` (i.e. dynamic) and `bytes<N>` (i.e. fixed-sized) arrays are packed, while `uint8[]` (i.e. dynamic) and `uint8[N]` (i.e. fixed-sized) are not.
As a concrete example, `bytes32` is encoded/packed into a 32 byte sequence (i.e. a single word) in Solidity calldata (and memory), while `uint8[32]` is encoded into a 1024 byte sequence (i.e. 32 words, with 32 bytes used for each element).

**NOTE:** [`abi.encodePacked()` is not part of the Solidity ABI spec](https://docs.soliditylang.org/en/latest/abi-spec.html#non-standard-packed-mode), so it's not a transparent optimization (i.e. interacting contracts have to be aware of it usage).

## Takeaways

- For Rust/ink!, byte and 8-bit unsigned integer sequences are semantically equivalent
- Representation/encoding differences only matter at the Solidity ABI/interoperability boundary

## Goals
- Because byte and 8-bit unsigned integer sequences are semantically equivalent in Rust/ink!, any abstractions over them should be zero-cost (i.e. it should be cheap/free to convert between any two semantically equivalent representations)
- It should be possible for ink! smart contract authors to keep abstracts at the interface boundary (i.e. only deal with close to the signature of an ink! message and mostly ignore them body)

## Design

The default mappings will be as follows:
- `u8` is mapped to `uint8`
- `u8` sequences/collections are mapped to equivalent Solidity fixed-size arrays (i.e. `uint8[N]` where `N` is the array size) and dynamic arrays (i.e. `uint8[]`)

We then introduce a newtype wrapper `AsBytes<T>` that:
- Can only be applied to `u8` sequences/collections with equivalent Solidity byte types (enforced by a sealed trait bound on T)
- Encapsulates logic for encoding/decoding `u8` sequences/collections as their Solidity bytes types (e.g. `AsBytes<[u8; 32]>` is mapped to `bytes32`)
- Implements core/standard traits for cheaply/freely using/passing the wrapper type in place of the underlying type (e.g. `Deref`, `Borrow`, `AsRef` e.t.c)

This is roughly translates to:
```
ink! <=> Solidity
AsBytes<[u8; 1]> == bytes1
AsBytes<[u8; 2]> == bytes2
...
AsBytes<[u8; 32]> == bytes32
AsBytes<Vec<u8>> == bytes
```

This allows ink! developers to largely keep the `AsBytes<T>` wrapper at the interface boundary (e.g. an ink! message with `AsBytes<[u8; 32]>` input and output types can immediately deref the input to `[u8; 32]` and deal with only the underlying type in the function body, and then cheaply wrap it back up to `AsBytes<[u8;32]>` at the return place).

## Alternatives

The following alternatives were considered and rejected/abandoned:

- Introducing a `Byte` type as a semantically distinct `u8` equivalent was abandoned because conversions from `Byte` sequences/collections to `u8` equivalents would be expensive
- Mapping `u8` to `uint8`, and `u8` sequences/collections to `bytes` and `bytesN` equivalents by default, and then providing two wrappers (e.g. `AsBytes<T>` and `AsInt<T>`) to override the preferred representation/encoding was abandoned due to:
  - Perceived higher user-level cognitive load (e.g. `u8` would map to `uint8` by default, while `[u8; N]` for `1 <= N <= 32` would map to `bytes<N>` by default, and `[u8; N]` for `N > 32` would *always* map to `uint8[N]`)
  - Generic implementation complexity due to Rust's limited support for specialization

### LINKED ISSUE #2429: Abstractions for mapping arbitrary Rust types to Solidity ABI compatible types
## Background

Our Solidity ABI compatibility infra currently relies on traits from [`alloy-sol-types`](https://docs.rs/alloy-sol-types/latest/alloy_sol_types/).
Mainly, the [`SolValue` trait](https://docs.rs/alloy-sol-types/latest/alloy_sol_types/trait.SolValue.html) which is the basis for our ABI encoding/decoding.
But also the lower level [`SolType` trait](https://docs.rs/alloy-sol-types/latest/alloy_sol_types/trait.SolType.html) which is used to handle "trickier" parts of decoding (e.g. due to complications from this [`From` predicate on `SolValue::abi_decode`](https://github.com/alloy-rs/core/blob/d0a0be46f80bc05cc6adc0157ccaf2fe629c7f18/crates/sol-types/src/types/value.rs#L145-L147) when dealing with a tuple of call args - Note though that this isn't the motivation for this issue i.e. it's already a solved problem).

## Motivation

The `SolValue` trait, in combination with the [`sol!` macro](https://docs.rs/alloy-sol-macro/latest/alloy_sol_macro/macro.sol.html), work(s) well enough by essentially:
1. mapping primitives types from Rust to equivalent Solidity types (except for `u8` whose mapping is semantically ambiguous because Solidity has dedicated `bytes` types)
2. introducing new types (from [`alloy-primitives`](https://docs.rs/alloy-primitives/latest/alloy_primitives/)) for any elementary Solidity types that don't have a semantically equivalent Rust type (e.g. `address`, `uint256`, `int256`, `bytes` and `bytes<N>`, `unit40` e.t.c)
3. allowing mapping of custom types to compositions from the above primitives with the `sol!` macro.

From the overview above, we can draw the following conclusions:

(1.) is mostly fine except for the missing `SolValue` for `u8` implementation.
We can reliably detect a `u8` type in the signature of an ink! message in the ink! proc-macro and use the [`Uint<8>: SolType` type](https://docs.rs/alloy-sol-types/latest/alloy_sol_types/sol_data/struct.Uint.html) to encode/decode (i.e. the `SolType` that would have been used in it's `SolValue` implementation). However, we can't reliably do the same for a type alias for a tuple that includes a `u8` type (e.g. `type Foo = (bool, u8)`).
We also cannot implement `SolValue` for `u8` because of Rust's coherence/orphan rules.

For (2.), `pallet-revive`'s equivalent Rust types for missing Solidity primitives like `H160` for `address` and `U256` for `uint256` are from `sp-core`/[`primitive-types`](https://docs.rs/primitive-types/latest/primitive_types/). We can't implement `SolValue` for any of these types, meaning we need to convert from `pallet-revive` primitives to equivalent `alloy-primitives` types in a way that (ideally) doesn't expose any similarly named primitives from both (e.g. both have `U256` types) to the end user in any public interfaces (as that would be quite confusing IMO).

For (3.), the `SolValue` trait actually relies on the `sol!` macro to implement a few necessary underlying type conversions, and [recommends using the macro instead of implementing the trait directly (presumably for this reason)](https://docs.rs/alloy-sol-types/latest/alloy_sol_types/trait.SolValue.html#implementers-guide). The `sol!` macro itself uses Solidity-like syntax for defining items, while direct `SolValue` implementations are composed with `SolType` types like `Uint<32>` as opposed equivalent primitive Rust types like u32.

All these things put together lead to:
- complex and/or fragile implementations on the ink! proc-macro side
- non-idiomatic (less "Rusty"?) interfaces (likely requiring high cognitive load) on the user side

## Design

TBD.

But the general idea is to introduce ink! specific trait(s) that encapsulate the requirements for a Rust type to be Solidity ABI compatible as well as a complementary `Derive` macro.

Some things to note:
- The user-facing `SolValue` replacing trait can't simply involve a blanket implementation of `SolValue` as that would still run afoul of coherence/orphan rules for foreign types that don't already implement `SolValue`.
- We'll largely still be using `alloy-sol-types` (and related crates) under the hood, but just wrapping them in a simpler (more Rusty?) interface.
- For existing interfaces that have a `T: SolValue` predicate, the goal will be to simply replace `SolValue` with the new user-facing trait and have everything just work.

davidsemakula:
## Summary
Closes #2429 
Closes #2428 

## Description

- Adds `SolEncode` and `SolDecode` traits to `ink_primitives` as replacements for `alloy_sol_types::SolValue` in public interfaces (and analogs to `scale::Encode` and `scale::Decode` but for Solidity ABI encoding/decoding). `SolEncode` and `SolDecode` essentially describe how to map an arbitrary Rust/ink! type to a Solidity ABI type equivalent (i.e. a type that implements `SolTypeEncode` and/or `SolTypeDecode`) for Solidity ABI encoding/decoding
- Adds `SolTypeEncode` and `SolTypeDecode` sealed traits which describe a Rust/ink! equivalent of a Solidity ABI type, and implement logic for Solidity ABI encoding/decoding. These traits are implemented for all Rust primitive types that have an equivalent Solidity ABI type, as well as `Address` and `U256` ink! primitives. However  `SolTypeEncode` is additionally implemented for references (and a few smart pointers) of types that implement `SolTypeEncode` (i.e. `&T`, `&mut T`, `Box<T>` and `Cow<T>` for `T: SolTypeEncode`). The rustdoc for both traits includes a table showing the mapping of Rust/ink! primitives to equivalent Solidity ABI types. All `SolTypeEncode` and `SolTypeDecode` types also implement `SolEncode` and `SolDecode`
- Adds `SolBytes<T>` type as a newtype wrapper that encodes/decodes `u8` sequences/collections as their equivalent Solidity bytes representations. `T` has a `SolBytesType` bound, and `SolBytesType` is a sealed trait that's only implemented for `u8`, `[u8; N]` where `1 <= N <= 32` and `Vec<u8>`
- Adds sanity checks to ensure our `SolTypeEncode`, `SolTypeDecode`, `SolEncode` and `SolDecode` implementations match `alloy_sol_types::SolValue` equivalents
- Removes all usages of `alloy_sol_types` outside `ink_primitives` and instead use the new Solidity compatibility abstractions above defined in `ink_primitives`
- Implement `SolEncode` and `SolDecode` for `AccountId`, `Hash`, `H160` and `H256` types from ink! primitives
- Converts `Address` into a type alias for `H160`, and replaces `H160` with `Address` in the public interfaces where `H160` semantically represents an address

## Related/future work - to be done in a follow up PR(s)
- [ ] Provide a `Derive` macro for implementing `SolEncode` and `SolDecode` for arbitrary types https://github.com/use-ink/ink/issues/2462
- [ ] Provide Rust/ink! types for Solidity integer types (apart from [U256](https://docs.rs/primitive-types/latest/primitive_types/struct.U256.html)) that don't have a Rust/ink!/`pallet-revive` equivalent (e.g. `uint248`) https://github.com/use-ink/ink/issues/2461
- [ ] Documentation for `SolEncode` and `SolDecode` in `ink` crate rustdoc and ink-docs website
  - [ ] `ink` crate rustdoc https://github.com/use-ink/ink/pull/2463
  - [ ] ink-docs website 
- [ ] Update `cargo-contract` Solidity metadata generation to properly support `SolEncode` and `SolDecode` implementing arbitrary types https://github.com/use-ink/cargo-contract/issues/1996


crates/primitives/src/sol.rs
```diff
@@ -0,0 +1,351 @@
...
+        // `&self.0`.
+        AsSolBytes(self.0)
+    }
+}
+
+// H160 <-> bytes20
+// TODO: (@davidsemakula) Evaluate if it's worth removing this mapping, or changing it to
+// `address` before v6 release.
+// Rationale: while this mapping is technically correct, it may be confusing for ink!
+// devs, or just needless increase the cognitive load.
```
davidsemakula, line 326: @cmichi @peetzweg I think doing the technically "correct" thing here (i.e. mapping `H160` to `bytes20`) is actually not desirable as described in this comment. WDYT?

If we decide to map `H160` to Solidity `address` though, it removes the need for us to have our own address type, and it can instead just be a type alias to H160.
One other thing that speaks in favor of this approach is `pallet-revive` contract events will still use `H160` as the address in any case.

crates/ink/tests/ui/contract/pass/message/message-selector-encoding-sol.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unexpected_cfgs)]
```
cmichi, line 1: Do you remember why this was necessary here?

crates/primitives/src/sol.rs
```diff
@@ -0,0 +1,440 @@
...
+use crate::types::{
+    AccountId,
+    Address,
+    Hash,
+};
+
+/// Maps an arbitrary Rust/ink! type to a Solidity ABI type equivalent for Solidity
+/// ABI decoding.
+///
+/// # Note
```
cmichi, line 56:
```suggestion
/// # Note
///
```

crates/primitives/src/sol.rs
```diff
@@ -0,0 +1,440 @@
...
+    i8, i16, i32, i64, i128,
+    // unsigned integers
+    u8, u16, u32, u64, u128, U256,
+    // string
+    String,
+    // address
+    Address,
+}
+
+// Rust array <-> Solidity fixed-sized array (i.e. `T[N]`).
```
cmichi, line 208:
```suggestion
// Rust array ⭤ Solidity fixed-sized array (i.e. `T[N]`).
```

Looks nicer in the rendered docs.

crates/primitives/src/sol.rs
```diff
@@ -0,0 +1,440 @@
...
+
+impl<'a, T: SolEncode<'a>, const N: usize> SolEncode<'a> for [T; N] {
+    type SolType = [T::SolType; N];
+
+    fn to_sol_type(&'a self) -> Self::SolType {
+        self.each_ref().map(<T as SolEncode>::to_sol_type)
+    }
+}
+
+// Rust `Vec` <-> Solidity dynamic size array (i.e. `T[]`).
```
cmichi, line 225:
```suggestion
// Rust `Vec` ⭤ Solidity dynamic size array (i.e. `T[]`).
```

crates/primitives/src/sol.rs
```diff
@@ -0,0 +1,440 @@
...
+                    self.iter().map(<T as SolEncode>::to_sol_type).collect()
+                }
+            }
+        )*
+    };
+}
+
+impl_slice_ref_encode!(&'a [T], &'a mut [T], Box<[T]>);
+
+// AccountId <-> bytes32
```
cmichi, line 330:
```suggestion
// AccountId ⭤ bytes32
```

crates/primitives/src/sol.rs
davidsemakula: > Why do you consider the [bytes20] approach to be the technically correct one instead?

@cmichi In the `primitive-types` crate, `H160` is defined as ["Fixed-size uninterpreted hash type with 20 bytes (160 bits) size."](https://docs.rs/primitive-types/latest/primitive_types/), along with other similar types (e.g. `H128`, `H256` e.t.c) for different sizes (we're actually also mapping `H256` to `byte32`).
A contrast is [`alloy-primitives`](https://docs.rs/alloy-primitives/latest/alloy_primitives) which has a dedicated types for [`Address`](https://docs.rs/alloy-primitives/latest/alloy_primitives/struct.Address.html) and [fixed-sized byte arrays](https://docs.rs/alloy-primitives/latest/alloy_primitives/struct.FixedBytes.html)

> Could you elaborate a bit here? Why might the current, correct mapping be confusing for ink! devs?

The wrinkle is that `pallet-revive` basically uses `H160` as an address type (e.g. [`AddressMapper`](https://docs.rs/pallet-revive/latest/pallet_revive/trait.AddressMapper.html), [Events](https://docs.rs/pallet-revive/latest/pallet_revive/pallet/enum.Event.html), [`InstantiateReturnValue`](https://docs.rs/pallet-revive/latest/pallet_revive/struct.InstantiateReturnValue.html)), and to me it looks inevitable that an ink! dev will at some point interact with `pallet-revive` (or its docs) and likely make the association `H160` is an `address`, and then will be surprised if it doesn't behave this way in ink!, especially across the Solidity ABI boundary.

So it looks (to me) like the reasonable thing to do (after all) is to behave like `pallet-revive` and treat `H160` as an address.

crates/ink/tests/ui/contract/pass/message/message-selector-encoding-sol.rs
davidsemakula: > Do you remember why this was necessary here?
It's because ink! codegen uses `cfg(feature = "ink-as-dependency")`, [`rustc` >= 1.81 checks that `cfg`s match expected names and values](https://blog.rust-lang.org/2024/05/06/check-cfg.html), which are declared with the `--check-cfg` flag. Cargo automatically sets this flag for all features for the target crate ... but in this case, these are ui tests run via `trybuild` with no proper `Cargo.toml` defining the feature. 

That's the short of it anyway 🙂  

> Can you check if it still is?
May be `trybuild` has a way to set the flag, will check.

crates/primitives/src/sol.rs
cmichi: 👍 We're on the same page then, I agree with you!

cmichi (approved):

crates/primitives/src/sol.rs
cmichi: @davidsemakula Can you create a follow-up issue in [our board](https://github.com/orgs/use-ink/projects/4)?

davidsemakula: @cmichi Actually, it's relatively trivial fix, I'll just handle it in this one already.

PR #2487, opened 2025-04-30
Define static distributed events slice in `ink` crate

davidsemakula:
## Summary
Closes #_

## Description

When generating ink! contract metadata, we use [linkme](https://crates.io/crates/linkme) to collect event specs from across the dependency graph (because events can be defined in separate crates) into a single "static distributed slice" defined in the `ink_metadata` crate.

However, having multiple versions of the `ink_metadata` crate in the dependency graph can lead to "duplication" of this "static distributed slice" (e.g. [see here](https://github.com/use-ink/ink/actions/runs/14665273888/job/41158790910?pr=2486#step:4:2210)).

AFAICT, this "duplicate #[distributed_slice]" issue doesn't typically happen when pulling both `ink` and `ink_e2e` (and hence `ink_metadata`) from [crates.io](http://crates.io/) because a single `ink_metadata` version can typically be resolved. 

However, it's a common issue when the parent crates (i.e. `ink` and `ink_e2e`) pull `ink_metadata` from different sources, or somehow have it pinned to different git commits. This is especially common in development branches because `ink_e2e` transitively depends on `ink_metadata` through the `contract-build` crate from `cargo-contract`, making our CI tests related to events metadata very flaky.

The best solution would be to define the "static distributed events slice" in the root contract package, however, that would require ink! proc-macros to be somehow aware of the root contract package.

So we go with the next best solution (for now), which is to move the "static distributed events slice" definition as high up the dependency graph as possible, into the `ink` crate. This at least bypasses the complications introduced by `ink_e2e` dependency on `cargo-contract` crates (i.e. `contract-build`) which can potentially introduce a duplicate definition of the events slice via a separate version of `ink_metadata` in the dependency graph.


integration-tests/public/events/Cargo.toml
```diff
@@ -28,3 +28,10 @@ std = [
 ]
 ink-as-dependency = []
 e2e-tests = []
+
+[profile.test]
+# Need this for linkme crate to work for the event metadata collection unit test.
+# See https://github.com/dtolnay/linkme/issues/61#issuecomment-1503653702
+# NOTE: `cargo-contract` also sets this flag when generating metadata for the same reason.
+# See https://github.com/use-ink/cargo-contract/pull/1200
+lto = "thin"
```
davidsemakula, line 37: NOTE: This is just being re-introduced, it was present for this test for previous ink! versions (e.g. [see here](https://github.com/use-ink/ink/blob/84d1158de43e0f7d8f0eb87843e931d33643f897/integration-tests/public/events/Cargo.toml#L32-L35)) and was erroneously remove by https://github.com/use-ink/ink/commit/03a8eeeab690d291d0ac9bd75c84bb015c8b6532#diff-a00395296d1808215704bc7cca551820d8232fab091345f72ca021aaccb239cf

I added a few more comments here to clarify it's purpose

crates/ink/src/lib.rs
```diff
@@ -105,3 +105,24 @@ pub use ink_primitives::{
...
+
+#[cfg(feature = "std")]
+use ink_metadata::EventSpec;
+
+/// Any event which derives `#[derive(ink::EventMetadata)]` and is used in the contract
+/// binary will have its implementation added to this distributed slice at linking time.
+#[cfg(feature = "std")]
+#[linkme::distributed_slice]
+#[linkme(crate = linkme)]
+pub static CONTRACT_EVENTS: [fn() -> EventSpec] = [..];
```
davidsemakula, line 121: The name change is mostly for vanity reasons, but it has the benefit of not requiring a coordinated update of the `contract-build` from `cargo-contract`.

crates/ink/Cargo.toml
```diff
@@ -28,6 +28,7 @@ scale-info = { workspace = true, default-features = false, features = ["derive"]
 derive_more = { workspace = true, features = ["from"] }
 const_format = { workspace = true }
 keccak-const = { workspace = true }
+linkme = { workspace = true }
```
evilrobot-01, line 31: Just checking for my understanding as new to the codebase. The `ink_metadata` dep was optional as above. If `linkme` was only used by the metadata crate previously, should it not perhaps be optional here as well? I see that this was due to `ink_metadata/std`, but wonder if this will still result in an unnecessary build of the crate in no-std scenarios. Better to be explicit?

Usages seem to be feature gated by `std`

evilrobot-01 (approved):
Looks great!

Just left a single question.

crates/ink/Cargo.toml
davidsemakula: Yes, it should be optional, thanks for catching this ... fixed! 🙂

PR #2489, opened 2025-05-01
fix: dead link in integration-tests/README.md

gap-editor:
updated with a new one. it is worling good now.


evilrobot-01:
As `master` is currently targeting currently unreleased ink v6, it should perhaps be https://use.ink/docs/v6/contract-testing/end-to-end-e2e-testing/ but will leave that up to someone more informed to advise.

cmichi (approved):

PR #2490, opened 2025-05-01
ci: fix contract size submission

evilrobot-01:
## Summary
Closes #_

Updates CI workflow to use run identifier for measurements file name when not being run on the master branch, whilst preserving the branch name when run on master. This ensures that a later step can still download results from master branch runs for contract size comparison.

## Description
Added a global environment variable to the ci workflow which defines the identifier used for the `measurements-` file throughout the workflow, affixing either `master` when run on the `master` branch or `github.run_id` otherwise. The latter is preferred over using the branch name, as a branch name could contain characters which are problematic when used for file names.


davidsemakula (approved):
LGTM!

PR #2525, opened 2025-06-18
Support Solidity ABI encoded constructor dispatch

### LINKED ISSUE #2518: Handle Solidity ABI encoded constructor dispatch
## Overview
Solidity allows at most one constructor, [doesn't have "on-chain" constructors](https://docs.soliditylang.org/en/latest/contracts.html#creating-contracts) and so doesn't have "constructor selectors"

## Design
- Only allow one constructor when the ABI is "sol"
- Either only one constructor or require one of the constructors to be [marked as the default](https://use.ink/docs/v6/macros-attributes/default) when ABI is "all"
- Update ink! dispatch codegen to 
   - Treat all constructor call data as args when ABI is "sol"
   - Fallback to treating all constructor call data as args when ABI is "all" and no selector is matched
- Update `CreateBuilder` and `ExecutionInput` to allow execution input with no selector and re-introduce a dedicated `ink_env::build_create_solidity` function

davidsemakula:
## Summary
Closes #2518

## Description

See #2518 for details

### DIFF
--- a/crates/env/src/call/execution.rs
+++ b/crates/env/src/call/execution.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use core::marker::PhantomData;
-
 use ink_prelude::vec::Vec;
 use ink_primitives::{
     abi::{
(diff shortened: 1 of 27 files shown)


crates/env/src/call/execution.rs
```diff
@@ -97,12 +96,30 @@ impl<Abi> ExecutionInput<EmptyArgumentList<Abi>, Abi> {
...
+}
 
+impl ExecutionInput<EmptyArgumentList<Sol>, Sol> {
+    /// Creates a new execution input with no selector.
+    ///
+    /// # Note
+    ///
+    /// Should only be used for Solidity ABI encoded constructors/instantiation.
+    #[inline]
+    pub fn no_selector() -> Self {
```
cmichi, line 113: Happy to see that you implemented this here, we'll also need it for calling the pre-compiles.

PR #2540, opened 2025-07-03
feat(primitives): export weight with sol encoding

evilrobot-01:
## Summary
Closes #_
- [x] Does it introduce breaking changes? N
- [x] Is it dependent on a specific version of `cargo-contract` or `pallet-revive`? N

Exports `Weight` from `sp-weights`, implementing  `SolEncode`/`SolDecode`.


crates/primitives/Cargo.toml
```diff
@@ -34,6 +34,7 @@ sp-core = { workspace = true }
 
 sp-io = { version = "40.0.0", default-features = false, features = ["disable_panic_handler", "disable_oom", "disable_allocator"] }
 sp-runtime-interface = { version = "29.0.1", default-features = false, features = ["disable_target_static_assertions"] }
+sp-weights.workspace = true
```
evilrobot-01, line 37: Seemed tolerable to add a dependency based on other existing sp_* usages. Whitespace changes are auto-formatting, mostly tabs being converted to 4 spaces.

crates/primitives/src/lib.rs
```diff
@@ -66,6 +66,8 @@ pub use primitive_types::{
     U256,
 };
 
+pub use sp_weights::Weight;
```
evilrobot-01, line 69: Exports in line with the approach as above.

crates/primitives/src/sol.rs
```diff
@@ -446,3 +447,20 @@ impl SolEncode<'_> for H256 {
...
+impl SolDecode for Weight {
+    type SolType = (u64, u64);
+
+    fn from_sol_type(value: Self::SolType) -> Self {
+        Weight::from_parts(value.0, value.1)
+    }
+}
+
+impl<'a> SolEncode<'a> for Weight {
+    type SolType = (u64, u64);
```
evilrobot-01, line 461: Unable to use references due to struct fields being private, requiring use of the getters.

davidsemakula: `Copy` and primitive types, so performance should be the same anyway

davidsemakula (approved):
LGTM!

PR #2545, opened 2025-07-11
Implement `SolEncode` and `SolDecode` for `Option<T>`

davidsemakula:
## Summary
Closes #_

## Description

For Solidity ABI encoding and decoding, `Option<T>` is represented as `(bool, T)` for `T: SolEncode` and `T: SolDecode` respectively.

`bool` is a "flag" indicating the variant i.e. `false` for `None` and `true` for `Some` such that:
 - `Option::None` is mapped to `(false, <default_value>)` where `<default_value>` is the zero bytes or empty representation of `T` (e.g. `0u8` for `u8` or `Vec::new()` for `Vec<T>`)
 - `Option::Some(value)` is mapped to `(true, value)`

The resulting type in Solidity can be represented as a struct with a field for the "flag" and another for the data.
Note that [`enum` in Solidity is encoded as `uint8` in Solidity ABI encoding][sol-abi-type], while the encoding for `bool` is equivalent to the encoding of `uint8` with `true` equivalent to `1` and `false` equivalent to `0`. Therefore, the `bool` "flag" can be safely interpreted as a `bool` or `enum` (or even `uint8`) in Solidity code.

[sol-abi-type]: https://docs.soliditylang.org/en/latest/abi-spec.html#mapping-solidity-to-abi-types

## Follow ups/ related TODOs

- [ ] Update ink! v6 docs


cmichi (approved):
LGTM! Please also check what needs to be documented in `ink-docs`.

PR #2577, opened 2025-08-04
Item name/identifier overrides for overloading, selector computation and metadata

### LINKED ISSUE #2576: Solidity ABI: Support function and event overloading
## Overview
Solidity allows functions and events with shared names but different signatures.
Rust doesn't allow in-scope items in the same namespace to share names.

## Considerations
- For messages/functions, traits can technically be used to construct overloaded interfaces, but this solution may be unergonomic for some use cases
- For events/structs, multiple events with the same name can be defined in different scopes (e.g. via modules), and be disambiguated in the calling context either using full paths or use/import aliases, but this solution may also be unergonomic for some use cases
- Custom selectors (i.e. precomputed manually provided selectors) can't be used as a solution because Solidity metadata specs used by Solidity tools only accept function/event signatures not pre-computed selectors
- Unlike functions and events, Solidity custom errors can't be overloaded and hence aren't considered in this issue

## Design

While the primary motivation for adding support for item `name` overrides (via an attribute argument) is enabling developers to easily define overloaded interfaces for Solidity ABI interoperability, the `name` override mechanism is generalized to support for both the ink! and Solidity ABIs, because it can be used as a more transparent alternative to custom selectors for the "name-changing while maintaining the same selector" use case for the ink! ABI.

For ink! ABI, message and constructor selectors are computed from the item name/identifier alone, so multiple messages with the same `name` override will have the same computed selector, which results in a compilation error about the overlapping selectors.

So in general, `name` overrides should mainly be used to: 
- Implement contracts compliant with Solidity standards (i.e. ERCs) that require/use overloaded interfaces
- Define interfaces (e.g. via trait definitions) for interacting with existing Solidity ABI-only contracts with overloaded interfaces
- As a more transparent alternative to custom selectors for the "name-changing while maintaining the same selector" use case for ink! ABI contracts
 
For new contracts (especially contracts that support the ink! ABI i.e. ABI mode "ink" or "all"), more semantic interfaces with descriptive and unique item names/identifiers are encouraged.

- Support an optional attribute argument `name` for `#[ink(message)]`, `#[ink(constructor)]`, `#[ink::event]` and `#[ink(event)]` attributes that's used as the name of the message, constructor or event for computing the ABI-specific selector for the item i.e: 
  - For ink! ABI, if provided, the `name` override is used as the name/identifier when computing selectors for messages, constructors and events
    - Note that the `selector` attribute takes precedence over the `name` override for ink! ABI selectors
  - For Solidity ABI, if provided, the `name` override is used as the name/identifier when computing selectors for messages and events. 
    - Note that Solidity ABI spec doesn't have selectors for constructors
    - Note that the `selector` attribute is ignored when computing Solidity ABI selectors because Solidity metadata specs and hence tools only accept function/event signatures not pre-computed selectors
  - For both ink! and Solidity ABI metadata, if provided, the `name` override is used in the generated metadata
- If provided, a `name` attribute argument must have a string value (provided using `name="value"` syntax) 
- If provided, the `name` value must be an "identifier-like" string i.e:
  - Must begin with an alphabetic character, underscore or dollar sign (supported by Solidity)
  - Must only contain alphanumeric characters, underscores or dollar signs (supported by Solidity)

Edit: Design was edited to reflect discussions/decisions about generalizing the `name` override mechanism to support both ink! ABI and Solidity ABI use cases.

## References
- https://docs.soliditylang.org/en/latest/contracts.html#function-overloading
- https://docs.soliditylang.org/en/latest/contracts.html#inheriting-different-kinds-of-members-of-the-same-name
- https://docs.soliditylang.org/en/latest/contracts.html#custom-errors
- https://doc.rust-lang.org/reference/names/namespaces.html
- https://docs.soliditylang.org/en/latest/abi-spec.html
- https://docs.soliditylang.org/en/latest/grammar.html#a4.SolidityLexer.Identifier
- https://doc.rust-lang.org/reference/identifiers.html

davidsemakula:
## Summary
Closes #2576 

## Description

- Introduces a `name` attribute argument for `#[ink(message)]`, `#[ink(constructor)]`, `#[ink::event]` and `#[ink(event)]` attributes that's used as the name/identifier of the message, constructor or event for overloading, selector computation and metadata
- See issue description (https://github.com/use-ink/ink/issues/2576) for more details about motivation and suitable use cases
- Ancillary improvements to custom signature topic and event config parsing and validation including 
  - Refactoring for reusability/DRY-ness
  - More unit tests

## Follow ups/ related TODOs

- [x] Update ink! v6 docs (https://github.com/use-ink/ink-docs/pull/483)

### DIFF
--- a/crates/ink/ir/src/ir/attrs.rs
+++ b/crates/ink/ir/src/ir/attrs.rs
@@ -513,7 +513,7 @@ impl core::fmt::Display for AttributeArgKind {
             Self::Implementation => write!(f, "impl"),
             Self::HandleStatus => write!(f, "handle_status"),
             Self::Default => write!(f, "default"),
-            Self::Name => write!(f, "sol_name = N:string"),
+            Self::Name => write!(f, "name = N:string"),
         }
     }
 }


cmichi:
We should clarify the relationship of the attribute `sol_name` (or `name`, if we generalize it) and `selector`. I think the intuitive behavior would be for `selector` to be the attribute that has the final say on the selector. 

Could you add some tests in the PR that make sure we have a defined behavior for interactions of those attributes?

davidsemakula:
> We should clarify the relationship of the attribute `sol_name` (or `name`, if we generalize it) and `selector`. I think the intuitive behavior would be for `selector` to be the attribute that has the final say on the selector.

For Solidity ABI, `selector` is always a non-factor, because Solidity metadata specs and hence tools only accept function/event signatures not pre-computed selectors, so name override will always be used in that case.

For ink! ABI, it does make sense for `selector` having higher precedence over the name override if we support both.

> Could you add some tests in the PR that make sure we have a defined behavior for interactions of those attributes?

Ok, so I'm gonna generalize the name override attribute to `name`, make it available for both ABIs, and add a few more tests for expected behavior when both `name` and `selector` are specified for different ABI modes

davidsemakula:
@cmichi Refactor is complete, also see https://github.com/use-ink/ink/issues/2576 for updated issue description

I'll update the v6 docs after this is reviewed/merged

crates/ink/tests/compile_tests.rs
```diff
@@ -271,13 +277,23 @@ fn trybuild_wrapper_test(
...
     cmd.env("CARGO", wrapper);
+
+    // Set ABI `cfg` flags that will be passed to `cargo` by the `trybuild` wrapper.
     let abi_cfg = format!("--cfg\x1fink_abi=\"{abi}\"\x1f--check-cfg\x1fcfg(ink_abi,values(\"ink\",\"sol\",\"all\"))");
     cmd.env("TRYBUILD_WRAPPER_ENCODED_FLAGS", abi_cfg);
 
+    // Enable `std` and `unstable-hostfn` features (needed by events and metadata tests).
+    cmd.env(
+        "TRYBUILD_WRAPPER_CARGO_ARGS",
+        "--features std,unstable-hostfn",
```
cmichi, line 294: Why do you need to pass `unstable-hostfn` here?

PR #2580, opened 2025-08-06
Emit Solidity ABI encoded events

davidsemakula:
## Summary
Closes #_
- [y] y/n | Is it dependent on a specific version of `cargo-contract` or `pallet-revive`?


crates/ink/src/env_access.rs
```diff
@@ -396,12 +397,26 @@ where
         ink_env::minimum_balance::<E>()
     }
 
-    /// Emits an event.
+    /// Emits an event using the "default" ABI for interactions for the ink! project.
```
cmichi, line 400: Could you document the behavior for `abi = "all"` and add a test for it?

davidsemakula: There's a couple of lines just after this one i.e.
```rust
/// The "default" ABI for interactions is "ink", unless the ABI is set to "sol"
/// in the ink! project's manifest file (i.e. `Cargo.toml`).
```

So, at the moment (i.e. decision not final), in "all" ABI mode, this will emit an ink! ABI encoded event.
But may be the current doc isn't clear because it doesn't mention the "all" ABI mode explicitly?

In general, I plan to put more effort into this type of doc during the final pass for deciding "default" behavior and public interface for this and similar utilities in "all" ABI mode.

I'm gonna add an integration test for the current behavior though.

cmichi: Yeah I read those lines, it's what got me to comment this here.

Without knowing more about the implementation my expectation as a dev would have been that for `abi = "all"`, the events are emitted in both SCALE and Solidity encoding. Is that also your intuition?

davidsemakula: Yeah, the intuition tracks, I'll change the behavior to emitting both.
Probably something to revisit for a final decision before release, or when we switch to "all" as the "default" ABI, will add a note.

crates/primitives/src/types.rs
```diff
@@ -323,6 +323,7 @@ pub trait Environment: Clone {
     ///
     /// The value must match the maximum number of supported event topics of the used
     /// runtime.
+    // TODO: (@davidsemakula) Remove, no longer configurable in `pallet-revive`.
```
cmichi, line 326: There's also a bunch of places in `cargo-contract` where we need to remove it.

davidsemakula: Gotcha, will add it to the list for the follow up issue

cmichi (approved):
So in general the PR looks great! The thing worrying me is that we have an increase in code size for some contracts:

image

(first dropdown in the bot statistic)

Let's see how it change with your recent commits. It shouldn't change much for those contracts (as they still use `abi = "ink"`).

davidsemakula:
> So in general the PR looks great! The thing worrying me is that we have an increase in code size for some contracts:
> image
> 
> (first dropdown in the bot statistic)
> 
> Let's see how it change with your recent commits.

Good point, gonna try to do some code size optimizations before merging.

PR #2584, opened 2025-08-10
fix(test): correct typo "falliable" -> "fallible" in call-builder test

radik878:
Corrected assertion message in integration-tests/internal/lang-err/call-builder/lib.rs:
changed "Call to falliable constructor failed, when it should have succeeded." → "Call to fallible constructor failed, when it should have succeeded."


davidsemakula (approved):
Thanks!

PR #2641, opened 2025-09-16
Feature gate `xcm`

davidsemakula:
## Summary
Closes #_

## Description

- Introduces an `xcm` feature in `ink` and `ink_env`
- Makes `staging-xcm` dependency optional in `ink` and `ink_env` and only enabled by the `xcm` feature gate
- Place `xcm` utilities behind the `xcm` feature gate


davidsemakula:
> Nice! Can you check if we need to update `ink-docs`?

Will do!

PR #2659, opened 2025-09-29
refactor: ink sandbox crate

AlexD10S:
## Summary
Closes #_
- [ ] y/n | Is it dependent on a specific version of `cargo-contract` or `pallet-revive`?


crates/e2e/macro/src/lib.rs
```diff
@@ -57,10 +57,10 @@ use syn::Result;
 /// You can switch the E2E test to use the [DRink!](https://use.ink/basics/contract-testing/drink)
 /// testing framework with this syntax:
 ///
-/// ```
+/// ```no_compile
```
AlexD10S, line 60: It doesn't have `ink_sandbox` as dependency so it doesn't compile

crates/e2e/macro/src/config.rs
```diff
@@ -64,22 +63,21 @@ impl Node {
...
 
 /// The runtime emulator that should be used within `TestExternalities`
-#[cfg(any(test, feature = "sandbox"))]
 #[derive(Clone, Eq, PartialEq, Debug, darling::FromMeta)]
-pub enum RuntimeOnly {
-    #[darling(word)]
-    #[darling(skip)]
-    Default,
-    Sandbox(syn::Path),
+pub struct RuntimeOnly {
```
AlexD10S, line 67: This was a bit tricky: since we don’t want to import `ink_sandbox` in this crate, we can’t provide it as a default. It always has to be specified explicitly by the user.

crates/e2e/macro/src/lib.rs
cmichi: You could make `ink_sandbox` a dev-dependency?

cmichi (approved):
Thanks a lot for this refactor! Besides this one tiny nit and adding a changelog entry it looks good to me.

I think it makes the whole testing story way clearer. Could you also take care of updating `ink-docs` afterwards (also the migration guide)?

crates/e2e/macro/src/config.rs
davidsemakula: The `ink_e2e` machinery already has a good look at the manifest (or more specifically cargo metadata) to do stuff like compiling the root contract and it's contract dependencies (if any) (see [here](https://github.com/use-ink/ink/blob/96434d9ff94f9bc13372c238a76905a866d9e2f7/crates/e2e/src/contract_build.rs#L46-L58)).  
So we can do something there to bring back defaults (after this PR), while also providing user friendly errors when `ink_sandbox` is missing. 
I'm already planning to do something similar for when the user runs `cargo test` (not ABI aware) instead of `cargo contract test` (ABI aware), so I can look at this as well while doing that.

PR #2662, opened 2025-09-30
chore: update crates dependencies

AlexD10S:
This PR continues the work from https://github.com/use-ink/ink/pull/2657 and https://github.com/use-ink/ink/pull/2659 after removing all `pallet_revive` and extract the sandbox functionality into its own crate (`ink_sandbox`) we can now use the dependencies from crates.io for most of the crates and tests.


crates/env/Cargo.toml
```diff
@@ -32,9 +32,6 @@ xcm = { workspace = true, optional = true }
 hex-literal = "1"
 const-crypto = "0.3.0"
 
-sp-io = { git = "https://github.com/use-ink/polkadot-sdk.git", rev = "a71ec19a94702ea71767ba5ac97603ea6c6305c1", default-features = false, features = ["disable_panic_handler", "disable_oom", "disable_allocator"] }
```
AlexD10S, line 35: No reason to be in this crate

davidsemakula (approved):
LGTM!

PR #2687, opened 2025-10-22
Implement XCM API with `XcmPrecompile`

cmichi:
This PR implements an XCM API using the Polkadot SDK `XcmPrecompile`. It also stabilizes the API behind the feature `xcm`.

The `XcmPrecompile` interface can be found [here](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/pallet-xcm/precompiles/src/interface/IXcm.sol). The precompile implementation [here](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/pallet-xcm/precompiles/src/lib.rs).

Note that the `xcm_send` unfortunately no longer returns a `XcmHash`, which we still had in ink! v5.

### Room for Optimization

We could introduce an API function `xcm_weigh_and_execute`. This function would be more efficient than what needs to be done now: calling `xcm_weigh` and passing the result to `xcm_execute`. The function could deduplicate operations that are executed twice in this sequence (e.g. encoding the `msg`).

### Testing 

We can't test the `contract-xcm` example in our CI currently. This is because the `ink-node --chain ink-parachain-local` configuration is currently broken (doesn't produce blocks).

Here's how I've tested the contract locally:
```bash
cd polkadot-sdk/
cargo install --path polkadot/ --locked
cargo install --path substrate/bin/utils/chain-spec-builder/
cargo install --path cumulus/polkadot-parachain/

cargo build --release --manifest-path cumulus/parachains/runtimes/assets/asset-hub-westend/Cargo.toml
export runtime=target/release/wbuild/asset-hub-westend-runtime/asset_hub_westend_runtime.wasm 
chain-spec-builder -c "/dev/stdout" create --relay-chain "dev" --para-id 1000 -r $runtime named-preset development > ../westend.json

polkadot --dev &
polkadot-parachain --chain /tmp/westend.json --dev -lerror,runtime::revive=trace,xcm=trace &

cd ink/integration-tests/public/contract-xcm/
CONTRACTS_NODE_URL=ws://127.0.0.1:9944 cargo test --all-features 
```

### DIFF
--- a/crates/env/src/engine/on_chain/pallet_revive.rs
+++ b/crates/env/src/engine/on_chain/pallet_revive.rs
@@ -270,7 +270,6 @@ fn solidity_encode_bytes(
 /// of 32 for the given number).
 ///
 /// # Developer Note
-///
 // The implementation does not use `.div_ceil()`, as that function is more complex
 // and would use up more stack space.
 #[allow(clippy::manual_div_ceil)]


crates/env/src/engine/on_chain/pallet_revive.rs
```diff
@@ -108,7 +110,10 @@ impl CryptoHash for Blake2x128 {
         let sel = const { solidity_selector("hashBlake128(bytes)") };
         buffer[..4].copy_from_slice(&sel[..4]);
 
-        let n = solidity_encode_bytes(input, 32, &mut buffer[4..]);
+        // we pass `offset = 32`, as the `bytes` payload starts there (right after the
+        // offset word) we pass `data_pos = 32`, as the payload starts right after
+        // the offset word
```
davidsemakula:
```suggestion
        // offset word), we pass `offset_pos = 0`, as the callee takes only one argument.
```

crates/env/src/engine/on_chain/pallet_revive.rs
```diff
@@ -1275,52 +1319,125 @@ impl TypedEnvBackend for EnvInstance {
...
-        */
+        let sel = const { solidity_selector("execute(bytes,(uint64,uint64))") };
+        buffer[..4].copy_from_slice(&sel[..4]);
+
+        // 96 because 64 for `Weight` and 32 for `bytes` offset
+        let n = solidity_encode_bytes(enc_msg, 96, 0, &mut buffer[4..]);
+
+        let mut weight_bytes = [0u8; 64];
+        weight_bytes[24..32].copy_from_slice(&weight.ref_time().to_be_bytes());
+        weight_bytes[56..64].copy_from_slice(&weight.proof_size().to_be_bytes());
```
davidsemakula: nit: can write the weight bytes to `buffer` directly

davidsemakula (approved):
LGTM!

A few comments and suggestions below.

crates/env/src/engine/on_chain/pallet_revive.rs
```diff
@@ -195,47 +204,73 @@ const SOL_BYTES_ENCODING_OVERHEAD: usize = 64;
...
+    // Write the `input` at `offset_pos`, after the 32 byte `len` word
+    out[offset + BYTES_LEN_WORD..offset + BYTES_LEN_WORD + input_len]
+        .copy_from_slice(input);
 
     64 + padded_len
 }
 
 /// Returns the Solidity word padded length for the given input length (i.e. next multiple
 /// of 32 for the given number).
+// `.div_ceil()` would allocate on the heap
```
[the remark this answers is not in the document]
davidsemakula: Gotcha, that makes more sense ... given significantly worse `codegen` (i.e. it uses a lot more instructions)

crates/env/src/engine/on_chain/pallet_revive.rs
cmichi: That's a good idea! I've changed it.

PR #2709, opened 2025-11-03
test: assets precompile integration test with node backend

Daanvdplas:
Shows how to test a contract using a precompile using node as backend.

I have made it look as similar as the tests using the runtime as backend. Ultimately we would want a world where the developer just switches from:
```
ink_e2e::test(runtime) to ink_e2e::test(node)
```
and it just works.

With all these refactors we should revisit how unit testing works and make that all look the same as well.


crates/e2e/src/subxt_client.rs
```diff
@@ -96,6 +96,27 @@ use subxt::{
...
+    fn to_account_bytes(&self) -> [u8; 32];
+}
+
+impl ToAccountBytes for Keypair {
+    fn to_account_bytes(&self) -> [u8; 32] {
+        self.public_key().0
+    }
+}
+
+impl ToAccountBytes for ink::primitives::AccountId {
```
cmichi: Do we really need to introduce a new trait for this?  Please check if one of the traits that `AccountId` already implements is also implented by `Keypair`: https://github.com/use-ink/ink/blob/master/crates/primitives/src/types.rs#L102-L106.

cmichi (commented):
Thanks, Daan! I've got some nits and the CI is failing. Let's fix that and then we can merge it.

PR #2710, opened 2025-11-03
Refactor contract ref generation and add automatic re-exporting

davidsemakula:
## Summary
Closes #_

## Description

- Only generate contract refs in test, std or dependency mode
- Automatically re-export contract refs in test, std or dependency mode
  - Currently, adding a contract as a dependency is not enough to allow the root contract to access it's contract ref, because it's also required that the author of the imported contract also manually re-exported its contract ref. This PR removes this requirement by automatically re-exporting the contract ref when a contract is compiled in "dependency" mode.
- Implement `::ink::codegen::ContractCallBuilder` for contract refs
  - Allows contract refs to be used in place of the contract type in e2e tests (e.g. when using [ink_e2e::create_call_builder](https://use-ink.github.io/ink/ink_e2e/fn.create_call_builder.html), [InstantiationResult::call_builder](https://use-ink.github.io/ink/ink_e2e/struct.InstantiationResult.html) e.t.c)

## Follow ups
- [x] Update ink! docs (https://github.com/use-ink/ink-docs/pull/531)
- [x] Update examples (https://github.com/use-ink/ink-examples/pull/90)

### DIFF
--- a/crates/ink/macro/src/contract_ref.rs
+++ b/crates/ink/macro/src/contract_ref.rs
@@ -68,7 +68,7 @@ pub fn analyze_or_err(
         #trait_def_impl
 
         // Type alias for contract ref.
-        type #contract_ref_name =
+        pub type #contract_ref_name =
             <<::ink::reflect::TraitDefinitionRegistry<#env> as #trait_name>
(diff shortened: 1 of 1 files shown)


cmichi (approved):
Coolio! Are you taking care of `ink-docs`?

